PR #37, opened 2017-10-11
Better error check for NCCL init

alsrgv:
We currently allow `ncclCommInitRank` to update the value in the map directly, which causes an issue if there is an error - `ncclCommInitRank` may return corrupted value.

Thanks to @sjeaugey for the suggestion!


alsrgv:
Will release together with #26, so there's no version bump.

PR #88, opened 2017-11-11
Modify LRWarmupCallback to scale LR from `initial_lr / hvd.size()` to `initial_lr`

alsrgv:
Currently, `LRWarmupCallback` scales the learning up from `initial_lr` to `initial_lr * hvd.size()`. While this allows using the default optimizer constructor for a variety of situations, the scripts which resume training from an epoch that falls outside of warmup range need to manually adjust the learning rate to `initial_lr * hvd.size()`.

With this change, `LRWarmupCallback` will scale LR from `initial_lr / hvd.size()` to `initial_lr`, allowing training to resume from a later checkpoint with the correct (scaled up) learning rate.

**Note**: Since this change modifies semantic of `LRWarmupCallback`, users should modify their scripts accordingly: multiply an initial learning rate by `hvd.size()`. To make sure users notice the change, `LRWarmupCallback` will be renamed to `LearningRateWarmupCallback`.

```python
# Adjust learning rate based on number of GPUs.
opt = keras.optimizers.Adadelta(1.0 * hvd.size())
```


examples/keras_mnist_advanced.py
```diff
@@ -68,7 +68,8 @@
 model.add(Dropout(0.5))
 model.add(Dense(num_classes, activation='softmax'))
 
-opt = keras.optimizers.Adadelta()
+# Adjust learning rate based on number of GPUs.
+opt = keras.optimizers.Adadelta(lr=1.0 * hvd.size())
```
alsrgv: Note that this is now aligned with the simple example.

PR #136, opened 2018-01-02
Dockerize Travis as TensorFlow will require new GLIBC

alsrgv:
TensorFlow upgraded base image to Ubuntu 16 (https://github.com/tensorflow/tensorflow/issues/15777).  Travis CI does not support Ubuntu 16 (https://github.com/travis-ci/travis-ci/issues/5821), so we need to dockerize our tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,17 +4,36 @@ python:
   - "2.7"
   - "3.4"
   - "3.6"
+services:
+  - docker
 
 before_install:
-  - sudo apt-get -qq update
-
-  - if [[ $TRAVIS_PYTHON_VERSION == 2* ]]; then pip install --upgrade pip; fi
-  - if [[ $TRAVIS_PYTHON_VERSION == 3* ]]; then pip3 install --upgrade pip; fi
-
-  # upgrade numpy to make sure `import tensorflow` does not cause numpy version error
-  # RuntimeError: module compiled against API version 0xb but this version of numpy is 0xa
-  - if [[ $TRAVIS_PYTHON_VERSION == 2* ]]; then pip install -U numpy; fi
-  - if [[ $TRAVIS_PYTHON_VERSION == 3* ]]; then pip3 install -U numpy; fi
+  # force latest Debian for Python 3.6 and nightly TensorFlow which requires new glibc
+  - |
+    if [[ $TRAVIS_PYTHON_VERSION == "3.6" || $TF_PACKAGE == "tf-nightly" ]]; then
+      export DEBIAN=sid
+    else
+      export DEBIAN=jessie
+    fi
+  - docker pull debian:$DEBIAN
+  # run docker container for an hour
+  - docker run -v `pwd`:/horovod debian:$DEBIAN /bin/sh -c "sleep 3600" &
+  # wait for docker to start
+  - sleep 5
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -30,69 +49,57 @@ matrix:
       env: TF_PACKAGE=tensorflow==1.1.0 KERAS_PACKAGE=keras==2.0.0 MPI=MPICH
     - python: "3.6"
       env: TF_PACKAGE=tensorflow==1.1.0 KERAS_PACKAGE=keras==2.0.0 MPI=MPICH
-    - python: "3.6"
+    - python: "3.4"
```
jeevandev: why do we run tf-nightly with 3.4 now?

.travis.yml
```diff
@@ -30,69 +49,57 @@ matrix:
...
 
 install:
   - |
-    if [ $MPI == "OpenMPI" ]; then
-      wget -O /tmp/openmpi-3.0.0-bin.tar.gz https://github.com/uber/horovod/files/1439728/openmpi-3.0.0-bin.tar.gz
-      pushd /usr/local && sudo tar -zxf /tmp/openmpi-3.0.0-bin.tar.gz && popd
-      export PATH=/usr/local/openmpi/bin:$PATH
+    if [[ $MPI == "OpenMPI" ]]; then
+      docker exec $CONTAINER /bin/sh -c "wget -O /tmp/openmpi-3.0.0-bin.tar.gz https://github.com/uber/horovod/files/1596799/openmpi-3.0.0-bin.tar.gz"
+      docker exec $CONTAINER /bin/sh -c "cd /usr/local && tar -zxf /tmp/openmpi-3.0.0-bin.tar.gz && ldconfig"
```
jeevandev: does ldconfig take care of the PATH?

.travis.yml
alsrgv: This is exclusion matrix.  We used to run it with 3.4.  Now we run it with 3.6 since the base image is upgraded and it doesn't have 3.4 anymore.

.travis.yml
```diff
@@ -30,69 +49,57 @@ matrix:
...
     else
       # installs mpich version 3.0.4
-      sudo apt-get install mpich
+      docker exec $CONTAINER /bin/sh -c "apt-get install -y mpich"
     fi
 
   # TensorFlow
-  - if [[ $TRAVIS_PYTHON_VERSION == 2* ]]; then pip install "$TF_PACKAGE"; fi
-  - if [[ $TRAVIS_PYTHON_VERSION == 3* ]]; then pip3 install "$TF_PACKAGE"; fi
+  - docker exec $CONTAINER /bin/sh -c "pip install $TF_PACKAGE"
```
jeevandev: so the force upgrade, upgrades pip to pip3 incase of python3? Hence we don't have to do explicit pip3 installs for python3.x?

.travis.yml
alsrgv: I changed `openmpi-3.0.0-bin.tar.gz` to be installable directly into `/usr/local`. `mpirun` ends up being installed as `/usr/local/bin/mpirun`.

.travis.yml
alsrgv: Purpose of force upgrade is to create `pip` binary even for Python3, which `apt install python3-pip` fails to do.

PR #143, opened 2018-01-08
Fix/Update Keras github url in .travis.yml

ziky90:
They have recently changed keras github url. Redirect from `fchollet` to `keras-team` still works, but this PR better changes the url to the new one in the .travis.yml.


alsrgv:
@ziky90, thanks for the PR! LGTM.

PR #185, opened 2018-02-28
Abstract common Horovod operations from any specific framework

alsrgv:
In preparation for support of PyTorch, CNTK, or any other framework.


horovod/common/hashes.h
```diff
@@ -33,6 +34,26 @@ template <typename T> struct hash<std::vector<T>> {
   }
 };
 
+template <typename U, typename V> struct hash<std::tuple<U, V>> {
+  typedef std::tuple<U, V> argument_type;
+  typedef std::size_t result_type;
+
+  result_type operator()(argument_type const& in) const {
+    size_t seed = 0;
+    seed ^= std::hash<U>()(std::get<0>(in)) + 0x9e3779b9 + (seed << 6) +
```
jeevandev: why this 0x9e3779b9 specific value? can we define it as a constant and not use the magic number?

horovod/common/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def local_size():
+    """A function which returns the number of Horovod processes within the
+    node the current process is running on.
+
+    Returns:
+      An integer scalar containing the number of local Horovod processes.
+    """
+    local_size = MPI_LIB_CTYPES.horovod_local_size()
```
jeevandev: Where is this initialized in refactored code? I did not see

horovod/common/timeline.h
```diff
@@ -20,26 +20,19 @@
 #include <fstream>
 #include <iostream>
 #include <unordered_map>
+#include <mutex>
+#include <atomic>
```
jeevandev: Why is mutex/atomic included?

horovod/common/__init__.py
alsrgv: It's a function in `operations.cc`

horovod/common/operations.cc
```diff
@@ -0,0 +1,1577 @@
...
+        CUDA_CHECK(
+            entries, "cudaStreamSynchronize",
+            cudaStreamSynchronize(horovod_global.streams[first_entry.device]))
+      }
+#endif
+      ACTIVITY_END_ALL(entries, timeline)
+    } else {
+      auto e = first_entry;
+      ACTIVITY_START_ALL(entries, timeline, "MPI_ALLREDUCE")
+      const void* sendbuf = e.tensor->data() == e.output->data()
```
alsrgv: Here we've added support for in-place allreduce.

horovod/common/hashes.h
```diff
@@ -0,0 +1,71 @@
...
+
+} // namespace
+
+template <typename T> struct hash<std::vector<T>> {
+  typedef std::vector<T> argument_type;
+  typedef std::size_t result_type;
+
+  result_type operator()(argument_type const& in) const {
+    size_t size = in.size();
+    size_t seed = 0;
```
thepaulm: for consistency this should really be `result_type seed = 0;`

horovod/common/hashes.h
```diff
@@ -0,0 +1,71 @@
...
+    return seed;
+  }
+};
+
+template <typename U, typename V> struct hash<std::tuple<U, V>> {
+  typedef std::tuple<U, V> argument_type;
+  typedef std::size_t result_type;
+
+  result_type operator()(argument_type const& in) const {
+    size_t seed = 0;
```
thepaulm: this guy too

horovod/common/operations.cc
```diff
@@ -0,0 +1,1577 @@
+// Copyright 2016 The TensorFlow Authors. All Rights Reserved.
```
thepaulm: How is this copyright tensorflow? did this file start from there? We should have Uber copyright?

horovod/common/operations.cc
```diff
@@ -0,0 +1,1577 @@
...
+  } else if (message_type == MPIRequest::BROADCAST) {
+    response.set_response_type(MPIResponse::BROADCAST);
+  }
+  response.set_devices(devices);
+
+  // Clear all queued up requests for this name. They are now taken care of
+  // by the constructed MPI response.
+  message_table->erase(it);
+
+  return response;
```
thepaulm: A valid C++ pattern is to have like this:

```
return response;

error_return:
    return Failure;
```

and then instead of all of these early exit `if error` you can just do `goto error_return;`

horovod/common/operations.cc
```diff
@@ -0,0 +1,1577 @@
...
+
+Status CheckInitialized() {
+  if (!horovod_global.initialization_done) {
+    return Status::PreconditionError(
+        "Horovod has not been initialized; use hvd.init().");
+  }
+  return Status::OK();
+}
+
+extern "C" void horovod_init() { InitializeHorovodOnce(); }
```
thepaulm: You can make one big `extern "C"` block here ...

horovod/common/operations.cc
alsrgv: We do have Uber copyright there as well.

PR #188, opened 2018-03-02
MPI request fusion

alsrgv:
Fuse all available MPI requests together into a single request to save total # of MPI communications.  Improves performance on a very large scale (512+ GPUs).


alsrgv:
Will rebase once previous PR lands.

PR #201, opened 2018-03-10
Allow workers to send SHUTDOWN request to coordinator

alsrgv:
If a worker process fails due to an application exception, allows it to signal to coordinator to request a shutdown, so that whole MPI application can be terminated, rather than hang.


horovod/common/operations.cc
```diff
@@ -1237,15 +1245,17 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
         }
       }
     } else {
-      std::string encoded_message;
-      MPIRequestList message_list;
-      while (!message_queue.empty()) {
-        message_list.add_requests(message_queue.front());
-        message_queue.pop();
+      if (!message_queue.empty()) {
```
alsrgv: Before we were always sending this empty message list, this is not necessary.

PR #206, opened 2018-03-12
Pin Dockerfile to a specific cuDNN version.

### LINKED ISSUE #192: an issue in the dockerfile, seems the cudnn version does't match.
tensorflow/stream_executor/cuda/cuda_dnn.cc:378] Loaded runtime CuDNN library: 7101 (compatibility version 7100) but source was compiled with 7004 (compatibility version 7000).  If using a binary install, upgrade your CuDNN library to match.  

maybe it is because the nvidia/cuda image has been updated.

### LINKED ISSUE #203: dockerfile need to downgrade cuDNN
if there is anyone who are faced with the error below,

> "Loaded runtime CuDNN library: 7101 (compatibility version 7100) but source was compiled with 7004 (compatibility version 7000). "

The thread below can be pointer to resolve this issue, 
https://github.com/keras-team/keras/issues/9567

and the command lines on the reply below could be added to the dockerfile of the current horovod until Keras supports new version cuDNN.
https://github.com/keras-team/keras/issues/9567#issuecomment-371566669

rongou:
The current Dockerfile is broken because of mismatched cuDNN versions (I filed a bug against tensorflow: https://github.com/tensorflow/tensorflow/issues/17566). Test image pushed to rongou/horovod on dockerhub.

Fixes #192, fixes #203.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,9 @@
-FROM nvidia/cuda:9.0-cudnn7-devel-ubuntu16.04
+FROM nvidia/cuda:9.0-devel-ubuntu16.04
 
 # TensorFlow version is tightly coupled to CUDA and cuDNN so it should be selected carefully
 ENV TENSORFLOW_VERSION=1.6.0
+ENV CUDNN_VERSION=7.0.5.15-1+cuda9.0
+ENV NCCL_VERSION=2.1.15-1+cuda9.0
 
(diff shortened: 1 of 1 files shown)


alsrgv:
Thanks for the PR!  I made #204 along the same lines, but I do like your version better.  Fixes #192, fixes #203.

Issue #205, opened 2018-03-12
Unable to run distributed training with two server that both using Anaconda


traveller59:
## Environment
Ubuntu 16.04
cuda9.0, cudnn7.0
openmpi with cuda

## Problems
When I run horovod in local environment, every thing is fine, but then I try to run horovod in a remote machine, ```print(sys.version)``` print 'python 2.7' but my Anaconda environment is python 3.6, I have add anaconda path to .bashrc and when I directly use ssh to access remote machine and run python, the python environment is normal(3.6.4).
## command
```
mpirun -np 2  --allow-run-as-root\
    -H server1:1,server2:1 \
    -bind-to none -map-by slot \
    -mca pml ob1 -mca btl ^openib \
    python train_horovod.py
```

alsrgv:
@traveller59, instead of using `python`, can you try using `/path/to/anaconda/bin/python`?  If you're using a custom environment, you may need to activate it in the command, too.

ink1:
You can try
```
mpirun -np 2 -H server1:1,server2:1 which python
```
to find out which python is being picked up.

I'm playing with horovod and really pleased with it. I'm also using anaconda and simply pass env vars
```
mpirun -np 2 -H server1:1,server2:1 -x LD_LIBRARY_PATH -x PATH which python
```

I'm not even using Open MPI with CUDA support, just the one I'd built previously.
Ethernet 10 Gbps, two servers with 4x P100 in each give me 1712s (342ms/step) on the first ImageNet epoch running keras_imagenet_resnet50.py (also cuda 9.0, cudnn 7.0 & nccl 2.1)

Guys, great library!

alsrgv:
Thanks, @ink1!  It's a good idea to pass `-x PATH`, we shall update our docs to include it in sample commands.

traveller59:
Thanks! Now the problem is solved, I use that cnn benchmark with two cloud server(1xP100, 3Gbps) and get 130->180 speed up.

PR #207, opened 2018-03-12
Add -x PATH to command line examples

alsrgv:
See #205


docs/docker.md
```diff
@@ -24,8 +24,9 @@ root@c278c88dd552:/examples# mpirun -np 4 -H localhost:4 python keras_mnist_adva
 ```
 
 You may notice that this command does not have a few options recommended in other parts of documentation:
```
sblotner: rephrase this now to "This command does not have options recommended in other parts of the documentation. `-bind-to none -map-by slot -x NCCL_DEBUG=INFO` options are already ...

Issue #216, opened 2018-03-16
Multiple MPI ranks per GPU process


mkovalski:
Hello,

I wanted to know if there's a way to use multiple processes per GPU rank. For example, say I wanted there to be 4 total processes per GPU, where 3 of those processes simply do data augmentation and the other to get the data from the other 3 processes and feed to the training. I was able to set this up using a single GPU using MPI Groups without horovod, but now I wanted to scale this up. Would something like this be possible?

alsrgv:
@mkovalski, would it work if we had a way to set up a communicator across a specific list of ranks as part of `hvd.init()` - do you know that information up front, or would you need MPI to discover topology?

mkovalski:
@alsrgv Ideally, I'd use MPI first to find topology but I can always grab this from some environment variables I use instead (for my use, grab whatever slurm environment variables I can)

alsrgv:
@mkovalski, that seems like sub-par experience.  I guess we could also add `mpi_init=False` to `hvd.init()` in case you ran `MPI_Init_thread()` before using mpi4py.  Would those two things be ideal?

mkovalski:
@alsrgv Yeah that would work really well then, that in addition to specifying ranks to hvd init would be great!

alsrgv:
OK, excellent.  This should be pretty straightforward.  I will mark this as a "contribution welcome" task in case you want to work on it before we get to it :-)

alsrgv:
@mkovalski, BTW, another way to solve this was mentioned in #105, I wonder if it may be relevant to you, too.  The idea was to be able to specify a subset of ranks per operation.  The benefit is that you can do separately an allreduce within node and an allreduce between nodes.  The downside is that all workers would still participate in global MPI coordination, but those are lightweight messages and it should not really be a problem.

PR #218, opened 2018-03-17
Support for not calling MPI_Init_thread()/MPI_Finalize() in Horovod lifecycle

lkmokadam:
enhancement #216 and fix for potential bug due to this enhancement. 

import all module after disabling MPI.Init and MPI.Finalize and 
   i)hvd.init(mpi_init=True, mpi_finalize=False) (run_1.py)
        or 
   ii) calling MPI.Init (run_2.py) 

New function:
hvd.close()

closes horovod if mpi_finalize=False. Provides mechanism to make sure that horovod thread will not 
call MPI_Finalize while other MPI threads still running.
if mpi_finalize=true, this function has no effect

### DIFF
--- a/horovod/common/__init__.py
+++ b/horovod/common/__init__.py
@@ -37,10 +37,10 @@ MPI_COMMON_LIB_CTYPES = \
                              'mpi_lib' + get_ext_suffix()), mode=ctypes.RTLD_GLOBAL)
 
 
-def init(auto_mpi_init=True ,auto_mpi_finalize=True):
+def init(mpi_init=True ,mpi_finalize=True):
     """A function that initializes Horovod.
     """
-    return MPI_COMMON_LIB_CTYPES.horovod_init(ctypes.c_int(auto_mpi_init),ctypes.c_int(auto_mpi_finalize))
+    return MPI_COMMON_LIB_CTYPES.horovod_init(ctypes.c_bool(mpi_init),ctypes.c_bool(mpi_finalize))
 
 def close():  
     """A function that initializes Horovod.
--- a/horovod/common/__init__.py
+++ b/horovod/common/__init__.py
@@ -17,6 +17,7 @@
 import ctypes
 import os
 import sysconfig
+import atexit
 
 
 def get_ext_suffix():
@@ -42,11 +43,12 @@ def init(mpi_init=True ,mpi_finalize=True):
     """
     return MPI_COMMON_LIB_CTYPES.horovod_init(ctypes.c_bool(mpi_init),ctypes.c_bool(mpi_finalize))
 
-def close():  
-    """A function that initializes Horovod.
+def _close():  
+    """A function used by horovod to close Horovod during normal interpreter termination.
     """
     return MPI_COMMON_LIB_CTYPES.horovod_close()
 
+atexit.register(close)
 
 def size():
     """A function that returns the number of Horovod processes.
--- a/horovod/common/__init__.py
+++ b/horovod/common/__init__.py
@@ -37,10 +37,15 @@ MPI_COMMON_LIB_CTYPES = \
                              'mpi_lib' + get_ext_suffix()), mode=ctypes.RTLD_GLOBAL)
 
 
-def init():
+def init(auto_mpi_init=True ,auto_mpi_finalize=True):
     """A function that initializes Horovod.
     """
-    return MPI_COMMON_LIB_CTYPES.horovod_init()
+    return MPI_COMMON_LIB_CTYPES.horovod_init(ctypes.c_int(auto_mpi_init),ctypes.c_int(auto_mpi_finalize))
+
+def close():  
+    """A function that initializes Horovod.
+    """
+    return MPI_COMMON_LIB_CTYPES.horovod_close()
 
 
 def size():
--- a/horovod/common/__init__.py
+++ b/horovod/common/__init__.py
@@ -48,7 +48,7 @@ def _close():
     """
     return MPI_COMMON_LIB_CTYPES.horovod_close()
 
-atexit.register(close)
+atexit.register(_close)
 
 def size():
     """A function that returns the number of Horovod processes.
--- a/horovod/common/__init__.py
+++ b/horovod/common/__init__.py
@@ -38,17 +38,20 @@ MPI_COMMON_LIB_CTYPES = \
                              'mpi_lib' + get_ext_suffix()), mode=ctypes.RTLD_GLOBAL)
 
 
-def init(mpi_init=True ,mpi_finalize=True):
+def init(mpi_init=True, mpi_finalize=True):
     """A function that initializes Horovod.
+        mpi_init : if True, Automatic MPI initialization will be done by horovod at import (default: True).
+        mpi_finalize : if True, Automatic MPI finalization will be done by horovod at exit(default: True).
     """
-    return MPI_COMMON_LIB_CTYPES.horovod_init(ctypes.c_bool(mpi_init),ctypes.c_bool(mpi_finalize))
+    atexit.register(finalize)
+    return MPI_COMMON_LIB_CTYPES.horovod_init(ctypes.c_bool(mpi_init), ctypes.c_bool(mpi_finalize))
 
-def _close():  
-    """A function used by horovod to close Horovod during normal interpreter termination.
+
+def finalize():
+    """A function that will close Horovod and .
     """
-    return MPI_COMMON_LIB_CTYPES.horovod_close()
+    return MPI_COMMON_LIB_CTYPES.horovod_finalize()
 
-atexit.register(_close)
 
 def size():
     """A function that returns the number of Horovod processes.
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -1171,6 +1171,7 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
   // be used together with Horovod if multi-threaded MPI is installed.
   int provided;
 
+  // Calls MPI_Init if horovod_global.auto_mpi_init is set otherwise prints warning
   if(horovod_global.auto_mpi_init){
     int is_mpi_initialized= 0;
     MPI_Initialized(&is_mpi_initialized);
@@ -1481,7 +1482,11 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
     (*it)(SHUT_DOWN_ERROR);
   }
 
-  while(true){
+  // Calls MPI_Finalize if horovod_global.auto_mpi_finalize is set otherwise prints warning
+  if(!horovod_global.auto_mpi_finalize){
+    std::cerr << "WARNING: Make sure to call MPI_Finalize." << std::endl
+  }
+  while(true){ 
     if(horovod_global.auto_mpi_finalize){
       int is_mpi_finalized = 0;
       MPI_Finalized(&is_mpi_finalized);
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -1179,7 +1179,7 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
       MPI_Init_thread(NULL, NULL, MPI_THREAD_MULTIPLE, &provided);
   }else{
     MPI_Query_thread(&provided);
-    std::cerr << "WARNING: Make sure to call MPI_Init before calling"
+    std::cerr << "WARNING: Make sure to call MPI_Init before calling "
                  "hvd.init otherwise horovod will crash." << std::endl;
   }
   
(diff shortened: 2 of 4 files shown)


alsrgv:
Thanks for the PR!  Before we dive deep into the details, I wanted to figure out the design.

Wouldn't it be sufficient to just add `def init(mpi_init=True, mpi_finalize=True)` that would do four things:

1. If `mpi_init=True`, do existing `MPI_Init_thread()`
2. If `mpi_init=False`, check for `MPI_Initialized()` and `MPI_Query_thread() == MPI_THREAD_MULTIPLE` and print corresponding warnings if they're not set.  We should also check in that case whether code crashes or hangs waiting for init to finish.  It should crash.
3. If `mpi_finalize=True`, do existing `MPI_Finalize()`
4. If `mpi_finalize=False`, just allow background thread to exit and rely on outer MPI (e.g. `mpi4py`) to call `MPI_Finalize()`.

lkmokadam:
Yeah. That is correct. But I am worrying about 4th scenario especially if one want to use as library in C/C++. If mpi_finalize=False and we rely on outer MPI (e.g. mpi4py) to call MPI_Finalize(), there can be a case when user might call MPI_Finalize even before the horovod hasn't processed all its collective operations. So, I think, we cannot rely on user to call MPI Finalize using current API. There is no way to know the user about the status of horovod's backgroundthreadloop. We can handle this by horovod.close()  or horovod.is_running() (return true if horovod thread is still running) or any thing like this. This will guide user for when to call MPI_Finalize while using horovod. What do you think about this?

alsrgv:
@lkmokadam, you bring a very good point about the race between MPI_Finalize() from the main program and MPI ops in the Horovod callback.

`horovod_finalize()` could help to resolve this race by nudging BackgroundThread to terminate the [same way the destructor does it](https://github.com/uber/horovod/blob/master/horovod/common/operations.cc#L192).

One situation that may still cause a problem is a user script throwing an exception:
```python
from mpi4py import MPI

hvd.init(mpi_init=False, mpi_finalize=False)
raise Exception('boo')
```

We should make sure that above program terminates with an error that clearly points at the exception, rather than at the fact that `MPI_Probe()` was called after `MPI_Finalize()` by `mpi4py`.  I'm not sure how to resolve this problem, do you have ideas?

alsrgv:
@lkmokadam, can you check what would happen if we add `hvd.finalize()` into [atexit](https://docs.python.org/2/library/atexit.html) handler in `hvd.init()`?

According to [atexit page](https://docs.python.org/2/library/atexit.html), if handlers were registered in order A, B, C it would call `atexit` in C, B, A order.  We know that `hvd.init(mpi_init=False)` was called after MPI was initialized - there would be a failure if it wasn't, so this seems to be covering all the scenarios and also doesn't require user to add `hvd.finalize()` to his code.

lkmokadam:
Yeah. It looks good to me.

horovod/common/__init__.py
```diff
@@ -37,11 +38,17 @@ def get_ext_suffix():
                              'mpi_lib' + get_ext_suffix()), mode=ctypes.RTLD_GLOBAL)
 
 
-def init():
+def init(mpi_init=True ,mpi_finalize=True):
```
alsrgv: Please run `autopep8` and `clang-format` for Python and C++ source files that you change.

horovod/common/__init__.py
```diff
@@ -37,11 +38,17 @@ def get_ext_suffix():
...
     """A function that initializes Horovod.
     """
```
alsrgv: Please add `mpi_init` and `mpi_finalize` to the docstring.

horovod/common/__init__.py
```diff
@@ -37,11 +38,17 @@ def get_ext_suffix():
...
-    return MPI_COMMON_LIB_CTYPES.horovod_init()
+    return MPI_COMMON_LIB_CTYPES.horovod_init(ctypes.c_bool(mpi_init),ctypes.c_bool(mpi_finalize))
 
+def _close():  
+    """A function used by horovod to close Horovod during normal interpreter termination.
+    """
+    return MPI_COMMON_LIB_CTYPES.horovod_close()
+
+atexit.register(_close)
```
alsrgv: Let's do this in `init()` rather than at the module import.

horovod/common/__init__.py
```diff
@@ -37,11 +38,17 @@ def get_ext_suffix():
...
 
 
-def init():
+def init(mpi_init=True ,mpi_finalize=True):
     """A function that initializes Horovod.
     """
-    return MPI_COMMON_LIB_CTYPES.horovod_init()
+    return MPI_COMMON_LIB_CTYPES.horovod_init(ctypes.c_bool(mpi_init),ctypes.c_bool(mpi_finalize))
 
+def _close():
```
alsrgv: Let's rename `close` to `finalize` to align better with MPI spirit, and make it public (i.e. no `_` prefix).

horovod/common/__init__.py
```diff
@@ -37,11 +38,17 @@ def get_ext_suffix():
...
 
-def init():
+def init(mpi_init=True ,mpi_finalize=True):
     """A function that initializes Horovod.
     """
-    return MPI_COMMON_LIB_CTYPES.horovod_init()
+    return MPI_COMMON_LIB_CTYPES.horovod_init(ctypes.c_bool(mpi_init),ctypes.c_bool(mpi_finalize))
 
+def _close():  
+    """A function used by horovod to close Horovod during normal interpreter termination.
```
alsrgv: close -> finalize.  Not sure what "normal interpreter termination" means - this will be called in case of an exception, too. Maybe just get rid of that part.

horovod/common/operations.cc
```diff
@@ -149,13 +149,16 @@ struct HorovodGlobalState {
...
-  // multi-threading is supported.
+  // The MPI rank, local rank, size, local size, flag indicating whether MPI
+  // multi-threading is supported and auto_mpi_finalize indicating whether to
+  // call MPI_Finalize automatically
   int rank = 0;
   int local_rank = 0;
   int size = 1;
   int local_size = 1;
   bool mpi_threads_supported = false;
+  bool auto_mpi_init = true;
```
alsrgv: Let's rename these just `mpi_init` and `mpi_finalize` to align with parameter names.

horovod/common/operations.cc
```diff
@@ -1167,8 +1170,20 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
   // We will ask for multiple threads, so other libraries like mpi4py can
   // be used together with Horovod if multi-threaded MPI is installed.
   int provided;
-  MPI_Init_thread(NULL, NULL, MPI_THREAD_MULTIPLE, &provided);
 
+  // Calls MPI_Init if horovod_global.auto_mpi_init is set otherwise prints warning
+  if(horovod_global.auto_mpi_init){
+    int is_mpi_initialized= 0;
+    MPI_Initialized(&is_mpi_initialized);
+    if(!is_mpi_initialized)
```
alsrgv: Let's make logic simpler - if `mpi_init=True`, let's just add a: WARNING: MPI has already been initialized, please use `hvd.init(mpi_init=False, mpi_finalize=False)` if you're using other frameworks that initialize MPI prior to Horovod.

horovod/common/operations.cc
```diff
@@ -1167,8 +1170,20 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
...
+      MPI_Init_thread(NULL, NULL, MPI_THREAD_MULTIPLE, &provided);
+  }else{
+    MPI_Query_thread(&provided);
```
alsrgv: Here, let's check that MPI has been initialized & provided thread level == MPI_THREAD_MULTIPLE.  If those are both true, we don't need a warning.  If they are false, we should provide a warning and say that process will now likely crash.

horovod/common/operations.cc
```diff
@@ -1467,14 +1482,38 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
...
 }
 
 // Start Horovod background thread. Ensure that this is
 // only done once no matter how many times this function is called.
-void InitializeHorovodOnce() {
+void InitializeHorovodOnce(bool auto_mpi_init,bool auto_mpi_finalize) {
   // Ensure background thread is only started once.
   if (!horovod_global.initialize_flag.test_and_set()) {
+
+    horovod_global.auto_mpi_init = auto_mpi_init;
```
alsrgv: We can actually just pass these as arguments to BackgroundThreadLoop.

horovod/common/operations.cc
```diff
@@ -1496,7 +1535,11 @@ Status CheckInitialized() {
 
 extern "C" {
 
-void horovod_init() { InitializeHorovodOnce(); }
+void horovod_init(bool auto_mpi_init,bool auto_mpi_finalize) { 
+
```
alsrgv: Remove extra spaces (style)

horovod/common/operations.cc
```diff
@@ -1467,14 +1482,38 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
     (*it)(SHUT_DOWN_ERROR);
   }
 
-  MPI_Finalize();
+  // Calls MPI_Finalize if horovod_global.auto_mpi_finalize is set otherwise prints warning
+  while(true){
```
alsrgv: I'm not sure this actually works.  What needs to happen is `horovod_finalize()` should have similar logic to `~HorovodGlobalState()` - we need to set `shut_down` flag and wait for the background thread to terminate.

alsrgv:
@lkmokadam, any luck with the fixes?  Would be great to get this landed once you make modifications.

lkmokadam:
@alsrgv will complete this by monday EOD.

horovod/common/__init__.py
```diff
@@ -37,10 +38,19 @@ def get_ext_suffix():
                              'mpi_lib' + get_ext_suffix()), mode=ctypes.RTLD_GLOBAL)
 
 
-def init():
+def init(mpi_init=True, mpi_finalize=True):
     """A function that initializes Horovod.
+        mpi_init : if True, Automatic MPI initialization will be done by horovod at import (default: True).
```
alsrgv: Please align style with https://github.com/uber/horovod/blob/master/horovod/keras/__init__.py#L70

horovod/common/__init__.py
```diff
@@ -37,10 +38,19 @@ def get_ext_suffix():
...
+        mpi_finalize : if True, Automatic MPI finalization will be done by horovod at exit(default: True).
     """
-    return MPI_COMMON_LIB_CTYPES.horovod_init()
+    atexit.register(finalize)
+    return MPI_COMMON_LIB_CTYPES.horovod_init(ctypes.c_bool(mpi_init), ctypes.c_bool(mpi_finalize))
+
+
+def finalize():
+    """A function that will close Horovod and .
```
alsrgv: Finish the sentence?

horovod/common/operations.cc
```diff
@@ -149,8 +149,9 @@ struct HorovodGlobalState {
   // Whether MPI_Init has been completed on the background thread.
   bool initialization_done = false;
 
-  // The MPI rank, local rank, size, local size and flag indicating whether MPI
-  // multi-threading is supported.
+  // The MPI rank, local rank, size, local size, flag indicating whether MPI
+  // multi-threading is supported and mpi_finalize indicating whether to
```
alsrgv: These got removed

horovod/common/operations.cc
```diff
@@ -1471,16 +1494,18 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
     (*it)(SHUT_DOWN_ERROR);
   }
 
-  MPI_Finalize();
+  if (mpi_finalize)
+    MPI_Finalize();
```
alsrgv: Please surround single-statement conditions with { }

horovod/common/operations.cc
```diff
@@ -1164,14 +1165,35 @@ void CheckForStalledTensors(HorovodGlobalState& state) {
...
+  MPI_Initialized(&is_mpi_initialized);
+
+  // Calls MPI_Init if mpi_init is set otherwise prints warning
+  if (mpi_init) {
+    if (is_mpi_initialized) {
+      std::cerr << " WARNING: MPI has already been initialized, please use "
+                   "hvd.init(mpi_init=False, mpi_finalize=False) if you're "
+                   "using other "
+                   "frameworks that initialize MPI prior to Horovod."
+                << std::endl;
```
alsrgv: If this happens, do you see MPI errors due to double finalization?  If so, we should mention it here.

alsrgv: Let's do the check here for thread support and if multithreading is not supported, print another warning similar to below that process may crash.

horovod/common/operations.cc
```diff
@@ -1164,14 +1165,35 @@ void CheckForStalledTensors(HorovodGlobalState& state) {
...
+    } else
+      MPI_Init_thread(NULL, NULL, MPI_THREAD_MULTIPLE, &provided);
+  } else {
+    MPI_Query_thread(&provided);
+    if (!is_mpi_initialized || provided != MPI_THREAD_MULTIPLE) {
+      std::cerr << " WARNING: MPI has not initialized with MPI_THREAD_MULTIPLE "
```
alsrgv: Let's split into two issues.  If MPI was not initialized, mention that Horovod was requested not to initialize MPI, but it wasn't initialized before so the process WILL crash.  If MPI was initialized without multi-threading, we can say that MPI was initialized outside of Horovod without MPI_THREAD_MULTIPLE, and the process MAY crash.

horovod/common/operations.cc
```diff
@@ -1536,6 +1563,19 @@ int horovod_mpi_threads_supported() {
...
   return horovod_global.mpi_threads_supported ? 1 : 0;
 }
+
+int horovod_finalize() {
+  if (!horovod_global.initialization_done) {
+    return -1;
+  }
+
+  // Sets the sutdown flag and wait for background thread to terminate.
+  if (horovod_global.background_thread.joinable()) {
```
alsrgv: If you do this in multiple threads, you could have a funny business.  I suggest making `shut_down` an `std::atomic_flag`, and doing similar checks to `horovod_init()` to make sure this will only happen once.

horovod/common/operations.h
```diff
@@ -37,7 +37,11 @@ Status CheckInitialized();
 extern "C" {
 
 // C interface to initialize Horovod.
-void horovod_init();
+void horovod_init(bool auto_mpi_init_flag, bool auto_mpi_finalize_flag);
```
alsrgv: Align parameter names with .cc file code.

alsrgv:
@lkmokadam, any luck with the new review comments?

wgale:
Hi, has there been any progress with this pull request? It would be incredibly useful to me as I need to call some MPI functions before launching horovod. Thank you!

alsrgv:
Superseded by #394.

PR #222, opened 2018-03-23
Fix void pointer arithmetic warning

alsrgv:
cc @xunzhang


horovod/common/operations.cc
```diff
@@ -820,8 +820,9 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
         // Copy memory into the fusion buffer.
         int64_t offset = 0;
         for (auto it = entries.begin(); it != entries.end(); it++) {
+          const void* buffer_data_at_offset = (uint8_t*)buffer_data + offset;
```
xunzhang: Here, why don’t we directly define as a void *, but I’m ok with it.

alsrgv: Sure, that's fine idea.

PR #231, opened 2018-04-03
Compile OpenMPI with CUDA

OmriShiv:
Compile OpenMPI with CUDA

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ RUN mkdir /tmp/openmpi && \
     wget https://www.open-mpi.org/software/ompi/v3.0/downloads/openmpi-3.0.0.tar.gz && \
     tar zxf openmpi-3.0.0.tar.gz && \
     cd openmpi-3.0.0 && \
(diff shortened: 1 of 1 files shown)


alsrgv:
@OmriShiv, thanks for the pull request.  What's the desire to use CUDA-aware Open MPI?  [NCCL developer guide](http://docs.nvidia.com/deeplearning/sdk/nccl-developer-guide/index.html#intergpu) mentions the possibility of deadlock with NCCL + CUDA-aware MPI combination that few Horovod customers reported as well, so it may not be a great idea to enable it in the general purpose container.

OmriShiv:
@alsrgv We had not run into that issue, thankfully, but also had not seen that in the developer guide; thanks for the reference! We'll keep an eye out to see if affects us.

alsrgv:
@OmriShiv, sounds good.  I think we should keep it as a private change for now, until NCCL fixes deadlocks if that's possible.

OmriShiv:
@alsrgv Agreed; feel free to close.

PR #275, opened 2018-05-21
 Support TensorFlow Optimizers that override compute_gradients/apply_gradients

### LINKED ISSUE #273: DistributedOptimizer cannot be used when underlying optimizer overwrites compute_gradients and apply_gradients
In our use case we have a TensorFlow optimizer that explicitly defines `compute_gradients` and `apply_gradients` functions. The current version of `DistributedOptimizer` cannot be used in our case because it does not overwrite `apply_gradients` to use` self._optimizer.apply_gradients` and it calls (lines 178-179 of `horovod/tensorflow/__init__.py`)
```python
gradients = super(DistributedOptimizer, self).compute_gradients(*args, **kwargs)
```
instead of `gradients = self._optimizer.compute_gradients(*args, **kwargs)`.

Is it possible to update `DistributedOptimizer` to work in our use case and similar situations when `compute_gradients` and `apply_gradients` are redefined? As far as I can see simply changing lines 178-179 of `horovod/tensorflow/__init__.py` to 
```python
gradients = self._optimizer.compute_gradients(*args, **kwargs)
```
and adding additional method:
```python
  def apply_gradients(self, grads_and_vars, global_step=None, name=None):
    """Calls this same method on the underlying optimizer."""
    return self._optimizer.apply_gradients(grads_and_vars, global_step, name)
```
should make it work.

alsrgv:
Previously, code relied on `tf.train.Optimizer` base class to provide `compute_gradients()` and `apply_gradients()` implementation.  This does not work if the user wishes to override those with a custom optimizer.  Additionally, should TensorFlow change implementation of `compute_gradients()` or `apply_gradients()` to use a different set of private methods than what was forwarded before, subtle bugs would happen.

Fixes #273.


alsrgv:
cc @Kipok

PR #276, opened 2018-05-21
Collective control plane & other low latency improvements

alsrgv:
On very large clusters (>10K GPUs) traditional point-to-point control plane does not work very well.  The main idea is to replace point-to-point communication with collective communications, which most MPI implementations will optimize to use tree-based implementations.

Additionally:
1. Making NCCL use high-priority CUDA stream.
2. Build FlatBuffers in a bottom-up fashion, as prescribed by the spec.
3. Adding ability to configure Horovod cycle time.

Kudos: these changes were inspired by Sean Treichler & Josh Romero from NVIDIA.
cc @streichler


docs/tensor-fusion.md
```diff
@@ -26,3 +26,9 @@ Setting the `HOROVOD_FUSION_THRESHOLD` environment variable to zero disables Ten
 ```bash
 $ HOROVOD_FUSION_THRESHOLD=0 mpirun -np 4 -x HOROVOD_FUSION_THRESHOLD python train.py
 ```
+
+Time between cycles, defined in milliseconds, can be tweaked using the `HOROVOD_CYCLE_TIME` environment variable:
```
sblotner: You can tweak time between cycles (defined in milliseconds) using the ...

horovod/common/wire/mpi_message.fbs
```diff
@@ -97,3 +92,9 @@ table MPIResponse {
     // indexed by the rank.
     tensor_sizes:[long];
 }
+table MPIResponseList {
+    responses:[MPIResponse];
+
+    // Flag indicating whether worker is requested to shutdown.
```
sblotner: whether --> if

horovod/common/mpi_message.cc
```diff
@@ -302,20 +297,89 @@ void MPIResponse::ParseFromString(MPIResponse& response,
...
+
+void MPIResponseList::add_responses(MPIResponse value) {
+  responses_.push_back(value);
+}
+
+void MPIResponseList::ParseFromString(MPIResponseList& response_list,
+                                     const std::string& input) {
+  auto obj =
+      flatbuffers::GetRoot<wire::MPIResponseList>((uint8_t*)input.c_str());
+  for (auto it = obj->responses()->begin(); it != obj->responses()->end(); it++) {
```
jiezhang: Why not use range based looping?

horovod/common/mpi_message.h
```diff
@@ -156,6 +152,24 @@ class MPIResponse {
   std::vector<int64_t> tensor_sizes_;
 };
 
+class MPIResponseList {
+public:
+  const std::vector<MPIResponse>& responses() const;
+  void set_responses(const std::vector<MPIResponse>& value);
+  void add_responses(MPIResponse value);
```
jiezhang: Pass MPIResponse by reference.

PR #305, opened 2018-06-11
Use Lecun's MNIST dataset

alsrgv:
TensorFlow MNIST dataset got broken, switching to Lecun's official MNIST.

See https://github.com/tensorflow/tensorflow/issues/19914

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,6 +105,12 @@ script:
   # hack TensorFlow MNIST example to be smaller
   - docker exec ${CONTAINER} /bin/sh -c "sed -i \"s/last_step=20000/last_step=100/\" /horovod/examples/tensorflow_mnist.py"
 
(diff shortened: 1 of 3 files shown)


girving:
Note that the original URL is insecure http-only.  The Google mirror will likely come back soon, and is safer to use.

alsrgv:
Fair point, this also makes example uglier and does not work with TensorFlow 1.1.0 (yes, we're backward-compatible:-) ).  Changing to only use Lecun's URL in the integration test, since we depend on it for [PyTorch MNIST example](https://github.com/uber/horovod/blob/master/examples/pytorch_mnist.py) anyway.

alsrgv:
The upstream issue has been resolved.  Will close this for now, and re-apply if it happens again.

PR #309, opened 2018-06-13
Add PyTorch ImageNet example

alsrgv:
Converges to 76%+ Top-1 accuracy on 128 GPUs.

### DIFF
--- a/examples/pytorch_imagenet_resnet50.py
+++ b/examples/pytorch_imagenet_resnet50.py
@@ -128,7 +128,6 @@ optimizer = optim.SGD(model.parameters(), lr=args.base_lr * hvd.size(),
 optimizer = hvd.DistributedOptimizer(
     optimizer, named_parameters=model.named_parameters())
 
-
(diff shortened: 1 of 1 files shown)


bapriddy:
Just wanted to let you know it runs without errors although I had to comment out the TensorboardX stuff.  Nice work!!  I was in the process of converting Soumith's version to use with Horovod but you guys beat me.  Thanks!  I'm hoping to out-perform G**gle's tf.

alsrgv:
@bapriddy, thanks!  `tensorboardX` is very useful, you can install it via `pip install tensorboardX`.

bapriddy:
in the line
'''kwargs = {'num_workers': 4, 'pin_memory': True} if args.cuda else {}'''
if i'm running on say 8 or 16 nodes, each node is a mother + cpu + gpu, should 'num workers'.  according to the notes thses are dataloaders but not sure about the details.

Second question...In the out put is it/s iterations per second...?

Issue #312, opened 2018-06-13
Tweaking Horovod: Using nn.DataParallel with Horovod


SrivastavaKshitij:
Hi @alsrgv: 

As you know Summit Supercomputer architecture has 6 GPUs per node. My understanding is if I want to run Horovod with `n1024 a6 g6' command as described in [308](https://github.com/uber/horovod/issues/308) , this would lead to creation of 6144 MPI ranks in Horovod that will lead to a lot of communication overhead. 

I looked at nn.DataParallel source code for Pytorch at [data parallel](https://github.com/pytorch/pytorch/blob/master/torch/nn/parallel/data_parallel.py) and found out that in order for pytorch to spawn six models on each GPU it doesnt use any mpi communicator , so it would not conflict with horovod communicator. Although it uses NCCL so I am not sure if thats going to create a problem with Horovod. 

So , I tried spawning just one model per node using Horovod  and used nn.DataParallel so that within each node, pytorch will spawn 6 GPUs on 6 nodes. That means, instead of `n1024 a6 g6`, i tried `n1024 a1 g6` which will lead to creation of 1024 MPI ranks instead of 6144. 

My first epoch goes well and is actually faster than the naive Horovod implementation but during my second epoch, pytorch data loader throws a segmentation fault. 
`RuntimeError: DataLoader worker (pid 33818) is killed by signal: Segmentation fault`

alsrgv:
Hi @SrivastavaKshitij,

Thanks for the detailed analysis.  You mentioned you're not using NCCL2 with Horovod, is it still the case?  Do you still install Horovod without `HOROVOD_GPU_ALLREDUCE=MPI`?

I believe what ends up happening is that in Horovod + DataParallel mode you use MPI for cross-node allreduce and NCCL1 for intra-node allreduce. This is actually semi-optimal setting at this amount of nodes (some folks found that having a separate ring per 3 GPUs NVLink mesh adds a bit more performance), because the single ring does not scale to this number of GPUs nicely and MPI uses hierarchical `O(log N)` allreduce algorithm.

As of Horovod 0.13.3, we use [collective control plane](https://github.com/uber/horovod/pull/276) for coordination messages, which uses a similar `O(log N)` algorithms.  This is a first step to support huge supercomputers.

The second phase is to support MPI + NCCL allreduce natively.  Would you be interested in trying a private branch with such support?

Could you share results of your performance tests?  Would be really interesting to see:

1. Horovod 0.13.2 native
[...]

PR #323, opened 2018-06-19
Add support for IBM PowerAI DDL gpu allreduce

nvcastet:
* Include making FinalizerThread more generic using
  a queue of events to wait for.

### DIFF
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -31,6 +31,10 @@
 #include <nccl.h>
 #endif
 
+#if HAVE_DDL
+#include <ddl.hpp>
+#endif
+
 #define OMPI_SKIP_MPICXX
 #include "hashes.h"
 #include "mpi.h"
@@ -204,9 +208,13 @@ struct HorovodGlobalState {
   std::unordered_map<std::vector<int32_t>, ncclComm_t> nccl_comms;
 #endif
 
+  // Will be set to true after initialization when ddl is used
+  bool ddl_initialized = false;
+  int32_t ddl_local_device_id = 0;
+
 // We reuse CUDA events as it appears that their creation carries non-zero cost.
 // Event management code is only used in NCCL path.
-#if HAVE_NCCL
+#if HAVE_NCCL || HAVE_DDL
   std::unordered_map<int, std::queue<cudaEvent_t>> cuda_events;
   std::mutex cuda_events_mutex;
 #endif
@@ -240,6 +248,15 @@ const Status SHUT_DOWN_ERROR = Status::Aborted(
     "one of the ranks finished execution. If the shutdown was caused by an "
     "exception, you should see the exception in the log before the first "
     "shutdown message.");
+
+#define OP_ERROR(entries, error_message)                                       \
+  {                                                                            \
+      for (auto& e : (entries)) {                                              \
+        timeline.End(e.tensor_name, nullptr);                                  \
+        e.callback(Status::UnknownError(error_message));                       \
+      }                                                                        \
+      return;                                                                  \
+  }
 
 // Store the MPIRequest for a name, and return whether the total count of
 // MPIRequests for that tensor is now equal to the MPI size (and thus we are
@@ -525,6 +542,18 @@ ncclDataType_t GetNCCLDataType(const std::shared_ptr<Tensor> tensor) {
 }
 #endif
 
+#if HAVE_DDL
+DDL_Type GetDDLDataType(const std::shared_ptr<Tensor> tensor) {
+  switch (tensor->dtype()) {
+  case HOROVOD_FLOAT32:
+    return DDL_TYPE_FLOAT;
+  default:
+    throw std::logic_error("Type " + MPIDataType_Name(tensor->dtype()) +
+                           " is not supported in DDL mode.");
+  }
+}
+#endif
+
 #define MPI_CHECK(entries, op_name, op)                                        \
   {                                                                            \
     auto mpi_result = (op);                                                    \
@@ -564,8 +593,20 @@ ncclDataType_t GetNCCLDataType(const std::shared_ptr<Tensor> tensor) {
     }                                                                          \
   }
 
+#define DDL_CHECK(entries, op_name, op)                                        \
+  {                                                                            \
+    auto ddl_result = (op);                                                    \
+    if (ddl_result != DDL_SUCCESS) {                                           \
+      for (auto& e : (entries)) {                                              \
+        timeline.End(e.tensor_name, nullptr);                                  \
+        e.callback(Status::UnknownError(std::string(op_name) + " failed."));   \
+      }                                                                        \
+      return;                                                                  \
+    }                                                                          \
+  }
+
 // This event management code is only used in NCCL.
-#ifdef HAVE_NCCL
+#if HAVE_NCCL || HAVE_DDL
 cudaError_t GetCudaEvent(cudaEvent_t* event) {
   int device;
   auto status = cudaGetDevice(&device);
@@ -812,7 +853,8 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
     }
 #endif
 
-#if HOROVOD_GPU_ALLREDUCE == 'N' // 'N' stands for NCCL
+// 'N' stands for NCCL and 'D' for DDL
+#if HOROVOD_GPU_ALLREDUCE == 'N' || HOROVOD_GPU_ALLREDUCE == 'D'
     if (on_gpu) {
       auto stream = horovod_global.streams[first_entry.device];
       auto event_queue = std::queue<std::pair<std::string, cudaEvent_t>>();
@@ -827,6 +869,7 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
         nccl_device_map = response.devices();
       }
 
+#if HOROVOD_GPU_ALLREDUCE == 'N'
       // Ensure NCCL communicator is in the map before executing reduction.
       ncclComm_t& nccl_comm = horovod_global.nccl_comms[nccl_device_map];
       if (nccl_comm == nullptr) {
@@ -865,6 +908,20 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
 
         ACTIVITY_END_ALL(entries, timeline)
       }
+#elif HOROVOD_GPU_ALLREDUCE == 'D'
+      if (!horovod_global.ddl_initialized) {
+        // Initialize DDL
+        auto ddl_options = std::getenv("DDL_OPTIONS");
+        if (ddl_options == nullptr) {
+          OP_ERROR(entries, "DDL_OPTIONS env variable needs to be set to use DDL.")
+        }
+        DDL_CHECK(entries, "ddl_init", ddl_init(ddl_options))
+        horovod_global.ddl_initialized = true;
+        horovod_global.ddl_local_device_id = first_entry.device;
+      } else if (horovod_global.ddl_local_device_id != first_entry.device) {
+        OP_ERROR(entries, "DDL does not support more than one GPU device per process.")
+      }
+#endif
 
       if (timeline.Initialized()) {
         RECORD_EVENT(entries, event_queue, QUEUE, stream)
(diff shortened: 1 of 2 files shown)


horovod/common/operations.cc
```diff
@@ -843,22 +936,37 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
                                      cudaMemcpyDeviceToDevice, stream))
           offset += it->tensor->size();
         }
+
+#if HOROVOD_GPU_ALLREDUCE != 'D'
         if (timeline.Initialized()) {
```
alsrgv: Why do you side-step check for `timeline.Initialized()`?

horovod/common/operations.cc
```diff
@@ -873,81 +981,44 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
...
-        CUDA_CHECK(entries, "cudaSetDevice", cudaSetDevice(first_entry.device))
-        if (queue_end_event != nullptr) {
-          CUDA_CHECK(entries, "cudaEventSynchronize",
-                     cudaEventSynchronize(queue_end_event))
-          // All the work scheduled on NCCL stream before this allreduce
-          // is done at this point, end queueing activity.
-          ACTIVITY_END_ALL(entries, timeline)
-          RELEASE_EVENT(entries, queue_end_event);
-        }
+      RECORD_EVENT("ALLREDUCE_DONE", entries, done_event, stream, true)
```
alsrgv: Can we do this without changing the set of events we're tracking?

horovod/common/operations.cc
```diff
@@ -1480,7 +1551,11 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
 
   MPI_Comm_free(&horovod_global.mpi_comm);
 
+#if HAVE_DDL
+  ddl_finalize();
+#else
   MPI_Finalize();
```
alsrgv: Does `ddl_finalize()` do `MPI_Finalize()` as well?

setup.py
```diff
@@ -313,15 +313,29 @@ def get_nccl_dirs(build_ext, cuda_include_dirs, cuda_lib_dirs, cpp_flags):
 
     return nccl_include_dirs, nccl_lib_dirs
 
+def get_ddl_dirs():
+    # Default DDL home
+    ddl_home = '/opt/DL/ddl'
```
alsrgv: We should make this location configurable.

horovod/common/operations.cc
```diff
@@ -515,6 +528,18 @@ ncclDataType_t GetNCCLDataType(const std::shared_ptr<Tensor> tensor) {
 }
 #endif
 
+#if HAVE_DDL
+DDL_Type GetDDLDataType(const std::shared_ptr<Tensor> tensor) {
+  switch (tensor->dtype()) {
+  case HOROVOD_FLOAT32:
+    return DDL_TYPE_FLOAT;
+  default:
+    throw std::logic_error("Type " + MPIDataType_Name(tensor->dtype()) +
```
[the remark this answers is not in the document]
nvcastet: DDL supports half-precision (16 bits) too. I didn't find a HOROVOD_FLOAT16 in this file. Is there a way to integrate half-precision with Horovod?

horovod/common/operations.cc
nvcastet: The DDL allreduce is a synchronous operation and uses its own Cuda streams. So we need to record an event on Horovod stream and wait for it before calling DDL to make all the work is done up to this point on the Horovod stream.
That is also why i need to start the finalizer_thread before all the events get added to the Horovod stream otherwise your timeline would be messed up because ddl is syncing the operation in the middle of all the events. Does that make sense?

horovod/common/operations.cc
nvcastet: This event is the done_event that was there before. You mean without adding an extra activity to the timeline? The bool set to true at the end of the call means it is the final event. We could not record the event as an activity in the finalizer thread. Would you prefer that?

horovod/common/operations.cc
nvcastet: Yes and also cleanup some MPI communicators so we cannot call MPI_Finalize before.

setup.py
nvcastet: Yes we could. But It will most likely never change because our rpm and deb packages assume it gets setup to this location.

alsrgv: Is it always rpm and deb, or you also have tarball with the libraries?

horovod/common/operations.cc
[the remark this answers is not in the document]
alsrgv: There is support coming in `fp16` branch that is being evaluated right now.  If DDL support is there, it will be easy to extend to DDL then!

horovod/common/operations.cc
alsrgv: I see - makes sense.  I actually do some of the things along the similar lines in #329.  I think it makes sense to merge in refactorings from #329 first, and then your code change will actually be much simpler.

nvcastet: ok will wait on #329.

setup.py
nvcastet: Yes always rpm or dep.

alsrgv: OK, it's fine to keep like that then.

horovod/common/operations.cc
nvcastet: @alsrgv How is your PR #329 going?

alsrgv: Waiting to hear back from a user in #312 to determine whether it works and whether it's faster.

alsrgv: @nvcastet, the #329 has been pushed.  Please rebase/simplify this PR.

horovod/common/operations.cc
```diff
@@ -863,6 +896,18 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
 
         ACTIVITY_END_ALL(entries, timeline)
       }
+#else
```
alsrgv: Let's make it `elif GPU_ALLREDUCE == 'D'` for future extensibility.

horovod/common/operations.cc
```diff
@@ -562,8 +581,20 @@ ncclDataType_t GetNCCLDataType(const std::shared_ptr<Tensor> tensor) {
     }                                                                          \
   }
 
+#define DDL_CHECK(entries, op_name, op)                                        \
+  {                                                                            \
+    auto ddl_result = (op);                                                    \
+    if (ddl_result != DDL_SUCCESS) {                                           \
+      for (auto it = entries.begin(); it != entries.end(); it++) {             \
```
alsrgv: Can you change it to iterator, like the rest of new `*_CHECK()`

horovod/common/operations.cc
```diff
@@ -1342,7 +1395,11 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
   MPI_Comm_free(&state.local_comm);
   MPI_Comm_free(&state.cross_comm);
 
-  MPI_Finalize();
+#if HAVE_DDL
+  ddl_finalize();
+#else
+   MPI_Finalize();
```
alsrgv: Can you add a comment that `ddl_finalize()` performs MPI finalization as well?

Nit: remove one extra space before `MPI`

nvcastet:
@alsrgv Could you look at my last comment in #329? I will wait for the fix to merge if one is needed.

nvcastet:
@alsrgv We have not yet run Horovod + DDL on a lot of nodes.
For up to 4 nodes here is the images/sec for HPM resnet101:
     GPUs: 1               2              4               8              16         
+NCCL2: 213.84	427.53	843.80	1665.70	3210.00
+DDL:      214.37	425.34	847.75	1680.50	3257.20
We think the DDL version will start getting better as we scale the number of nodes.

alsrgv:
@nvcastet, the fix has been landed.  Thanks for letting us know about the issue!

nvcastet:
@alsrgv No problem. I updated this PR accordingly and added a copy from input to output buffer for this case because DDL only has in-place allreduce now.

horovod/common/operations.cc
```diff
@@ -865,6 +898,18 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
...
+#elif HOROVOD_GPU_ALLREDUCE == 'D'
+      if (!horovod_global.ddl_initialized) {
+        // Initialize DDL
+        auto ddl_options = std::getenv("DDL_OPTIONS");
+        if (ddl_options == nullptr) {
+          throw std::logic_error("DDL_OPTIONS env variable needs "
+                                 "to be set to use DDL.");
+        }
+        DDL_CHECK(entries, "ddl_init", ddl_init(ddl_options))
+        horovod_global.ddl_initialized = true;
```
alsrgv: One more question.  Does this pass `test_horovod_allreduce_multi_gpu()`?  In the case of NCCL, we dynamically create a communicator based on the set of GPUs that will participate in the operation.  How does DDL handle that?

nvcastet: It probably won't pass this test. Right now, DDL handles only one device per process.

nvcastet: @alsrgv I am adding a check for the "test_horovod_allreduce_multi_gpu" to fail gracefully. Just curious, what is the use case for Horovod to support several devices per process?

nvcastet: I pushed the check.

horovod/common/operations.cc
```diff
@@ -865,6 +899,21 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
...
+        horovod_global.ddl_local_device_id = first_entry.device;
+      } else if (horovod_global.ddl_local_device_id != first_entry.device) {
+        throw std::logic_error("DDL does not support more than one GPU device per process.");
```
alsrgv: Instead of exception (on a background thread), can this do something like what DDL_CHECK does:
```
for (auto& e : (entries)) {
        timeline.End(e.tensor_name, nullptr);                                  \
        e.callback(Status::UnknownError(std::string(op_name) + " failed."));   \
}
return
```

horovod/common/operations.cc
```diff
@@ -865,6 +899,21 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
 
         ACTIVITY_END_ALL(entries, timeline)
       }
+#elif HOROVOD_GPU_ALLREDUCE == 'D'
+      if (!horovod_global.ddl_initialized) {
+        // Initialize DDL
+        auto ddl_options = std::getenv("DDL_OPTIONS");
+        if (ddl_options == nullptr) {
+          throw std::logic_error("DDL_OPTIONS env variable needs "
```
alsrgv: Similar to new the other comment, can this use mechanism similar to `DDL_CHECK`?

horovod/common/operations.cc
```diff
@@ -865,6 +899,21 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
 
         ACTIVITY_END_ALL(entries, timeline)
       }
+#elif HOROVOD_GPU_ALLREDUCE == 'D'
+      if (!horovod_global.ddl_initialized) {
+        // Initialize DDL
+        auto ddl_options = std::getenv("DDL_OPTIONS");
```
alsrgv: Can you give an example of DDL_OPTIONS?  Are they system-wide?

horovod/common/operations.cc
```diff
@@ -525,6 +533,18 @@ ncclDataType_t GetNCCLDataType(const std::shared_ptr<Tensor> tensor) {
 }
 #endif
 
+#if HAVE_DDL
+DDL_Type GetDDLDataType(const std::shared_ptr<Tensor> tensor) {
+  switch (tensor->dtype()) {
+  case HOROVOD_FLOAT32:
+    return DDL_TYPE_FLOAT;
+  default:
+    throw std::logic_error("Type " + MPIDataType_Name(tensor->dtype()) +
```
alsrgv: `std::logic_error()` here is not desirable.

`GetMPIDataType()` and `GetNCCLDataType()` throw `std::logic_error()` because it indicates Horovod developer bug - they both cover all possible types allowed for GPU allreduce & other ops by TF and PyTorch, and if a user tries to use unsupported type he'll get a nice error before that in the stack.

Since here DDL is more restrictive, we should handle it nicer and similarly give a user error via a mechanism similar to `DDL_CHECK()`.

nvcastet:
@alsrgv I don't know if you see my comment of yesterday: "Just curious, what is the use case for Horovod to support several devices per process?"

horovod/common/operations.cc
nvcastet: I can do that

horovod/common/operations.cc
nvcastet: They are setup by ddlrun (or you can set it up yourself too). I would refer to the DDL documentation to get the specific about DDL_OPTIONS.

alsrgv:
@nvcastet, model parallelism. The user can manually pin tensors on multiple GPUs and Horovod will make sure it will create an NCCL ring across the right set of GPUs for each op.

PR #329, opened 2018-06-22
Hierarchical NCCL + MPI allreduce

alsrgv:
A simple version of hierarchical allreduce (NCCL reduce, MPI allreduce between nodes, NCCL broadcast).

Some simplifications + code movements and refactoring.

### DIFF
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -840,15 +840,17 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
       if (nccl_comm == nullptr) {
         ACTIVITY_START_ALL(entries, timeline, "INIT_NCCL")
 
-        int nccl_rank = horovod_global.hierarchical_allreduce
-                            ? horovod_global.local_rank
-                            : horovod_global.rank;
-        int nccl_size = horovod_global.hierarchical_allreduce
-                            ? horovod_global.local_size
-                            : horovod_global.size;
-        MPI_Comm nccl_id_bcast_comm = horovod_global.hierarchical_allreduce
-                                          ? horovod_global.local_comm
-                                          : horovod_global.mpi_comm;
+        int nccl_rank, nccl_size;
+        MPI_Comm nccl_id_bcast_comm;
+        if (horovod_global.hierarchical_allreduce) {
+          nccl_rank = horovod_global.local_rank;
+          nccl_size = horovod_global.local_size;
+          nccl_id_bcast_comm = horovod_global.local_comm;
+        } else {
+          nccl_rank = horovod_global.rank;
+          nccl_size = horovod_global.size;
+          nccl_id_bcast_comm = horovod_global.mpi_comm;
+        }
 
         ncclUniqueId nccl_id;
         if (nccl_rank == 0) {
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -223,21 +223,18 @@ struct HorovodGlobalState {
 };
 
 // All the Horovod state that must be stored globally per-process.
-static HorovodGlobalState horovod_global;
+HorovodGlobalState horovod_global;
 
 // For clarify in argument lists.
 #define RANK_ZERO 0
 
-// A tag used for all coordinator messaging.
-#define TAG_NOTIFY 1
-
 // Stall-check warning time
 #define STALL_WARNING_TIME std::chrono::seconds(60)
 
-static const Status NOT_INITIALIZED_ERROR = Status::PreconditionError(
+const Status NOT_INITIALIZED_ERROR = Status::PreconditionError(
     "Horovod has not been initialized; use hvd.init().");
 
-static const Status SHUT_DOWN_ERROR = Status::Aborted(
+const Status SHUT_DOWN_ERROR = Status::Aborted(
     "Horovod has been shut down. This has been caused by an exception on one "
     "of the rank or an attempt to allreduce, allgather or broadcast a tensor "
     "after one of the ranks has finished execution.");
@@ -247,7 +244,7 @@ static const Status SHUT_DOWN_ERROR = Status::Aborted(
 // ready to reduce the tensor).
 bool IncrementTensorCount(std::unique_ptr<MessageTable>& message_table,
                           MPIRequest msg, int mpi_size) {
-  auto name = msg.tensor_name();
+  auto& name = msg.tensor_name();
   auto& timeline = horovod_global.timeline;
   auto table_iter = message_table->find(name);
   if (table_iter == message_table->end()) {
@@ -327,9 +324,8 @@ MPIResponse ConstructMPIResponse(std::unique_ptr<MessageTable>& message_table,
   if (message_type == MPIRequest::ALLREDUCE ||
       message_type == MPIRequest::BROADCAST) {
     TensorShape tensor_shape;
-    for (auto it = requests[0].tensor_shape().begin();
-         it != requests[0].tensor_shape().end(); it++) {
-      tensor_shape.AddDim(*it);
+    for (auto dim : requests[0].tensor_shape()) {
+      tensor_shape.AddDim(dim);
     }
     for (unsigned int i = 1; i < requests.size(); i++) {
       if (error) {
@@ -337,9 +333,8 @@ MPIResponse ConstructMPIResponse(std::unique_ptr<MessageTable>& message_table,
       }
 
       TensorShape request_shape;
-      for (auto it = requests[i].tensor_shape().begin();
-           it != requests[i].tensor_shape().end(); it++) {
-        request_shape.AddDim(*it);
+      for (auto dim : requests[i].tensor_shape()) {
+        request_shape.AddDim(dim);
       }
       if (tensor_shape != request_shape) {
         error = true;
@@ -360,9 +355,8 @@ MPIResponse ConstructMPIResponse(std::unique_ptr<MessageTable>& message_table,
   std::vector<int64_t> tensor_sizes(requests.size());
   if (message_type == MPIRequest::ALLGATHER) {
     TensorShape tensor_shape;
-    for (auto it = requests[0].tensor_shape().begin();
-         it != requests[0].tensor_shape().end(); it++) {
-      tensor_shape.AddDim(*it);
+    for (auto dim : requests[0].tensor_shape()) {
+      tensor_shape.AddDim(dim);
     }
 
     if (tensor_shape.dims() == 0) {
@@ -380,9 +374,8 @@ MPIResponse ConstructMPIResponse(std::unique_ptr<MessageTable>& message_table,
       }
 
       TensorShape request_shape;
-      for (auto it = requests[i].tensor_shape().begin();
-           it != requests[i].tensor_shape().end(); it++) {
-        request_shape.AddDim(*it);
+      for (auto dim : requests[i].tensor_shape()) {
+        request_shape.AddDim(dim);
       }
       if (tensor_shape.dims() != request_shape.dims()) {
         error = true;
@@ -457,8 +450,8 @@ MPIResponse ConstructMPIResponse(std::unique_ptr<MessageTable>& message_table,
     }
   }
   std::vector<int32_t> devices(requests.size());
-  for (auto it = requests.begin(); it != requests.end(); it++) {
-    devices[it->request_rank()] = it->device();
+  for (auto& request : requests) {
+    devices[request.request_rank()] = request.device();
   }
 
   MPIResponse response;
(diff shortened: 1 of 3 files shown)


rahul003:
Hi @alsrgv, just to clarify, by hierarchical you mean, the separation of intra node and inter node communication? Is the motivation here that we expect MPI to be better than NCCL for multi node communication?

alsrgv:
@rahul003, that's right - the data exchange is hierarchical.  With a large number of nodes, MPI will use hierarchical collectives which should outperform NCCL at large scale.  In a sense, NCCL becomes leaf node in data exchange hierarchy.

horovod/common/operations.cc
```diff
@@ -789,25 +822,46 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
 #if HOROVOD_GPU_ALLREDUCE == 'N' // 'N' stands for NCCL
     if (on_gpu) {
       auto stream = horovod_global.streams[first_entry.device];
+      auto event_queue = std::queue<std::pair<std::string, cudaEvent_t>>();
+
+      // Determine GPU IDs of the devices participating in this communicator.
+      std::vector<int32_t> nccl_device_map;
+      if (horovod_global.hierarchical_allreduce) {
+        for (int rank : horovod_global.local_comm_ranks) {
+          nccl_device_map.push_back(response.devices()[rank]);
```
jiezhang: rank guaranteed to be less than size() here?

horovod/common/operations.cc
```diff
@@ -789,25 +822,46 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
...
+        nccl_device_map = response.devices();
+      }
 
       // Ensure NCCL communicator is in the map before executing reduction.
-      ncclComm_t& nccl_comm = horovod_global.nccl_comms[response.devices()];
+      ncclComm_t& nccl_comm = horovod_global.nccl_comms[nccl_device_map];
       if (nccl_comm == nullptr) {
         ACTIVITY_START_ALL(entries, timeline, "INIT_NCCL")
 
+        int nccl_rank = horovod_global.hierarchical_allreduce
```
jiezhang: One if block might be more readable than 3 ternary operator here.

horovod/common/operations.cc
```diff
@@ -1487,11 +1337,251 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
...
+
+        bool reduce = IncrementTensorCount(state.message_table,
+                                           received_message, state.size);
+        if (reduce) {
+          ready_to_reduce.push_back(received_name);
+        }
+      }
+      if (received_message_list.shutdown()) {
+        // Received SHUTDOWN request from one of the workers.
+        state.shut_down = true;
```
jiezhang: break?

horovod/common/operations.cc
alsrgv: Yes, by definition.

horovod/common/operations.cc
alsrgv: We still process last tensors received in this batch.

horovod/common/operations.cc
```diff
@@ -1242,220 +1288,26 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
     state.cycle_time = std::atof(horovod_cycle_time);
   }
 
+  // Set flag for hierarchical allreduce. Ignore if Horovod is running on a
+  // single node.
+  auto horovod_hierarchical_allreduce =
+      std::getenv("HOROVOD_HIERARCHICAL_ALLREDUCE");
+  if (horovod_hierarchical_allreduce != nullptr &&
+      std::atoi(horovod_hierarchical_allreduce) > 0 && cross_size > 1) {
```
tgaddair: Wondering if [strtol](https://en.cppreference.com/w/c/string/byte/strtol) might be safer here.

horovod/common/operations.cc
```diff
@@ -1487,11 +1339,251 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
...
+//      If instead of "DONE" they receive "SHUTDOWN", they exit their background
+//      loop.
+bool RunLoopOnce(HorovodGlobalState& state, bool is_coordinator) {
+  // The coordinator sends a SHUTDOWN message to trigger shutdown.
+  bool should_shut_down = false;
+
+  // This delay determines thread frequency and MPI message latency
+  auto sleep_duration =
+      state.last_cycle_start +
+      std::chrono::microseconds(long(state.cycle_time * 1000.)) -
```
tgaddair: Why is state.last_cyle_start in microseconds, but state.cycle_time in seconds?

horovod/common/operations.cc
```diff
@@ -1487,11 +1339,251 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
...
+        ready_to_reduce.push_back(message.tensor_name());
+      }
+    }
+
+    // Rank zero has put all its own tensors in the tensor count table.
+    // Now, it should count all the tensors that are coming from other
+    // ranks at this tick.
+
+    // 1. Get message lengths from every rank.
+    auto recvcounts = new int[state.size];
```
tgaddair: There seem to be a lot of mallocs in this function.  How often is this routine executed?  (I see it sleeps some duration between ticks).  Might it be worth considering keeping some of these lists as globals that dynamically grow (not sure which resource is more constrained here, memory or compute)?

horovod/common/operations.cc
```diff
@@ -1487,11 +1339,251 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
...
+    delete[] buffer;
+
+    // At this point, rank zero should have a fully updated tensor count
+    // table and should know all the tensors that need to be reduced or
+    // gathered, and everyone else should have sent all their information
+    // to rank zero. We can now do reductions and gathers; rank zero will
+    // choose which ones and in what order, and will notify the other ranks
+    // before doing each reduction.
+    std::deque<MPIResponse> responses;
+    for (auto it = ready_to_reduce.begin(); it != ready_to_reduce.end(); it++) {
```
tgaddair: Preincrement with `++it` will save you a copy of the iterator.

horovod/common/operations.cc
```diff
@@ -595,25 +609,44 @@ cudaError_t ReleaseCudaEvent(cudaEvent_t event) {
...
+      CUDA_CHECK(entries, "ReleaseCudaEvent", ReleaseCudaEvent(event))         \
+    }                                                                          \
+  }
 #endif
 
 #define ACTIVITY_START_ALL(entries, timeline, activity)                        \
   {                                                                            \
-    for (auto it = entries.begin(); it != entries.end(); it++) {               \
-      timeline.ActivityStart(it->tensor_name, activity);                       \
+    for (auto it = (entries).begin(); it != (entries).end(); it++) {           \
```
tgaddair: Preincrement with `++it` will save you a copy of the iterator.

horovod/common/operations.cc
```diff
@@ -595,25 +609,44 @@ cudaError_t ReleaseCudaEvent(cudaEvent_t event) {
...
+      (timeline).ActivityStart(it->tensor_name, activity);                     \
     }                                                                          \
   }
 
 #define ACTIVITY_END_ALL(entries, timeline)                                    \
   {                                                                            \
-    for (auto it = entries.begin(); it != entries.end(); it++) {               \
-      timeline.ActivityEnd(it->tensor_name);                                   \
+    for (auto it = (entries).begin(); it != (entries).end(); it++) {           \
```
tgaddair: `++it`

horovod/common/operations.cc
```diff
@@ -817,21 +873,19 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
         ACTIVITY_END_ALL(entries, timeline)
       }
 
-      ACTIVITY_START_ALL(entries, timeline, "SCHEDULE")
-
-      cudaEvent_t queue_end_event = nullptr;
       if (timeline.Initialized()) {
-        RECORD_EVENT(entries, queue_end_event, stream);
+        RECORD_EVENT(entries, event_queue, "QUEUE", stream)
```
tgaddair: Can we make "QUEUE" a constant at the top?

horovod/common/operations.cc
```diff
@@ -843,26 +897,80 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
                                      cudaMemcpyDeviceToDevice, stream))
           offset += it->tensor->size();
         }
+        buffer_len = (size_t)offset;
         if (timeline.Initialized()) {
-          RECORD_EVENT(entries, after_memcpy_in_event, stream)
+          RECORD_EVENT(entries, event_queue, "MEMCPY_IN_FUSION_BUFFER", stream)
```
tgaddair: MEMCPY_IN_FUSION_BUFFER constant?

horovod/common/operations.cc
```diff
@@ -843,26 +897,80 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
...
+      }
+
+      void* host_buffer = nullptr;
+      if (horovod_global.hierarchical_allreduce) {
+        NCCL_CHECK(entries, "ncclReduce",
+                   ncclReduce(buffer_data, buffer_data, (size_t)num_elements,
+                              GetNCCLDataType(first_entry.tensor), ncclSum, 0,
+                              nccl_comm, stream))
+        if (timeline.Initialized()) {
+          RECORD_EVENT(entries, event_queue, "NCCL_REDUCE", stream)
```
tgaddair: Generally, for all string constant, I think it would be more maintainable and safer to have them be defined with global variables.

horovod/common/operations.cc
```diff
@@ -843,26 +897,80 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
...
-        }
+      }
+      if (timeline.Initialized()) {
+        RECORD_EVENT(entries, event_queue, "NCCL_ALLREDUCE", stream)
+      }
 
+      if (entries.size() > 1) {
         // Copy memory out of the fusion buffer.
-        offset = 0;
+        int64_t offset = 0;
```
tgaddair: `++it`

horovod/common/operations.cc
```diff
@@ -843,26 +897,80 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
...
                                  nccl_comm, stream))
-        if (timeline.Initialized()) {
-          RECORD_EVENT(entries, after_reduce_event, stream)
-        }
+      }
+      if (timeline.Initialized()) {
+        RECORD_EVENT(entries, event_queue, "NCCL_ALLREDUCE", stream)
+      }
 
+      if (entries.size() > 1) {
```
tgaddair: What happens if `entries.size() == 1`?  Can you add a comment?

horovod/common/operations.cc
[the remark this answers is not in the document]
alsrgv: For the safety of macro-expansions.  Suggested by IDE/compiler.

horovod/common/operations.cc
alsrgv: `state.cycle_time` is actually in milliseconds.  Sleep duration is not a number but a time point, and so it doesn't have fixed resolution.

tgaddair: Ah, gotcha.  Minor thing for variable names, in general I think it's good to put units when they're otherwise implied.  So `cycle_time` -> `cycle_time_ms`.

alsrgv: Good idea.

horovod/common/operations.cc
alsrgv: These tiny mallocs are most likely not an issue, but [this malloc](https://github.com/uber/horovod/pull/329/files#diff-98391bc0c5b3bf87519e745b4ec3b9fdR923) could be quite expensive.  Determining tradeoff between such vanilla malloc, and cudaHostAlloc with reusable buffer is in the scope of future work.

alsrgv:
Shower thoughts are usually most insightful.  I think depending on how #323 integration goes, we can indeed refactor further.

horovod/common/operations.cc
```diff
@@ -776,9 +766,9 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
 
     // Tensors may have different first dimension, so we need to use
     // MPI_Allgatherv API that supports gathering arrays of different length.
-    ACTIVITY_START_ALL(entries, timeline, "MPI_ALLGATHER")
-    int* recvcounts = new int[tensor_sizes.size()];
-    int* displcmnts = new int[tensor_sizes.size()];
+    ACTIVITY_START_ALL(entries, timeline, MPI_ALLGATHER)
+    auto* recvcounts = new int[tensor_sizes.size()];
```
tgaddair: Is `auto*` different from `auto` here?  I've never seen that.

alsrgv: Should be the same, except it highlights ["pointerness"](https://stackoverflow.com/questions/34560365/c11-auto-declaration-with-and-without-pointer-declarator).

horovod/common/operations.cc
```diff
@@ -818,135 +864,143 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
...
+        for (auto& e : entries) {
+          num_elements += e.tensor->shape().num_elements();
         }
-        NCCL_CHECK(entries, "ncclAllReduce",
-                   ncclAllReduce(buffer_data, (void*)buffer_data,
-                                 (size_t)num_elements,
-                                 GetNCCLDataType(first_entry.tensor), ncclSum,
-                                 nccl_comm, stream))
+      } else {
+        buffer_data = (void*)first_entry.output->data();
```
nvcastet: @alsrgv It looks like a bug.
If (entries.size() == 1), you set output buffer "buffer_data" to "first_entry.output->data()" and then you call "ncclReduce(buffer_data, buffer_data, ...)". You need to copy input data "first_entry.tensor->data()" to "buffer_data" before the ncclReduce. Or, you need a different ncclReduce call taking the input buffer for this special case.

alsrgv: @nvcastet, great catch, thanks a lot!  Pushing through a PR #358 with the fix.

PR #331, opened 2018-06-22
Made Horovod ops differentiable in TensorFlow API

tgaddair:


horovod/tensorflow/mpi_ops.py
```diff
@@ -119,4 +149,15 @@ def broadcast(tensor, root_rank, name=None):
...
+    """Gradient for broadcast op.
+
+    Args:
+      op: An operation.
+      grad: `Tensor` gradient with respect to the output of the op.
+
+    Returns:
+      The gradient with respect to the input of the op.
+    """
+    return _allreduce(grad)
```
alsrgv: Since broadcast sends data from rank 0 and to everyone else, presumably gradients on ranks > 0 should be always zero?

horovod/tensorflow/mpi_ops.py
```diff
@@ -119,4 +149,15 @@ def broadcast(tensor, root_rank, name=None):
     return MPI_LIB.horovod_broadcast(tensor, name=name, root_rank=root_rank)
 
 
-ops.NotDifferentiable('HorovodBroadcast')
+@ops.RegisterGradient('HorovodBroadcast')
+def broadcast_grad(op, grad):
```
alsrgv: Let's make all these gradient methods start with '_'.

horovod/tensorflow/mpi_ops.py
```diff
@@ -99,7 +111,25 @@ def allgather(tensor, name=None):
...
+    Returns:
+      The gradient with respect to the input of the op.
+    """
+    grad = _allreduce(grad)
+
+    x = op.inputs[0]
+    d = tf.convert_to_tensor([x.shape[0]], dtype=tf.int32)
+    d = tf.reshape(allgather(d), [size()])
+
+    splits = tf.split(grad, num_or_size_splits=d, axis=0)
```
alsrgv: I wonder if there's more optimized version of this?

horovod/tensorflow/mpi_ops.py
tgaddair: Yes, I think that makes sense.  Please see the updated _broadcast_grad function.

horovod/tensorflow/mpi_ops.py
tgaddair: I've been wondering that, too.  I think the most expensive part is that we do both an allreduce and allgather here, when it seems we could condense it to a single network call.  But with the current API, we don't have any way (that I know of) to do both at once.

I think in general it will not be too much of an issue, however.  Even in batch normalization, I would expect the user to primarily be using allreduce, not allgather, so I don't currently know of a usecase for this gradient.

test/test_tensorflow.py
```diff
@@ -308,7 +308,7 @@ def test_horovod_allreduce_grad(self):
         size = hvd.size()
 
         with self.test_session(config=self.config) as session:
-            dtypes = [tf.int32, tf.int64, tf.float32, tf.float64]
+            dtypes = [tf.float32, tf.float64]
```
alsrgv: Let's add comment that this is not supported in TF

tgaddair:
Resolves #331

PR #338, opened 2018-06-26
Made synchronous pure functional Horovod ops differentiable in PyTorch

tgaddair:
Note: only made the synchronous functions that do not modify tensors in place differentiable.  The reasoning here is that modifying tensors in place and performing async ops doesn't lend itself conceptually well to the concept of a graph, but open to adding additional support.

Because the input tensor needs to have the property requires_grad for this to do anything, there should not be any additional memory or compute cost from this implementation.


tgaddair:
Resolves #331

PR #359, opened 2018-07-05
Added custom load_model function to wrap the model optimizer with a Horovod DistributedOptimizer

tgaddair:


examples/keras_imagenet_resnet50.py
```diff
@@ -53,55 +54,63 @@
...
+
+    # Horovod: add Horovod Distributed Optimizer.
+    opt = hvd.DistributedOptimizer(opt)
+
+    model.compile(loss=keras.losses.categorical_crossentropy,
+                  optimizer=opt,
+                  metrics=['accuracy', 'top_k_categorical_accuracy'])
+else:
+    # Restore from a previous checkpoint, if initial epoch is specified.
+    model = hvd.load_model(checkpoint_format.format(epoch=resume_from_epoch))
```
alsrgv: Only rank-0 should be required to read the file, as per:
```
Horovod: restore on the first worker which will broadcast weights to other workers.
```

examples/keras_imagenet_resnet50.py
```diff
@@ -110,32 +119,34 @@
...
-    # Horovod: after the warmup reduce learning rate by 10 on the 30th, 60th and 80th epochs.
-    hvd.callbacks.LearningRateScheduleCallback(start_epoch=warmup_epochs, end_epoch=30, multiplier=1.),
-    hvd.callbacks.LearningRateScheduleCallback(start_epoch=30, end_epoch=60, multiplier=1e-1),
-    hvd.callbacks.LearningRateScheduleCallback(start_epoch=60, end_epoch=80, multiplier=1e-2),
+    hvd.callbacks.LearningRateWarmupCallback(warmup_epochs=warmup_epochs,
+                                             verbose=verbose),
+
+    # Horovod: after the warmup reduce learning rate by 10 on the 30th, 60th
+    # and 80th epochs.
+    hvd.callbacks.LearningRateScheduleCallback(start_epoch=warmup_epochs,
```
alsrgv: We could keep this file at line limit of 110-120 chars as it is right now -- much fewer line breaks.

horovod/keras/__init__.py
```diff
@@ -144,3 +144,55 @@ def broadcast(value, root_rank, name=None):
...
+
+    # Arguments
+        filepath: One of the following:
+            - string, path to the saved model, or
+            - h5py.File object from which to load the model
+        custom_optimizers: Optional list of Optimizer subclasses to support
+            during loading.
+        custom_objects: Optional dictionary mapping names (strings) to custom
+            classes or functions to be considered during deserialization.
+        compile: Boolean, whether to compile the model after loading.
```
alsrgv: Mention compile is also optional and defaults to True.

examples/keras_imagenet_resnet50.py
tgaddair: Does that also broadcast the optimizer weights and state as well?

alsrgv: If they are TF global variables, it should - but would be great to check.

examples/keras_imagenet_resnet50.py
tgaddair: Done.

horovod/keras/__init__.py
tgaddair: This param is not available in Keras 2.0.0, so removed.

examples/keras_imagenet_resnet50.py
```diff
@@ -74,29 +74,35 @@
...
-opt = keras.optimizers.SGD(lr=learning_rate * hvd.size(), momentum=0.9)
-
-# Horovod: add Horovod Distributed Optimizer.
-opt = hvd.DistributedOptimizer(opt)
-
-model.compile(loss=keras.losses.categorical_crossentropy,
-              optimizer=opt,
-              metrics=['accuracy', 'top_k_categorical_accuracy'])
+# Restore from a previous checkpoint, if initial_epoch is specified.
+# Horovod: restore on the first worker which will broadcast weights to other workers.
```
alsrgv: Let's clarify here that it will broadcast both model and optimizer weights.

PR #371, opened 2018-07-11
 Added PyTorch support for restoring optimizer state on model load and broadcast

tgaddair:


horovod/torch/__init__.py
```diff
@@ -129,15 +131,43 @@ def broadcast_parameters(params, root_rank):
...
+        if len(params.state_dict()['state']) == 0:
+            for group in params.param_groups:
+                for p in group['params']:
+                    p.grad = torch.autograd.Variable(
+                        p.data.new(p.size()).zero_())
+            params.step()
+
+        new_params = []
+        occurrences = collections.defaultdict(int)
+        for group in params.state_dict()['param_groups']:
```
alsrgv: Is it possible that these groups will be ordered differently on different nodes?

test/test_torch.py
```diff
@@ -650,3 +653,54 @@ def test_horovod_broadcast_grad(self):
...
+
+        y_pred = model(x)
+        loss = F.mse_loss(y_pred, y, size_average=False)
+        optimizer.zero_grad()
+        loss.backward()
+        optimizer.step()
+
+        if hvd.rank() == 0:
+            _, fname = tempfile.mkstemp('.h5')
+            hvd.save_model(fname, model, optimizer)
```
alsrgv: Can we add a custom state to the test and make sure it gets deserialized?

horovod/torch/__init__.py
```diff
@@ -129,15 +131,43 @@ def broadcast_parameters(params, root_rank):
...
     if isinstance(params, dict):
         params = sorted(params.items())
-    else:
+    elif isinstance(params, list):
         # support both named_parameters() and regular parameters()
         params = [p if isinstance(p, tuple) else (None, p) for p in params]
+    elif isinstance(params, torch.optim.Optimizer):
+        # Newly created optimizers will not have their state initialized, so
+        # do that initialization here
+        if len(params.state_dict()['state']) == 0:
```
alsrgv: `params.state_dict()` is actually non-trivial op, we should cache it.  I'd recommend looking `optimizer.load_state_dict()`, `optimizer.state_dict()`.  It seems that there are a lot of corner cases, including CUDA placement.  I wonder if we could just serialize the state_dict, transmit via `hvd.broadcast()`, and load using `optimizer.load_state_dict()`?

horovod/torch/__init__.py
tgaddair: No, I'll add a comment on that.  The `params` list is ordered by the order of the layers in the model.

test/test_torch.py
tgaddair: Yes, will do that.

horovod/torch/__init__.py
tgaddair: To make sure I understand correctly, is the suggestion here to serialize the state dict, put it in a tensor, broadcast it with a unique name, then pull out the blob, deserialize it, and load it like so:

```
t = tensor_from_blob(serialize(optimizer.state_dict()))
hvd.broadcast(name, t)
if hvd.rank() > 0:
    state_dict = deserialize(blob_from_tensor(t))
    optimizer.load_state_dict(state_dict)
```

tgaddair: If we were to do that, how could we guarantee that the Tensors have the same dimension?

alsrgv: Something like this:
```
if hvd.rank == 0:
    t = tensor_from_blob(serialize(optimizer.state_dict()))
    sz = torch.tensor(t.size()).int()
    hvd.broadcast_(name+".sz", sz)
else:
    sz = torch.tensor(3).int() # how many dims?
    hvd.broadcast_(name+".sz", sz)
    t = torch.tensor(sz)
hvd.broadcast_(name, t)
if hvd.rank() > 0:
    state_dict = deserialize(blob_from_tensor(t))
    optimizer.load_state_dict(state_dict)
```

horovod/torch/__init__.py
```diff
@@ -129,15 +131,48 @@ def broadcast_parameters(params, root_rank):
...
+        for group in state_dict['param_groups']:
+            # The params list here is ordered by the layers in the model
+            for pid in group['params']:
+                param_state = state_dict['state'][pid]
+                for name, p in param_state.items():
+                    # Some parameter names may appear more than once, in which
+                    # case we ensure they have a unique identifier defined by
+                    # their order
+                    occurrences[name] += 1
+                    name = '%s.%d' % (str(name), occurrences[name])
```
alsrgv: Is there guarantee here that if there are two parameters named `xyz` in each process (on different groups/params), they will always be traversed in the same order?

horovod/torch/__init__.py
```diff
@@ -150,3 +199,45 @@ def broadcast_parameters(params, root_rank):
     # Wait for completion.
     for handle in handles:
         synchronize(handle)
+
+
+def save_model(filepath, model, optimizer=None, custom_state=None):
```
alsrgv: I think we can nuke these, as there's nothing Horovod-specific.  Just doing `torch.save()` of dict would be almost as short.  WDYT?

tgaddair: Makes sense.  Would you also then remove the `load_model` function as well?  It would break parity with the Keras API, but there's nothing Horovod specific there either.

alsrgv: Right, these are too simple.  Keras has that function because it has magical optimizer wrappers.  As long as you can still do similar things in both frameworks, there's parity. :-)

horovod/torch/__init__.py
```diff
@@ -122,22 +126,71 @@ def DistributedOptimizer(optimizer, named_parameters=None):
...
+        b = io.BytesIO()
+        pickle.dump(obj, b)
+        buf = bytearray(b.getvalue())
+        t = torch.ByteTensor(buf)
+        sz = torch.IntTensor([t.shape[0]])
+        broadcast_(sz, root_rank, name + '.sz')
+    else:
+        sz = torch.IntTensor([0])
+        broadcast_(sz, root_rank, name + '.sz')
+        t = torch.ByteTensor([0] * sz.tolist()[0])
```
alsrgv: What if the tensor is 5GB?  `*` may be quite inefficient.

test/test_torch.py
```diff
@@ -650,3 +653,64 @@ def test_horovod_broadcast_grad(self):
...
+        hvd.broadcast_parameters(model.state_dict(), 0)
+        state_dict = hvd.broadcast_object(optimizer.state_dict(), 0)
+        if hvd.rank() > 0:
+            optimizer.load_state_dict(state_dict)
+
+        opt_state_dict = optimizer.state_dict()
+        self.assertEqual(opt_state_dict['param_groups'][0]['lr'], 2.0)
+        self.assertEqual(opt_state_dict['param_groups'][0]['momentum'], 0.9)
+        self.assertEqual(len(opt_state_dict['state'].values()), 4)
+
```
alsrgv: Nit: nuke whitespace in the end.

horovod/torch/__init__.py
alsrgv: You can just specify the size, no need for zeroing out.

tgaddair: Fixed.

test/test_torch.py
tgaddair: Fixed.

PR #374, opened 2018-07-13
C++11 syntax enhancement

chengscott:
- another recommendation

in `horovod/common/timeline.cc`:
those escaped `"` can be rewritten as raw string literal


horovod/common/operations.cc
```diff
@@ -92,8 +92,8 @@ typedef struct {
   int device;
   // A callback to call with the status.
   StatusCallback callback;
-} TensorTableEntry;
-typedef std::unordered_map<std::string, TensorTableEntry> TensorTable;
+};
+using TensorTable = std::unordered_map<std::string, TensorTableEntry>;
```
alsrgv: There are two other typedefs in that file, can you change those as well?

PR #394, opened 2018-07-21
Initialize Horovod with the provided MPI ranks or MPI communicator

vloncar:
This enables users to initialize Horovod on a specified list of ranks or a given communicator. It allows multiple Horovod instances to be embedded in a larger MPI application. Very similar to the existing (stale) #218

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,8 +254,6 @@ provided that the MPI was built with multi-threading support.
 
 You can check for MPI multi-threading support by querying the `hvd.mpi_threads_supported()` function.
 
-**Note**: Make sure that MPI library will **NOT** re-initialize MPI.  For example:
-
 ```python
 import horovod.tensorflow as hvd
 
@@ -264,10 +262,6 @@ hvd.init()
 
 # Verify that MPI multi-threading is supported.
 assert hvd.mpi_threads_supported()
-
-# Make sure MPI is not re-initialized.
-import mpi4py.rc
-mpi4py.rc.initialize = False
 
 from mpi4py import MPI
 assert hvd.size() == MPI.COMM_WORLD.Get_size()
--- a/horovod/common/__init__.py
+++ b/horovod/common/__init__.py
@@ -60,7 +60,7 @@ def init(comm=None):
     if comm is None:
         comm = []
 
-    atexit.register(shutdown, finalize=True)
+    atexit.register(shutdown)
 
     if not isinstance(comm, list):
         from mpi4py import MPI
@@ -78,8 +78,8 @@ def init(comm=None):
             (ctypes.c_int * comm_size)(*comm), ctypes.c_int(comm_size))
 
 
-def shutdown(finalize=False):
-    return MPI_COMMON_LIB_CTYPES.horovod_shutdown(ctypes.c_bool(finalize))
+def shutdown():
+    return MPI_COMMON_LIB_CTYPES.horovod_shutdown()
 
 
 def size():
--- a/horovod/common/__init__.py
+++ b/horovod/common/__init__.py
@@ -47,19 +47,31 @@ MPI_COMMON_LIB_CTYPES = \
     ctypes.CDLL(os.path.join(os.path.dirname(__file__),
                              'mpi_lib' + get_ext_suffix()), mode=ctypes.RTLD_GLOBAL)
 
-
-def init(ranks=None):
+def init(comm=None):
     """A function that initializes Horovod.
 
     Args:
-      ranks: List specifying ranks for the communicator, relative to the WORLD communicator.
+      comm: List specifying ranks for the communicator, relative to the WORLD communicator OR
+        the MPI communicator to use.
         If None, horovod will use WORLD Communicator.
 
     """
-    if ranks is None:
-        ranks=[]
-    nranks = len(ranks)
-    return MPI_COMMON_LIB_CTYPES.horovod_init((ctypes.c_int * nranks)(*ranks), ctypes.c_int(nranks))
+    if comm is None:
+        comm=[]
+
+    if not isinstance(comm, list):
+        from mpi4py import MPI
+        if MPI._sizeof(MPI.Comm) == ctypes.sizeof(ctypes.c_int):
+            MPI_Comm = ctypes.c_int
+        else:
+            MPI_Comm = ctypes.c_void_p
+            MPI_COMMON_LIB_CTYPES.horovod_init_comm.argtypes = [MPI_Comm]
+
+        comm_obj = MPI_Comm.from_address(MPI._addressof(comm))
+        return MPI_COMMON_LIB_CTYPES.horovod_init_comm(comm_obj)
+    else:
+        comm_size = len(comm)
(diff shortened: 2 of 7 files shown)


horovod/common/__init__.py
```diff
@@ -46,12 +47,36 @@ def check_extension(ext_name, ext_env_var, pkg_path, *args):
     ctypes.CDLL(os.path.join(os.path.dirname(__file__),
                              'mpi_lib' + get_ext_suffix()), mode=ctypes.RTLD_GLOBAL)
 
-
-def init():
+def init(comm=None):
```
alsrgv: Can you run `autopep8` on Python files?

horovod/common/operations.cc
```diff
@@ -160,15 +160,17 @@ struct HorovodGlobalState {
   // Whether MPI_Init has been completed on the background thread.
   bool initialization_done = false;
 
-  // The MPI rank, local rank, size, local size and flag indicating whether MPI
-  // multi-threading is supported.
+  // The MPI rank, local rank, size, local size, flag indicating whether MPI
```
alsrgv: Can you run `clang-format` on C++ files?

horovod/common/__init__.py
```diff
@@ -46,12 +47,36 @@ def check_extension(ext_name, ext_env_var, pkg_path, *args):
...
+    Args:
+      comm: List specifying ranks for the communicator, relative to the MPI_COMM_WORLD
+        communicator OR the MPI communicator to use. Given communicator will be duplicated.
+        If None, Horovod will use MPI_COMM_WORLD Communicator.
+
+    """
+    if comm is None:
+        comm=[]
+
+    atexit.register(terminate, finalize=True)
```
alsrgv: Wouldn't forcing `finalize=True` cause problems for mpi4py, which will try to finalize as well?

alsrgv: Idea: should we do `MPI_Finalize()` only if we did `MPI_Init_thread()`, and not provide `finalize` flag altogether, until somebody comes up with use case where it is necessary?

horovod/keras/__init__.py
```diff
@@ -18,6 +18,7 @@
 
 import horovod.tensorflow as hvd
 from horovod.common import init
+from horovod.common import terminate
```
alsrgv: Can you add to TensorFlow and PyTorch as well?

horovod/common/__init__.py
```diff
@@ -46,12 +47,36 @@ def check_extension(ext_name, ext_env_var, pkg_path, *args):
...
+            MPI_Comm = ctypes.c_void_p
+            MPI_COMMON_LIB_CTYPES.horovod_init_comm.argtypes = [MPI_Comm]
+
+        comm_obj = MPI_Comm.from_address(MPI._addressof(comm))
+        return MPI_COMMON_LIB_CTYPES.horovod_init_comm(comm_obj)
+    else:
+        comm_size = len(comm)
+        return MPI_COMMON_LIB_CTYPES.horovod_init((ctypes.c_int * comm_size)(*comm), ctypes.c_int(comm_size))
+
+def terminate(finalize=False):
```
alsrgv: `terminate` sounds a bit fatalistic and like something that can only be done once.  I can see why `finalize` may be loaded name here, but perhaps `shutdown` is better?

horovod/common/__init__.py
vloncar: Using `finalize=True` doesn't cause problems for mpi4py as it has a check if MPI is already finalized. If I am not mistaken mpi4py does [this](https://github.com/mpi4py/mpi4py/blob/527867aa8415d7086b1d387eb2f76c42ee060efe/src/mpi4py/MPI/atimport.pxi#L205). If you would prefer we only finalize if we called `MPI_Init_thread()` I can do that, it doesn't affect our use case.

alsrgv: It's easy to add parameters, but hard to remove. :-)

I think we should finalize only if we initialized, and provide knobs if somebody asks for it.

vloncar:
With these changes, importing mpi4py before or after Horovod has been initialized makes no difference. Both will try to initialize threads with 'multiple' support. So in that sense, disabling mpi4py initialization as shown in the example is not needed. However, in case we want to use `HOROVOD_MPI_THREADS_DISABLE`, then mpi4py should not initialize threads. Maybe we can remove the mpi4py exmaple from main `readme.md` and instead mention it along with `HOROVOD_MPI_THREADS_DISABLE` in `running.md`?

alsrgv:
@vloncar, mpi4py won't actually work without multi-threaded MPI because background Horovod thread and Python thread are running concurrently.  At the same time, many users are not even aware of mpi4py which could give them a lot of benefits with an incremental learning curve. I think it's fine to simplify mpi4py example in README but still mention that it requires multi-threading.

README.md
```diff
@@ -265,10 +263,6 @@ hvd.init()
 # Verify that MPI multi-threading is supported.
 assert hvd.mpi_threads_supported()
 
-# Make sure MPI is not re-initialized.
-import mpi4py.rc
-mpi4py.rc.initialize = False
-
 from mpi4py import MPI
```
alsrgv: I think this will actually fail because `hvd.init()` will initialize MPI. We should bring MPI import before `hvd.init()`

horovod/common/operations.cc
```diff
@@ -1305,29 +1311,52 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
...
+    std::strtol(mpi_threads_disable, nullptr, 10) > 0) {
     required = MPI_THREAD_FUNNELED;
   }
   int provided;
-  MPI_Init_thread(NULL, NULL, required, &provided);
+  int is_mpi_initialized = 0;
+  MPI_Initialized(&is_mpi_initialized);
+  if (is_mpi_initialized) {
+    MPI_Query_thread(&provided);
+  } else {
```
alsrgv: Let's add a warning - if we did not initialize mpi, and multi threading is not supported, this are very likely going to fail with segfault.

README.md
vloncar: I have tried it, it doesn't fail because mpi4py tests if MPI has already been initialized.

wgale:
Thank you! This will be incredibly useful.

georgezakinih:
Hi, 
Do you have an example that shows how I can pass a communicator created in C to initialize Horovod?
Thanks,
George.

alsrgv:
@georgezakinih, I haven't tried myself, but if you have ctypes object of the communicator, you should be able to pass it to `hvd.init(comm=xyz)`

alsrgv:
@vloncar, do you have an example?

vloncar:
I don't have an example, but like @alsrgv said, you can make a ctypes object and use that. There is an [example](https://github.com/mpi4py/mpi4py/tree/master/demo/wrap-ctypes) in mpi4py.

georgezakinih:
@alsrgv Thanks. this worked for me for OpenMPI. for other implementation, the pointer should be an integer: 
horovod_communicator_pointer = ...
comm_pointer = ctypes.c_void_p
mpi4py_comm = MPI.Intracomm()
handle = comm_pointer.from_address(MPI._addressof(mpi4py_comm))
handle.value = horovod_comm_int
hvd.init(mpi4py_comm)

PR #408, opened 2018-07-29
Add compatibility with PyTorch 0.4.1

alsrgv:
PyTorch 0.4.1 breaks the compatibility in two ways:
1. All `TH<xxx>Tensor` types have been aliased to `THTensor` which breaks templating.
2. Computing gradients for non-floating types has been prohibited.

This change is adding an integration test running over latest PyTorch release as well.


horovod/torch/tensor_util.h
```diff
@@ -35,171 +35,211 @@ namespace torch {
...
-  template <> void TensorUtil::Free<THTensor>(THTensor * tensor);              \
   template <>                                                                  \
-  void TensorUtil::ResizeNd<THTensor>(THTensor * tensor, int nDimension,       \
-                                      int64_t* size, int64_t* stride);         \
+  THTensor* TensorUtil::New<HorovodType, DeviceType, THTensor>(int device);    \
+  template <>                                                                  \
+  void TensorUtil::Free<HorovodType, DeviceType, THTensor>(THTensor * tensor); \
+  template <>                                                                  \
+  void TensorUtil::ResizeNd<HorovodType, DeviceType, THTensor>(                \
+      THTensor * tensor, int nDimension, int64_t* size, int64_t* stride);      \
```
tgaddair: Nit: Why the inconsistent style with `THTensor *` and `int64_t*`?

setup.py
```diff
@@ -459,20 +463,43 @@ def build_tf_extension(build_ext, options):
     return [flag for flag in tf_compile_flags if '_GLIBCXX_USE_CXX11_ABI' in flag]
 
 
+def parse_version(version_str):
+    m = re.match('^(\d+)(?:\.(\d+))?(?:\.(\d+))?(?:\.(\d+))?', version_str)
+    if m is None:
+        return None
+
+    # turn version string to long integer
+    version = int(m.group(1)) * 10 ** 9
```
tgaddair: I believe this is just going to give you a 32-bit integer.  You might want to do this instead:

`version = long(m.group(1)) * 10L ** 9`

horovod/torch/tensor_util.h
alsrgv: No idea - `clang-format` does it :-)

setup.py
alsrgv: Doesn't look like Python cares about it:
```
>>> int(1000)*10**9
1000000000000
```

tgaddair: I see, apparently they unified the `int` and `long` types into some time back.

horovod/torch/tensor_util.h
tgaddair: May as well roll with it then.

PR #419, opened 2018-08-02
Added command line args to Keras ImageNet examples

tgaddair:
Intent is to have parity with those in the PyTorch examples

### DIFF
--- a/examples/keras_imagenet_resnet50.py
+++ b/examples/keras_imagenet_resnet50.py
@@ -31,6 +31,8 @@ parser.add_argument('--log-dir', default='./logs',
                     help='tensorboard log directory')
 parser.add_argument('--checkpoint-format', default='./checkpoint-{epoch}.pth.tar',
                     help='checkpoint file format')
+
+# Default settings from https://arxiv.org/abs/1706.02677.
 parser.add_argument('--batch-size', type=int, default=32,
                     help='input batch size for training')
 parser.add_argument('--val-batch-size', type=int, default=32,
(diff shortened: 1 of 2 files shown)


examples/keras_imagenet_resnet50.py
```diff
@@ -29,20 +56,21 @@
...
 K.set_session(tf.Session(config=config))
 
-# Settings from https://arxiv.org/abs/1706.02677.
-batch_size = 32
-learning_rate = 0.0125
-warmup_epochs = 5
-weight_decay = 0.00005
-epochs = 90
+# Default settings from https://arxiv.org/abs/1706.02677.
+train_batch_size = args.batch_size
```
alsrgv: Let's inline these, and move the comment about default settings to where args are defined.

tgaddair: Done.

examples/keras_imagenet_resnet50.py
```diff
@@ -13,13 +13,43 @@
...
 
+parser = argparse.ArgumentParser(description='Keras ImageNet Example',
+                                 formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+parser.add_argument('--train-dir', default=os.path.expanduser('~/imagenet/train'),
+                    help='path to training data')
+parser.add_argument('--val-dir', default=os.path.expanduser('~/imagenet/validation'),
+                    help='path to validation data')
+parser.add_argument('--log-dir', default='./logs',
+                    help='tensorboard log directory')
+parser.add_argument('--checkpoint-format', default='./checkpoint-{epoch}.pth.tar',
```
alsrgv: Should be `./checkpoint-{epoch}.h5`

PR #432, opened 2018-08-09
Upgrade TensorFlow to 1.10.0 in Dockerfile, add build script

alsrgv:


build-docker-images.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
tgaddair: This should probably be added to the README.md at some point.

alsrgv: I'm thinking to add it to the `DEVELOPMENT.md` in future.

PR #453, opened 2018-08-17
Deep gradient compression implementation in the common layer using CUDA

sgpyc:
This is a CUDA implementation of the deep gradient compression (DGC, https://arxiv.org/abs/1712.01887). The entry point of DGC is https://github.com/sgpyc/horovod/blob/dgc/horovod/common/operations.cc#L1070, replacing ncclAllReduce() call when horovod_global::use_dgc is true.

Some changes have been made to support DGC:
1. setup.py: added nvcc compilation path, to support CUDA kernel compilation
2. externals/cub: added CUB (https://github.com/NVlabs/cub/tree/1.8.0/cub) as dependency, to have GPU sorting capability; tried thrust, but it has a bug leading to illegal memory access
3. using environmental variables to pass in configurations to horovod and DGC.

Main parts of DGC:
horovod/common/{dgc.h, dgc.cu, dgc_kernel.cu.cc} : main DGC routines
horovod/common/device_intrinsics.cuh : basic device functions
horovod/common/types.h: type definitions for DGC

The code has been tested using RestNet50 on ImageNet, with upto 28 DGX-1s connected via Infiniband.
Further testing and modifications are may still required for other models / datasets / enviroments


rongou:
Thanks @sgpyc!

@alsrgv I know it's a lot of code :), would be happy to discuss how best to test/integrate this.

anpark:
@sgpyc  could you share the benchmark results of this patch? any details for this design?

sgpyc:
Preliminary results show about 5% to 10% speedup in terms of images / sec, with similar accuracies, for RestNet50 on ImageNet, using 4 DGX-1V connected via 100Gbps infinibands.
I writing a document on the design, should be able to share within a few days.

sgpyc:
@anpark I wrote up the design doc for this PR (https://docs.google.com/document/d/14_MnkjGgmHS3Q9rHuAoIpI_yMIpjpbx69JHrYxI01Q8/edit?usp=sharing ), comments are welcomed.

The Travis CI tests (https://travis-ci.org/uber/horovod/builds/424688209) failed 2 out of 10, but the issue seems coming from not being able to get PyTorch Nightly build from https://pypi.org/simple/torch-nightly/ . Let me know if the error is indeed caused by the PR. Thanks

anpark:
@sgpyc  Great, I'll check it

Lyken17:
Hi, is this PR merged into master branch? This is a cool feature allowing users to train with much cheaper network.

tTianyu:
Hi，I've read the PDF of this PR and I curious about that how do u check the communication volume. some software or program ?

horovod/common/dgc.cu
```diff
@@ -0,0 +1,2272 @@
...
+    __device__ (const int &layer) {
+      coefficients[layer] = clipping_threshold /
+        (sqrt(sums[layer]) * total_num_layers + 1e-6);
+    });
+
+  // Update gradients
+  loop_kernel<<<grid_size, block_size, 0, stream>>>(start_counter,
+    [layer_starts, gradients, coefficients, num_layers]
+    __device__ (const uint32_t &i) {
+      int layer = binarySearch(layer_starts, 0, num_layers, i);
```
shahzadlone: ```layer``` should be const if it isn't getting mutated / modified.

horovod/common/operations.cc
```diff
@@ -1049,7 +1065,48 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
...
+          state.layer_offset_bytes.clear();
+          state.step_counters.clear();
+          state.memory_table.clear();
+        }
+        config.context = entries[0].context;
+        config.device  = entries[0].device;
+
+        // Get names and sizes of each layer, in order
+        std::vector<std::pair<std::string, uint64_t> > layers;
+        layers.clear();
```
shahzadlone: Why do you clear it as soon as you make the vector.

Instead consider reserving that same size as entries: 
```diff
-    layers.clear();
+    layers.reserve(entries.size());
```

PR #458, opened 2018-08-22
Migrate PyTorch to C++ bindings

alsrgv:
Migrate PyTorch from deprecated `cffi` API to new C++ extensibility interface https://pytorch.org/tutorials/advanced/cpp_extension.html.  In order to minimize change surface, some Python APIs are left intact and can be modified once we fully deprecate `cffi`.


horovod/torch/mpi_ops_v2.cc
```diff
@@ -0,0 +1,324 @@
...
+  m.def("horovod_torch_allreduce_async_torch_IntTensor", &DoAllreduce);
+  m.def("horovod_torch_allreduce_async_torch_LongTensor", &DoAllreduce);
+  m.def("horovod_torch_allreduce_async_torch_FloatTensor", &DoAllreduce);
+  m.def("horovod_torch_allreduce_async_torch_DoubleTensor", &DoAllreduce);
+#if HOROVOD_GPU_ALLREDUCE
+  m.def("horovod_torch_allreduce_async_torch_cuda_IntTensor", &DoAllreduce);
+  m.def("horovod_torch_allreduce_async_torch_cuda_LongTensor", &DoAllreduce);
+  m.def("horovod_torch_allreduce_async_torch_cuda_FloatTensor", &DoAllreduce);
+  m.def("horovod_torch_allreduce_async_torch_cuda_DoubleTensor", &DoAllreduce);
+#elif !HOROVOD_GPU_ALLREDUCE && HAVE_CUDA
```
tgaddair: Nit: Isn't checking `!HOROVOD_GPU_ALLREDUCE` here redundant given that `HOROVOD_GPU_ALLREDUCE` evaluated to false above?

Same for allgather and broadcast.

alsrgv: Good point :-)

horovod/torch/ready_event.cc
```diff
@@ -57,7 +56,6 @@ TorchReadyEvent::TorchReadyEvent(int device) : device_(device) {
   }
   auto stream = THCState_getCurrentStreamOnDevice(state, device_);
   THCudaCheck(cudaEventRecord(cuda_event_, stream));
-  THCudaCheck(cudaSetDevice(restoreDevice));
```
soumith: for streams and events, you can use the aten Stream API now. For example: `at::cuda::getCurrentCUDAStream()`. See https://github.com/pytorch/pytorch/pull/9435 for reference.

Events are also going to be exposed on ATen, PR is out: https://github.com/pytorch/pytorch/pull/9726

pnbrown:
I had the same error and was able to build from this branch. Thank you!

alsrgv:
I will merge this in, and migrate `ready_event.c` to `at::cuda::` when public headers are available.

PR #480, opened 2018-09-06
Implement  a custom  hierarchical ring which is compare to hierarchical_allreduce.

wu-yy:
in the code file "horovod/common/operation.cc", I find that the GPU's data must copy from rank '0'  and  use MPI_ALLreduce  when hierarchical_allreduce.

At the `hierarchical_allreduce_ring`,  I define  a nccl_across_ring_comm ,So I can use `ncclAllReduce`  instead of 'MPI_Allreduce' across the  hierarchical ranks.  Memory replication is eliminated.

 Even ,the ring's size can be  define any Integer  compare with the ' hierarchical_allreduce'.


wu-yy:
sure.

alsrgv:
Superseded by #481.

Frank00001:
Hello, I am working on synchronization algorithms in distributed machine learning. Following the introduction of Ring All-reduce by Baidu for the first time in deep learning, Tencent and Sony have respectively developed Hierarchical All-reduce and 2D-Torus All-reduce synchronization algorithms to further improve parallel efficiency. Based on the above work, we propose a new allreduce method, which greatly reduces the communication time.
I see your issue mentioned on horovod, you have implemented Hierarchical All-reduce yourself, and I feel admired for your work. We kindly ask you to leave your mailbox, hoping to communicate with you about the implementation of the synchronization algorithm.
    Thank you again for taking time to read my thoughts and I look forward to communicating with you!

PR #481, opened 2018-09-06
Implement a custom hierarchical ring which is compare to hierarchical_allreduce.

wu-yy:
I  squash some commits .

In the code file "horovod/common/operation.cc", I find that the GPU's data must copy from rank '0' and use MPI_ALLreduce when hierarchical_allreduce.

At the hierarchical_allreduce_ring, I define a nccl_across_ring_comm ,So I can use ncclAllReduce instead of 'MPI_Allreduce' across the hierarchical ranks. Memory replication is eliminated.

Even ,the ring's size can be defined any Integer compared with the ' hierarchical_allreduce'.

I am sorry about that my english is poor.

### DIFF
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -574,8 +574,6 @@ ncclDataType_t GetNCCLDataType(const std::shared_ptr<Tensor> tensor) {
     return ncclFloat32;
   case HOROVOD_FLOAT64:
     return ncclFloat64;
-  case HOROVOD_FLOAT16:
-    return ncclFloat16;
   default:
     throw std::logic_error("Type " + MPIDataType_Name(tensor->dtype()) +
                            " is not supported in NCCL mode.");
@@ -740,7 +738,6 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
       // We should never fail at finding this key in the tensor table.
       auto iter = tensor_table.find(name);
       assert(iter != tensor_table.end());
-	    //printf("operation.cc PerformOperation response.tensor_names name :%s,response Type:%d\n",name.c_str(),response.response_type());
 	
       assert(response.response_type() == MPIResponse::ALLREDUCE ||
              response.response_type() == MPIResponse::ALLGATHER ||
@@ -833,8 +830,7 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
   if (response.response_type() == MPIResponse::ALLGATHER) {
     assert(entries.size() == 1);
     auto e = entries[0];
-	  printf("operation.cc PerformOperation MPIResponse:ALLGATHER -->entry.size()%d\n",entries.size());
-    // Copy tensor sizes from the MPI response into a vector of int64_t
+	 // Copy tensor sizes from the MPI response into a vector of int64_t
     // and compute total size.  This is size of first dimension.
     std::vector<int64_t> tensor_sizes;
     int64_t total_dimension_size = 0;
@@ -1076,7 +1072,7 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
           num_elements += e.tensor->shape().num_elements();
         }
       } else {
-        //printf("Line 989 ...\n");
+        
         fused_input_data = first_entry.tensor->data();
         buffer_data = (void*)first_entry.output->data();
         num_elements = first_entry.tensor->shape().num_elements();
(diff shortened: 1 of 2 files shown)


alsrgv:
@wu-yy, thanks!  In future, please update existing pull request rather than create a new one.  I will close #480.  Is this similar in functionality to https://github.com/uber/horovod/tree/ring-of-rings?  Which scenarios does this provide speed up in?

wu-yy:
@alsrgv  thanks very much.

Wow, I find that this is similar in functionality to https://github.com/uber/horovod/tree/ring-of-rings.    I never find that.
 Memory replication is eliminated. So I think these could speed up in trainning deep learning models on distributed machins.

wu-yy:
@alsrgv Is it possible to use DPDK on MPI  to solve network bandwidth problems when use lots of machines ?

alsrgv:
@wu-yy, we have tested the ring-of-rings approach on a large number of servers, but never got it to work better than both hierarchical or single ring approaches.  I would recommend trying `HOROVOD_HIERARCHICAL_ALLREDUCE=1` on a large number of servers.

wu-yy:
@alsrgv  Thank you very much!

xiaozhouX:
Hello, I'm reading horovod code of all reduce implement and have some question about the issue. 
I think this PR is about using `ncclAllreduce` in place of `MPI_Allreduce`, which maybe will reduce memory copy between GPU and CPU. But it still a `HIERARCHICAL all reduce algorithm` .
The `ring-of-rings` branch seems to be different. I had read the [code](https://github.com/uber/horovod/blob/ring-of-rings/horovod/common/operations.cc#L882). It do a reduce inner node to make `rank 0` have the all sum tensor within node, and do a allReduce between every `rank 0` , than `rank 0` do a `ncclBcast` in node.
What do you think ?  @alsrgv  @wu-yy

horovod/common/operations.cc
```diff
@@ -863,22 +914,37 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
...
       auto event_queue = std::queue<std::pair<std::string, cudaEvent_t>>();
 
       // Determine GPU IDs of the devices participating in this communicator.
       std::vector<int32_t> nccl_device_map;
+
+      //TODO:wuyongyu Determine the GPU IDs for ring devices participating in this commmunicator
+      std::vector<int32_t> nccl_across_ring_device_map;
+
       if (horovod_global.hierarchical_allreduce) {
         for (int rank : horovod_global.local_comm_ranks) {
```
shahzadlone: if enters the ```if``` then don't forget to reserve ```nccl_across_ring_device_map``` before this ```for``` loop:
```
nccl_device_map.reserve(horovod_global.local_comm_ranks.size());
```

horovod/common/operations.cc
```diff
@@ -863,22 +914,37 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
...
           nccl_device_map.push_back(response.devices()[rank]);
         }
-      } else {
+      } else if(horovod_global.hierarchical_allreduce_ring){
+          for (int rank : horovod_global.ring_comm_ranks) {
```
shahzadlone: if enters the ```else if``` then don't forget to reserve ```nccl_device_map``` before this ```for``` loop:
```
nccl_device_map.reserve(horovod_global.ring_comm_ranks.size());
```

horovod/common/operations.cc
```diff
@@ -863,22 +914,37 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
...
+             nccl_device_map.push_back(response.devices()[rank]);
+          }
+          //TODO:wuyongyu define the nccl_ring _device_map
+          for ( int rank : horovod_global.across_ring_comm_ranks){
```
shahzadlone: please reserve ```nccl_across_ring_device_map``` before this ```for``` loop:
```
nccl_across_ring_device_map.reserve(horovod_global.across_ring_comm_ranks.size());
```

alsrgv:
We do support `HOROVOD_HIERARCHICAL_ALLREDUCE=1` which does a hybrid approach where allreduce in-node is done by NCCL, and allreduce across nodes is done by MPI.  NCCL 2.4 will introduce a native tree based algorithm as well.

Having a benchmark of this PR against Horovod's hierarchical ring & NCCL 2.4 showing an advantage would be necessary to proceed forward.

tgaddair:
Hey @Frank00001, happy to chat with you about what you have in mind, would be great to see your idea contributed to Horovod!  One thing worth mentioning is that recent versions of NCCL handle switching between ring and tree allreduce automatically, so this hierarchical allreduce option isn't recommended in most cases.

PR #493, opened 2018-09-13
Add python[23]-opencv to Travis CI script

alsrgv:
Recommended by https://unix.stackexchange.com/questions/444697/cannot-run-python-file-asks-to-relink-libraries.


alsrgv:
These attempts did not work, need to investigate further.  Closing this train of thought for now.

PR #523, opened 2018-09-26
PyTorch 1.0 readiness updates

alsrgv:
Accommodations for https://github.com/pytorch/pytorch/pull/11510

Removing PyTorch 0.3.1 backward compatibility to simplify the code.

cc @goldsborough


setup.py
```diff
@@ -507,6 +507,10 @@ def dummy_import_torch():
 def check_torch_version():
     try:
         import torch
+        if torch.__version__ < '0.4.0':
```
tgaddair: Generally safer to use `distutils.version`: https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python

alsrgv: Good idea!

PR #529, opened 2018-09-28
FP16 support for GPU tensors in all frameworks

tgaddair:
Versions of PyTorch < 1.0.0 are not supported.

Any attempt to reduce a CPU resident tensor will result in a runtime exception.


examples/keras_imagenet_resnet50.py
```diff
@@ -31,6 +31,8 @@
                     help='tensorboard log directory')
 parser.add_argument('--checkpoint-format', default='./checkpoint-{epoch}.h5',
                     help='checkpoint file format')
+parser.add_argument('--fp16', action='store_true', default=False,
```
alsrgv: Users will likely expect `--fp16` to mean FP16 computation.  Let's rename to `--fp16_allreduce`?

horovod/torch/__init__.py
```diff
@@ -75,21 +77,28 @@ def hook(*ignore):
...
+            self._handles[p] = (handle, compressor)
         return hook
 
     def synchronize(self):
-        for handle in self._handles.values():
-            synchronize(handle)
+        for p, value in self._handles.items():
+            handle, compressor = value
+            output = synchronize(handle)
+            p.grad.data = compressor.decompress(output)
```
alsrgv: I wonder if we should do `.set_()` instead of assignment.  Should look into pros/cons if any.

horovod/torch/mpi_ops.py
```diff
@@ -62,6 +65,12 @@ def _allreduce_function_factory(tensor):
 
 
 def _allreduce_async(tensor, output, average, name):
+    if tensor.dtype == torch.float16 and \
+            LooseVersion(torch.__version__) < LooseVersion('0.4.2'):
```
alsrgv: Is this parsing fast enough to do 1000 times per second?

test/test_torch.py
```diff
@@ -201,7 +201,7 @@ def test_horovod_allreduce_multi_gpu(self):
 
         iter = 0
         dtypes = [torch.cuda.IntTensor, torch.cuda.LongTensor,
-                  torch.cuda.FloatTensor, torch.cuda.DoubleTensor]
+                  torch.cuda.HalfTensor, torch.cuda.FloatTensor, torch.cuda.DoubleTensor]
```
alsrgv: Let's add `torch.cuda.HalfTensor` to other tests as well?  It implies GPU by itself.

horovod/torch/mpi_ops.py
tgaddair: I'll cache it on import.

horovod/torch/__init__.py
tgaddair: I was wondering about that myself.  For the non-fp16 case, this will be a no-op, as it'll just assign the tensor to itself.  But in the case of fp16, we always need to create a new tensor, because there is no way to do an in-place type cast.  

It may be possible to call `set_()` and pass in a tensor of a different type (fp16).  According to the documentation on [`set_()`](https://pytorch.org/docs/stable/tensors.html), the two tensors will share the same underlying storage, which seems problematic if the current type is fp32, but the parameter tensor type is fp16.

alsrgv: Nah, we'd still need to convert, but I'm curious about tradeoff between `.set_()` vs `=`.  I'm wondering if any other thing may have a pointer to `p.grad.data`.

tgaddair: I managed to track the implementation down to [here](https://github.com/torch/torch7/blob/master/lib/TH/generic/THTensor.c#L765) in Torch7 (I believe ATen is similar).

The difference, as I can tell, is that `set_` will delete the underlying storage of the previous tensor data, and set it to the provided tensor's storage (setting it for retention).  Whereas `=` will just replace the pointers, so any properties of the old tensor will be lost.

Overall, I think using `set_` is the safest option here.  I don't think it will be any faster, necessarily (though cleanup will happen faster, avoiding garbage collection).

alsrgv:
Since there's no 0.4.2, can you update the check to `>= 1.0.0`?

PR #532, opened 2018-09-28
Fully de-initialize Horovod upon shutdown

alsrgv:


horovod/common/operations.cc
```diff
@@ -1902,11 +1902,11 @@ void horovod_init_comm(MPI_Comm comm) {
 }
 
 void horovod_shutdown() {
-
   if (horovod_global.background_thread.joinable()) {
     horovod_global.shut_down = true;
     horovod_global.background_thread.join();
     // Reset the initialization flag to allow restarting with horovod_init(...)
+    horovod_global.initialization_done = false;
```
alsrgv: The absence of this flag causes new ops to be enqueued follow up by PyTorch's wait_and_clear() that hangs forever.

PR #535, opened 2018-09-29
Added fp16 to pytorch_imagenet_resnet50 example

tgaddair:


bapriddy:
@alsrgv @tgaddair

I'm curious about something so I'll ask...Do you guys have a way to estimate the size of the graph that lives on the GPU.  I'm referring to the pytorch_imagenet_resnet50.py code in this case.  As you probably know, a P100 has 16Gb of memory.  I'm wondering at what point will the size of the model become to large and assuming that some memory is allocated for other requirements.  I've been experimenting with resnet18, resnet50, resnet101.  I guess it might be easier to do this with a fully connected network.  Then the problems becomes a bipartite graph counting exercise (i Think).

alsrgv:
@bapriddy, `--batch-size=32` is set just to match original ImageNet in 1 hour paper.  I'm sure P100 will fit batch 64, if not more.  Batch normalization is quite sensitive to batch size, so you may get different results with `--batch-size=64`.

PR #542, opened 2018-10-06
Horovod support for MXNet framework

ctcyang:
This is a PR for adding Horovod support to do distributed training with the MXNet deep learning framework. The core Allreduce and Broadcast functionality passes unit tests. @ctcyang worked on the design and implementing the core functionalities. @yuxihu and @apeforest worked on addressing reviewers’ comments, simplifying and improving code logic, fixing build issues, implementing error handling mechanisms and improving two examples using MXNet with Hovorod for distributed training. We can consistently train ResNet-50 to top-1 accuracy between 75% and 77% with different training settings.

Our performance results are here showing throughput (scaling efficiency) with and without hierarchical allreduce (HA) on ResNet-50 with float32:
```
# gpus | Without HA |   With HA
---------------------------------
   8   |  3072  (NA)|  3078  (NA)
  16   |  6027 (98%)|  5859 (95%)
  32   | 12030 (98%)| 11675 (95%)
  64   | 22346 (83%)| 23166 (94%)
 128   | 40938 (84%)| 45972 (93%)
 256   | 64998 (66%)| 89858 (91%)
```

We have completed and merged the PR to make necessary changes on the MXNet for Horovod support. You can see the PR here: https://github.com/apache/incubator-mxnet/pull/12666

[Design document](https://cwiki.apache.org/confluence/display/MXNET/Horovod-MXNet+Integration)


NanYan1119:
It seems that the link of 'Installation instructions' has 404 error.

ctcyang:
@NanYan1119 Thanks, I've added the correct link

horovod/mxnet/mpi_ops.py
```diff
@@ -0,0 +1,248 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
NanYan1119: Here missing a '#' symbol.

ctcyang:
@NanYan1119 How many epochs into training does this happen? Are there any error messages before this?

NanYan1119:
I have run 90 epochs, the problem started from 15th epoch and from then occurred random until end. There are no errors and the training can still run next epoch.

ctcyang:
@NanYan1119 Hmm, that's strange. Have you been able to fix the problem? How many GPUs are on the two servers? Is it always the rank 15 GPU that is stuck waiting for tensors? What training/validation accuracy do you reach after 90 epochs?

NanYan1119:
There are 16 gpus on two servers. The problem occurred random and the missing ranks are also random such as 

allreduce.0 [missing ranks: 15]
allreduce.4 [missing ranks: 15]
allreduce.8 [missing ranks: 15]
allreduce.11 [missing ranks: 15]
allreduce.12 [missing ranks: 15]
allreduce.19 [missing ranks: 15]
allreduce.22 [missing ranks: 15]
allreduce.24 [missing ranks: 15]
allreduce.27 [missing ranks: 15]
allreduce.26 [missing ranks: 15]
allreduce.29 [missing ranks: 15]
allreduce.28 [missing ranks: 15]

The train-accuracy is 0.64 and the validation-accuracy is 0.675 after 90 epochs.

NanYan1119:
Can you tell me the meaning of 'hierarchical allreduce'? What should I do the training with and without hierarchical allreduce (HA)?

ctcyang:
Hierarchical allreduce uses NCCL to do the allreduce intra-node and MPI inter-node. Once the allreduce from other nodes is received after the MPI allreduce, NCCL is used to broadcast the result. You can read more about it in the PRs that added it:
https://github.com/uber/horovod/pull/329
https://github.com/uber/horovod/pull/411

examples/mxnet_imagenet_resnet50.py
```diff
@@ -0,0 +1,290 @@
...
+    opt = mx.optimizer.create('sgd', sym=out, **optimizer_params)
+    opt = hvd.DistributedOptimizer(opt)
+
+    # Create initializer and initializer parameters
+    initializer = mx.init.Xavier(rnd_type='gaussian', factor_type="in", magnitude=2)
+    mod.bind(data_shapes=train_data.provide_data, label_shapes=train_data.provide_label)
+    mod.init_params(initializer, arg_params=arg_params, aux_params=aux_params)
+
+    # Fetch and broadcast parameters
+    (arg_params, aux_params) = mod.get_params(copy_to_cpu=False)
```
yuxihu: you may need to remove copy_to_cpu argument once the changes of removing this argument in mxnet repo is validated.

horovod/mxnet/handle_manager.cc
```diff
@@ -0,0 +1,63 @@
...
+bool HandleManager::PollHandle(int handle) {
+  std::lock_guard<std::mutex> guard(mutex_);
+  if (results_.find(handle) == results_.end()) {
+    throw std::invalid_argument("Handle " + std::to_string(handle) +
+                                " was not created or has been cleared.");
+  }
+  return results_[handle] != nullptr;
+}
+
+void HandleManager::ExecuteCallback(int handle) {
```
yuxihu: Please sync with my fix: https://github.com/ctcyang/horovod/commit/98699886f8a678ef6f80a8958372fe4fc30955cd.

horovod/mxnet/handle_manager.cc
```diff
@@ -0,0 +1,63 @@
...
+}
+
+std::shared_ptr<Status> HandleManager::ReleaseHandle(int handle) {
+  std::lock_guard<std::mutex> guard(mutex_);
+  if (results_.find(handle) == results_.end()) {
+    throw std::invalid_argument("Handle " + std::to_string(handle) +
+        " was not created or has been cleared.");
+  }
+  auto status = results_[handle];
+  results_.erase(handle);
```
yuxihu: Please sync with my fix: https://github.com/ctcyang/horovod/commit/98699886f8a678ef6f80a8958372fe4fc30955cd

ctcyang: Done

horovod/mxnet/handle_manager.cc
ctcyang: Done

setup.py
```diff
@@ -214,6 +230,19 @@ def get_tf_flags(build_ext, cpp_flags):
         return compile_flags, link_flags
 
 
+def get_mx_flags(build_ext, cpp_flags, lib_dirs):
+    compile_flags = []
+    link_flags = []
+    mx_lib_dirs = ['/home/ubuntu/master/lib']
```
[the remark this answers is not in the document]
ctcyang: fixed by @apeforest

horovod/mxnet/mpi_ops.py
ctcyang: fixed

examples/mxnet_imagenet_resnet50.py
ctcyang: this parameter has been removed

ctcyang:
Reverted merge branch `prod/mxnet`. @yuxihu and @apeforest will submit another PR using that branch.

alsrgv:
@ctcyang, thanks for the PR.  Please let me know when it's ready to be looked at.

ctcyang:
@alsrgv Thank you for your note. We are waiting for Amazon legal team to look at this. Will let you know when it's ready to be looked at.

ctcyang:
@alsrgv Amazon legal team has given us the okay, so feel free to start looking over the code.

Dockerfile
```diff
@@ -1,15 +1,17 @@
 FROM nvidia/cuda:9.0-devel-ubuntu16.04
 
 # TensorFlow version is tightly coupled to CUDA and cuDNN so it should be selected carefully
-ENV TENSORFLOW_VERSION=1.11.0
```
alsrgv: Can unintended changes like these be reverted?

horovod/mxnet/adapter.cc
```diff
@@ -0,0 +1,144 @@
...
+#endif
+
+#include <mxnet/base.h>
+
+#include "adapter.h"
+#include "cuda_util.h"
+#include "tensor_util.h"
+
+namespace horovod {
+namespace MX {
```
apeforest: @ctcyang Could you please merge my commit 86c89c72bfe4a170e18436ca5acefc66cd2c0568 so the namespace will be consistent with tensorflow and tf? Thanks!

ctcyang: Done. Thanks!

horovod/mxnet/mpi_ops.cc
```diff
@@ -0,0 +1,411 @@
...
+}
+
+} // namespace
+
+int DoAllreduce(NDArray* tensor, NDArray* output, int average, const std::string& name, Callback cb) {
+  ThrowIfError(common::CheckInitialized());
+ 
+  auto handle = handle_manager.AllocateHandle(cb);
+  auto device = TensorUtil::GetDevice(tensor);
+  auto hvd_tensor = std::make_shared<MXTensor<NDArray>>(tensor);
```
alsrgv: You may need ready event here as well in case you do any async CUDA calls in MXNet.  In both PyTorch and TensorFlow we've found that you need to wait for current CUDA threads to finish before data is ready for NCCL and MPI.

horovod/mxnet/mpi_ops.cc
```diff
@@ -0,0 +1,411 @@
...
+    NDArray* input, NDArray* output, int average, char* name) {
+
+  std::string new_name = GetOpName("allreduce", name);
+  auto allreduce_async_fn = [input, output, new_name, average](
+      RunContext rctx, Engine::CallbackOnComplete cb) mutable {
+      DoAllreduce(input, output, average, new_name, cb);
+  };
+  auto allreduce_async_cpu_fn = [input, output, new_name, average](
+      RunContext rctx, Engine::CallbackOnComplete cb) mutable {
+      DoAllreduceCudaOnCPU(input, output, average, new_name, cb);
```
alsrgv: Does this compile on CPU-only?

horovod/mxnet/mpi_ops.cc
```diff
@@ -0,0 +1,411 @@
...
+      });
+  ThrowIfError(enqueue_result);
+
+  return handle;
+}
+#endif
+
+// Do AllReduce on GPU only if src and dst are on GPU
+// Otherwise do AllReduce on CPU
+extern "C" int horovod_mxnet_allreduce_async(
```
alsrgv: Should this logic live outside of `DoAllreaduce*` and other functions?

horovod/mxnet/mpi_ops.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    c_in = tensor.handle
+    c_out = tensor.handle
+    if isinstance(name, string_types):
+        check_call(MPI_MXNET_LIB_CTYPES.horovod_mxnet_broadcast_async(c_in, c_out, ctypes.c_int(root_rank), c_str(name)))
+    else:
+        check_call(MPI_MXNET_LIB_CTYPES.horovod_mxnet_broadcast_async(c_in, c_out, ctypes.c_int(root_rank), name))
+    return tensor
+      
+# TODO(@ctcyang):
```
alsrgv: Please implement TODOs

setup.py
```diff
@@ -379,6 +437,7 @@ def get_common_options(build_ext):
     cpp_flags = get_cpp_flags(build_ext)
     link_flags = get_link_flags(build_ext)
     mpi_flags = get_mpi_flags()
+    mxnet_include_dirs = os.environ.get('INCLUDES')
```
alsrgv: I'm not sure this environment variable will be set often.  Is there any other way?

setup.py
```diff
@@ -734,6 +822,18 @@ def build_extensions(self):
                     built_plugins.append(False)
                 else:
                     raise
+        if not os.environ.get('HOROVOD_WITHOUT_MXNET'):
+            try:
+                build_mx_extension(self, options)
+                built_plugins.append(True)
+            except Exception as e:
+                print(e)
```
alsrgv: Remove print?

setup.py
```diff
@@ -734,6 +822,18 @@ def build_extensions(self):
...
+                if not os.environ.get('HOROVOD_WITHOUT_MXNET'):
```
alsrgv: Should be `HOROVOD_WITH_MXNET`

horovod/mxnet/mpi_ops.cc
```diff
@@ -0,0 +1,411 @@
...
+  auto hvd_tensor = std::make_shared<MXTensor<NDArray>>(tensor);
+  auto hvd_context = std::make_shared<MXOpContext<NDArray>>(device, output);
+  auto hvd_output = std::make_shared<MXTensor<NDArray>>(output);
+
+  auto enqueue_result = EnqueueTensorAllreduce(
+      hvd_context, hvd_tensor, hvd_output, nullptr,
+      GetOpNameHandle("allreduce", name, handle), device,
+      [handle, average, output](const Status& status) {
+        handle_manager.MarkDone(handle, status);
+        handle_manager.ExecuteCallback(handle);
```
alsrgv: Why don't just do `cb()` here?  Do they need to be synchronized to each other as you do in `ExecuteCallback`?

setup.py
apeforest: Updated

setup.py
apeforest: Removed

examples/mxnet_imagenet_resnet50.py
```diff
@@ -0,0 +1,290 @@
...
+    # Create callback to sum validation accuracy
+    eval_end_callbacks = [LogValidationMetricsCallback()]
+
+    # Train model
+    mod.fit(train_data,
+            eval_data=val_data,
+            num_epoch=args.num_epochs,
+            kvstore=None,
+            batch_end_callback=mx.callback.Speedometer(batch_size, 20),
+            #eval_end_callback=eval_end_callbacks,
```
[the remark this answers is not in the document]
apeforest: Removed

examples/mxnet_imagenet_resnet50.py
```diff
@@ -0,0 +1,290 @@
...
+epoch_size = int(math.ceil(int(num_training_samples // num_workers)/batch_size))
+
+# Adjust learning rate
+lr_decay = args.lr_decay
+lr_decay_epoch = [int(i) for i in args.lr_decay_epoch.split(',')]
+steps = [epoch_size * x for x in lr_decay_epoch]
+lr_scheduler = MultiFactorScheduler(step=steps, factor=lr_decay, base_lr=args.lr, warmup_steps=(args.warmup_epochs * epoch_size), warmup_begin_lr=args.warmup_lr)
+
+# Two functions for reading data from record file or raw images
+def get_data_rec(rec_train, rec_train_idx, rec_val, rec_val_idx, batch_size, data_nthreads):
```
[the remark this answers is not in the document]
apeforest: Link added

examples/mxnet_imagenet_resnet50.py
```diff
@@ -0,0 +1,290 @@
...
+parser.add_argument('--batch-size', type=int, default=128,
+                    help='training batch size per device (CPU/GPU).')
+parser.add_argument('--dtype', type=str, default='float32',
+                    help='data type for training. default is float32')
+parser.add_argument('--gpus', type=str, default='0',
+                    help='number of gpus to use.')
+parser.add_argument('--num-epochs', type=int, default=90,
+                    help='number of training epochs.')
+parser.add_argument('--lr', type=float, default=0.1,
+                    help='learning rate. default is 0.1.')
```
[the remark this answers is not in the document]
apeforest: Updated

horovod/common/operations.cc
```diff
@@ -2022,6 +2022,10 @@ int horovod_mpi_threads_supported() {
   }
   return horovod_global.mpi_threads_supported ? 1 : 0;
 }
+
+int horovod_barrier() {
+  return MPI_Barrier(horovod_global.mpi_comm);
```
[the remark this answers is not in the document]
ctcyang: I thought they might've been necessary/handy to expose this method, but we were able to get it to work without it and haven't seen any usecases, so I'll remove it.

horovod/mxnet/mpi_ops.cc
ctcyang: Yeah, we were seeing segfaults when they were not synchronized with each other. @apeforest and @yuxihu tracked the segfaults to calling `cb()` here.

The specific commit: https://github.com/uber/horovod/commit/b2ef69eefbffb93bf61562bccf85af2164d56f22#diff-141f1184be39395e8f8708ff51a6fba7

examples/mxnet_imagenet_resnet50.py
```diff
@@ -0,0 +1,290 @@
...
+    aux_params = None
+
+    # Create model
+    mod = mx.mod.Module(softmax, context=context)
+
+    # Create optimizer
+    optimizer_params = {'wd': args.wd,
+                        'momentum': args.momentum,
+                        'lr_scheduler': lr_scheduler,
+                        'rescale_grad': 1.0/batch_size/num_workers,
```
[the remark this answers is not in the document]
ctcyang: The reason is that we have seen some performance degradation with doing gradient averaging in the DistributedOptimizer. It seems to happen, because of a `wait_to_read()` on the NDArray [here](https://github.com/uber/horovod/blob/645cbd0125b3556da531f8d5c1989a1308fb5c4b/horovod/mxnet/mpi_ops.py#L86). This `wait_to_read()` is necessary, because we noticed that the elementwise operations (that follows it on Line 87) get captured in the task graph generated by the scheduler. The two enter a race condition, which generates wrong results half the time.

More details in this commit: [0517fa0](https://github.com/uber/horovod/pull/542/commits/0517fa0289b0fdc12e3b7237b5ec4e5b39e98a1b)

setup.py
apeforest: We will update this soon once the include path is added to MXNet pip package.

horovod/mxnet/mpi_ops.cc
ctcyang: After discussion, we came to the conclusion that all async CUDA calls in MXNet have a `ctx.get_stream<gpu>()->Wait();` at the end of the call pushing them to MXNet engine. For example when copying data from CPU to GPU asynchronously, we use such a call [here](https://github.com/apache/incubator-mxnet/blob/5a83b6b563211f430688e41eab4752c6de4ecf22/src/ndarray/ndarray.cc#L1245) to ensure data is ready before the next CUDA kernel happens. This should guarantee that the current CUDA threads finish before data is ready for NCCL and MPI.

setup.py
yuxihu: Use an API to get the path now.

horovod/mxnet/mpi_ops.cc
ctcyang: At the very least the `Engine::Get()->PushAsync` component has to live outside `DoAllreduce*` and other functions. With MXNet the way synchronization is done with the engine is by calling the MXNet engine from Horovod, passing it a lambda (`DoAllreduce*`) and telling it what are the mutated variables. The engine creates a callback function and promises that no operations with the mutated variables as inputs can start before the callback is executed. @apeforest and @yuxihu are working on a PR that fixes CPU compilation and makes this part of the code tidier.

horovod/mxnet/mpi_ops.cc
ctcyang: @apeforest and @yuxihu are working on a PR that lets it compile on CPU-only.

horovod/mxnet/mpi_ops.py
ctcyang: I took a closer look at `allreduce_async` in the PyTorch branch and think it's sufficiently different from the MXNet execution model that we don't need the `poll` and `synchronize` to check the allreduce has completed.

horovod/mxnet/mpi_ops.cc
apeforest: Updated.

horovod/mxnet/mpi_ops.cc
apeforest: This comment is outdated. I have removed it.

apeforest:
@alshopov Hi Alex, we have adddressed your comments. Could you please take a look at this PR again at your convenience? Thanks!

yuxihu:
@alsrgv I have rebased and the CI passed. Please take a look at the PR and let us know if any questions. Thanks.

horovod/mxnet/tensor_util.cc
```diff
@@ -0,0 +1,172 @@
...
+  }
+  return shape;
+}
+
+// Return data of tensor
+const void* TensorUtil::GetData(NDArray* tensor) {
+  // The following returns an error:
+  // return tensor->data().dptr<void>();
+  switch (tensor->dtype()) {
+    case 0:
```
alsrgv: Can we name those constants?

.travis.yml
```diff
@@ -109,7 +112,7 @@ script:
   
 
   # run unit tests
-  - docker exec ${CONTAINER} /bin/sh -c "pip install pytest && cd /horovod/test && ${MPIRUN} pytest -v"
+  - docker exec ${CONTAINER} /bin/sh -c "pip install pytest && cd /horovod/test && ${MPIRUN} pytest -v --ignore=test_mxnet.py && ${MPIRUN} pytest -v test_mxnet.py"
```
alsrgv: No order in which MXNet can co-exist with TensorFlow and PyTorch?

examples/mxnet_imagenet_resnet50.py
```diff
@@ -0,0 +1,287 @@
...
+        self.cur_iter = 0
+        self.max_iter = max_iter
+        self.dtype = dtype
+        label = np.random.randint(0, num_classes, [self.batch_size,])
+        data = np.random.uniform(-1, 1, data_shape)
+        self.data = mx.nd.array(data, dtype=self.dtype, ctx=mx.Context('cpu_pinned', hvd.local_rank()))
+        self.label = mx.nd.array(label, dtype=self.dtype, ctx=mx.Context('cpu_pinned', hvd.local_rank()))
+    def __iter__(self):
+        return self
+    @property
```
alsrgv: Style: can you format according to PEP-8?  There should be a space between member declarations, two spaces around classes, etc.

examples/mxnet_imagenet_resnet50.py
```diff
@@ -0,0 +1,287 @@
...
+                    help='weight decay rate, default is 0.0001.')
+parser.add_argument('--lr-decay', type=float, default=0.1,
+                    help='decay rate of learning rate, default is 0.1.')
+parser.add_argument('--lr-decay-epoch', type=str, default='40,60',
+                    help='epoches at which learning rate decays, default is 40,60.')
+parser.add_argument('--warmup-lr', type=float, default=0.001,
+                    help='starting warmup learning rate, default is 0.001')
+parser.add_argument('--warmup-epochs', type=int, default=5,
+                    help='number of warmup epochs, default is 5.')
+parser.add_argument('--model', type=str, required=True,
```
alsrgv: Should this default to `resnet50` due to name of the python script?  Where to get this `vision_model`?

examples/mxnet_imagenet_resnet50.py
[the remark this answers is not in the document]
alsrgv: It's not very user-friendly to push it down on user though.  How large is the degradation?

I'm still reading the code, but could this averaging be done in the callback, like it's done for PyTorch?

horovod/mxnet/handle_manager.cc
```diff
@@ -0,0 +1,50 @@
...
+}
+
+void HandleManager::ExecuteCallback(int handle) {
+  std::unique_lock<std::mutex> lock(mutex_);
+  if (callbacks_.find(handle) == callbacks_.end()) {
+    return;
+  }
+  auto cb_ptr = callbacks_[handle];
+  lock.unlock();
+   if (cb_ptr != nullptr) {
```
alsrgv: Style: extra space before `if`

.travis.yml
yuxihu: We spent quite a bit time on this and let me try to explain what we found.

pytest loads all the tests by the alphabetic order of the filename. So the tests will run in the following order:
```
test_keras.py
test_mxnet.py
test_tensorflow.py
test_tensorflow_keras.py
test_torch.py
```
All the unit tests in each test file passed but we encountered an fatal error in MPI during pytest exit.
```
==================== 65 passed, 2 warnings in 12.94 seconds ====================

[c7949752ef7f:651] *** An error occurred in MPI_Gather
[c7949752ef7f:651] *** reported by process [2061107201,1]
[c7949752ef7f:651] *** on a NULL communicator
[c7949752ef7f:651] *** Unknown error
[c7949752ef7f:651] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[c7949752ef7f:651] ***    and potentially your MPI job)
[c7949752ef7f:651] [0] func:/usr/local/lib/libopen-pal.so.40(opal_backtrace_buffer+0x20) [0x7f6851c42090]
[c7949752ef7f:651] [1] func:/usr/local/lib/libmpi.so.40(ompi_mpi_abort+0x258) [0x7f68521cb4f8]
[c7949752ef7f:651] [2] func:/usr/local/lib/libmpi.so.40(ompi_mpi_errors_are_fatal_comm_handler+0xfd) [0x7f68521bbf0d]
[c7949752ef7f:651] [3] func:/usr/local/lib/libmpi.so.40(PMPI_Gather+0x269) [0x7f68521ec5a9]
[c7949752ef7f:651] [4] func:/usr/local/lib/python2.7/dist-packages/horovod/mxnet/mpi_lib.so(+0x2e050) [0x7f68391fd050]
[c7949752ef7f:651] [5] func:/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0xb6990) [0x7f68ecd78990]
[c7949752ef7f:651] [6] func:/lib/x86_64-linux-gnu/libpthread.so.0(+0x8064) [0x7f68f4866064]
[c7949752ef7f:651] [7] func:/lib/x86_64-linux-gnu/libc.so.6(clone+0x6d) [0x7f68f3c7862d]
```
I think there is an implicit requirement that we need to run all unit tests using the same backend framework together without interruption. Since Keras uses Tensorflow as backend, we need to run test_keras/test_tensorflow/test_tensorflow_keras together. With test_mxnet, it runs after test_keras and changes the order of running unit tests which causes the fatal error during exit.

It workes without test_mxnet because test_torch runs in the last without interrupting the tests using tensorflow backend. If we change test_torch to test_pytorch, the order of running unit tests is as follows:
```
test_keras.py
test_pytorch.py
test_tensorflow.py
test_tensorflow_keras.py
```
All the unit tests in each test file also passed but the same fatal error happened.
```
==================== 59 passed, 2 warnings in 11.53 seconds ====================
[c7949752ef7f:5213] *** An error occurred in MPI_Gather
[c7949752ef7f:5213] *** reported by process [1812398081,0]
[c7949752ef7f:5213] *** on a NULL communicator
[c7949752ef7f:5213] *** Unknown error
[c7949752ef7f:5213] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[c7949752ef7f:5213] ***    and potentially your MPI job)
[c7949752ef7f:5213] [0] func:/usr/local/lib/libopen-pal.so.40(opal_backtrace_buffer+0x20) [0x7f3be0698090]
[c7949752ef7f:5213] [1] func:/usr/local/lib/libmpi.so.40(ompi_mpi_abort+0x258) [0x7f3be0c214f8]
[c7949752ef7f:5213] [2] func:/usr/local/lib/libmpi.so.40(ompi_mpi_errors_are_fatal_comm_handler+0xfd) [0x7f3be0c11f0d]
[c7949752ef7f:5213] [3] func:/usr/local/lib/libmpi.so.40(PMPI_Gather+0x269) [0x7f3be0c425a9]
[c7949752ef7f:5213] [4] func:/usr/local/lib/python2.7/dist-packages/horovod/torch/mpi_lib_v2.so(+0x2f709) [0x7f3be00de709]
[c7949752ef7f:5213] [5] func:/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0xb6990) [0x7f3c77508990]
[c7949752ef7f:5213] [6] func:/lib/x86_64-linux-gnu/libpthread.so.0(+0x8064) [0x7f3c8323c064]
[c7949752ef7f:5213] [7] func:/lib/x86_64-linux-gnu/libc.so.6(clone+0x6d) [0x7f3c8264e62d]
```
We tried to rename test_mxnet to test_apache_mxnet which helped us get around this ordering issue locally because test_apache_mxnet run at the very beginning without interrupting the tests using tensorflow backend. However, it is flaky in Travis CI where we see unit test reordering happens (where test_keras started to run before we finished running all unit tests in test_apache_mxnet).

Running test_mxnet in a separate pytest session is the most stable way we found so far.

horovod/mxnet/tensor_util.cc
apeforest: updated

examples/mxnet_imagenet_resnet50.py
apeforest: updated

examples/mxnet_imagenet_resnet50.py
apeforest: updated

horovod/mxnet/handle_manager.cc
apeforest: updated

.travis.yml
alsrgv: OK, I think we can land this workaround for now until we figure out the root cause of NULL communicator issue - it does not seem to be related to MXNet.

horovod/mxnet/mpi_ops.cc
alsrgv: Does it cover a situation where the data is originally on GPU, and no copy is needed?  NCCL CUDA stream is different from the original CUDA stream, so we still need to synchronize on original stream before NCCL stream starts.

test/test_mxnet.py
```diff
@@ -0,0 +1,324 @@
...
+import unittest
+import numpy as np
+import horovod.mxnet as hvd
+
+
+# Currently, when we build and install Horovod with MXNet pip package, we need
+# to do it on the same OS which MXNet pip package is built. Otherwise, we will
+# hit segmentation fault when running MXNet unit tests, which we suspect is due
+# to different versions of libc library. While we are investigating the root
+# cause and remove this requirement, we will skip unit tests for MXNet
```
alsrgv: We should figure it out before landing, as the majority of users will not be on later versions of OS.

examples/mxnet_imagenet_resnet50.py
```diff
@@ -0,0 +1,325 @@
...
+    # Create initializer and initializer parameters
+    initializer = mx.init.Xavier(rnd_type='gaussian', factor_type="in",
+                                 magnitude=2)
+    mod.bind(data_shapes=train_data.provide_data,
+             label_shapes=train_data.provide_label)
+    mod.init_params(initializer, arg_params=arg_params, aux_params=aux_params)
+
+    # Fetch and broadcast parameters
+    (arg_params, aux_params) = mod.get_params()
+    mx.nd.waitall()
```
alsrgv: Why do we need these waitalls here?  Is it because we don't have ReadyEvent in Horovod adapter?

horovod/mxnet/mpi_ops.cc
```diff
@@ -0,0 +1,335 @@
...
+    // In-place
+  } else {
+    Engine::Get()->PushAsync(allreduce_async_fn, input->ctx(), {},
+                             {output->var()}, FnProperty::kNormal, 0,
+                             "HorovodAllreduce");
+  }
+#endif
+
+  if (average) {
+    *output /= horovod_size();
```
alsrgv: Is it truly async function in a sense that control would be returned before CUDA kernels are done?  If so, would `*output /= horovod_size()` would also be enqueued operation and return immediately?  How do we guarantee that it will execute after allreduce, since allreduce is using NCCL CUDA stream?

In PyTorch, we put synchronous averaging in callback that is passed to `EnqueueTensorAllreduce()`.

horovod/mxnet/mpi_ops.cc
```diff
@@ -0,0 +1,335 @@
...
+      CPU_DEVICE_ID, hvd_cpu_buffer->tensor());
+
+  auto handle = handle_manager.AllocateHandle(cb);
+  auto enqueue_result = EnqueueTensorAllreduce(
+      hvd_context, hvd_cpu_buffer, hvd_cpu_buffer, ready_event,
+      GetOpNameHandle("allreduce", name, handle), CPU_DEVICE_ID,
+      [handle, hvd_cpu_buffer, output](const Status& status) {
+        TensorUtil::CopyCPUToCuda(hvd_cpu_buffer->tensor(), output);
+        handle_manager.MarkDone(handle, status);
+        handle_manager.ExecuteCallback(handle);
```
alsrgv: Do we need this handle manager, or can we just call callback directly, perhaps with little synchronization?  Handle manager was useful in PyTorch since it was exposed to user.

horovod/mxnet/mpi_ops.cc
apeforest: Yes, `ctx.get_strea<gpu>()->Wait()` is also called when copying from GPU to GPU in MXNet. Please see [here](https://github.com/apache/incubator-mxnet/blob/5a83b6b563211f430688e41eab4752c6de4ecf22/src/ndarray/ndarray.cc#L1257)

horovod/mxnet/mpi_ops.cc
alsrgv: Can we just introduce a helper function inside this file that does locking and runs callback, and use it directly w/o need for handle manager?  Would we miss any functionality this way?

.travis.yml
alsrgv: I'm testing https://github.com/uber/horovod/pull/730 which should fix this better.

yuxihu: Great to know the root cause. Had the same feeling that we should run tests independently, at least for each framework.

horovod/mxnet/mpi_ops.cc
ctcyang: Yes, it is truly async in that sense. The `/=` operator calls down to a `PushSync` operation that pushes it to the engine, and the allreduce calls a `PushAsync` operation. Despite the naming, both operations are asynchronous in the control returning sense you are talking about as described here [1]:

> All Push APIs are asynchronous. The API call returns immediately regardless of whether the pushed Fn is finished or not.

We can guarantee that it will execute after allreduce, because the callback will be called after the Horovod MPI call finishes and only then will the engine release the read lock on output.

[1] https://mxnet.incubator.apache.org/architecture/overview.html

examples/mxnet_mnist.py
```diff
@@ -0,0 +1,106 @@
...
+
+    return fc3
+
+
+# Step 4: fit the model
+net = mlp()
+# Softmax with cross entropy loss
+loss = mx.sym.SoftmaxOutput(data=net, name='softmax')
+mlp_model = mx.mod.Module(symbol=loss, context=context)
+optimizer_params = {'learning_rate': 0.01}
```
alsrgv: Scale LR by `hvd.size()`?

apeforest: Updated

examples/mxnet_mnist.py
```diff
@@ -0,0 +1,106 @@
...
+import argparse
+import logging
+import mxnet as mx
+import horovod.mxnet as hvd
+
+hvd.init()
+
+# CLI
+parser = argparse.ArgumentParser(description='MXNet MNIST Example')
+parser.add_argument('--batch-size', type=int, default=64,
```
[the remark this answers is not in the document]
apeforest: Updated. They will be used [here](https://github.com/uber/horovod/pull/542/files#diff-e047a613860988cb493db16e369c15caR87)

horovod/mxnet/mpi_ops.cc
ctcyang: I think what @apeforest is saying is that even when the data is originally on GPU, it will have been the result of _some_ MXNet operation. My understanding is that for practically all MXNet operations using the Push API, it will have a `ctx.get_stream<gpu>()->Wait()` after the current CUDA threads have finished for that stream. Therefore, we can be assured that the data is ready and synchronized on original stream before NCCL starts.

test/test_mxnet.py
yuxihu: We have been investigating this issue for a while but have not figured out the root cause yet.

The MXNet pip package is built on Ubuntu 14.04 with gcc version 4.8.4. If we install MXNet using pip package and then pip install horovod on a OS with gcc version higher than 5 (Ubuntu 16.04: 5.4.0, Ubuntu 18.04: 7.3.0, Debian/stretch: 6.3.0), we hit the segfault while running MXNet unit tests and examples. We suspect that it may be related to different gcc versions (and libstdc++) we used to build MXNet and Horovod on different OSes.

If we build MXNet from source and then install horovod on a newer OS (Ubuntu 16.04, Ubuntu 18.04, Debian/Stretch, Debian/Buster), we do not see any issue running MXNet unit tests and examples (maybe because we compile MXNet and horovod using the same gcc).

We do not want to block this PR while we are investigating this issue. That's why we skip unittests and example on OSes other than jessie in the CI. Another option is to build MXNet from source in the CI but it will take longer which might not be a very good idea. Please suggest.

.travis.yml
```diff
@@ -155,3 +158,9 @@ script:
 
   # run PyTorch MNIST example
   - docker exec ${CONTAINER} /bin/sh -c "${MPIRUN} python /horovod/examples/pytorch_mnist.py"
+
+  # run MXNet MNIST example
+  - |
+    if [[ ${DEBIAN} == "jessie" ]]; then
+      docker exec ${CONTAINER} /bin/sh -c "${MPIRUN} python /horovod/examples/mxnet_mnist.py"
```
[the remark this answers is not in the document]
yuxihu: Same reason for skipping unit tests on OSes other than jessie.

horovod/mxnet/mpi_ops.cc
```diff
@@ -16,64 +16,53 @@
...
 
   auto enqueue_result =
       EnqueueTensorAllreduce(hvd_context, hvd_tensor, hvd_output, nullptr,
-                             GetOpNameHandle("allreduce", name, handle), device,
-                             [handle](const Status& status) {
-                               handle_manager.MarkDone(handle, status);
-                               handle_manager.ExecuteCallback(handle);
+                             name, device,
+                             [cb](const Status& status) {
+                               cb();
```
alsrgv: I remember in prior code you claimed that it's necessary to have a lock around callback - has it been determined to not be the case?

alsrgv: Also, how does errors like shape mismatch get signalled back to user, if we ignore status?

apeforest: The mutex was to protect concurrent access to the unordered_map in handle_manager. shared_ptr was to track the ownership of the callback. We no longer need the lock if we don't return the handle.

yuxihu: The lock was needed to synchronize access the unordered_map where callback was stored in the handle manager. We did not need the lock to invoke the callback. We ran tests with this change and did not see issues.

apeforest: If there is error like shape mismatch, exception will be thrown from MXNet engine [here](https://github.com/uber/horovod/pull/542/files/cf0c7c8c3a6633c32146db4b0f4ee4472726e730..24b3f59bb6a821ee2e9a1c926be1a58bfffc5222#diff-79b78539e15db1d75d73029490eed708L237)

alsrgv: For some reason, your link doesn't work... Could you try linking against [a line in PR](https://github.com/uber/horovod/pull/542/files) directly? Shape mismatch error originates from Horovod, so `status` would have an error here, but it's getting ignored:
```c++
[cb](const Status& status) {
  cb();
});
```

yuxihu:
@alsrgv could you please share the test.py program you used here: gdb -ex r --args python test.py ? I will run with a debug build.

alsrgv:
@yuxihu,
```python
import mxnet as mx
import horovod.mxnet as hvd
hvd.init()
my_var = mx.symbol.Variable('myconst',shape=(),init=mx.init.Constant(value=[1.5]))
hvd.broadcast_(my_var, root_rank=0)
```

apeforest:
@alsrgv Thanks for testing MXNet. The example you wrote was using symbol variable, which is equivalent to placeholder in tensorflow. You need to assign (we call bind) values to it before broadcasting. MXNet also supports imperitive programming using NDArray. Here is a simple example using NDArray:
Please let me know if you have any question.
````
import mxnet as mx
import horovod.mxnet as hvd
hvd.init()
my_var = mx.ndarray.array([1,5])
hvd.broadcast(my_var, root_rank=0)
```

alsrgv:
@apeforest, got it!  Do you have an example with bind() that will work with `hvd.broadcast_`?

setup.py
```diff
@@ -739,6 +839,17 @@ def build_extensions(self):
...
+            except:
+                if not os.environ.get('HOROVOD_WITH_MXNET'):
+                    print('INFO: Unable to build MXNet plugin, will skip it.\n\n'
+                          '%s' % traceback.format_exc(), file=sys.stderr)
+                    built_plugins.append(False)
+                else:
+                    raise
         if not built_plugins:
             raise DistutilsError(
                 'Both TensorFlow and PyTorch plugins were excluded from build. Aborting.')
```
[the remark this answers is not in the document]
apeforest: Added

alsrgv:
OK, I ran your test code with `debian:buster` and got a different segfault:
```
Thread 142 "python" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffebc0c7700 (LWP 6011)]
0x00007fff6cf9ee07 in std::_Function_base::_Base_manager<horovod::mxnet::horovod_mxnet_broadcast_async(horovod::mxnet::NDArray*, horovod::mxnet::NDArray*, int, char*)::<lambda(mxnet::RunContext, horovod::mxnet::Callback)> >::_M_get_pointer (__source=...) at /usr/include/c++/8/bits/std_function.h:295
295	      _M_invoke(const _Any_data& __functor, _ArgTypes&&... __args)
(gdb) bt
#0  0x00007fff6cf9ee07 in std::_Function_base::_Base_manager<horovod::mxnet::horovod_mxnet_broadcast_async(horovod::mxnet::NDArray*, horovod::mxnet::NDArray*, int, char*)::<lambda(mxnet::RunContext, horovod::mxnet::Callback)> >::_M_get_pointer (__source=...) at /usr/include/c++/8/bits/std_function.h:295
#1  std::_Function_handler<void(mxnet::RunContext, mxnet::engine::CallbackOnComplete), horovod::mxnet::horovod_mxnet_broadcast_async(horovod::mxnet::NDArray*, horovod::mxnet::NDArray*, int, char*)::<lambda(mxnet::RunContext, horovod::mxnet::Callback)> >::_M_invoke(const std::_Any_data &, mxnet::RunContext &&, mxnet::engine::CallbackOnComplete &&) (__functor=..., __args#0=..., __args#1=...)
    at /usr/include/c++/8/bits/std_function.h:297
#2  0x00007fff535ef454 in ?? () from /usr/local/lib/python2.7/dist-packages/mxnet/libmxnet.so
#3  0x00007fff535f3371 in ?? () from /usr/local/lib/python2.7/dist-packages/mxnet/libmxnet.so
#4  0x00007fff535efb2b in ?? () from /usr/local/lib/python2.7/dist-packages/mxnet/libmxnet.so
#5  0x00007ffff42d8e20 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff7fb0fa3 in start_thread (arg=<optimized out>) at pthread_create.c:486
#7  0x00007ffff7b3688f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb)
```

Could you run it with debug MXNet (built in `debian:jessie`) to see the full stack?

apeforest:
@alsrgv I have added example for using MXNet Symbol. I also noticed there was a bug in my earlier NDArray example. Please try either of them now. Thanks!

yuxihu:
@alsrgv I ran the first example provided by @apeforest with mxnet debug build (built on Ubuntu 14.04) on debian:jessie (python 2.7)/stretch (python 3.5)/buster (python 3.6) respectively.

It ran through without segfault on debian:jessie.

Got segfault on debian:stretch with the following bt
```
(gdb) bt
#0  0x00007fff33d03b97 in std::_Function_base::_Base_manager<horovod::mxnet::horovod_mxnet_broadcast_async(horovod::mxnet::NDArray*, horovod::mxnet::NDArray*, int, char*)::<lambda(mxnet::RunContext, horovod::mxnet::Callback)> >::_M_get_pointer (__source=...) at /usr/include/c++/6/functional:1553
#1  std::_Function_handler<void(mxnet::RunContext, mxnet::engine::CallbackOnComplete), horovod::mxnet::horovod_mxnet_broadcast_async(horovod::mxnet::NDArray*, horovod::mxnet::NDArray*, int, char*)::<lambda(mxnet::RunContext, horovod::mxnet::Callback)> >::_M_invoke(const std::_Any_data &, <unknown type in /usr/local/lib/python3.5/dist-packages/horovod/mxnet/mpi_lib.cpython-35m-x86_64-linux-gnu.so, CU 0x41aaaf, DIE 0x451046>, <unknown type in /usr/local/lib/python3.5/dist-packages/horovod/mxnet/mpi_lib.cpython-35m-x86_64-linux-gnu.so, CU 0x41aaaf, DIE 0x45109e>) (__functor=...,
    __args#0=<unknown type in /usr/local/lib/python3.5/dist-packages/horovod/mxnet/mpi_lib.cpython-35m-x86_64-linux-gnu.so, CU 0x41aaaf, DIE 0x451046>,
    __args#1=<unknown type in /usr/local/lib/python3.5/dist-packages/horovod/mxnet/mpi_lib.cpython-35m-x86_64-linux-gnu.so, CU 0x41aaaf, DIE 0x45109e>)
    at /usr/include/c++/6/functional:1731
#2  0x00007fff97224fc9 in std::function<void (mxnet::RunContext, mxnet::engine::CallbackOnComplete)>::operator()(mxnet::RunContext, mxnet::engine::CallbackOnComplete) const (this=0x55555674d078, __args#0=..., __args#1=...) at /usr/include/c++/4.8/functional:2471
#3  0x00007fff97231d1f in mxnet::engine::ThreadedEngine::ExecuteOprBlock (this=0x55555674c0b0, run_ctx=..., opr_block=0x55555674f028)
    at src/engine/./threaded_engine.h:364
#4  0x00007fff9723576b in mxnet::engine::ThreadedEnginePerDevice::CPUWorker<(dmlc::ConcurrentQueueType)0> (this=0x55555674c0b0, ctx=...,
    block=0x55555676b500, ready_event=std::shared_ptr (count 2, weak 0) 0x55555713b668) at src/engine/threaded_engine_perdevice.cc:296
#5  0x00007fff97232f47 in mxnet::engine::ThreadedEnginePerDevice::PushToExecute(mxnet::engine::OprBlock*, bool)::{lambda()#1}::operator()() const::{lambda(std::shared_ptr<dmlc::ManualEvent>)#1}::operator()(dmlc::ManualEvent) const (__closure=0x5555571b3dc0,
    ready_event=std::shared_ptr (count 2, weak 0) 0x55555713b668) at src/engine/threaded_engine_perdevice.cc:116
#6  0x00007fff972391f0 in std::_Function_handler<void (std::shared_ptr<dmlc::ManualEvent>), mxnet::engine::ThreadedEnginePerDevice::PushToExecute(mxnet::engine::OprBlock*, bool)::{lambda()#1}::operator()() const::{lambda(std::shared_ptr<dmlc::ManualEvent>)#1}>::_M_invoke(std::_Any_data const&, std::shared_ptr<dmlc::ManualEvent>) (__functor=..., __args#0=std::shared_ptr (empty) 0x0) at /usr/include/c++/4.8/functional:2071
#7  0x00007fff97240d6e in std::function<void (std::shared_ptr<dmlc::ManualEvent>)>::operator()(std::shared_ptr<dmlc::ManualEvent>) const (
    this=0x555557199ea0, __args#0=std::shared_ptr (empty) 0x0) at /usr/include/c++/4.8/functional:2471
#8  0x00007fff97240c08 in std::_Bind_simple<std::function<void (std::shared_ptr<dmlc::ManualEvent>)> (std::shared_ptr<dmlc::ManualEvent>)>::_M_invoke<0ul>(std::_Index_tuple<0ul>) (this=0x555557199e90) at /usr/include/c++/4.8/functional:1732
#9  0x00007fff97240a69 in std::_Bind_simple<std::function<void (std::shared_ptr<dmlc::ManualEvent>)> (std::shared_ptr<dmlc::ManualEvent>)>::operator()() (
    this=0x555557199e90) at /usr/include/c++/4.8/functional:1720
#10 0x00007fff9724096e in std::thread::_Impl<std::_Bind_simple<std::function<void (std::shared_ptr<dmlc::ManualEvent>)> (std::shared_ptr<dmlc::ManualEvent>)> >::_M_run() (this=0x555557199e78) at /usr/include/c++/4.8/thread:115
#11 0x00007ffff30c9200 in std::execute_native_thread_routine_compat (__p=<optimized out>) at ../../../../../src/libstdc++-v3/src/c++11/thread.cc:110
#12 0x00007ffff7bc3494 in start_thread (arg=0x7fff1fcd9700) at pthread_create.c:333
#13 0x00007ffff6db6acf in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:97
```
on debian:buster, we got exactly the same ones:
```
(gdb) bt
#0  0x00007fffc71ee237 in std::_Function_base::_Base_manager<horovod::mxnet::horovod_mxnet_broadcast_async(horovod::mxnet::NDArray*, horovod::mxnet::NDArray*, int, char*)::<lambda(mxnet::RunContext, horovod::mxnet::Callback)> >::_M_get_pointer (__source=...) at /usr/include/c++/8/bits/std_function.h:295
#1  std::_Function_handler<void(mxnet::RunContext, mxnet::engine::CallbackOnComplete), horovod::mxnet::horovod_mxnet_broadcast_async(horovod::mxnet::NDArray*, horovod::mxnet::NDArray*, int, char*)::<lambda(mxnet::RunContext, horovod::mxnet::Callback)> >::_M_invoke(const std::_Any_data &, mxnet::RunContext &&, mxnet::engine::CallbackOnComplete &&) (__functor=..., __args#0=..., __args#1=...) at /usr/include/c++/8/bits/std_function.h:297
#2  0x00007fff98c8bfc9 in std::function<void (mxnet::RunContext, mxnet::engine::CallbackOnComplete)>::operator()(mxnet::RunContext, mxnet::engine::CallbackOnComplete) const (this=0x1592078, __args#0=..., __args#1=...) at /usr/include/c++/4.8/functional:2471
#3  0x00007fff98c98d1f in mxnet::engine::ThreadedEngine::ExecuteOprBlock (this=0x159d9e0, run_ctx=..., opr_block=0x1594028)
    at src/engine/./threaded_engine.h:364
#4  0x00007fff98c9c76b in mxnet::engine::ThreadedEnginePerDevice::CPUWorker<(dmlc::ConcurrentQueueType)0> (this=0x159d9e0, ctx=..., block=0x10dac20,
    ready_event=std::shared_ptr<dmlc::ManualEvent> (use count 2, weak count 0) = {...}) at src/engine/threaded_engine_perdevice.cc:296
#5  0x00007fff98c99f47 in mxnet::engine::ThreadedEnginePerDevice::PushToExecute(mxnet::engine::OprBlock*, bool)::{lambda()#1}::operator()() const::{lambda(std::shared_ptr<dmlc::ManualEvent>)#1}::operator()(dmlc::ManualEvent) const (__closure=0xe94ee0,
    ready_event=std::shared_ptr<dmlc::ManualEvent> (use count 2, weak count 0) = {...}) at src/engine/threaded_engine_perdevice.cc:116
#6  0x00007fff98ca01f0 in std::_Function_handler<void (std::shared_ptr<dmlc::ManualEvent>), mxnet::engine::ThreadedEnginePerDevice::PushToExecute(mxnet::engine::OprBlock*, bool)::{lambda()#1}::operator()() const::{lambda(std::shared_ptr<dmlc::ManualEvent>)#1}>::_M_invoke(std::_Any_data const&, std::shared_ptr<dmlc::ManualEvent>) (__functor=..., __args#0=std::shared_ptr<dmlc::ManualEvent> (empty) = {...}) at /usr/include/c++/4.8/functional:2071
#7  0x00007fff98ca7d6e in std::function<void (std::shared_ptr<dmlc::ManualEvent>)>::operator()(std::shared_ptr<dmlc::ManualEvent>) const (this=0x1050890,
    __args#0=std::shared_ptr<dmlc::ManualEvent> (empty) = {...}) at /usr/include/c++/4.8/functional:2471
#8  0x00007fff98ca7c08 in std::_Bind_simple<std::function<void (std::shared_ptr<dmlc::ManualEvent>)> (std::shared_ptr<dmlc::ManualEvent>)>::_M_invoke<0ul>(std::_Index_tuple<0ul>) (this=0x1050880) at /usr/include/c++/4.8/functional:1732
#9  0x00007fff98ca7a69 in std::_Bind_simple<std::function<void (std::shared_ptr<dmlc::ManualEvent>)> (std::shared_ptr<dmlc::ManualEvent>)>::operator()() (
    this=0x1050880) at /usr/include/c++/4.8/functional:1720
#10 0x00007fff98ca796e in std::thread::_Impl<std::_Bind_simple<std::function<void (std::shared_ptr<dmlc::ManualEvent>)> (std::shared_ptr<dmlc::ManualEvent>)> >::_M_run() (this=0x1050868) at /usr/include/c++/4.8/thread:115
#11 0x00007ffff41e2e20 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#12 0x00007ffff7faffa3 in start_thread (arg=<optimized out>) at pthread_create.c:486
#13 0x00007ffff7af688f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```
This is not broadcast specific. We got segfault at the same place for allreduce/allgather unit tests. For the allreduce case, the segfault seems related to [this line](https://github.com/uber/horovod/pull/542/files#diff-79b78539e15db1d75d73029490eed708R183). If I comment out that line, it does not segfault. Can it be related to name mangling with different gcc versions?

horovod/mxnet/mpi_ops.cc
apeforest: I was referring to [this](https://github.com/uber/horovod/pull/542/files#diff-79b78539e15db1d75d73029490eed708R208) line. However, now I am not sure if I understood your question correctly. Are you suggesting shape mismatch is captured by Horovod and the error status would be written to the `status` variable? Thanks!

apeforest:
@alsrgv I haven't seen a usage pattern like this. I would say it's very unlikely and I don't see a performance benefit doing this way.

alsrgv:
Excellent, let's not worry about it, then!  I'm debugging the error on `debian:buster`.

horovod/mxnet/mpi_ops.cc
```diff
@@ -0,0 +1,304 @@
...
+                             FnProperty::kNormal, 0, "HorovodAllgather");
+    // In-place
+  } else {
+    Engine::Get()->PushAsync(allgather_async_cpu_fn, Context::CPU(0), {},
+                             {output->var()}, FnProperty::kNormal, 0,
+                             "HorovodAllgather");
+  }
+#else
+  if (input->var() != output->var()) {
+    Engine::Get()->PushAsync(allgather_async_fn, Context::CPU(0),
```
alsrgv: Wanted to double check - in case of allreduce you use `input->ctx()`, why is it always hardcoded to CPU here?

apeforest:
Thanks a ton! Building mxnet from source on cpu doesn’t take too long. I will add it to integration test.

yuxihu:
@alsrgv thanks for digging into this and found the root cause. As you pointed out, the _M_invoke interface has been changed from GCC [4.X](https://github.com/gcc-mirror/gcc/blob/gcc-4_8_4-release/libstdc++-v3/include/std/functional#L2069) to GCC [5.X](https://github.com/gcc-mirror/gcc/blob/gcc-5_4_0-release/libstdc++-v3/include/std/functional#L1854). GCC 6/7/8.X has the same interface with GCC 5.X. I built a MXNet pip package (mxnet-horovod) using GCC 5.4.0 and modified the [integration test](https://github.com/uber/horovod/pull/542/commits/2130612fba666ad8f8d784aba7dd98fdd142bc7f#diff-354f30a63fb0907d4ad57269548329e3R117) to use this package on debian:stretch/buster.  It passed.

However, due to different [flavors](https://mxnet.incubator.apache.org/install/) when installing mxnet on Linux, maintainability can be cumbersome if we build MXNet for each flavor using GCC 5.4.0 at this moment. We can use this mxnet-horovod package in the CI but we still instruct users to build from source in the README if they are installing Horovod with MXNet on a OS with GCC 5.X+. Alternatively, we can build MXNet from source on debian:stretch/buster in the integration test, which I am testing in my local container now. Please let us know which approach you prefer.

setup.py
```diff
@@ -739,27 +839,39 @@ def build_extensions(self):
...
+                else:
+                    raise
         if not built_plugins:
             raise DistutilsError(
-                'Both TensorFlow and PyTorch plugins were excluded from build. Aborting.')
+                'TensorFlow, PyTorch and MXNet plugins were excluded from build. Aborting.')
         if not any(built_plugins):
             raise DistutilsError(
-                'Neither TensorFlow nor PyTorch plugins were built. See errors above.')
+                'None of TensorFlow or PyTorch or MXNet plugins were built. See errors above.')
```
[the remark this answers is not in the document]
apeforest: Fixed.

horovod/mxnet/mpi_ops.cc
apeforest: You are right. Updated.

horovod/mxnet/mpi_ops.cc
apeforest: We did save the Status into enqueue_result and signal errors [here](https://github.com/uber/horovod/pull/542/files#diff-79b78539e15db1d75d73029490eed708R60)

apeforest:
@alsrgv We have addressed all your comments and updated PR. Please let us know if it looks good to you now. Thanks a lot for rootcausing issue for us last midnight. Have a good weekend!

yuxihu:
@alsrgv I also tested building MXNet from source in local debian:stretch/buster containers. If you prefer to build from source in the integration test, I can push a commit and trigger the CI. Please let me know. Thanks for your help. Have a good weekend!

horovod/mxnet/mpi_ops.cc
alsrgv: That's insufficient - it will only catch a portion of errors.  It will not catch errors discovered by Horovod coordinator (mismatch of shapes between ranks, etc).  Can you try porting tests from PyTorch tests (e.g. `test_horovod_broadcast_type_error`), and seeing the impact?  We should port all the error condition tests.

alsrgv:
@yuxihu, let's try and see how much it will affect the CI time.

alsrgv:
@yuxihu, looks like it's taking too long.  Let's use your prebuild packages for GCC 4 (jessie) and GCC 5  (stretch/buster)?

yuxihu:
@alsrgv Yes, I think it is more stable to user prebuilt packages. It usually took 5 mins to build from source on an AWS instance with 36 CPUs. Not sure about the hardware type in the CI. And it can be worse if multiple containers run on a single instance. I will revert my previous commit.

alsrgv:
@yuxihu, great!  My biggest concern remains error handling - we need to make sure that errors from coordinator get propagated to users, without segfaults (see the comment above).

Aside from that, we need to make sure that tests run well on GPU and ImageNet example that you built converges to good accuracy - did you verify those two things?

README.md
```diff
@@ -200,6 +201,45 @@ Horovod supports Estimator API and regular TensorFlow in similar ways.
 
 See a full training [example](examples/tensorflow_mnist_estimator.py).
 
+## MXNet
+
+Horovod supports MXNet and regular TensorFlow in similar ways.
+
+See full training [mnist](examples/mxnet_mnist.py) and [imagenet](examples/mxnet_imagenet_resnet50.py) examples.
```
alsrgv: Nit: MNIST and ImageNet

yuxihu: fixed

yuxihu:
@alsrgv The unit tests passed on GPU. We are working on improving the error handling which requires a change in MXNet. We are also training the imagenet example and verify the accuracy.

alsrgv:
@yuxihu, thanks - left one comment in https://github.com/ctcyang/horovod/pull/19

examples/mxnet_imagenet_resnet50.py
```diff
@@ -0,0 +1,337 @@
...
+        arg_params = None
+    aux_params = None
+
+    # Create model
+    mod = mx.mod.Module(softmax, context=context)
+
+    # Create optimizer
+    optimizer_params = {'wd': args.wd,
+                        'momentum': args.momentum,
+                        'rescale_grad': 1.0 / batch_size,
```
alsrgv: What does this rescaling accomplish?  Should LR be decreased to get the same effect?

examples/mxnet_mnist.py
```diff
@@ -0,0 +1,121 @@
...
+parser.add_argument('--lr', type=float, default=0.001,
+                    help='learning rate (default: 0.001)')
+parser.add_argument('--momentum', type=float, default=0.5,
+                    help='SGD momentum (default: 0.5)')
+args = parser.parse_args()
+
+logging.basicConfig(level=logging.INFO)
+logging.info(args)
+
+context = mx.cpu() if args.gpus is None or args.gpus == '0' \
```
alsrgv: Add comment: `# Horovod: pin GPU to local rank.`

examples/mxnet_imagenet_resnet50.py
```diff
@@ -0,0 +1,337 @@
...
+    lr_sched = lr_scheduler.CosineScheduler(
+        args.num_epochs * epoch_size,
+        base_lr=(args.lr * num_workers),
+        warmup_steps=(args.warmup_epochs * epoch_size),
+        warmup_begin_lr=args.warmup_lr
+    )
+else:
+    raise ValueError('Invalid lr mode')
+
+context = mx.cpu() if args.no_cuda else mx.gpu(hvd.local_rank())
```
alsrgv: Add comment: `# Horovod: pin GPU to local rank.`

examples/mxnet_mnist.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+# Step 4: fit the model
+net = mlp()
+# Softmax with cross entropy loss
+loss = mx.sym.SoftmaxOutput(data=net, name='softmax')
+mlp_model = mx.mod.Module(symbol=loss, context=context)
+optimizer_params = {'learning_rate': args.lr * hvd.size()}
+opt = mx.optimizer.create('sgd', sym=net, **optimizer_params)
+opt = hvd.DistributedOptimizer(opt)
```
alsrgv: Add comment: `# Horovod: Wrap optimizer with DistributedOptimizer.`

examples/mxnet_imagenet_resnet50.py
```diff
@@ -0,0 +1,337 @@
...
+
+    # Create optimizer
+    optimizer_params = {'wd': args.wd,
+                        'momentum': args.momentum,
+                        'rescale_grad': 1.0 / batch_size,
+                        'lr_scheduler': lr_sched}
+    if args.dtype == 'float16':
+        optimizer_params['multi_precision'] = True
+    opt = mx.optimizer.create('nag', sym=out, **optimizer_params)
+    opt = hvd.DistributedOptimizer(opt)
```
alsrgv: Add comment: `# Horovod: Wrap optimizer with DistributedOptimizer.`

examples/mxnet_imagenet_resnet50.py
```diff
@@ -0,0 +1,337 @@
...
+
+    # Create initializer and initializer parameters
+    initializer = mx.init.Xavier(rnd_type='gaussian', factor_type="in",
+                                 magnitude=2)
+    mod.bind(data_shapes=train_data.provide_data,
+             label_shapes=train_data.provide_label)
+    mod.init_params(initializer, arg_params=arg_params, aux_params=aux_params)
+
+    # Fetch and broadcast parameters
```
alsrgv: Add `Horovod` to comment since it's Horovod specific.

examples/mxnet_mnist.py
```diff
@@ -0,0 +1,121 @@
...
+
+initializer = mx.init.Xavier(rnd_type='gaussian', factor_type="in",
+                             magnitude=2)
+mlp_model.bind(data_shapes=train_iter.provide_data,
+               label_shapes=train_iter.provide_label)
+mlp_model.init_params(initializer)
+
+# Fetch and broadcast parameters
```
alsrgv: Add `Horovod` to comment since it's Horovod specific.

examples/mxnet_imagenet_resnet50.py
yuxihu: Rescale_grad is used to multiply the gradients before updating. When using MXNet Module API and calling [fit](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/module/base_module.py#L440), you can give a name string or an Optimizer object for the optimizer. If a name string is given, the rescale_grad is normalized to 1.0 / batch_size [intermally](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/module/module.py#L517) and users do not need to specify it in the optimizer_params.  However, in this example, we need to give an Optimizer object to the fit API because we need to create the optimizer first and wrap it with the DistributedOptimizer. In this case, we need to specify this rescale_grad when creating the [optimizer](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/optimizer/optimizer.py#L46). 

Decreasing LR (to LR/batch_size) cannot get the same effect directly. I did one experiment by removing this rescale_grad and decreasing the LR. I got 69% accuracy in 90 epoches while the same training with rescale_grad specified achieved 76% accuracy. And specifying rescale_grad can also keep the same user experience for current MXNet users because they can use the same LR value they used in the Parameter Server approach with Horovod.

examples/mxnet_mnist.py
yuxihu: Done

examples/mxnet_imagenet_resnet50.py
yuxihu: Done

examples/mxnet_imagenet_resnet50.py
yuxihu: Done

examples/mxnet_mnist.py
yuxihu: Done

examples/mxnet_imagenet_resnet50.py
yuxihu: Done

examples/mxnet_mnist.py
yuxihu: Done

examples/mxnet_imagenet_resnet50.py
alsrgv: I see, so `rescale_grad` is not Horovod-specific?

yuxihu: It's not. Even without Horovod, we have the option to create Optimizer first (need to specify rescale_grad) and pass it to the fit API but we usually do not do that.

yuxihu:
@alsrgv is the tensorflow_mnist_eager example flaky? I just triggered the CI twice and it failed in different runs: [1](https://travis-ci.org/uber/horovod/jobs/479243281#L6066), [2](https://travis-ci.org/uber/horovod/jobs/479250204#L6144) and [3](https://travis-ci.org/uber/horovod/jobs/479250203#L6079). All MXNet related tests passed.

examples/mxnet_mnist.py
```diff
@@ -0,0 +1,127 @@
+# Step 1: import required packages
```
alsrgv: Replace with `Step 0` or increase count everywhere?

examples/mxnet_mnist.py
```diff
@@ -0,0 +1,127 @@
...
+logging.basicConfig(level=logging.INFO)
+logging.info(args)
+
+# Horovod: pin GPU to local rank.
+context = mx.cpu() if args.gpus is None or args.gpus == '0' \
+                   else mx.gpu(hvd.local_rank())
+
+# Step 2: data loading
+
+# Download mnist data set using get_mnist_ubyte()
```
alsrgv: Is it possible to have a directory per rank that would eliminate data contention?

Or maybe if the directory is not found you could print a message with a full `python -c "..."` command to run to download MNIST files?

alsrgv:
Looks like https://storage.googleapis.com/tensorflow/tf-keras-datasets/mnist.npz is flaky. Normally, we don't see it.

alsrgv:
Update: I think #756 should fix the eager test.

examples/mxnet_mnist.py
yuxihu: Done

yuxihu:
@alsrgv is this torch [test](https://travis-ci.org/uber/horovod/builds/479735024?utm_source=github_status&utm_medium=notification) flaky? You added it recently. It failed on one run and passed on all others. MXNet tests passed without problem.

alsrgv:
Fixing in #762

yuxihu:
@alsrgv please review this error handling [PR](https://github.com/ctcyang/horovod/pull/24) at your earliest convenience. We tried a few approaches and came up with this PR which had little performance impact.

alsrgv:
@yuxihu, thanks!  I've left a comment in there.

yuxihu:
@alsrgv The dependent PR has been merged into MXNet. We will need today's mxnet nightly build to pass the CI. I will update the package version in the CI once it is available. In the meanwhile, could you please review and let us know if any further suggestions?

alsrgv:
@yuxihu, great!  I like new error handling much better.

@tgaddair, do you want to do a quick pass over as well?  I've seen this code for a while now and may get blindsided.

apeforest:
@alsrgv CI tests have passed. Thanks!

alsrgv:
Merged.  Thanks for the PR, and all the work to make it great!

PR #546, opened 2018-10-06
PR: Enable local gradient accumulation

### LINKED ISSUE #455: PyTorch gradient aggregation + Horovod
In stand-alone PyTorch it is possible to accumulate gradients from multiple batches before applying them by not calling `optimizer.step()`. However in Horovod, when we call `loss.backwards()`, we begin to send the update asynchronously in the background right away, thus not calling `optimizer.step()` does not reduce communication. What is the simplest way to emulate the stand-alone PyTorch method in Horovod?

As a follow up question, since we will be summing gradients rather than averaging them on every worker, would modifying [this line ](https://github.com/uber/horovod/blob/4f0bbbb063023f01af6c56374cf6b5ff58fa2168/horovod/torch/__init__.py#L78) be the only change needed to have Horovod sum gradietns across workers rather than average them when performing all_reduce?

### LINKED ISSUE #543: AssertionError: Local gradient accumulation
Hi,
I'm trying to implement gradient accumulation as described on this [PyTorch discussion thread](https://discuss.pytorch.org/t/how-to-implement-accumulated-gradient/3822/8?u=andfoy) to train a large model by breaking down a larger input batch into smaller batches, accumulating the resulting gradients and then updating the optimizer. 

```python
partial_loss = 0
start_time = time.time()
iter_size = math.ceil(candidates.size(0) / args.num_candidates)
for i in range(0, candidates.size(0), args.num_candidates):
    candidates_x = candidates[i: i + args.num_candidates]
    labels_x = labels[i: i + args.num_candidates]
    candidates_x = candidates_x.to(device)
    labels_x = labels_x.to(device)
    scores = net(img, phrase, candidates_x)
    loss = criterion(scores, labels_x) / iter_size
    partial_loss += loss
    loss.backward()
optimizer.step()
```

While this approach works on a single GPU, it fails when using Horovod, as it displays the following error stacktrace. We've also experienced this error before when trying to train GANs in a distributed manner, also using Horovod.

```python 
bcv005:3290:3535 [0] NCCL INFO NET : Using interface eno1:157.253.243.11<0>
bcv005:3290:3535 [0] NCCL INFO NET/IB : Using interface eno1 for sideband communication
bcv005:3290:3535 [0] NCCL INFO Using internal Network Socket
bcv005:3290:3535 [0] NCCL INFO NET : Using interface eno1:157.253.243.11<0>
bcv005:3290:3535 [0] NCCL INFO NET : Using interface ztbto2ybf2:10.241.99.233<0>
bcv005:3290:3535 [0] NCCL INFO NET : Using interface br-515dfe9b9b35:172.18.0.1<0>
bcv005:3290:3535 [0] NCCL INFO NET : Using interface br-e84ee09ce056:172.19.0.1<0>
bcv005:3290:3535 [0] NCCL INFO NET/Socket : 4 interfaces found
NCCL version 2.3.5+cuda9.0
bcv005:3290:3535 [0] NCCL INFO rank 0 nranks 2
bcv005:3291:3534 [1] NCCL INFO NET : Using interface eno1:157.253.243.11<0>
bcv005:3291:3534 [1] NCCL INFO NET/IB : Using interface eno1 for sideband communication
bcv005:3291:3534 [1] NCCL INFO Using internal Network Socket
bcv005:3291:3534 [1] NCCL INFO rank 1 nranks 2
bcv005:3290:3535 [0] NCCL INFO comm 0x7f5308380cd0 rank 0 nranks 2
bcv005:3290:3535 [0] NCCL INFO Could not find real path of /sys/class/net/ztbto2ybf2/device
bcv005:3290:3535 [0] NCCL INFO Could not find real path of /sys/class/net/br-515dfe9b9b35/device
bcv005:3290:3535 [0] NCCL INFO Could not find real path of /sys/class/net/br-e84ee09ce056/device
bcv005:3290:3535 [0] NCCL INFO CUDA Dev 0, IP Interfaces : eno1(SOC) ztbto2ybf2(SOC) br-515dfe9b9b35(SOC) br-e84ee09ce056(SOC) 
  File "/home/eamargffoy/anaconda3/envs/parallel/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/eamargffoy/anaconda3/envs/parallel/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/media/SSD3/score-textseg/ref_score_net/train.py", line 483, in <module>
    train_loss = train(epoch)
  File "/media/SSD3/score-textseg/ref_score_net/train.py", line 359, in train
    loss.backward()
  File "/home/eamargffoy/anaconda3/envs/parallel/lib/python3.6/site-packages/torch/tensor.py", line 93, in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph)
  File "/home/eamargffoy/anaconda3/envs/parallel/lib/python3.6/site-packages/torch/autograd/__init__.py", line 89, in backward
    allow_unreachable=True)  # allow_unreachable flag
  File "/home/eamargffoy/anaconda3/envs/parallel/lib/python3.6/site-packages/horovod/torch/__init__.py", line 77, in hook
    assert p not in self._handles
AssertionError
bcv005:3291:3534 [1] NCCL INFO comm 0x7fb190371510 rank 1 nranks 2
bcv005:3291:3534 [1] NCCL INFO NET : Using interface eno1:157.253.243.11<0>
bcv005:3291:3534 [1] NCCL INFO NET : Using interface ztbto2ybf2:10.241.99.233<0>
bcv005:3291:3534 [1] NCCL INFO NET : Using interface br-515dfe9b9b35:172.18.0.1<0>
bcv005:3291:3534 [1] NCCL INFO NET : Using interface br-e84ee09ce056:172.19.0.1<0>
bcv005:3291:3534 [1] NCCL INFO NET/Socket : 4 interfaces found
bcv005:3291:3534 [1] NCCL INFO Could not find real path of /sys/class/net/ztbto2ybf2/device
bcv005:3291:3534 [1] NCCL INFO Could not find real path of /sys/class/net/br-515dfe9b9b35/device
bcv005:3291:3534 [1] NCCL INFO Could not find real path of /sys/class/net/br-e84ee09ce056/device
bcv005:3291:3534 [1] NCCL INFO CUDA Dev 1, IP Interfaces : eno1(SOC) ztbto2ybf2(SOC) br-515dfe9b9b35(SOC) br-e84ee09ce056(SOC) 
bcv005:3290:3535 [0] NCCL INFO Using 256 threads
bcv005:3290:3535 [0] NCCL INFO Min Comp Cap 6
bcv005:3290:3535 [0] NCCL INFO Ring 00 :    0   1
bcv005:3290:3535 [0] NCCL INFO Ring 00 : 0[0] -> 1[1] via P2P/IPC
bcv005:3291:3534 [1] NCCL INFO Ring 00 : 1[1] -> 0[0] via P2P/IPC
bcv005:3290:3535 [0] NCCL INFO Launch mode Parallel
  File "/home/eamargffoy/anaconda3/envs/parallel/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/eamargffoy/anaconda3/envs/parallel/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/media/SSD3/score-textseg/ref_score_net/train.py", line 483, in <module>
    train_loss = train(epoch)
  File "/media/SSD3/score-textseg/ref_score_net/train.py", line 359, in train
    loss.backward()
  File "/home/eamargffoy/anaconda3/envs/parallel/lib/python3.6/site-packages/torch/tensor.py", line 93, in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph)
  File "/home/eamargffoy/anaconda3/envs/parallel/lib/python3.6/site-packages/torch/autograd/__init__.py", line 89, in backward
    allow_unreachable=True)  # allow_unreachable flag
  File "/home/eamargffoy/anaconda3/envs/parallel/lib/python3.6/site-packages/horovod/torch/__init__.py", line 77, in hook
    assert p not in self._handles
AssertionError
```

I would like to know if it is possible to perform local gradient accumulation on a single GPU, and then all-reduce it on the other ones.

andfoy:
Fixes #543 
Fixes #455


horovod/torch/__init__.py
```diff
@@ -72,22 +77,38 @@ def _register_hooks(self):
...
+            handle, ctx = None, None
+            if self._reduce_gradients:
+                handle, ctx = self._all_reduce_grad(p)
             self._handles[p] = (handle, ctx)
         return hook
 
     def synchronize(self):
         for p, value in self._handles.items():
             handle, ctx = value
+            if not self._reduce_gradients:
```
alsrgv: It'd be better to check for `handle is None`, since flag may have been flipped after some gradients were submitted.

horovod/torch/__init__.py
```diff
@@ -72,22 +77,38 @@ def _register_hooks(self):
...
+                warnings.warn("Attempting to synchronize an optimizer that "
```
alsrgv: We should mention why it's bad (by default gradients are allreduced asynchronously).  Not sure we should fallback / mention it, since it may create an illusion of "no problem".

horovod/torch/__init__.py
```diff
@@ -63,6 +65,9 @@ def __init__(self, params, named_parameters, compression):
         if size() > 1:
             self._register_hooks()
 
+    def ignore_gradients(self, state):
```
alsrgv: I'll think a bit more about a better name for this method.

andfoy: @alsrgv What about ``disable/enable_gradient_updates``

tgaddair: Thanks for putting this together @andfoy!

This is a very useful capability, but personally, I would rather not add this method to the optimizer API.  I think it forces the user to make a lot of modifications on their end to experiment with local gradient accumulation, when it could be simplified instead to a single parameter to the DistributedOptimizer constructor:

```
opt = hvd.DistributedOptimizer(opt, aggregation_delay=5)
```

We could then manage all of the state internally without having to expose the optimizer API as a state machine, which I think would be more in line with our goals of making Horovod easy to plug into existing single-GPU scripts.  The `step()` function would then be a no-op in the case that we're currently aggregating locally.  

We lose some versatility, but I think it would greatly simplify things and make the feature less intimidating for people who want to try it out without modifying their training script (and less prone to error).

@alsrgv thoughts?

alsrgv: Assuming the `aggregation_delay` is static, this sounds like a neat idea!

andfoy: Thanks for your suggestions, it sounds good, but I have a corner case (Like in our research), what would happen if the  aggregation_delay is dynamic and not static. e.g., In variable-sized batches

tgaddair: Good question.  I was wondering if it would suffice to have a `opt.set_aggregation_delay(delay)` method?  That way advanced users could set it dynamically based on some criteria, and beginning users could just set it via the constructor.  Would that address the corner case you had in mind?

andfoy: @tgaddair, that sounds perfect! So, let me introduce the changes on the PR, and discuss ir further

alsrgv: @andfoy, if the delay is dynamic, I'm wondering about a couple of things:
1. Will the number of synchronizations be the same across all workers?
2. Do you always know the aggregation delay before you start aggregating?

If not, we may still want to expose finer-grained controls.

andfoy: @alsrgv Addressing your two questions:
1. In my specific case, there should be a single synchronization step across all workers once they have finished their individual gradient accumulations.
2. Before starting an accumulation loop, the number of steps is known by each worker.

The general idea would be to synchronize once all workers have finished, which is limited by the worker that has the largest gradient accumulation iteration number. I don't know in which cases, the more general scenario (Unknown number of iterations, different synchronization scheme) would arise, but I think that in that case, it would be preferable to disable/enable gradient synchronization as it is needed.

andfoy:
@alsrgv @tgaddair I've pushed a version of this PR using ``set_aggregation_delay``, but I can go back to ``ignore_gradients`` if you decide to.

horovod/torch/__init__.py
```diff
@@ -72,23 +79,42 @@ def _register_hooks(self):
...
+        handle = allreduce_async_(tensor_compressed, average=True, name=name)
+        return handle, ctx
+
     def _make_hook(self, p):
         def hook(*ignore):
-            assert p not in self._handles
+            if p in self._handles and self._handles[p][0] is not None:
+                if self._parameter_update_delay[p] <= 0:
+                    raise AssertionError(
+                        "Gradients were accumulated twice without a "
```
alsrgv: With other comments in mind, we can say something like `Gradients were computed more than backward_passes_per_step times without call to step().  Please increase backward_passes_per_step to accumulate gradients locally.`

horovod/torch/__init__.py
```diff
@@ -72,23 +79,42 @@ def _register_hooks(self):
...
     def synchronize(self):
         for p, value in self._handles.items():
             handle, ctx = value
+            if handle is None:
+                warnings.warn("Attempting to synchronize an optimizer that "
+                              "is still waiting for more gradients. This may "
+                              "have a performance impact or cause "
+                              "synchronization failures")
+                handle, ctx = self._all_reduce_grad(p)
+                self.ignore_gradients(False)
```
alsrgv: Don't need this.

horovod/torch/__init__.py
alsrgv: With new `backward_passes_per_step`, we can say something like `Called step on optimizer before computing gradients backwards_passes_per_step times.  This will cause performance degradation, as gradient exchange is interleaved with computation only on the backwards_passes_per_step'th pass.`

alsrgv:
@andfoy, by the way - would you want to divide by # of `backward_passes_per_step` before allreduce?

andfoy:
> would you want to divide by # of backward_passes_per_step before allreduce?

@alsrgv are you referring to scale the gradients by ``backward_passes_per_step``

alsrgv:
@andfoy, yes - divide gradients by `backward_passes_per_step`.

alsrgv:
@andfoy, sorry, that was the question. :-)  Would such feature be helpful to you or other use cases you can imagine?

andfoy:
@alsrgv I'm having a strange error on Travis, do you know what is about?

alsrgv:
@andfoy, I just fixed the breaking change from PyTorch 1.0 that has been causing unit tests to fail. 
 If you rebase on `master`, the break should be fixed. 

> Besides, in some cases (Like ours), we would like to scale the gradient by other factors, which is more user-case specific, rather than a single solution for all.

Would you scale gradients before or after averaging them?  Will there be use cases where you'd scale them before averaging?

We had more discussions with @tgaddair and are now thinking about some sort of extensibility model for `DistributedOptimizer` that would allow to decouple functionality like this into a separate class.

andfoy:
> Would you scale gradients before or after averaging them? Will there be use cases where you'd scale them before averaging?

Well, the idea would be to average the loss itself, such that it propagates to the gradients. However, it might be possible that someone prefers to average the gradient values, due to numerical precision or other issues. I don't know how to tackle this on a holistic point-of-view, what are your thoughts @alsrgv  @tgaddair ?

andfoy:
> We had more discussions with @tgaddair and are now thinking about some sort of extensibility model for DistributedOptimizer that would allow to decouple functionality like this into a separate class.

Would you like to introduce this new ``DistributedOptimizer`` wrapper on this PR?

whr94621:
I have a question on the implementation of  ```DistributedOptimizer```. Why we should use ```hook``` to implement the ```allreduce``` operation of gradients? Can we manually do this opetation in ```step``` method, i.e. :
``` python
    def step(self, closure=None):
        # just for example
        # for p in all parameters: 
        #     all_reduce_grad_(p)
        self.synchronize()
        return super(self.__class__, self).step(closure)

```

whr94621:
> @whr94621 We allow users to synchronize their gradients, even if they forgot to perform backward `backward_passes_per_step` times

All right now I see that line of code. But I am still curious about the 'performance degradation' mentioned in warning information.

If we always do ```handle, ctx = self._all_reduce_grad(p)``` in ```step```,  like fix ```backward_passes_per_step``` to a very large value, why it will become slower than implementation in this PR.

andfoy:
@whr94621 Allreduce instructions are sent to MPI distribution nodes long before they update their parameters, so you would expect that by the time ``synchronize`` gets executed, it has already allreduce'd all the gradients across all nodes.

horovod/torch/__init__.py
```diff
@@ -40,7 +41,8 @@
 
 
 class _DistributedOptimizer(torch.optim.Optimizer):
-    def __init__(self, params, named_parameters, compression):
+    def __init__(self, params, named_parameters, compression,
+                 backward_passes_per_step=1):
```
alsrgv: Let's pass this via `def DistributedOptimizer()` as well.

horovod/torch/__init__.py
```diff
@@ -72,23 +76,44 @@ def _register_hooks(self):
                     grad_acc.register_hook(self._make_hook(p))
                     self._grad_accs.append(grad_acc)
 
+    def _all_reduce_grad(self, p):
```
alsrgv: Nit: lets make `s/all_reduce/allreduce/g` for consistency.

horovod/torch/__init__.py
```diff
@@ -72,23 +76,44 @@ def _register_hooks(self):
...
+        tensor_compressed, ctx = self._compression.compress(tensor)
+
+        handle = allreduce_async_(tensor_compressed, average=True, name=name)
+        return handle, ctx
+
     def _make_hook(self, p):
         def hook(*ignore):
-            assert p not in self._handles
+            if p in self._handles and self._handles[p][0] is not None:
+                if self._parameter_update_delay[p] <= 0:
```
alsrgv: This `if self._parameter_update_delay[p] <= 0` should be turned into another assertion.

horovod/torch/__init__.py
```diff
@@ -72,23 +76,44 @@ def _register_hooks(self):
...
+                    raise AssertionError(
+                        "Gradients were computed more than "
+                        "backward_passes_per_step times call to step(). "
```
alsrgv: `s/times call/times before call/`

horovod/torch/__init__.py
```diff
@@ -72,23 +76,44 @@ def _register_hooks(self):
...
+            if self._parameter_update_delay[p] == 0:
+                handle, ctx = self._all_reduce_grad(p)
             self._handles[p] = (handle, ctx)
         return hook
 
     def synchronize(self):
         for p, value in self._handles.items():
             handle, ctx = value
+            if handle is None:
+                warnings.warn("Called step on optimizer before computing "
```
alsrgv: `s/step/step() or synchronize()/`

horovod/torch/__init__.py
```diff
@@ -72,23 +76,44 @@ def _register_hooks(self):
...
+                              "gradients backwards_passes_per_step times. "
+                              "This will cause performance degradation, as "
+                              "gradient exchange is interleaved with "
+                              "computation only on the "
+                              "backwards_passes_per_step'th pass.")
+                handle, ctx = self._all_reduce_grad(p)
             output = synchronize(handle)
+            self._parameter_update_delay[p] = self.backward_passes_per_step
```
alsrgv: `s/_parameter_update_delay/_allreduce_delay/g`

whr94621:
@alsrgv Ok now I understand the usage. Thank you for your explanation!

horovod/torch/__init__.py
```diff
@@ -72,23 +81,45 @@ def _register_hooks(self):
...
+            if self._allreduce_delay[p] == 0:
+                handle, ctx = self._allreduce_grad(p)
             self._handles[p] = (handle, ctx)
         return hook
 
     def synchronize(self):
         for p, value in self._handles.items():
             handle, ctx = value
+            if handle is None:
+                warnings.warn("Called synchronize() on optimizer before "
```
alsrgv: step()/synchronize()  (mention both, most likely users don't call synchronize())

horovod/torch/__init__.py
```diff
@@ -127,12 +160,15 @@ def DistributedOptimizer(optimizer, named_parameters=None, compression=Compressi
         compression: Compression algorithm used during allreduce to reduce the amount
                      of data sent during the each parameter update step.  Defaults to
                      not using compression.
+        backward_passes_per_step: Number of expected backward passes to perform before
+                                  executing allreduce
```
alsrgv: ... before calling step()/synchronize().  This allows accumulating gradients over multiple mini-batches before executing allreduce and updating parameters.

examples/pytorch_imagenet_resnet50.py
```diff
@@ -156,13 +161,24 @@ def train(epoch):
...
+            batch_output = []
+            for i in range(0, data.size(0), sub_batch_size):
+                data_step = data[i:i + sub_batch_size]
+                target_step = target[i:i + sub_batch_size]
+                output = model(data_step)
+                batch_output.append(output)
+                loss = F.cross_entropy(output, target_step)
+                batch_loss += loss / args.accum_grad
+                loss.backward()
+            # Gradient is averaged by args.backward_steps by the optimizer
```
alsrgv: It's not really averaged by `args.backwards_steps`

examples/pytorch_imagenet_resnet50.py
```diff
@@ -156,13 +161,24 @@ def train(epoch):
...
             optimizer.step()
 
-            train_loss.update(loss)
-            train_accuracy.update(accuracy(output, target))
+            batch_output = torch.cat(batch_output, dim=0)
+            train_loss.update(batch_loss.item())
+            train_accuracy.update(accuracy(batch_output, target))
```
alsrgv: Can push up accuracy computation into the loop as well, then no need to collect outputs and concat the tensor.

examples/pytorch_imagenet_resnet50.py
```diff
@@ -156,13 +161,24 @@ def train(epoch):
...
+            sub_batch_size = int(data.size(0) // args.backward_steps)
+            batch_loss = 0
+            batch_output = []
+            for i in range(0, data.size(0), sub_batch_size):
+                data_step = data[i:i + sub_batch_size]
+                target_step = target[i:i + sub_batch_size]
+                output = model(data_step)
+                batch_output.append(output)
+                loss = F.cross_entropy(output, target_step)
+                batch_loss += loss / args.accum_grad
```
alsrgv: Should be `backwards_steps`?

examples/pytorch_imagenet_resnet50.py
```diff
@@ -156,13 +161,24 @@ def train(epoch):
             if args.cuda:
                 data, target = data.cuda(), target.cuda()
             optimizer.zero_grad()
-            output = model(data)
-            loss = F.cross_entropy(output, target)
-            loss.backward()
+            # Split batch into sub-batches and sum gradients for all samples
+            sub_batch_size = int(data.size(0) // args.backward_steps)
```
alsrgv: Can we just use `args.batch_size`?

examples/pytorch_imagenet_resnet50.py
```diff
@@ -40,6 +41,9 @@
                     help='SGD momentum')
 parser.add_argument('--wd', type=float, default=0.00005,
                     help='weight decay')
+parser.add_argument('--backward-steps', type=float, default=1,
```
alsrgv: Rename to `backward-passes-per-step` to teach users the parameter name?

examples/pytorch_imagenet_resnet50.py
andfoy: You're right! Haha

examples/pytorch_imagenet_resnet50.py
```diff
@@ -169,7 +169,8 @@ def train(epoch):
             loss.backward()
             batch_loss += loss
             train_accuracy.update(accuracy(output, target))
-            if (batch_idx + 1) % args.backward_passes_per_step == 0:
+            if ((batch_idx + 1) % args.backward_passes_per_step == 0 or
+                (batch_idx + 1) >= len(train_loader)):
```
andfoy: @alsrgv Shall we restrict users to set-up ``backward_passes_per_step`` to be a factor of the total number of training examples? If the value is not set correctly, the ``step()`` warning would appear.

alsrgv: Good idea!

alsrgv:
@andfoy, I think a version with an inner loop of sub-batches was easier to understand, can we get back to it?

examples/pytorch_imagenet_resnet50.py
```diff
@@ -156,13 +161,25 @@ def train(epoch):
             if args.cuda:
                 data, target = data.cuda(), target.cuda()
             optimizer.zero_grad()
-            output = model(data)
-            loss = F.cross_entropy(output, target)
-            loss.backward()
+            # Split batch into sub-batches and sum gradients for all samples
+            # Note: the performance may degrade if backward_passes_per_step is
```
alsrgv: We could just make it an error that throws a `ValueError` exception.  Let's make an explicit check at line ~55, and introduce `sub_batch_size` (or `backward_pass_batch_size` for clarity) at that point.

examples/pytorch_imagenet_resnet50.py
```diff
@@ -156,13 +161,25 @@ def train(epoch):
...
+                # Average gradients by the total number of sub-batches
+                loss = loss / args.backward_passes_per_step
+                batch_loss += loss
+                loss.backward()
+            # Gradient is updated across all ranks
             optimizer.step()
-
-            train_loss.update(loss)
-            train_accuracy.update(accuracy(output, target))
+            train_loss.update(batch_loss.item())
```
alsrgv: We can move this train_loss update into backward pass loop as well.

alsrgv:
@andfoy, actually, I think we need to re-think the positioning a little bit.  As far as I understand, BN parameters will be updated every forward pass in the training mode.  So we should change wording to say that `step()` applies gradients (instead of updates parameters), replace `--backward-passes-per-batch` with `--batches-per-allreduce` which run multiple full batches.  In help for `--batches-per-allreduce` we should mention that it multiplies total batch size.

andfoy:
@alsrgv Shall we also increase ``val_batch_size``?

alsrgv:
@andfoy, we only allreduce metrics for validation, so it does not make sense.

examples/pytorch_imagenet_resnet50.py
```diff
@@ -49,6 +54,8 @@
 args = parser.parse_args()
 args.cuda = not args.no_cuda and torch.cuda.is_available()
 
+actual_batch_size = args.batch_size * args.batches_per_allreduce
```
alsrgv: `allreduce_batch_size`

examples/pytorch_imagenet_resnet50.py
```diff
@@ -127,7 +135,8 @@
 # Horovod: wrap optimizer with DistributedOptimizer.
 optimizer = hvd.DistributedOptimizer(optimizer,
                                      named_parameters=model.named_parameters(),
-                                     compression=compression)
+                                     compression=compression,
+                                     batches_per_allreduce=args.backward_steps)
```
alsrgv: This line definitely won't work :-)

examples/pytorch_imagenet_resnet50.py
```diff
@@ -156,13 +165,20 @@ def train(epoch):
...
+            # of size batch_size
+            for i in range(0, actual_batch_size, args.batch_size):
+                data_step = data[i:i + args.batch_size]
+                target_step = target[i:i + args.batch_size]
+                output = model(data_step)
+                train_accuracy.update(accuracy(output, target_step))
+                loss = F.cross_entropy(output, target_step)
+                # Average gradients by the total number of sub-batches
+                loss = loss / args.batches_per_allreduce
+                train_loss.update(loss.item())
```
alsrgv: Since we're now always updating loss (which averages instead of summing), we should not need this division by # of batches?

examples/pytorch_imagenet_resnet50.py
```diff
@@ -156,13 +165,20 @@ def train(epoch):
...
+                loss.backward()
+            # Gradient is updated across all ranks
```
alsrgv: s/updated/applied/

horovod/torch/__init__.py
```diff
@@ -127,12 +161,18 @@ def DistributedOptimizer(optimizer, named_parameters=None, compression=Compressi
         compression: Compression algorithm used during allreduce to reduce the amount
                      of data sent during the each parameter update step.  Defaults to
                      not using compression.
+        backward_passes_per_step: Number of expected backward passes to perform
+                                  before calling step()/synchronize(). This
+                                  allows accumulating gradients over multiple
+                                  mini-batches before executing allreduce and
+                                  updating parameters.
```
alsrgv: `s/executing allreduce and updating parameters/averaging and applying them/`?

examples/pytorch_imagenet_resnet50.py
```diff
@@ -40,6 +41,10 @@
                     help='SGD momentum')
 parser.add_argument('--wd', type=float, default=0.00005,
                     help='weight decay')
+parser.add_argument('--batches-per-allreduce', type=float, default=1,
+                    help='number of backward iteration steps '
+                         'executed before applying gradients. '
```
alsrgv: nit: `s/. It/; it/`

examples/pytorch_imagenet_resnet50.py
andfoy: Actually, at the end, stats should be averaged by `allreduce_batch_size``, so I think we should multiply instead by ``batch_size`` and then divide by ``allreduce_batch_size``

alsrgv:
@andfoy, I think your idea to rescale loss is actually correct (if accompanied by raising learning rate).  The reason is - it's not equivalent to multiply gradient and to multiply learning rate.  Operations like weight decay would produce different results: https://github.com/pytorch/pytorch/blob/master/torch/optim/sgd.py#L93

If my napkin math is correct, rescaling loss and raising learning rate would actually be equivalent to having more independent GPUs.

examples/pytorch_imagenet_resnet50.py
```diff
@@ -173,6 +176,7 @@ def train(epoch):
                 output = model(data_batch)
                 train_accuracy.update(accuracy(output, target_batch))
                 loss = F.cross_entropy(output, target_batch)
+                loss.div_(allreduce_batch_size).mul_(args.batch_size)
```
alsrgv: Why not just `.div_(args.batches_per_allreduce)`?  Saves one operation.

alsrgv: Let's add a comment saying that this re-scaling averages gradients among sub-batches.

examples/pytorch_imagenet_resnet50.py
```diff
@@ -40,6 +41,10 @@
                     help='SGD momentum')
 parser.add_argument('--wd', type=float, default=0.00005,
                     help='weight decay')
+parser.add_argument('--batches-per-allreduce', type=int, default=1,
```
alsrgv: nit: let's bring this up after `--fp16-allreduce`, since this help section is for model parameters, while `--batches-per-allreduce` is more of performance optimization parameter.

examples/pytorch_imagenet_resnet50.py
```diff
@@ -40,6 +41,10 @@
...
+                    help='number of backward iteration steps '
```
alsrgv: Mayber rephrase to `number of batches processed locally before executing allreduce across workers; it multiplies total batch size.`

andfoy:
@alsrgv Done!

horovod/torch/__init__.py
```diff
@@ -72,23 +81,46 @@ def _register_hooks(self):
...
+            if self._allreduce_delay[p] == 0:
+                handle, ctx = self._allreduce_grad(p)
             self._handles[p] = (handle, ctx)
         return hook
 
     def synchronize(self):
         for p, value in self._handles.items():
             handle, ctx = value
+            if handle is None:
+                warnings.warn("Called step()/synchronize() on optimizer "
```
alsrgv: Let's remove this warning (and `import warnings`) since this will routinely happen due to DistributedSampler.

examples/pytorch_imagenet_resnet50.py
```diff
@@ -156,13 +168,20 @@ def train(epoch):
             if args.cuda:
                 data, target = data.cuda(), target.cuda()
             optimizer.zero_grad()
-            output = model(data)
-            loss = F.cross_entropy(output, target)
-            loss.backward()
+            # Split batch_size * batches_per_allreduce batch into sub-batches
+            # of size batch_size
+            for i in range(0, allreduce_batch_size, args.batch_size):
```
alsrgv: Turns out DistributedSampler just makes sure that every worker has the same # of batches, but does not guarantee that it's divisible by the batch size.

So, we should do something like this:
```
# Split `data` into batches of size `batch_size`.
for i in range(0, len(data), args.batch_size):
    ...
```

examples/pytorch_imagenet_resnet50.py
```diff
@@ -156,13 +168,20 @@ def train(epoch):
...
+                data_batch = data[i:i + args.batch_size]
+                target_batch = target[i:i + args.batch_size]
+                output = model(data_batch)
+                train_accuracy.update(accuracy(output, target_batch))
+                loss = F.cross_entropy(output, target_batch)
+                # Average gradients among sub-batches
```
alsrgv: Train loss should be updated before re-scaling loss, since it does its own averaging.  Also, we need to scale by proper number of batches.

```
train_loss.update(loss)
# Average gradients among batches in this data slice.
loss.div_(math.ceil(float(len(data)) / args.batch_size)
```

horovod/torch/__init__.py
```diff
@@ -72,23 +81,46 @@ def _register_hooks(self):
...
+                              "before computing gradients "
+                              "backwards_passes_per_step times. "
+                              "This will cause performance "
+                              "degradation, as gradient exchange is "
+                              "interleaved with computation only on the "
+                              "backwards_passes_per_step'th pass.")
+                handle, ctx = self._allreduce_grad(p)
```
alsrgv: We need to do two passes.  In a first pass, we check all the handles and backfill missing allreduces.  In a second pass, we synchronize all the handles.  If we don't do that, we can get a race and hang forever.

andfoy:
@alsrgv Give it a try!

andfoy:
@alsgrv Any updates on this one?

alsrgv:
Bad news.  Both current version (with fixes that I mentioned earlier) and the original idea of not adjusting gradients don't converge correctly.

Training curves:
screen shot 2018-10-24 at 8 19 00 pm

Validation curves:
screen shot 2018-10-24 at 8 18 43 pm

Blue is the curve of Horovod 0.15.1 in FP16 mode run over 64 GPUs.  Orange is this PR run over 16 GPUs with batches-per-allreduce=4.  Gray is an older version of this PR that did not adjust gradients.

Any ideas why it doesn't converge as well?

andfoy:
> Any ideas why it doesn't converge as well?

Maybe we're not rescaling the gradients correctly? Or is the lr wrong? Shall we try without rescaling the lr? Let me check the maths regarding gradient accumulation and I'll get back to this

alsrgv:
@andfoy, did you get any insights?

andfoy:
@alsrgv According to the maths, we should divide the loss only by ``args.batches_per_allreduce
``

andfoy:
We need to update the gradients as it follows, where M are the total number of GPUs, N corresponds to the gradient accumulation steps, and B is the batch size

imagen

alsrgv:
@andfoy, sorry for the delay.  Isn't that what the current version is doing already?

andfoy:
@alsrgv Don't worry, right now I'm very busy with the CVPR deadline. I'll get back to this after Friday

andfoy:
@alsrgv What happens if we don't scale the learning by the total number of accumulations?

andfoy:
@alsrgv Any new updates/ideas on this one?

alsrgv:
@andfoy, I found the problem.  In learning rate warmup adjustment we did not multiply LR by `batches_per_allreduce`.  Could you rebase your branch, resolve conflicts and make this fix?  I'm running the convergence test, so far it matches baseline perfectly.

PR #563, opened 2018-10-16
FP16 allreduce on CPU

alsrgv:
Implements custom MPI operation to provide CPU-FP16 allreduce. Uses FP16<->FP32 conversion for the actual computation.


alsrgv:
Note to self: need to fix tests since PyTorch does not support any math on FP16 CPU tensors.

PR #569, opened 2018-10-17
PyTorch master compatibility, improved errors in case of .so import failure

alsrgv:
See pytorch/pytorch#12687 & pytorch/pytorch#12504

### DIFF
--- a/horovod/torch/mpi_ops.py
+++ b/horovod/torch/mpi_ops.py
@@ -22,7 +22,9 @@ from distutils.version import LooseVersion
 # Load all the necessary PyTorch C types.
 import torch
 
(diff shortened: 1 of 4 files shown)


horovod/torch/mpi_ops.py
```diff
@@ -22,19 +22,20 @@
 # Load all the necessary PyTorch C types.
 import torch
 
-from horovod.torch.compression import Compression
-try:
+if LooseVersion(torch.__version__) >= LooseVersion('0.4.2'):
```
tgaddair: I believe we concluded that because there's no version 0.4.2, we should compare against 1.0.0 instead (see test_torch.py).

alsrgv: Good point!  Updated.

PR #570, opened 2018-10-18
Added option to treat sparse gradients as dense tensors

### LINKED ISSUE #430: question: why nccl allgather is not supported?
In horovod implementation, only MPI allgather is supported. Why isn't nccl allgather supported? any specific reasons? Thanks.

### LINKED ISSUE #554: Segmentation fault when scaling up the nodes
Description:
Horovod crashes when I'm trying to scale up over 50 nodes. The model I'm trying to run horovod with is tensorflow/models/official/transformer/ . 

I've attached the gdb output of the coredump. 
[debug.txt](https://github.com/uber/horovod/files/2466318/debug.txt)

I'm running it on cpu's.
'num_cores': 40, 'cpu_info': 'Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz'

Environment:
horovod==0.15.0
tensorflow==1.10.1
Python 3.6.5 
(Open MPI) 3.1.2

mpirun command:

HOROVOD_FUSION_THRESHOLD=134217728 \\
mpirun -x HOROVOD_FUSION_THRESHOLD \\
$CODE_DIR/my_scripts/debug_run.sh

### LINKED ISSUE #566: Severely limited performance. Error: errno 38
I see almost no performance gain running tensorflow_mnist.py on 1 GPU vs 2 GPU using Horovod (logs for both runs attached) on a single node. I suspect, maybe there is some configuration issue.

I see the following message repeated hundreds of times when running on more that 1 GPU (please see the log) :

[gcrgpu1172:04092] Read -1, expected 102400, errno = 38
 
For some of my own models, I see this line repeated constantly through out the training. Can you provide suggestions on what might be the issue here? 

I am running horovod using the docker image built using the dockfile provided through this repo ( https://raw.githubusercontent.com/uber/horovod/master/Dockerfile) . My host OS is CoreOS which does not support nvidia-docker, so I am using the following command to run the image file

sudo docker run --privileged --cap-add=SYS_PTRACE --net=host --name=$container_name --device=/dev/nvidia0 --device=/dev/nvidia1 --device=/dev/nvidia2 --device=/dev/nvidia3 --device=/dev/nvidia4 --device=/dev/nvidia5 --device=/dev/nvidia6 --device=/dev/nvidia7 --device=/dev/nvidiactl --device=/dev/nvidia-uvm -v /var/drivers/nvidia/current:/usr/local/nvidia:ro -v /home/core/samyamr:/home/samyamr -v /tmp/hdfs-mount-dummy/:/hdfs/ -e PHILLY_CLUSTER=gcr -d -it horovod-python35:latest

[1-GPU-log-tensorflow-mnist.txt](https://github.com/uber/horovod/files/2485454/1-GPU-log-tensorflow-mnist.txt)
[2-GPU-log-tensorflow-mnist.txt](https://github.com/uber/horovod/files/2485455/2-GPU-log-tensorflow-mnist.txt)

alsrgv:
As shown in #554, #566, #430, and potentially other issues, sometimes sparse gradients are not so sparse.  This can cause bad performance and OOMs.

This change introduces option `hvd.DistributedOptimizer(..., sparse_as_dense=False)` that, if set to true, will force all sparse gradients to be densified.

Fixes #554 
Fixes #566 
Fixes #430

### DIFF
--- a/horovod/tensorflow/__init__.py
+++ b/horovod/tensorflow/__init__.py
@@ -137,7 +137,8 @@ class DistributedOptimizer(tf.train.Optimizer):
     average gradient values before applying gradients to model weights."""
 
     def __init__(self, optimizer, name=None, use_locking=False, device_dense='',
-                 device_sparse='', compression=Compression.none):
+                 device_sparse='', compression=Compression.none,
+                 sparse_as_dense=False):
         """Construct a new DistributedOptimizer, which uses another optimizer
         under the hood for computing single-process gradient values and
         applying gradient updates after the gradient values have been averaged
@@ -163,6 +164,10 @@ class DistributedOptimizer(tf.train.Optimizer):
             Compression algorithm used during allreduce to reduce the amount
             of data sent during the each parameter update step.  Defaults to
             not using compression.
+          sparse_as_dense:
+            Treat all sparse gradients as dense tensors.  This can help improve
+            performance and memory utilization if the original sparse gradient
+            has high density.  Defaults to false.
         """
         if name is None:
             name = "Distributed{}".format(type(optimizer).__name__)
@@ -171,6 +176,7 @@ class DistributedOptimizer(tf.train.Optimizer):
         self._device_dense = device_dense
         self._device_sparse = device_sparse
         self._compression = compression
+        self._sparse_as_dense = sparse_as_dense
         super(DistributedOptimizer, self).__init__(
             name=name, use_locking=use_locking)
 
@@ -188,6 +194,9 @@ class DistributedOptimizer(tf.train.Optimizer):
             with tf.name_scope(self._name + "_Allreduce"):
                 for grad, var in gradients:
                     if grad is not None:
+                        if self._sparse_as_dense and \
+                                isinstance(grad, tf.IndexedSlices):
+                            grad = tf.convert_to_tensor(grad)
                         avg_grad = allreduce(grad,
                                              device_dense=self._device_dense,
                                              device_sparse=self._device_sparse,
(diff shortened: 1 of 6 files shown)


alsrgv:
Note to self: after usability is confirmed, add tests.

horovod/tensorflow/__init__.py
```diff
@@ -188,6 +194,9 @@ def compute_gradients(self, *args, **kwargs):
             with tf.name_scope(self._name + "_Allreduce"):
                 for grad, var in gradients:
                     if grad is not None:
+                        if self._sparse_as_dense and \
```
tgaddair: Would it make more sense to add this parameter to `allreduce` and do the casting there?  It would make the API more versatile, and reduce duplication across TensorFlow and Keras DistributedOptimizer implementations.

alsrgv: I just feel weird about `hvd.allreduce()` turning sparse tensor into dense tensor, but I guess explicit `sparse_as_dense` flag would make it ok?

alsrgv: Actually, the only thing it'd safe is these two lines:
```
if self._sparse_as_dense and isinstance(grad, tf.IndexedSlices):
   grad = tf.convert_to_tensor(grad)
```

We'd still need to have optimizers take `sparse_as_dense` flag, plumb it through and have docs.  So, until somebody asks for it, I'm still leaning towards making it optimizer-specific thing.

tgaddair: I think that's fine.  I agree that it feels weird having `allreduce` turn a sparse tensor dense, but in fairness I also think it's weird that it will call `allgather` if the input is a `tf.IndexedSlices`.  But I agree, for now it's probably fine just to put it in the optimizer.

PR #573, opened 2018-10-19
MLSL integration

SmorkalovME:
This PR adds integration with Intel(R) MLSL library. Please note we've added it as a separate file as we expect the requirements implied by MLSL on background thread to keep changing with further MLSL-related pull requests.
Also adding the link to the [paper ](https://cug.org/proceedings/cug2018_proceedings/includes/files/pap146s2-file1.pdf  )covering initial performance results.


shirosankaku:
Hello!

@alsrgv , could you, please, review our PR?

wei-v-wang:
Hi Alex @alsrgv, could you please help review this PR from Intel? We appreciate your horovod work and have optimized it for Intel architectures via this PR. Thank you for your time in advance!

alsrgv:
@shirosankaku, @wei-v-wang, sorry - just saw your requests.  I was thinking this is still in development.  My main meta-comment is about duplication between `operations.cc` and `mlsl_operations.cc`.  Is it possible to merge them into one?

SmorkalovME:
Hi @alsrgv. Thanks for your feedback! The primary reason we have moved MLSL backend into a separate file is to accommodate different requirements for background thread MLSL implies (and we expect those to get relaxed further). Would it make sense to abstract a background thread in Horovod architecture to allow for different implementations of that depending on what is sufficient for certain communication backend? This way we could avoid duplication of the common logic, while still allow for different coordination techniques.

alsrgv:
@SmorkalovME, we're planning to refactor `operations.cc` to allow this sort of logic in the next couple of weeks.  We should factor in changes to support your use case.  Does your approach work with the vanilla `BackgroundThread` and a custom `PerformOperation`?

SmorkalovME:
@alsrgv Thanks for your suggestion! Even w/ customizable `PerformOperation` this PR will require changes in bg thread body as well, but let us estimate the amount of those. We will get back w/ more details soon.

horovod/common/operations.cc
```diff
@@ -2245,18 +2714,24 @@ int horovod_size() {
 }
 
 int horovod_local_size() {
+#if HAVE_MLSL
+  return horovod_size();
```
alsrgv: `local_size` = a number of processes that run on the same node or container / can use shared memory to communicate.  What is the way you typically run Horovod with MLSL, is it a single process per node or multiple processes per node?

horovod/common/operations.cc
```diff
@@ -75,6 +108,63 @@ namespace horovod {
 namespace common {
 
 namespace {
+    
+#if HAVE_MLSL
+#define COMMUNICATOR_TABLE_MAX 1000
+#define NULL_VAL nullptr
```
alsrgv: Let's just use `nullptr` literal everywhere instead of a macro.

horovod/common/operations.cc
```diff
@@ -75,6 +108,63 @@ namespace horovod {
...
+
+#define ERROR 0
+#define INFO  1
+#define DEBUG 2
+#define TRACE 3
+
+#define GET_TID() syscall(SYS_gettid)
+
+#define MLSL_LOG(log_lvl, fmt, ...)					\
```
alsrgv: There's another PR that adds logging - #671, we should switch to use it once it's merged.

horovod/common/operations.cc
```diff
@@ -267,6 +389,141 @@ const Status DUPLICATE_NAME_ERROR = Status::InvalidArgument(
...
+            tensor_table[i].comm = NULL_VAL;
+            MLSL_LOG(DEBUG, "NOW collective operation done %s %p %p %d %p %p %ld\n",
+                     tensor_table[i].tensor_name.c_str(), tensor_table[i].tensor->data(),
+                     tensor_table[i].comm, tensor_table[i].tensor->dtype(), tensor_table[i].request,
+                     NULL_VAL, tensor_table[i].tensor->size());
+        }
+        else if (tensor_table[i].request != NULL_VAL && tensor_table[i].comm == NULL_VAL)
+        {
+            bool testflag = 0;
+            MLSL::Environment::GetEnv().Test(tensor_table[i].request, &testflag);
```
alsrgv: What does this test mean, and what if it fails?

horovod/common/operations.cc
```diff
@@ -75,6 +108,63 @@ namespace horovod {
...
+{
+   time_t timer;
+   struct tm* timeInfo = 0;
+   time(&timer);
+   timeInfo = localtime(&timer);
+   assert(timeInfo);
+   strftime(buf, bufSize, "%Y:%m:%d %H:%M:%S", timeInfo);
+}
+
+enum HvdRequestType
```
alsrgv: We already have request type in `MPIRequest::RequestType`, can we use that?

horovod/common/operations.cc
```diff
@@ -267,6 +389,141 @@ const Status DUPLICATE_NAME_ERROR = Status::InvalidArgument(
...
+                    delete tensor_table[i].size_vec;
+                    tensor_table[i].size_vec = NULL;
+                    tensor_table[i].comm = NULL_VAL;
+                    MLSL_LOG(DEBUG, "Test success for intermediate AllGather call %s %p\n",
+                             tensor_table[i].tensor_name.c_str(), tensor_table[i].tensor->data());
+                }
+                else
+                {
+                    // Remove request from request_table and execute tensor callback.
+                    tensor_table[i].callback(Status::OK());
```
alsrgv: Should it be a failure?

test/test_tensorflow.py
```diff
@@ -253,6 +261,11 @@ def test_horovod_allreduce_error(self):
         rank = hvd.rank()
         size = hvd.size()
 
+        # There is no checking if tensor sizes from different ranks are equal in MLSL.
```
alsrgv: Why not use existing mechanism we have in Horovod to ensure shape compatibility?  This way, UX is the same across backends.

horovod/common/operations.cc
XiaotaoChen: @SmorkalovME  Can you explain where the `callback` function is executed or why doesn't it need to be executed in the case of: `tensor_table[i].request == NULL_VAL && tensor_table[i].comm != NULL_VAL`

XiaotaoChen: @SmorkalovME By the way, Can you show the performance improvement between `horovod allreduce/broadcast/allgather` and `mlsl allreduce/bcast/allgather` ?  thanks.

horovod/common/operations.cc
shirosankaku: Done

horovod/common/operations.cc
shirosankaku: @alsrgv Yes, we can. Done

shirosankaku:
I am sorry for bothering, but does anybody know why my CLA agreement signing is still pending? I filled up the form with my data, and I can see them when clicking on "CLA | not signed yet" along with the message "You have signed the CLA for uber/horovod", but the form is not editable and the status is still pending.

horovod/common/operations.cc
shirosankaku: @XiaotaoChen Because here we work with non-blocking collective operations. In this piece of code tensor_table[i].request == NULL_VAL && tensor_table[i].comm != NULL_VAL we start collective operations. Here tensor_table[i].request != NULL_VAL && tensor_table[i].comm == NULL_VAL we check whether the operation was terminated. If it was, then, in case of the called operation being ALLGATHER of the first stage, we call another non-blocking ALLGATHER (the second stage) and wait for its completion, but in other cases we set op_completed = true and call callback function

horovod/common/operations.cc
shirosankaku: @alsrgv I am sorry, but could you please, clarify a little bit, why do you think it should be a failure?

horovod/common/operations.cc
XiaotaoChen: @shirosankaku Thanks for your detailed explanation. got that.  So the main difference is that `mlsl` work with non-blocking collective operations, but the default implementation work with blocking mode. Non-blocking mode can make better use of bandwidth theoretically. Can you tell the performance improvement between the two implementations in some concrete cases.

test/test_tensorflow.py
shirosankaku: @alsrgv Thank you for your comment. Unfortunately, we cannot use `ConstructMPIResponse` function as is because in MLSL implementation we don't have and don't need the data types used in this function and would like to avoid extra data copies. The second point is that in our implementation coordinator doesn't know anything about the tensors of other processes (one sided communication). All the other processes can compare their tensors with the one of coordinator, but the decision about error will be made in a decentralized way that can lead to the situation when coordinator is still working with troubled tensor but others are waiting it to go to Finalize phase. We just wouldn't like to add some extra communications between coordinator and the other processes. MPI will report about different sizes and types anyway. And if such kind of situation exist then something should be changed in the way which data collective operations are called with. I unserstand your desire to keep UX the same, but the question is, how important it is to implement error reporting feature inside our piece of code?

horovod/common/operations.cc
shirosankaku: @alsrgv We can change our logging system to the one from the PR you referred to anytime (after it will be merged). Does this mean that our PR has to wait until [https://github.com/uber/horovod/pull/671](url) is merged (as far as I can see there are still some unsuccessful checks)?

alsrgv:
@shirosankaku, `git log` shows me that currently your commits were made with email `yshchyok@nnlmpihsw05.inn.intel.com` which doesn't match your GitHub account.  You can either use your regular email, or you can use `shirosankaku@users.noreply.github.com`.

horovod/common/operations.cc
alsrgv: Let's rename all these testflags to `is_completed` (or something along those lines) for clarity.

test/test_tensorflow.py
alsrgv: I'd like to understand this problem deeper.  We should not diverge from approach used by original coordinator implementation to keep code base simple, unless there's potential for significant performance improvements - in which case we may in fact want to implementation of the original code base.  Sent you email to discuss over V/C.  Let's keep this open to summarize our discussion here afterwards.

horovod/common/operations.cc
alsrgv: #671 has been merged, please incorporate its functionality.

alsrgv:
As discussed, let's split this into three phases:

Phase 1 (short term): Keep sequential orchestration logic as-is, just add MLSL collective ops, like in https://github.com/uber/horovod/pull/323. Do the benchmark of these changes to measure the benefits on large systems.

Phase 2 (medium term): Benchmark Phase 1 implementation vs optimized orchestration implementation on the same system. If benefit is large, let's brainstorm again to find the best way to incorporate such orchestration mechanisms into Horovod.

Phase 3 (long term): If Phase 2 goes through with large improvements, we'll take brainstorm results and refactor Horovod orchestration logic to allow for pluggable optimizations like yours. You can then implement your change.

horovod/common/operations.cc
```diff
@@ -2013,6 +2292,125 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
 //      If instead of "DONE" they receive "SHUTDOWN", they exit their background
 //      loop.
 bool RunLoopOnce(HorovodGlobalState& state, bool is_coordinator) {
+#if HAVE_MLSL
+    bool should_shut_down = false;
+
+    // Lock and create private copy of tensor table
+    {
+        std::lock_guard<std::mutex> guard(state.mutex);
+        for (auto& i : state.tensor_table)
```
shahzadlone: should it perhaps be ```const auto& i```?

fightbob:
Hi @SmorkalovME , could you provide horovod-mlsl perfromace data on Xeon platform? Thanks.

SmorkalovME:
> Hi @SmorkalovME , could you provide horovod-mlsl perfromace data on Xeon platform? Thanks.

Hi @fightbob, we are collecting/validating the data right now - will be posted here anytime soon.

fightbob:
> > Hi @SmorkalovME , could you provide horovod-mlsl perfromace data on Xeon platform? Thanks.
> 
> Hi @fightbob, we are collecting/validating the data right now - will be posted here anytime soon.

@SmorkalovME cool!

fightbob:
@SmorkalovME Laskt week, I trained my model with horovod-mlsl on one Xeon servers. Compared with uber original horovod, it's only 3%~5% perf boosted with the same mpirun parameters. Could you give me some guides for more pef boosted?
Here is the command lines:

OMP_NUM_THREADS=16 KMP_AFFINITY="granularity=fine,compact,1,0" I_MPI_PORT_RANGE=50000:60000 mpirun -n 2 python3 test.py

On one Xeon server, train model with  two ranks.

SmorkalovME:
> @SmorkalovME Laskt week, I trained my model with horovod-mlsl on one Xeon servers. Compared with uber original horovod, it's only 3%~5% perf boosted with the same mpirun parameters. Could you give me some guides for more pef boosted?
> Here is the command lines:
> 
> OMP_NUM_THREADS=16 KMP_AFFINITY="granularity=fine,compact,1,0" mpirun -n 2 python3 test.py
> 
> On one Xeon server, train model with two ranks.

@fightbob Could you please a bit elaborate on your configuration? How many cores each socket has? The general BKM is to place horovod background thread and MLSL servers to cores different than ones used for compute/OpenMP, so we can suggest specific parameters based on your system configuration.

fightbob:
> > @SmorkalovME Laskt week, I trained my model with horovod-mlsl on one Xeon servers. Compared with uber original horovod, it's only 3%~5% perf boosted with the same mpirun parameters. Could you give me some guides for more pef boosted?
> > Here is the command lines:
> > OMP_NUM_THREADS=16 KMP_AFFINITY="granularity=fine,compact,1,0" mpirun -n 2 python3 test.py
> > On one Xeon server, train model with two ranks.
> 
> @fightbob Could you please a bit elaborate on your configuration? How many cores each socket has? The general BKM is to place horovod background thread and MLSL servers to cores different than ones used for compute/OpenMP, so we can suggest specific parameters based on your system configuration.

@SmorkalovME Here is the server lscpu info:
 
_Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                32
On-line CPU(s) list:   0-31
Thread(s) per core:    2
Core(s) per socket:    8
Socket(s):             2
NUMA node(s):          2
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 79
Model name:            Intel(R) Xeon(R) CPU E5-2620 v4 @ 2.10GHz
Stepping:              1
CPU MHz:               1200.000
CPU max MHz:           2100.0000
CPU min MHz:           1200.0000
BogoMIPS:              4206.69
Virtualization:        VT-x
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              20480K
NUMA node0 CPU(s):     0-7,16-23
NUMA node1 CPU(s):     8-15,24-31_

1. Could you provide specific parameters based on single node?
2. Could you provide specific parameters based on cluster?(On the test cluster, there is no any RDMA  device)

PR #574, opened 2018-10-20
Add AVX intrinsics to vectorize & speed up FP16-CPU computations

alsrgv:
Use CPU intrinsics to do fast FP16 conversion & vectorization. Observe 10-12% speedup over two nodes.

### DIFF
--- a/horovod/common/half.cc
+++ b/horovod/common/half.cc
@@ -1,0 +1,78 @@
+// Copyright 2018 Uber Technologies, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


horovod/common/half.cc
```diff
@@ -0,0 +1,78 @@
...
+#endif
+
+// float16 custom data type summation operation.
+void float16_sum(void* invec, void* inoutvec, int* len,
+                 MPI_Datatype* datatype) {
+  // cast invec and inoutvec to your float16 type
+  auto* in = (unsigned short*)invec;
+  auto* inout = (unsigned short*)inoutvec;
+
+  int i = 0;
```
tgaddair: Why not initialize `i` within the for loop?

alsrgv: Because if we have AVX, we will process all the blocks divisible by 8, and the rest will be processed as "leftovers" using the slow algorithm.  If we don't have AVX, `i` will start at 0.

PR #596, opened 2018-10-29
Use torch:: instead of at:: namespace in Torch backend

goldsborough:
ATen is still undergoing major churn and many things are moving from the `at::` namespace to the `c10::` namespace. Both namespaces are exported into the `torch::` namespace so you can save yourself some trouble by using `torch::` everywhere and never `at::`.

### DIFF
--- a/horovod/torch/adapter_v2.h
+++ b/horovod/torch/adapter_v2.h
@@ -16,7 +16,7 @@
 #ifndef HOROVOD_TORCH_ADAPTER_V2_H
 #define HOROVOD_TORCH_ADAPTER_V2_H
 
-#include <torch/extension.h>
+#include <torch/torch.h>
 
 #include "../common/common.h"
 
@@ -33,24 +33,24 @@ public:
 
 private:
   int device_ = CPU_DEVICE_ID;
-  at::Tensor tensor_;
+  torch::Tensor tensor_;
 };
 
 class TorchTensor : public Tensor {
 public:
-  TorchTensor(at::Tensor tensor);
+  TorchTensor(torch::Tensor tensor);
   virtual const MPIDataType dtype() const override;
   virtual const TensorShape shape() const override;
   virtual const void* data() const override;
   virtual int64_t size() const override;
 
 protected:
-  at::Tensor tensor_;
+  torch::Tensor tensor_;
 };
 
 class TorchOpContext : public OpContext {
 public:
-  TorchOpContext(int device, at::Tensor output);
+  TorchOpContext(int device, torch::Tensor output);
   virtual Status
   AllocatePersistent(int64_t size,
                      std::shared_ptr<PersistentBuffer>* tensor) override;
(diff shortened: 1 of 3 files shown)


alsrgv:
@goldsborough, many thanks for the PR!  If we don't need to use `torch/extension.h`, can we get rid of https://github.com/uber/horovod/blob/master/setup.py#L676?

goldsborough:
@alsrgv yep you can get rid of that! Are the two travis failures legit?

alsrgv:
@goldsborough, they are.

I can see this warning:
```
In file included from horovod/torch/adapter_v2.h:19,
                   from horovod/torch/mpi_ops_v2.cc:22:
  /usr/local/lib/python2.7/dist-packages/torch/lib/include/torch/csrc/api/include/torch/torch.h:14:2: warning: #warning "Including torch/torch.h for C++ extensions is deprecated. Please include torch/extension.h" [-Wcpp]
   #warning "Including torch/torch.h for C++ extensions is deprecated. Please include torch/extension.h"
    ^~~~~~~
  In file included from horovod/torch/mpi_ops_v2.cc:22:
```

That warning was the reason for switching to `torch/extension.h`.  Are you un-deprecating it?

The errors are because of the name clash.  Could you prefix PyTorch's namespace with `::`, e.g. `::torch::empty()`?

goldsborough:
@alsrgv Oh, my bad, I didn't realize you were building this code as a C++ extension. In that case `torch/extension.h` is the correct include. I'll revert the change of that line.

horovod/torch/mpi_ops_v2.cc
```diff
@@ -39,7 +39,7 @@ std::string GetOpName(const std::string& prefix, const std::string& name,
   return prefix + ".noname." + std::to_string(handle);
 }
 
-int GetDeviceID(const at::Tensor& tensor) {
+int GetDeviceID(const torch::Tensor& tensor) {
```
alsrgv: Need to add `::` here, too.

horovod/torch/adapter_v2.h
```diff
@@ -17,6 +17,7 @@
 #define HOROVOD_TORCH_ADAPTER_V2_H
 
 #include <torch/extension.h>
+#include <torch/torch.h>
```
alsrgv: I still see a deprecation warning with `torch.h`:
```
   #warning "Including torch/torch.h for C++ extensions is deprecated. Please include torch/extension.h"
```

Is it going to continue working for the foreseeable future?  Is the plan to move those missing declarations to `torch/extension.h`, at which point we can remove `torch/torch.h` include with a version guard?

goldsborough: I'm going to make a PR very soon to make `torch/extension.h` the canonical header for all C++ extensions and includes everything from `torch/torch.h` too. The warnings will then go away too.

alsrgv: Sounds good.  Thanks for making the changes!

goldsborough: https://github.com/pytorch/pytorch/pull/13482

Issue #564, opened 2018-10-16
Pytorch gradients with control flow


jeff-engineml:
Environment: Python 2.7, Horovod 0.15.0, Pytorch 0.4.0, CUDA 9.0, Cudnn 7

```python
def calc_loss(pred, gt)
  if some_condition():
    loss = torch.sum(torch.abs(pred - gt))
  else:
    loss = 0.0
  return loss

optimizer = torch.optim.SGD(model.parameters(), lr=0.1)
optimizer = hvd.DistributedOptimizer(optimizer, model.named_parameters())

pred = model(x)
loss = calc_loss(pred, ground_truth)
loss.backward()
optimizer.step()
```

This example hangs if `some_condition` evaluates to `False` for some ranks, and `True` for others. For the ranks that `some_condition` evaluate to `True`, pytorch will send gradients for all weights in model (as expected). For ranks that evaluate to `False`, pytorch will not send **any** gradients. When this happens I see messages like:

```allreduce.model.conv_1.0.weight [missing ranks: 0, 1, 2, 3, 4, 6, 7]```

I can fix this by making the `False` case depend on the input `pred`, but this seems a bit hacky.

```python
def calc_loss(pred, gt)
  if some_condition():
    loss = torch.sum(torch.abs(pred - gt))
  else:
    loss = 0.0 * torch.sum(pred)
  return loss
```

Is there a better way to do this?

alsrgv:
I think we could solve this by forcing allreduce for all model parameters in `step()` that were not being allreduced already in gradient hooks.  Would you like to contribute that change?

jeff-engineml:
I'll take a look and see what I can do

PR #597, opened 2018-10-30
 Added force allreduce of all gradients in step()

steve-engineml:
Addresses Issue #564
1) Added a set, ```self._requires_update```, to ensure that all gradients are passed to allreduce even if a gradient is not calculated during ```loss.backward()```. If a gradient is not calculated, assume the gradient should be set to 0.

2) Added a function ```_allreduce_grad(self, p)``` to avoid redundant code because this is called both when making hooks and when forcing allreduce on missing gradients.

3) Created arbitrary parameter names even if ```model.named_parameters()``` is not passed into ```DistributedOptimizer()```. This ensures that the gradients which were forced to allreduce on one replica know which gradients to communicate with on the other replicas. Otherwise replicas will have communication problems about which gradients to perform allreduce on.

### DIFF
--- a/horovod/torch/__init__.py
+++ b/horovod/torch/__init__.py
@@ -57,6 +57,11 @@ class _DistributedOptimizer(torch.optim.Optimizer):
 
         self._parameter_names = {v: k for k, v
                                  in sorted(named_parameters)}
+
+        if len(self._parameter_names) == 0:
+            self._parameter_names = {v: str(i) for param_group in self.param_groups
+                                     for i, v in enumerate(param_group['params'])}
+
         self._handles = {}
         self._grad_accs = []
         self._requires_update = set()
--- a/horovod/torch/__init__.py
+++ b/horovod/torch/__init__.py
@@ -55,11 +55,11 @@ class _DistributedOptimizer(torch.optim.Optimizer):
                              'tuples (name, parameter), usually produced by '
                              'model.named_parameters().')
 
-        self._parameter_names = {v: k for k, v
-                                 in sorted(named_parameters)}
-
-        if len(self._parameter_names) == 0:
-            self._parameter_names = {v: str(i) for param_group in self.param_groups
+        if len(named_parameters) > 0:
+            self._parameter_names = {v: k for k, v
+                                     in sorted(named_parameters)}
+        else:
+            self._parameter_names = {v: 'allreduce.noname.%s' % i for param_group in self.param_groups
                                      for i, v in enumerate(param_group['params'])}
 
         self._handles = {}
--- a/horovod/torch/__init__.py
+++ b/horovod/torch/__init__.py
@@ -204,7 +204,7 @@ def broadcast_optimizer_state(optimizer, root_rank):
         for group in optimizer.param_groups:
             for p in group['params']:
                 p.grad = p.data.new(p.size()).zero_()
-        if type(optimizer).__module__.startswith('horovod'):
+        if optimizer.__module__ == DistributedOptimizer.__module__:
             super(optimizer.__class__, optimizer).step()
         else:
             optimizer.step()
--- a/horovod/torch/__init__.py
+++ b/horovod/torch/__init__.py
@@ -205,6 +205,11 @@ def broadcast_optimizer_state(optimizer, root_rank):
         for group in optimizer.param_groups:
             for p in group['params']:
                 p.grad = p.data.new(p.size()).zero_()
+        # This function accepts a torch.optim.Optimizer or a DistributedOptimizer
+        # wrapped around a torch optimizer. Calling step() with a DistributedOptimizer
+        # forces allreduce on all model parameters, which will result in deadlock
+        # unless every rank calls step(). Therefore, to finish state initialization
+        # only call optimizer.step() with a torch.optim.Optimizer.
         if optimizer.__module__ == DistributedOptimizer.__module__:
             super(optimizer.__class__, optimizer).step()
         else:
--- a/horovod/torch/__init__.py
+++ b/horovod/torch/__init__.py
@@ -99,7 +99,6 @@ class _DistributedOptimizer(torch.optim.Optimizer):
     def synchronize(self):
         missing_p = self._requires_update - set(self._handles.keys())
         for p in missing_p:
-            # p.grad = torch.zeros(p.shape, dtype=p.dtype, device=p.device)
             self._allreduce_grad(p)
 
         for p, value in self._handles.items():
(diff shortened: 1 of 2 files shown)


horovod/torch/__init__.py
```diff
@@ -55,33 +55,42 @@ def __init__(self, params, named_parameters, compression):
                              'tuples (name, parameter), usually produced by '
                              'model.named_parameters().')
 
-        self._parameter_names = {v: k for k, v
-                                 in sorted(named_parameters)}
+        if len(named_parameters) > 0:
+            self._parameter_names = {v: k for k, v
+                                     in sorted(named_parameters)}
+        else:
+            self._parameter_names = {v: 'allreduce.noname.%s' % i for param_group in self.param_groups
```
alsrgv: To make line length less crazy, suggest adding `\n` before `for`.

horovod/torch/__init__.py
```diff
@@ -93,6 +102,10 @@ def synchronize(self):
         self._handles.clear()
 
     def step(self, closure=None):
+        missing_p = self._requires_update - set(self._handles.keys())
+        for p in missing_p:
+            p.grad = torch.zeros(p.shape, dtype=p.dtype, device=p.device)
```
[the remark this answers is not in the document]
steve-engineml: For some reason, I thought the missing p gradients were set to None during loss.backward(). You are correct, opt.zero_grad() has already set them to 0.

alsrgv:
@steve-engineml, can you fix the broken test, and add a new test for this functionality in `test_torch.py`?

horovod/torch/__init__.py
```diff
@@ -190,7 +204,7 @@ def broadcast_optimizer_state(optimizer, root_rank):
         for group in optimizer.param_groups:
             for p in group['params']:
                 p.grad = p.data.new(p.size()).zero_()
-        optimizer.step()
+        super(optimizer.__class__, optimizer).step()
```
alsrgv: Any idea why does it fail with the original `optimizer.step()`?  Can you check the stack trace of the failure?

steve-engineml: It fails because of how ```test_broadcast_state()```  in test/torch.py works. At line 827, a new model and uninitialized optimizer are created. If the rank is 0, the previous optimizer state is loaded, then broadcast to other ranks. For ranks other than 0, the optimizer remains empty; therefore ```len(state_dict['state']) == 0``` , which causes it to call ```optimizer.step()``` in torch/__init__.py (the line I altered, line 207). In this case, the only purpose of calling ```optimizer.step()``` is to initialize an uninitialized optimizer. Since I force allreduce on all model parameters in ```_DistributedOptimizer().step()```, If ```optimizer.step()``` is not called on all ranks, resource starvation occurs.

alsrgv: Ah, makes sense.  One concern I have is that some users may do this broadcast on the original optimizer before wrapping it in DistributedOptimizer.  Let's only do super if the optimizer is a subclass of _DistributedOptimizer?

horovod/torch/__init__.py
```diff
@@ -190,7 +204,10 @@ def broadcast_optimizer_state(optimizer, root_rank):
         for group in optimizer.param_groups:
             for p in group['params']:
                 p.grad = p.data.new(p.size()).zero_()
-        optimizer.step()
+        if type(optimizer).__module__.startswith('horovod'):
```
alsrgv: That kind of burns my eyes... I guess we could make it a bit more robust though: 
```
if optimizer.__module__ == DistributedOptimizer.__module__:
```

alsrgv:
@steve-engineml, looks good!  Let's add tests and call it a day :-)

steve-engineml:
One of the first comments you made was to remove ```p.grad = torch.zeros(p.shape, dtype=p.dtype, device=p.device)``` because opt.zero_grad() already does this. While writing the unittest, I realized why I had that. If loss.backward() only activates the hooks of some gradients on the very first step, then while trying to force allreduce on the missing gradients you get this error:
```
  File "test/temp_test.py", line 92, in <module>
    test_force_allreduce()
  File "test/temp_test.py", line 87, in test_force_allreduce
    optimizer.step()
  File "/Users/stevethomas/anaconda3/envs/horovod-pytorch/lib/python3.5/site-packages/horovod/torch/__init__.py", line 111, in step
    self.synchronize()
  File "/Users/stevethomas/anaconda3/envs/horovod-pytorch/lib/python3.5/site-packages/horovod/torch/__init__.py", line 106, in synchronize
    output = synchronize(handle)
  File "/Users/stevethomas/anaconda3/envs/horovod-pytorch/lib/python3.5/site-packages/horovod/torch/mpi_ops.py", line 436, in synchronize
  File "/Users/stevethomas/anaconda3/envs/horovod-pytorch/lib/python3.5/site-packages/torch/utils/ffi/__init__.py", line 197, in safe_call
    result = torch._C._safe_call(*args, **kwargs)
torch.FatalError: Horovod has been shut down. This was caused by an exception on one of the ranks or an attempt to allreduce, allgather or broadcast a tensor after one of the ranks finished execution. If the shutdown was caused by an exception, you should see the exception in the log before the first shutdown message.
  File "test/temp_test.py", line 92, in <module>
    test_force_allreduce()
  File "test/temp_test.py", line 87, in test_force_allreduce
    optimizer.step()
  File "/Users/stevethomas/anaconda3/envs/horovod-pytorch/lib/python3.5/site-packages/horovod/torch/__init__.py", line 111, in step
    self.synchronize()
  File "/Users/stevethomas/anaconda3/envs/horovod-pytorch/lib/python3.5/site-packages/horovod/torch/__init__.py", line 102, in synchronize
    self._allreduce_grad(p)
  File "/Users/stevethomas/anaconda3/envs/horovod-pytorch/lib/python3.5/site-packages/horovod/torch/__init__.py", line 84, in _allreduce_grad
    tensor = p.grad.data
AttributeError: 'NoneType' object has no attribute 'data'
-------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
-------------------------------------------------------
--------------------------------------------------------------------------
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[38159,1],1]
  Exit code:    1
```
After the optimizer for any given rank has had at least 1 pass when loss.backward() activates every gradient hook, then everything works fine. However, if some hooks aren't called on the first pass, then...

alsrgv:
@steve-engineml, what if we do `zero_grad()` before that first step?

horovod/torch/__init__.py
```diff
@@ -190,7 +205,10 @@ def broadcast_optimizer_state(optimizer, root_rank):
         for group in optimizer.param_groups:
             for p in group['params']:
                 p.grad = p.data.new(p.size()).zero_()
-        optimizer.step()
+        if optimizer.__module__ == DistributedOptimizer.__module__:
```
tgaddair: Could you add a comment here explaining why this is necessary?

PR #608, opened 2018-11-02
Fixed broadcast_optimizer_state to properly convert nested types

### LINKED ISSUE #605: pytorch + horovod 0.15.1 distributed optimizer not working anymore
I just upgraded horovod 0.15.0 -> 0.15.1 on a ubuntu image `4.4.0-137-generic #163-Ubuntu SMP Mon Sep 24 13:14:43 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`. When using the DistributedOptimizer from horovod.torch I now encounter the error

```
  File "train.py", line 641, in <module>
    train_images(hps)
  File "train.py", line 444, in train_images
    train_step(batch, batch_idx, epoch, hps, model, opt, train_logger)
  File "train.py", line 457, in train_step
    opt.step()
  File "/opt/conda/lib/python3.6/site-packages/horovod/torch/__init__.py", line 97, in step
    return super(self.__class__, self).step(closure)
  File "/opt/conda/lib/python3.6/site-packages/torch/optim/adamax.py", line 75, in step
    exp_avg.mul_(beta1).add_(1 - beta1, grad)
TypeError: mul_() received an invalid combination of arguments - got (numpy.float32), but expected one of:
 * (Tensor other)
      didn't match because some of the arguments have invalid types: (numpy.float32)
 * (float other)
      didn't match because some of the arguments have invalid types: (numpy.float32)
```

Downgrading to 0.15.0 fixes the issue. The behavior is independent of CPU, GPU or MultipleGPU training.

tgaddair:
Fixes #605

### DIFF
--- a/horovod/torch/__init__.py
+++ b/horovod/torch/__init__.py
@@ -227,6 +227,22 @@ def broadcast_optimizer_state(optimizer, root_rank):
     callbacks = {}
     occurrences = collections.defaultdict(int)
 
+    # Returns the full type structure of the possibly nested objects for recursive casting back
+    def _get_types(x):
+        if isinstance(x, collections.Iterable):
+            return type(x), [_get_types(xi) for xi in x]
+        else:
+            return type(x)
+
+    # Casts an object encoded in a tensor back into its original type and subtypes
+    def _recursive_cast(x, dtype):
+        if isinstance(dtype, tuple):
+            t, dtypes = dtype
+            x = t(x)
+            return t([_recursive_cast(x[i], dtypes[i]) for i in range(len(x))])
+        else:
+            return dtype(x)
+
     # Some optimizer parameters may be represented as scalars instead of
     # tensors.  In such cases, we need to wrap the scalar in a tensor, then
     # broadcast, then update the appropriate value in the state_dict with the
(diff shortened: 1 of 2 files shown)


horovod/torch/__init__.py
```diff
@@ -227,6 +227,22 @@ def broadcast_optimizer_state(optimizer, root_rank):
...
+            return type(x), [_get_types(xi) for xi in x]
+        else:
+            return type(x)
+
+    # Casts an object encoded in a tensor back into its original type and subtypes
+    def _recursive_cast(x, dtype):
+        if isinstance(dtype, tuple):
+            t, dtypes = dtype
+            x = t(x)
+            return t([_recursive_cast(x[i], dtypes[i]) for i in range(len(x))])
```
alsrgv: Does this work with both lists, set, and dictionaries?

tgaddair: It will work with lists, sets, and tuples.  However, it will not work with dictionaries as is, but I can make that change.

alsrgv: Let's do it - maybe there will be dictionaries, too.

tgaddair: The one catch here is that dictionaries cannot be converted to tensors using `torch.Tensor`.  So in order to support dictionaries, we'd need to make changes to our encoding as well as our decoding.

alsrgv: Oh, I see.  I guess we could wait until this becomes necessary then.

tgaddair: Sounds good.

PR #615, opened 2018-11-08
Automatic parameter tuning for hierarchical allreduce, fusion buffer, and cycle time

tgaddair:
Adds support for automatic parameter tuning with HOROVOD_AUTOTUNE environment variable.

Adds third party libraries for Eigen (linear algebra) and LBFGS (optimization).

The high level strategy for parameter tuning:

1. Spend a few cycles warming up as buffers are allocated and cached.
2. Begin collecting samples of how many bytes and microseconds it took to allreduce a batch of tensors (approximated by checking for duplicate tensor names) -- defined as a "cycle".
3. After a fixed number of cycles, compute the sample score by dividing the total number of bytes by the total number of microseconds.
4. Collect a fixed number of such samples, then take the median sample score to amortize the noise in each sample.
5. Compute the next set of parameter values to try subject to the median sample score, then repeat from (2).

The parameter optimization uses a top-level strategy of a hierarchical optimization framework, where independent parameters are arranged in a chain, and parameters lower in the chain are optimized before parameters higher in the chain.  

This process proceeds as follows:

1. Choose the initial values for all parameters.
2. From the "leaf" parameter, select a new value as each median sample score is observed.
3. When the stopping criteria is reached (convergence, exhausted all values), then push the best score up to the next parameter in the chain.
4. Inductively, select a new value for the current parameter value until stopping criteria is reached, then repeat (3).

For example, we set the joint Bayesian parameter of (tensor fusion threshold, cycle time) as the leaf parameter, and optimize all values for a fixed number of iterations, then push the best score up to the hierarchical allreduce parameter.  Then we repeat optimizing the joint parameter for hierarchical allreduce separately.


tgaddair:
@alsrgv It may simplify things to add third party libraries in a separate PR.  Let me know if you'd prefer to do that and I'll break out the diff.

horovod/common/operations.cc
```diff
@@ -1839,7 +1843,7 @@ bool RunLoopOnce(HorovodGlobalState& state, bool is_coordinator) {
           if (response.response_type() == new_response.response_type() &&
               response.devices() == new_response.devices() &&
               entry.tensor->dtype() == new_entry.tensor->dtype() &&
-              tensor_size + new_tensor_size <= state.tensor_fusion_threshold) {
+              tensor_size + new_tensor_size <= TensorFusionThresholdBytes()) {
```
alsrgv: Should this be sourced from Parameter Manager?

horovod/common/operations.cc
```diff
@@ -709,6 +698,27 @@ cudaError_t ReleaseCudaEvent(cudaEvent_t event) {
     }                                                                          \
   }
 
+int64_t TensorFusionThresholdBytes() {
```
alsrgv: Let's rename to InitialTensorFusionThresholdBytes() to avoid confusion

horovod/common/operations.cc
tgaddair: Yeah, I was just thinking about that.  That way we could also do some clever caching to avoid recomputing it every time as well.  I'll make that change.

horovod/common/operations.cc
tgaddair: Sounds good.

alsrgv: Nevermind -- I thought that this actually just read environment variables.  This looks good as-is.

horovod/common/operations.cc
alsrgv: Nevermind, you already read Parameter Manager Threshold within TensorFusionThresholdBytes

horovod/common/fusion_buffer_manager.cc
```diff
@@ -0,0 +1,54 @@
...
+namespace common {
+
+Status FusionBufferManager::InitializeBuffer(int64_t threshold, int device, std::shared_ptr<OpContext> context,
+                                             std::function<void()> on_start_init,
+                                             std::function<void()> on_end_init) {
+  auto& elem = tensor_fusion_buffers_[std::make_tuple(device, context->framework())];
+  auto& buffer = elem.first;
+  int64_t& size = elem.second;
+  if (size != threshold) {
+    buffer.reset();
```
alsrgv: Can you comment what does `buffer.reset()` do precisely?

If this does not set buffer to nullptr, would you re-initialize it with AllocatePersistent?

tgaddair: It will (1) delete the currently held pointer, then (2) set that current pointer reference to `nullptr`.  One thing I can do is explicitly reset it to `nullptr` like: `buffer.reset(nullptr)`.  Would that be preferred?

http://www.cplusplus.com/reference/memory/unique_ptr/reset/

rahul003:
Would it be better to include the Eigen and LBFGS libraries using the `git submodule` functionality? https://blog.github.com/2016-02-01-working-with-submodules/ 

I feel like that would be easier to maintain than copying the files over into Horovod repository. This way we can easily pull in changes from those repositories by commit from the submodules.

tgaddair:
> Would it be better to include the Eigen and LBFGS libraries using the `git submodule` functionality? https://blog.github.com/2016-02-01-working-with-submodules/
> 
> I feel like that would be easier to maintain than copying the files over into Horovod repository. This way we can easily pull in changes from those repositories by commit from the submodules.

It's a good suggestion, but I'm torn.  On the one hand, it makes tracking easier if we decide to keep the dependencies up to date with the upstream repos.  But on the other hand, it introduces complexity for contributors (every update, including this one, will require a manual `git submodule update` for all contributors).  Definitely something to discuss though, thanks for bringing it up!

horovod/common/optim/bayesian_optimization.cc
```diff
@@ -0,0 +1,163 @@
...
+
+  VectorXd x_next = VectorXd::Zero(d_);
+  double fx_min = std::numeric_limits<double>::max();
+  for (int i = 0; i < n_restarts; i++) {
+    VectorXd x = VectorXd::Zero(d_);
+    for (unsigned int j = 0; j < d_; j++) {
+      x[j] = dists_[j](gen_);
+    }
+
+    VectorXd x0 = x;
```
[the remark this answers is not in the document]
tgaddair: Removed.

horovod/common/fusion_buffer_manager.cc
```diff
@@ -0,0 +1,54 @@
...
+
+  if (buffer == nullptr) {
+    on_start_init();
+    size = threshold;
+
+    // Lazily allocate persistent buffer for Tensor Fusion and keep it
+    // forever per device.
+    Status status = context->AllocatePersistent(threshold, &buffer);
+    if (status.ok()) {
+      on_end_init();
```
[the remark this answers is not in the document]
tgaddair: Done.

horovod/common/operations.cc
```diff
@@ -1590,47 +1605,35 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
...
-    int64_t div =
-        state.local_size * mpi_double_size * FUSION_BUFFER_ATOMIC_UNIT;
-    state.tensor_fusion_threshold =
-        ((proposed_fusion_threshold + div - 1) / div) * div;
-  } else {
-    state.tensor_fusion_threshold = proposed_fusion_threshold;
+  // Enable auto-tuning.
+  auto horovod_tune_params = std::getenv(HOROVOD_TUNE_PARAMS);
+  if (horovod_tune_params != nullptr && std::strtol(horovod_tune_params, nullptr, 10) > 0) {
+    auto horovod_params = std::getenv(HOROVOD_PARAMS_OUTPUT);
```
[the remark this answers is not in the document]
tgaddair: Renamed to HOROVOD_AUTOTUNER and HOROVOD_AUTOTUNER_LOG.

horovod/common/parameter_manager.cc
```diff
@@ -0,0 +1,419 @@
...
+  if (iteration_ < test_points_.size()) {
+    value = FilterTestPoint(iteration_);
+  } else {
+    value = bayes_->NextSample();
+  }
+}
+
+bool ParameterManager::BayesianParameter::IsDoneTuning() const {
+  unsigned long d = bayes_->Dim();
+  return iteration_ > 20 * d;
```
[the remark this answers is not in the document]
tgaddair: Max steps of bayesian optimization. Added constant.

horovod/common/parameter_manager.cc
```diff
@@ -0,0 +1,419 @@
...
+  int j = 0;
+  for (auto var : variables_) {
+    if (fixed_values_.find(var.variable) == fixed_values_.end()) {
+      bounds.push_back(var.bounds);
+      index_[var.variable] = j;
+      j++;
+    }
+  }
+
+  bayes_.reset(new BayesianOptimization(bounds, 0.2));
```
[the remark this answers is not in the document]
tgaddair: Gaussian process noise parameter, bigger means we assume more noise in the underlying process.  Added a constant.

horovod/common/parameter_manager.h
```diff
@@ -0,0 +1,229 @@
...
+  // threshold will be fused.
+  int64_t TensorFusionThresholdBytes() const;
+  void SetTensorFusionThresholdBytes(int64_t threshold, bool fixed=false);
+
+  // Background thread cycle time in milliseconds.  Fractional numbers are
+  // permitted.
+  double CycleTimeMs() const;
+  void SetCycleTimeMs(double cycle_time_ms, bool fixed=false);
+
+  // Observes that the given tensors have been allreduced over the given number of microseconds.
```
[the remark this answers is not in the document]
tgaddair: The Parameter Manager itself is general enough to work for both allreduce and allgather.  I can update the docs to make it more generic.  Right now, none of these parameters affect allgather performance, which is why I don't make mention to it or have hooks into it, but if we were to do tensor fusion with allgather, then we can definitely hook it in.

horovod/common/parameter_manager.h
```diff
@@ -0,0 +1,229 @@
...
+  //
+  // Args:
+  //  tensor_names: The names of the tensors that have been allreduced.
+  //  bytes: The number of bytes that were allreduced per worker.
+  //  microseconds: The number of microseconds taken to allreduce the bytes on this worker.
+  void Update(const std::vector<std::string>& tensor_names, int64_t bytes, double microseconds);
+
+private:
+  void Tune(double score);
```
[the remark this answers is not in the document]
tgaddair: Renamed `ReadyTune()` to `Reset()` and added comments in the header.

horovod/common/parameter_manager.h
```diff
@@ -0,0 +1,229 @@
...
+    inline double BestScore() const override { return best_score_; };
+
+    inline bool IsTunable() const override { return tunable_; };
+
+  protected:
+    void SetCurrentValue(T value);
+    void Reinitialize(T value);
+
+  private:
+    void TuneNextParameter();
```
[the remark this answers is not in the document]
tgaddair: Refactored to an iterative approach, let me know what you think.

horovod/common/optim/gaussian_process.h
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) 2007–2018 The scikit-learn developers. All rights reserved.
```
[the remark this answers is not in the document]
tgaddair: One is, the other is apparently BSD.  Let me know how would you like to proceed.

horovod/common/operations.h
```diff
@@ -52,6 +52,8 @@ namespace common {
 // Horovod knobs.
 #define HOROVOD_MPI_THREADS_DISABLE "HOROVOD_MPI_THREADS_DISABLE"
 #define HOROVOD_TIMELINE "HOROVOD_TIMELINE"
+#define HOROVOD_AUTOTUNER "HOROVOD_AUTOTUNER"
```
alsrgv: HOROVOD_AUTOTUNER -> HOROVOD_AUTOTUNE

horovod/common/parameter_manager.cc
```diff
@@ -0,0 +1,435 @@
...
+  }
+}
+
+void ParameterManager::SetAutoTuning(bool active) {
+  if (active != active_) {
+    warmup_remaining_ = WARMUPS;
+  }
+  active_ = active;
+  if (!active_ && rank_ == root_rank_) {
+    std::cerr << "Horovod Tune: BEST [ "
```
alsrgv: HOROVOD_AUTOTUNER

PR #632, opened 2018-11-17
Simplify PyTorch continuous-integration example to run faster

alsrgv:


.travis.yml
```diff
@@ -127,5 +127,11 @@ script:
   # run Keras MNIST advanced example
   - docker exec ${CONTAINER} /bin/sh -c "${MPIRUN} python /horovod/examples/keras_mnist_advanced.py"
 
+  # hack PyTorch MNIST example to be smaller
```
tgaddair: Maybe we can add a flag to the `pytorch_mnist.py` example like `--small` that will use this network construction instead of the default.

alsrgv: @tgaddair, I'm hoping we can get a better CI environment that wouldn't require dealing with exclusively tiny models.

Issue #673, opened 2018-12-01
Many dataset APIs have been depreciated


ghltshubh:
Many tensorflow APIs for downloading datasets for running horovod examples have been depreciated.
This is one of the examples: https://www.tensorflow.org/api_docs/python/tf/contrib/learn/datasets

tgaddair:
Hey @ghltshubh, looking at the older versions of TensorFlow, it looks like `tf.estimator` has been around for some time, so I think this is a good and safe change to make.  Thanks for bringing it up!

PR #680, opened 2018-12-06
Removed depreciated tensorflow dataset APIs

abditag2:
Some of the examples were using depreciated tensorflow APIs to load the data and these changes fix the problem.

- updated the MNIST Dataset API
- updated ModeKey

Issue #673

### DIFF
--- a/examples/tensorflow_mnist.py
+++ b/examples/tensorflow_mnist.py
@@ -18,11 +18,17 @@ import shutil
 import os
 
 import tensorflow as tf
-import keras
 import horovod.tensorflow as hvd
 import numpy as np
 
-layers = tf.contrib.layers
+from distutils.version import LooseVersion
+
+if LooseVersion(tf.__version__) >= LooseVersion("1.4.0"):
+    from tensorflow import keras
+else:
+    from tensorflow.contrib import keras
+
+layers = tf.layers
 
 tf.logging.set_verbosity(tf.logging.INFO)
 
@@ -39,15 +45,15 @@ def conv_model(feature, target, mode):
 
     # First conv layer will compute 32 features for each 5x5 patch
     with tf.variable_scope('conv_layer1'):
-        h_conv1 = layers.conv2d(
-            feature, 32, kernel_size=[5, 5], activation_fn=tf.nn.relu)
+        h_conv1 = layers.conv2d(feature, 32, kernel_size=[5, 5],
+                                activation=tf.nn.relu, padding="SAME")
         h_pool1 = tf.nn.max_pool(
             h_conv1, ksize=[1, 2, 2, 1], strides=[1, 2, 2, 1], padding='SAME')
 
     # Second conv layer will compute 64 features for each 5x5 patch.
     with tf.variable_scope('conv_layer2'):
-        h_conv2 = layers.conv2d(
-            h_pool1, 64, kernel_size=[5, 5], activation_fn=tf.nn.relu)
+        h_conv2 = layers.conv2d(h_pool1, 64, kernel_size=[5, 5],
+                                activation=tf.nn.relu, padding="SAME")
         h_pool2 = tf.nn.max_pool(
             h_conv2, ksize=[1, 2, 2, 1], strides=[1, 2, 2, 1], padding='SAME')
         # reshape tensor into a batch of vectors
@@ -55,13 +61,11 @@ def conv_model(feature, target, mode):
 
     # Densely connected layer with 1024 neurons.
     h_fc1 = layers.dropout(
-        layers.fully_connected(
-            h_pool2_flat, 1024, activation_fn=tf.nn.relu),
-        keep_prob=0.5,
-        is_training=mode == tf.estimator.ModeKeys.TRAIN)
+        layers.dense(h_pool2_flat, 1024, activation=tf.nn.relu),
+        rate=0.5, training=mode == tf.estimator.ModeKeys.TRAIN)
 
     # Compute logits (1 per class) and compute loss.
-    logits = layers.fully_connected(h_fc1, 10, activation_fn=None)
+    logits = layers.dense(h_fc1, 10, activation=None)
     loss = tf.losses.softmax_cross_entropy(target, logits)
 
     return tf.argmax(logits, 1), loss
@@ -99,8 +103,8 @@ def main(_):
             x_test, y_test) = keras.datasets.mnist.load_data(
             'MNIST-data-%d' % hvd.rank())
 
-    x_train = np.reshape(x_train, (-1, 784)) / 255
-    x_test = np.reshape(x_test, (-1, 784)) / 255
+    x_train = np.reshape(x_train, (-1, 784)) / 255.0
+    x_test = np.reshape(x_test, (-1, 784)) / 255.0
 
     # Build model...
     with tf.name_scope('input'):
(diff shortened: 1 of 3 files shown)


examples/tensorflow_mnist.py
```diff
@@ -12,12 +12,17 @@
...
+
+import shutil
+import os
 
 import tensorflow as tf
+import keras
 import horovod.tensorflow as hvd
+import numpy as np
+
 layers = tf.contrib.layers
```
tgaddair: Is it possible to use `tf.layers` instead of `tf.contrib.layers` as well?

examples/tensorflow_mnist_estimator.py
```diff
@@ -18,10 +18,13 @@
 from __future__ import division
 from __future__ import print_function
 
+import os
+import shutil
+
+import keras
```
tgaddair: As above: `tf.keras`

examples/tensorflow_mnist.py
```diff
@@ -62,18 +67,46 @@ def conv_model(feature, target, mode):
...
+        # folder first and download the dataset again.
+        cache_dir = os.path.join(os.path.expanduser('~'), '.keras')
+        datadir_base = os.path.expanduser(cache_dir)
+        datadir = os.path.join(datadir_base, "datasets")
+        shutil.rmtree(datadir)
+        (x_train, y_train), (
+            x_test, y_test) = keras.datasets.mnist.load_data(
+            'MNIST-data-%d' % hvd.rank())
+
+    x_train = np.reshape(x_train, (-1, 784)) / 255
```
tgaddair: Is this safe in Python2, where integer division is the default?  Maybe we can divide by 255.0 to be safe?

examples/tensorflow_mnist.py
```diff
@@ -62,18 +67,46 @@ def conv_model(feature, target, mode):
...
     # Download and load MNIST dataset.
-    mnist = learn.datasets.mnist.read_data_sets('MNIST-data-%d' % hvd.rank())
+    try:
+        (x_train, y_train), (x_test, y_test) = \
+            keras.datasets.mnist.load_data('MNIST-data-%d' % hvd.rank())
+    except OSError as ex:
+        # When running tests, if dataset is previously downloaded, it may cause
+        # the tests to fail. In this case, we need to remove the dataset cache
+        # folder first and download the dataset again.
+        cache_dir = os.path.join(os.path.expanduser('~'), '.keras')
```
tgaddair: Hmmm, is it necessary to remove the dataset every time?  This seems expensive.

examples/tensorflow_mnist_estimator.py
```diff
@@ -131,11 +136,24 @@ def main(unused_argv):
...
-    mnist = learn.datasets.mnist.read_data_sets('MNIST-data-%d' % hvd.rank())
-    train_data = mnist.train.images  # Returns np.array
-    train_labels = np.asarray(mnist.train.labels, dtype=np.int32)
-    eval_data = mnist.test.images  # Returns np.array
-    eval_labels = np.asarray(mnist.test.labels, dtype=np.int32)
+    try:
+        (train_data, train_labels), (eval_data, eval_labels) = \
+            keras.datasets.mnist.load_data('MNIST-data-%d' % hvd.rank())
+    except OSError as ex:
+        # When running tests, if dataset is previously downloaded, it may cause
```
tgaddair: Same comment as above: is it necessary to remove all?

examples/tensorflow_mnist_estimator.py
```diff
@@ -131,11 +136,24 @@ def main(unused_argv):
...
+        cache_dir = os.path.join(os.path.expanduser('~'), '.keras')
+        datadir_base = os.path.expanduser(cache_dir)
+        datadir = os.path.join(datadir_base, "datasets")
+        shutil.rmtree(datadir)
+        (train_data, train_labels), (
+            eval_data, eval_labels) = keras.datasets.mnist.load_data(
+            'MNIST-data-%d' % hvd.rank())
+
+    # reshape the features and normalize them between 0 and 1
+    train_data = np.reshape(train_data, (-1, 784)) / 255
```
tgaddair: Same comment as above: division by int vs float.

examples/tensorflow_mnist.py
```diff
@@ -12,12 +12,17 @@
...
 # limitations under the License.
 # ==============================================================================
-#!/usr/bin/env python
+# !/usr/bin/env python
+
+import shutil
+import os
 
 import tensorflow as tf
+import keras
```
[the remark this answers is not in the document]
abditag2: I initially did that but the problem is `tf.keras` is not available in TF 1.1.x. I can check for the TF version in the code and import `keras` or `tf.keras.`.

examples/tensorflow_mnist_estimator.py
abditag2: same as above.

examples/tensorflow_mnist.py
abditag2: fixed.

examples/tensorflow_mnist_estimator.py
abditag2: fixed.

examples/tensorflow_mnist.py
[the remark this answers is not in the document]
tgaddair: I would do a version check again v1.4.0 similar to what we do for `horovod.tensorflow.keras`:

```
if LooseVersion(tf.__version__) >= LooseVersion("1.4.0"):
    from tensorflow import keras
else:
    from tensorflow.contrib import keras
```

examples/tensorflow_mnist.py
abditag2: Changed this as well.

examples/tensorflow_mnist.py
```diff
@@ -12,12 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-#!/usr/bin/env python
+# !/usr/bin/env python
```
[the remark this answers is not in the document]
abditag2: fixed.

examples/tensorflow_mnist.py
```diff
@@ -34,54 +44,73 @@ def conv_model(feature, target, mode):
...
+
+
 def main(_):
     # Horovod: initialize Horovod.
     hvd.init()
 
     # Download and load MNIST dataset.
-    mnist = learn.datasets.mnist.read_data_sets('MNIST-data-%d' % hvd.rank())
+    dataset_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                            'MNIST-data-%d' % hvd.rank())
```
alsrgv: Fix formatting

examples/tensorflow_mnist.py
```diff
@@ -14,10 +14,20 @@
 # ==============================================================================
 #!/usr/bin/env python
 
+import os
```
alsrgv: Move to the next import group

examples/tensorflow_mnist.py
```diff
@@ -14,10 +14,20 @@
...
+
 import tensorflow as tf
 import horovod.tensorflow as hvd
-layers = tf.contrib.layers
-learn = tf.contrib.learn
+import numpy as np
+
+from distutils.version import LooseVersion
+
+if LooseVersion(tf.__version__) >= LooseVersion("1.4.0"):
```
alsrgv: I actually think it's OK to assume reasonably fresh TF version in examples - so, just doing tf.keras should be OK (and cleaner).

examples/tensorflow_mnist.py
abditag2: fixed.

PR #606, opened 2018-11-02
Horovod in PySpark

alsrgv:
This PR adds convenient helper function `horovod.spark.run(fn)` that runs Horovod in PySpark environment.

Under the hood, we set up a little TCP service on Driver and each Task that perform few functions:
1. Horovod Driver asks Spark Driver to spin up Spark Tasks which run Horovod Tasks.
2. Horovod Tasks register with Horovod Driver. They provide information about their network interfaces and hostname.
3. After all the Horovod Tasks have registered, each of them contacts one other Horovod Task to determine which network interfaces are routable, and report discoveries back to Horovod Driver.
4. Horovod Driver computes the intersection of routable interfaces, reorders tasks ranks to optimize for host colocation and executes mpirun with the custom remote executor (via `-mca plm_rsh_agent`).  Custom remote executor allows us to avoid using ssh.
5. The custom remote executor contacts Horovod Task to execute the command requested by `mpirun`, in case of Open MPI it's `orted`.
6. Horovod Task executes orted and will terminate when it finishes.
7. `orted` establishes the handshake with `mpirun` and starts running `mpirun_exec_fn.py` script.  That script calls back to Horovod Driver, fetches user code definition, executes it, and sends Horovod Task execution result when user code finishes.
8. `mpirun` terminates when the program finishes.
9. Horovod Tasks report execution results to Spark Tasks.
10. Horovod Driver waits for Spark Driver to collect execution results from Spark Tasks.
11. The execution results are surfaced to the user in the rank order.

This flow is depicted on the following sequence diagram (also available in [editor](https://tinyurl.com/ybty927w)):

image

All external programs, such as `mpirun`, `orted` and `python` are executed in a `safe_shell_exec.py` wrapper which guarantees children termination if the parent process is killed.

One detail not depicted in the diagram above: if there are multiple tasks on the same host, `mpirun` will launch only one `orted` process per host - on the first task running on that host.  In order to accommodate that `mpirun` behavior, tasks other than the first task on the host poll the first task for its command completion.  This behavior does not cause RPC polling explosion since maximum cardinality is the number of tasks per host.

The data is expected to be saved in Parquet format and ingested using [Petastorm](https://github.com/uber/petastorm).  This has been found to be more reliable and scalable than directly ingesting Spark RDDs.

This PR is a work in progress.

Example:
```
>>> def fn(magic_number):
...   import horovod.torch as hvd
...   hvd.init()
[...]

PR #700, opened 2018-12-12
fix configuration issue with nccl 2.3.5

rongou:
With NCCL 2.3.5, setting `NCCL_SOCKET_IFNAME` causes the process to hang. See https://github.com/kubeflow/mpi-operator/issues/57. From talking to the NCCL guys, simply removing that line from `nccl.conf` should suffice. I've tested this on the mpi operator.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,8 +65,7 @@ RUN echo "hwloc_base_binding_policy = none" >> /usr/local/etc/openmpi-mca-params
     echo "btl_tcp_if_exclude = lo,docker0" >> /usr/local/etc/openmpi-mca-params.conf
 
 # Set default NCCL parameters
-RUN echo NCCL_DEBUG=INFO >> /etc/nccl.conf && \
-    echo NCCL_SOCKET_IFNAME=^docker0 >> /etc/nccl.conf
+RUN echo NCCL_DEBUG=INFO >> /etc/nccl.conf
 
 # Install OpenSSH for MPI to communicate between containers
 RUN apt-get install -y --no-install-recommends openssh-client openssh-server && \


alsrgv:
@rongou, thanks for the PR!

@sjeaugey, is there more details you could share?  Does it only fail with `^docker0` or it fails with included interfaces as well?  We have a separate PR #606 to add Spark support which makes use of `NCCL_SOCKET_IFNAME` as well.

sjeaugey:
Since 2.3.5, docker0 and lo are not used by default (unless there is no other interface). So that line should no longer be needed.

On the other hand, we realized that with the new code, setting `NCCL_SOCKET_IFNAME=^docker0` means we might pick the `lo` interface. That was on purpose, since `lo` used to be always disabled before and we felt it could be useful in some cases (like a desktop machine with no other interface).

The new code honors the user setting precisely (no more, no less). If nothing is set we use `ib*` cards, then any card which is not `docker0` nor `lo`, then `docker0`, then finally `lo`.

In the end, you should either set `NCCL_SOCKET_IFNAME=^docker0,lo` or not set anything.

alsrgv:
@sjeaugey, got it - thanks a lot!  Will merge this in.

rongou:
@alsrgv would it be possible to push a new docker hub image?

alsrgv:
@rongou, 0.16.0 is almost out of the door with few improvements (autotuning, spark, mxnet, allgather fusion, etc), so the plan was to push out a new docker image when it happens, likely this or next week.  Please let me know if it's urgent.

rongou:
No worries, I put a temporary fix in our image.

PR #704, opened 2018-12-14
Added parallelization of allreduce during eager execution

tgaddair:


examples/tensorflow_synthetic_benchmark.py
```diff
@@ -0,0 +1,121 @@
...
+    img_sec_mean = np.mean(img_secs)
+    img_sec_conf = 1.96 * np.std(img_secs)
+    log('Img/sec per %s: %.1f +-%.1f' % (device, img_sec_mean, img_sec_conf))
+    log('Total img/sec on %d %s(s): %.1f +-%.1f' %
+        (hvd.size(), device, hvd.size() * img_sec_mean, hvd.size() * img_sec_conf))
+
+
+if tf.executing_eagerly():
+    with tf.device(device):
+        run(lambda: opt.minimize(loss_function, var_list=model.trainable_variables))
```
alsrgv: Don't we need to broadcast?  Does broadcast_global_variables() include model variables?

examples/tensorflow_synthetic_benchmark.py
```diff
@@ -0,0 +1,121 @@
+from __future__ import absolute_import, division, print_function
```
alsrgv: What's performance of this benchmark compared to native TF benchmark?

examples/tensorflow_synthetic_benchmark.py
tgaddair: It does, apparently.

examples/tensorflow_synthetic_benchmark.py
tgaddair: On 16 GPUs (8 nodes), I got 146 im/s with PyTorch, 140 im/s with native TF benchmark, and 135 im/s with this Horovod benchmark.

alsrgv: How about single node?

tgaddair: 189 im/s with the Horovod benchmark vs 195 im/s with the TF native benchmark.

alsrgv: Sweet! I suggest to pull out those #s in the PR description.  They're really nice - TF native benchmark is way more bloated.

PR #711, opened 2018-12-19
Switch Horovod Shutdown to OutOfRange error

abditag2:
If one worker processes fewer batches than the other workers or equivalently if it finishes the training with a smaller number of steps than the others, a SHUT_DOWN_ERROR will be raised in rank 0. translating this error into a `tf.error.UnknownError` causes the `tf.estimator.Estimator.train` function to finish prematurely without checkpointing the final model.  As specified here https://github.com/tensorflow/tensorflow/blob/a6d8ffae097d0132989ae4688d224121ec6d8f35/tensorflow/python/estimator/estimator.py#L303, `tf.estimator.Estimator.train` catches `tf.errors.OutOfRange` and can safely store the final model before terminating the process. Changing the error type solves this problem. 

The down side of this approach is that the rank 0 output does not print any error messages or exceptions. In the case that the SHUT_DOWN_ERROR was caused by another issue, like an internal bug or an MPI related problem, this would prevent the error from being propagated properly to the output. One potential way to address this point is to print an error message on rank 0 when a SHUT_DOWN_ERROR is raised. 

@alsrgv any thoughts?

### DIFF
--- a/horovod/tensorflow/mpi_ops.cc
+++ b/horovod/tensorflow/mpi_ops.cc
@@ -46,6 +46,8 @@ Status ConvertStatus(const common::Status& status) {
     return Status::OK();
   case common::UNKNOWN_ERROR:
     return errors::Unknown(status.reason());
+  case common::OUT_OF_RANGE_ERROR:
+    return errors::OutOfRange(status.reason());
   case common::PRECONDITION_ERROR:
     return errors::FailedPrecondition(status.reason());
   case common::ABORTED:
--- a/horovod/torch/adapter.cc
+++ b/horovod/torch/adapter.cc
@@ -146,7 +146,9 @@ void ThrowIfError(Status status) {
     throw std::runtime_error(status.reason());
   case StatusType::INVALID_ARGUMENT:
     throw std::invalid_argument(status.reason());
-  default: // Includes UNKNOWN_ERROR and OUT_OF_RANGE_ERROR
+  case StatusType::OUT_OF_RANGE_ERROR:
+    throw std::runtime_error(status.reason());
+  default: // Includes UNKNOWN_ERROR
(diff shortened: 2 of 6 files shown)


horovod/tensorflow/mpi_ops.cc
```diff
@@ -46,6 +46,8 @@ Status ConvertStatus(const common::Status& status) {
     return Status::OK();
   case common::UNKNOWN_ERROR:
     return errors::Unknown(status.reason());
+  case common::OUT_OF_RANGE_ERROR:
```
alsrgv: Can you add the mapping for PyTorch (old and new style adapters)?

abditag2:
@alsrgv Under both Estimator and MonitoredTrainingSession, if a random exception is thrown in rank 1, rank 1 process dies, and MPI on rank0 kills rank 0 and prints an error that rank 1 process died. However, if rank 1 sends a shut_down_error, Estimator on rank 1 saves the model before exiting, but MonitoredTrainingSession only terminates and does not save the mode.

horovod/torch/adapter.cc
```diff
@@ -146,7 +146,7 @@ void ThrowIfError(Status status) {
     throw std::runtime_error(status.reason());
   case StatusType::INVALID_ARGUMENT:
     throw std::invalid_argument(status.reason());
-  default: // Includes UNKNOWN_ERROR
+  default: // Includes UNKNOWN_ERROR and OUT_OF_RANGE_ERROR
```
alsrgv: Let's just have an explicit case, like we do for ABORTED error.

abditag2: changed.

alsrgv:
Will close this one as not needed.

PR #744, opened 2019-01-10
Fixed auto-tuning process to correctly bubble up parameters

tgaddair:


horovod/common/parameter_manager.cc
```diff
@@ -182,8 +182,12 @@ void ParameterManager::Tune(double score) {
     // Only do the tuning on the coordinator to ensure consistency.
     if (rank_ == root_rank_) {
       bool finished_tuning = true;
+      double last_score = score;
```
alsrgv: Should we replace `s/best_score/new_best_score/` and `s/last_score/best_score/` for clarity?

tgaddair: Done.

VincentLeeMax:
Hi @tgaddair, I can't figure out the purpose of this change. Can you give me more information about this change? Thanks..

PR #751, opened 2019-01-14
Filter warnings to avoid Travis CI failure

alsrgv:
We have so many deprecation warnings because of old TensorFlow/PyTorch, that Travis CI fails, e.g. https://api.travis-ci.org/v3/job/479219817/log.txt

### DIFF
--- a/test/test_keras.py
+++ b/test/test_keras.py
@@ -26,6 +26,7 @@ import numpy as np
 import os
 import tempfile
 import tensorflow as tf
+import warnings
 
 import horovod.keras as hvd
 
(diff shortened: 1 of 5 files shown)


abditag2:
On second thought, this will also prevent us from seeing the deprecation warnings and "encouraging" us to update them. We could potentially use an environment variable and disable the warnings only when running Travis tests.

alsrgv:
It is supposed to show the warning once per module, so we'll know it's there - it just won't spam as much.

alsrgv:
Logs are [readable again](https://travis-ci.org/uber/horovod/jobs/479230981)! Happiness.

PR #772, opened 2019-01-20
Optimized vectors and post-increments. [Refactored for Performance]

shahzadlone:
Change all post-increments to pre-increments where necessary to improve performance.

For example consider the following:
```
class Integer {

  public:

  // Sample implementation of pre-incrementing.
  Integer & operator++() {
    myInt += 1;
    return *this;
  }
  // Sample implementation of post-incrementing.
  const Integer operator++(int) {
    Integer temporary = *this;
    ++*this;
    return temporary;
  }

  private:

  int myInt;
};
```

The additional copy (temporary) in post-incrementing step is unnecessary.

---------------------------------------------------------------------------
Also optimized vectors which were not being reserved even though we knew
the exact number of elements that were going into it, this saves on
reallocation costs (and copies).
---------------------------------------------------------------------------

### DIFF
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -328,7 +328,7 @@ MPIResponse ConstructMPIResponse(std::unique_ptr<MessageTable>& message_table,
   // Check that all data types of tensors being reduced, gathered or broadcasted
   // are identical.
   auto data_type = requests[0].tensor_type();
-  for (unsigned int i = 1; i < requests.size(); i++) {
+  for (unsigned int i = 1; i < requests.size(); ++i) {
     auto request_type = requests[i].tensor_type();
     if (data_type != request_type) {
       error = true;
@@ -342,7 +342,7 @@ MPIResponse ConstructMPIResponse(std::unique_ptr<MessageTable>& message_table,
 
   // Check that all requested operations are the same
   auto message_type = requests[0].request_type();
-  for (unsigned int i = 1; i < requests.size(); i++) {
+  for (unsigned int i = 1; i < requests.size(); ++i) {
     if (error) {
       break;
     }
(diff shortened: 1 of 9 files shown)


shahzadlone:
Is this in progress?

shahzadlone:
Yes you are right, on average that's true. But why not optimize it further to be even faster (or better code at-least). I
In terms of noticing a difference I haven't done any bench marking, but if it's a big vector it should be faster, otherwise no big difference.

Haha yes, yes every line of code is another place for possible mistake but this line isn't it's a pure improvement. This is seen commonly across all good code bases (infact you can't find anywhere they don't do it, where possible).

alsrgv:
Let’s keep iterator chnages, but do array preservation only for code paths that happen at least once per cycle (every 5ms). I see that you modified one list serialization operation in mpi messages, but there is another one where you could do similar improvement - both of those are called every cycle.

horovod/common/operations.cc
```diff
@@ -1926,18 +1930,20 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
...
+  //       it != horovod_global.nccl_comms.end(); ++it) {
   //    ncclCommDestroy(it->second);
   //  }
   //#endif
 
   // Notify all outstanding operations that Horovod has been shut down
   // and clear up the tensor table and message queue.
   std::vector<StatusCallback> callbacks;
+  // we know (# of elements put) == (size of tensor_table map).
+  callbacks.reserve(state.tensor_table.size());
```
alsrgv: This happens only once, so let's revert this.

horovod/common/operations.cc
```diff
@@ -1094,7 +1096,7 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
       delete[] recvcounts;
       delete[] displcmnts;
 
-      for (size_t ec = 0; ec < entries.size(); ec++) {
+      for (size_t ec = 0; ec < entries.size(); ++ec {
```
alsrgv: Missing closing parenthesis.  Please compile the code locally before submitting.

horovod/common/operations.cc
shahzadlone: done

PR #778, opened 2019-01-24
Added DistributedOptimizerExt which supports virtually increasing batch size for TensorFlow

fps7806:
DistributedOptimizerExt allows simulation of a larger batch size by accumulating gradients over many iterations.
It extends horovod.tensorflow.DistributedOptimizer which allows multiplying the batch_size in a distributed setting.

Notes:
The class will wait until apply_gradients is called to accumulate gradients (rather than doing so in compute_gradients).

### DIFF
--- a/horovod/tensorflow/__init__.py
+++ b/horovod/tensorflow/__init__.py
@@ -248,6 +248,110 @@ class DistributedOptimizer(tf.train.Optimizer):
         return self._optimizer.variables(*args, **kwargs)
 
 
+def _var_key(var):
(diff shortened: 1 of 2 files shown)


alsrgv:
Notes from before:
1) We can either make the user call compute_gradients/aggregate_gradients N times before calling apply_gradients.
2) Or apply_gradients only actually apply_gradients every N times it is called

Option 1 makes the user do extra work and it's more aligned with PyTorch. Option 2 is not super intuitive.

PR #780, opened 2019-01-25
Fix out of sync error in mxnet test_broadcast

apeforest:
At rank 1, the `if` [block](https://github.com/uber/horovod/blob/master/test/test_mxnet.py#L276) will be processed and engine will issue a blocking call to tensor.asnumpy() and root_tensor.asnumpy().

However, this may cause out of sync error, because rank0 will skip this if block and may have already finished executing the hvd.broad_cast(tensor, root_rank) operation, which pass value of tensor to broadcast_tensor in rank 1. Adding a broadcast_tensor.wait_to_all() before the `if` block on line 276 will solve this problem. 

@yuxihu @ctcyang  please help to review.

### DIFF
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -83,12 +83,8 @@ def broadcast_parameters(params, root_rank=0):
         raise ValueError('invalid params of type: %s' % type(params))
 
     # Run broadcasts.
-    ret_list = []
     count = 0
-    for name, p in params:
+    for _, p in params:
         int_name = str(count)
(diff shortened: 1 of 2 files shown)


alsrgv:
@apeforest, I'm not sure I fully understand why this is supposed to work (I did observe before that it helps as well), since you mentioned that `broadcast_tensor.asnumpy()` is supposed to block until `broadcast_tensor` is computed.  Could you clarify?

alsrgv:
I will close this as unnecessary.

PR #782, opened 2019-01-25
Optimize Horovod Timeline and add Cycle Markers

alsrgv:


test/test_timeline.py
```diff
@@ -0,0 +1,57 @@
...
+    Tests for ops in horovod.torch.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(TimelineTests, self).__init__(*args, **kwargs)
+        warnings.simplefilter('module')
+
+    def test_timeline(self):
+        with tempfile.NamedTemporaryFile() as t:
+            os.environ['HOROVOD_TIMELINE'] = t.name
```
tgaddair: Safer thing to do would be to use:
```
with patch.dict('os.environ', {'HOROVOD_TIMELINE': t.name, 'HOROVOD_TIMELINE_MARK_CYCLES': '1'}):
   ...
```

horovod/common/operations.cc
```diff
@@ -900,8 +904,8 @@ void PerformOperation(TensorTable& tensor_table, MPIResponse response) {
     }
 
     unsigned int rank_displacement = 0;
-    for (unsigned rc = 0; rc < horovod_global.size; rc++) {
-      for (unsigned ec = 0; ec < entries.size(); ec++) {
+    for (int rc = 0; rc < horovod_global.size; rc++) {
+      for (size_t ec = 0; ec < entries.size(); ec++) {
```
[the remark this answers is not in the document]
alsrgv: Underlying types are different. `horovod_global.size` is int, while `entries.size()` are size_t.  This change is to avoid compilation warnings.

docs/timeline.md
```diff
@@ -41,3 +41,20 @@ workers were early and which were late.
...
+
+### Adding cycle markers
+
+Horovod performs work in cycles.  These cycles are used to aid [Tensor Fusion](tensor-fusion.md).
+
+Horovod has the ability to record the moment when each cycle starts for debugging of Tensor Fusion.
+
+![Cycle Markers](https://user-images.githubusercontent.com/16640218/51659458-64806100-1f5f-11e9-9a27-ba934ceec75f.png)
+
+Since this information makes timeline view very crowded it is not enabled by default.
```
sblotner: add a comma after "crowded"

PR #793, opened 2019-01-29
MXNet: support aggregation in optimizer & few bugfixes

yuxihu:
Recently we add [aggregated updates](https://github.com/apache/incubator-mxnet/pull/13346) in MXNet for SGD optimizer. The same aggregation behavior will extend to other optimizers. This requires changes in Horovod's DistributedOptimizer wrapper for MXNet to support allreduce multiple tensors in a single update call. This PR is to add this support. We are also working to add CI in MXNet for Horovod to identify such breaking API changes.

In addition, we fixed two issues:
1. We add a sync point in broadcast_parameters to make sure tensors pushed to MXNet engine got processed to avoid potential duplicate name error.
2. We fix horovod_mxnet_broadcast_async function to do in-place and not-in-place broadcast correctly. Thanks @alsrgv for pointing it out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,7 @@ if arg_params:
     hvd.broadcast_parameters(arg_params, root_rank=0)
 if aux_params:
     hvd.broadcast_parameters(aux_params, root_rank=0)
+model.set_params(arg_params=arg_params, aux_params=aux_params)
 
 # Train model
 model.fit(train_data,
--- a/README.md
+++ b/README.md
@@ -230,7 +230,11 @@ optimizer_params = ...
 opt = mx.optimizer.create('sgd', sym=model, **optimizer_params)
 opt = hvd.DistributedOptimizer(opt)
 
-hvd.broadcast_parameters(model.get_params(), root_rank=0)
+(arg_params, aux_params) = model.get_params()
+if arg_params:
+    hvd.broadcast_parameters(arg_params, root_rank=0)
+if aux_params:
+    hvd.broadcast_parameters(aux_params, root_rank=0)
 
 # Train model
 model.fit(train_data,
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -93,3 +93,6 @@ def broadcast_parameters(params, root_rank=0):
     for _, p in params:
         broadcast_(p, root_rank, str(count))
         count += 1
+
+    # Make sure the tensors pushed to MXNet engine get processed
+    mx.nd.waitall()
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -93,6 +93,6 @@ def broadcast_parameters(params, root_rank=0):
     for _, p in params:
         broadcast_(p, root_rank, str(count))
         count += 1
-
-    # Make sure the tensors pushed to MXNet engine get processed
-    p.wait_to_read()
+        if count == len(params):
+            # Make sure the tensor pushed to MXNet engine gets processed
+            p.wait_to_read()
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -92,7 +92,7 @@ def broadcast_parameters(params, root_rank=0):
     count = 0
     for _, p in params:
         broadcast_(p, root_rank, str(count))
+        # Make sure the tensor pushed to MXNet engine gets processed
+        p.wait_to_read()
         count += 1
 
-    # Make sure the tensors pushed to MXNet engine get processed
-    mx.nd.waitall()
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -93,6 +93,8 @@ def broadcast_parameters(params, root_rank=0):
     for _, p in params:
         broadcast_(p, root_rank, str(count))
         count += 1
-        if count == len(params):
-            # Make sure the tensor pushed to MXNet engine gets processed
-            p.wait_to_read()
+
+    # Make sure tensors pushed to MXNet engine get processed such that all
+    # workers are synced before starting training.
+    for _, p in params:
+        p.wait_to_read()
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -92,7 +92,7 @@ def broadcast_parameters(params, root_rank=0):
     count = 0
     for _, p in params:
         broadcast_(p, root_rank, str(count))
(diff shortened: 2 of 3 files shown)


yuxihu:
@apeforest @ctcyang please review

yuxihu:
@alsrgv Is there any issue for Python 3.6 container in the CI? They are failing here and in another two PRs.

alsrgv:
@yuxihu, yes - I'm fixing it...

horovod/mxnet/__init__.py
```diff
@@ -87,3 +93,6 @@ def broadcast_parameters(params, root_rank=0):
     for _, p in params:
         broadcast_(p, root_rank, str(count))
         count += 1
+
+    # Make sure the tensors pushed to MXNet engine get processed
+    mx.nd.waitall()
```
apeforest: It seems waitall will prevent exceptions from being rethrown. Please see this: https://mxnet.apache.org/versions/master/architecture/exception_handling.html

yuxihu:
Then we need to put wait_to_read back in the loop

horovod/mxnet/__init__.py
yuxihu: Then we need to put wait_to_read back in the loop

alsrgv: Oh, I see.  What if you omit explicit op names and allow internals to generate operation IDs automatically (sequentially)?

yuxihu: That can be an option but it might not be a good idea to apply here. This function is used to make sure each worker have the same set of parameters to start. We’d better make sure these ops gets executed such that each worker are synced before starting training.

yuxihu:
@alsrgv looks like one torch test hangs in two python 2.7 runs. could you please take a look? thanks.

horovod/mxnet/__init__.py
alsrgv: OK - is it possible to do the sync after the loop?  That way most of the operation would have a chance to finish before we start syncing the outputs.

yuxihu:
@alsrgv please help merge if it looks good to you. thanks.

horovod/mxnet/__init__.py
alsrgv: @yuxihu, does engine always process tensors sequentially, or it could do graph execution, in which case it may be a better idea to wait for all of them?

horovod/mxnet/__init__.py
```diff
@@ -87,3 +93,6 @@ def broadcast_parameters(params, root_rank=0):
     for _, p in params:
         broadcast_(p, root_rank, str(count))
```
alsrgv: Nit: let's change tensor name to `'broadcast_parameters.%d' % count` for better debuggability?

horovod/mxnet/__init__.py
```diff
@@ -43,23 +43,29 @@ def __getattr__(self, item):
     def create_state_multi_precision(self, index, weight):
         return self._optimizer.create_state_multi_precision(index, weight)
 
+    def _do_allreduce(self, index, grad):
+        if isinstance(index, (tuple, list)):
+            for i in range(len(index)):
+                allreduce_(grad[i], average=True, name=str(index[i]))
```
alsrgv: Same, let's name operations `'allreduce.%d' % index[i]` for clarity?

horovod/mxnet/__init__.py
```diff
@@ -87,3 +93,6 @@ def broadcast_parameters(params, root_rank=0):
     for _, p in params:
```
alsrgv: Since we're changing this piece, could we make `hvd.broadcast_parameters()` support a tuple, so that https://github.com/uber/horovod/blob/master/examples/mxnet_mnist.py#L120 could be simplified as advertised in `README.md` to:
```
hvd.broadcast_parameters(model.get_params(), root_rank=0)
```
?

horovod/mxnet/__init__.py
yuxihu: The engine process tensors which have dependencies sequentially. For independent tensors, each of them is processed in a thread simultaneously. Given that params here are independent, it might be a good idea to wait all. Now we have two options:

1. waitall after the loop: this method looks cleaner but it suppresses exceptions.
2. wait_to_read for each tensor in the loop: not so efficient but have all we want: throwing exception, wait for all tensors. Since broadcast_parameters is one time action for each training, we might be able to tolerate the little inefficiency.

Please let us know what's your opinion on these two options.

horovod/mxnet/__init__.py
yuxihu: We already prefix the index in mpiops.cc. See [here](https://github.com/uber/horovod/blob/master/horovod/mxnet/mpi_ops.cc#L186)

horovod/mxnet/__init__.py
yuxihu: I prefer not to change it. I am working on supporting MXNet Gluon trainer API, where we have a different type of params to broadcast and it is not a tuple.

horovod/mxnet/__init__.py
alsrgv: Ah, right. Nevermind.

horovod/mxnet/__init__.py
alsrgv: OK, sounds good.  Can we fix README.md, for now, to use the correct procedure for the time being then?

horovod/mxnet/__init__.py
alsrgv: How about having two loops.  In the first loop, you enqueue all broadcast_ ops, and in the second loop you do all the waiting?

yuxihu: Two loops sound good.

horovod/mxnet/__init__.py
yuxihu: Updated

README.md
```diff
@@ -230,7 +230,11 @@ optimizer_params = ...
 opt = mx.optimizer.create('sgd', sym=model, **optimizer_params)
 opt = hvd.DistributedOptimizer(opt)
 
-hvd.broadcast_parameters(model.get_params(), root_rank=0)
+(arg_params, aux_params) = model.get_params()
+if arg_params:
+    hvd.broadcast_parameters(arg_params, root_rank=0)
+if aux_params:
+    hvd.broadcast_parameters(aux_params, root_rank=0)
```
alsrgv: Do you need to do `model.set_params()`, like you do in the MNIST example?

yuxihu: Done

PR #810, opened 2019-02-05
Check CUDA runtime availability in MXNet when building Horovod

apeforest:
This PR performs two checks:
1) If HOROVOD_GPU_ALLREDUCE/BROADCAST/ALLGATHER is set but MXNet is not built with CUDA runtime, horovod installation should error out
2) If HOROVOD_GPU_ALLREDUCE/BROADCAST/ALLGATHER is not set but MXNet is built with CUDA runtime, horovod should be built with CUDA runtime.

### DIFF
--- a/horovod/mxnet/mpi_ops.cc
+++ b/horovod/mxnet/mpi_ops.cc
@@ -1,0 +1,320 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =============================================================================
+
+#include <atomic>
+
+#include "../common/operations.h"
+#include "adapter.h"
+#include "cuda_util.h"
+#include "mpi_ops.h"
+#include "ready_event.h"
+#include "tensor_util.h"
+
+namespace horovod {
+namespace mxnet {
+
+namespace {
+
+std::atomic_int op_count;
+
+std::string GetOpName(std::string prefix, char* name) {
+  if (name != nullptr) {
+    return prefix + "." + std::string(name);
+  }
+
+  op_count.fetch_add(1);
+  return prefix + ".noname." + std::to_string(op_count);
+}
+} // namespace
+
+inline void InvokeCompleteCallback(Callback on_complete, const Status& status) {
+  if (status.ok()) {
+    on_complete();
+  } else {
+    auto error = dmlc::Error(status.reason());
+    on_complete(&error);
+  }
+}
+
+void DoAllreduce(NDArray* tensor, NDArray* output, const std::string& name,
+                 Callback on_complete) {
+  ThrowIfError(common::CheckInitialized());
+
+  auto device = TensorUtil::GetDevice(tensor);
+  auto hvd_tensor = std::make_shared<MXTensor<NDArray>>(tensor);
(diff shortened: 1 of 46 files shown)


apeforest:
@alsrgv Added CUDA check per your suggestion. This PR should address your TODO 1) and 3) in https://github.com/uber/horovod/blob/buildkite/horovod/mxnet/mpi_ops.cc#L294. Please help to review.

apeforest:
@yuxihu @ctcyang Please also help to review. Thanks

yuxihu:
Have you run the unittests on GPU instance?

setup.py
```diff
@@ -622,6 +634,11 @@ def build_mx_extension(build_ext, options):
     build_ext.build_extension(mxnet_mpi_lib)
 
 
+def is_mx_cuda():
+    from mxnet import context
+    return context.num_gpus() > 0
```
alsrgv: That will not work if we're building Horovod in Docker on the machine that does not have GPUs.  For PyTorch we had to implement dummy compilation to test if CUDA is enabled in the binary.  If MXNet does not offer diagnostic flag to tell if it was built with CUDA support, maybe we can also do a test compilation?

apeforest: @alsrgv Could you elaborate on how you are going to test on the machine without GPUs? Do you mean you will build a docker image with CUDA and run it on a machine without GPU?

alsrgv: The use case is to build a docker image with CUDA on the machine without GPU, but run on the machine with GPU.

alsrgv:
Bump. @apeforest, I'm hoping we can get it in before the release.

apeforest:
@alsrgv Got too busy with other projects last week. I will work on this tomorrow.

alsrgv:
@apeforest, we've moved Horovod to its own organization, which requires DCO (dev certificate of origin).  Could you sign commits as per https://github.com/horovod/horovod/pull/810/checks?check_run_id=62082468?

horovod/mxnet/mpi_ops.cc
```diff
@@ -290,8 +290,7 @@ extern "C" int horovod_mxnet_broadcast_async(NDArray* input, NDArray* output,
     DoBroadcast(input, output, root_rank, op_name, on_complete);
   };
 
-#if HAVE_CUDA && HOROVOD_GPU_BROADCAST != 'M'
-  ThrowIfError(common::CheckInitialized());
```
alsrgv: We should keep this check here since it's not done in `DoBroadcastCudaOnCPU`.

setup.py
```diff
@@ -622,6 +633,16 @@ def build_mx_extension(build_ext, options):
     build_ext.build_extension(mxnet_mpi_lib)
 
 
+def is_mx_cuda():
+    from mxnet import runtime
+    features = runtime.libinfo_features()
```
yuxihu: This API is nice. Can we have an alternative approach to detect the cuda if the API is not there? In this way we do not lose Horovod support in the upcoming MXNet 1.4.0 release. Have you tried to detect the feature from libmxnet.so dependencies?

horovod/mxnet/mpi_ops.cc
apeforest: I have moved this into DoBroadcastCudaOnCPU to be consistent with other APIs.

apeforest:
> Could you run tests on GPU instance?

Currently tests fail on GPU instances when HOROVOD_GPU_(ALLREDUCE|BROADCAST|ALLGATHER) is not set to NCCL. The reason is that the branch of DoXXXCudaOnCPU was never covered by current tests before (it was either HAVE_CUDA=0 or HOROVOD_GPU_ALLREDUCE=NCCL). I am working to fix these errors now.

apeforest:
Close this PR for now and will resend once all tests are fixed.

alsrgv:
@apeforest, any luck with the PR?  Tomorrow is the release day, would be lovely to merge in before that.

apeforest:
@alsrgv The change is non-trivial. Even if I could push it in today, I think it's a little bit risky for the release tomorrow. What's the importance of this feature for the release and if so would it be possible to postpone the release by a few days?

alsrgv:
@apeforest, I think we can release without it then, and follow up with 0.16.1 in a week or so.

PR #813, opened 2019-02-06
Improved autotuning scoring and parameter search process

tgaddair:
- Use time between samples instead of aggregated time spent in the background thread for computing sample score
- Normalized scores after each sample to focus the Bayesian optimization more on certain ranges of parameters
- Increased noise parameter from 0.2 -> 0.8 to reflect apriori entropy expectations in samples

### DIFF
--- a/horovod/common/optim/bayesian_optimization.cc
+++ b/horovod/common/optim/bayesian_optimization.cc
@@ -37,7 +37,7 @@ void GetSufficientStats(std::vector<double>& v, double* mu, double* sigma) {
   *mu = sum / v.size();
 
   std::vector<double> diff(v.size());
-  std::transform(v.begin(), v.end(), diff.begin(), std::bind2nd(std::minus<double>(), *mu));
+  std::transform(v.begin(), v.end(), diff.begin(), [mu](double& x) { return x - *mu; });
   double sq_sum = std::inner_product(diff.begin(), diff.end(), diff.begin(), 0.0);
   *sigma = std::sqrt(sq_sum / v.size());
 }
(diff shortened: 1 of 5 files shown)


horovod/common/optim/bayesian_optimization.cc
```diff
@@ -31,6 +32,16 @@ namespace common {
 
 const double NORM_PDF_C = std::sqrt(2 * M_PI);
 
+void GetSufficientStats(std::vector<double>& v, double* mu, double* sigma) {
+  double sum = std::accumulate(v.begin(), v.end(), 0.0);
+  *mu = sum / v.size();
+
+  std::vector<double> diff(v.size());
+  std::transform(v.begin(), v.end(), diff.begin(), std::bind2nd(std::minus<double>(), *mu));
```
alsrgv: Does it take lambda instead of `std::bind2nd`?  Seems more reader-friendly :-)

horovod/common/optim/bayesian_optimization.cc
```diff
@@ -68,7 +79,11 @@ VectorXd BayesianOptimization::NextSample() {
 
   MatrixXd y_sample(y_samples_.size(), 1);
   for (unsigned int i = 0; i < y_samples_.size(); ++i) {
-    y_sample.row(i) = y_samples_[i];
+    double norm_score = (y_samples_[i] - mu) / sigma;
+
+    VectorXd y_i(1);
+    y_i(0) = norm_score;
+    y_sample.row(i) = y_i;
```
alsrgv: If I read this correctly, all observations are re-normalized every time, right?

horovod/common/optim/bayesian_optimization.cc
tgaddair: Good idea.  Done!

PR #826, opened 2019-02-13
Refactor operations into separate components by framework

tgaddair:
This change pulls out the various framework-specific preprocessor conditional logic into a class hierarchy that selects the appropriate operations (CUDA, NCCL, DDL, Hierarchical) based on available libraries and parameter selections.

Additionally, it abstracts much of the MPI-specific concepts into a CommunicationContext interface that can be substituted for a different framework for orchestration as desired.  This PR still retains coupling between operations.cc and MPI, but subsequent changes in the roadmap will further decouple operations.cc to use the CommunicationContext interface.


horovod/common/message/mpi_message.cc
```diff
@@ -15,67 +15,14 @@
 // =============================================================================
 
 #include "mpi_message.h"
-#include "wire/mpi_message_generated.h"
+
 #include <iostream>
 
+#include "../wire/mpi_message_generated.h"
+
```
alsrgv: Do we need to have mpi_message.cc and message.cc?  Why not just move everything to message.cc?  If we add Gloo backend or for coordination, we can still serialize those generic messages.

horovod/common/ops/cuda_operations.cc
```diff
@@ -0,0 +1,274 @@
...
+                             : AllreduceOp(comm_context, global_state), cuda_context_(context) {}
+
+Status CUDAAllreduce::Execute(std::vector<TensorTableEntry>& entries, const HorovodResponse& response) {
+  if (OnGPU(entries)) {
+    InitCUDA(entries);
+  }
+  return AllreduceOp::Execute(entries, response);
+}
+
+CUDACustomAllreduce::CUDACustomAllreduce(CUDAContext* context,
```
alsrgv: AbstractCUDAAllreduce

horovod/common/ops/cuda_operations.cc
```diff
@@ -0,0 +1,274 @@
...
+                                         CommunicationContext* comm_context,
+                                         HorovodGlobalState* global_state)
+: CUDAAllreduce(context, comm_context, global_state) {}
+
+Status CUDACustomAllreduce::Execute(std::vector<TensorTableEntry>& entries, const HorovodResponse& response) {
+  if (!OnGPU(entries)) {
+    return AllreduceOp::Execute(entries, response);
```
alsrgv: DefaultAllreduceOp?

horovod/common/ops/cuda_operations.cc
```diff
@@ -0,0 +1,274 @@
...
+    }
+  } else {
+    fused_input_data = first_entry.tensor->data();
+    buffer_data = (void*)first_entry.output->data();
+    num_elements = first_entry.tensor->shape().num_elements();
+    buffer_len = (size_t)first_entry.output->size();
+  }
+
+  void* host_buffer = nullptr;
+  CustomAllreduce(entries, stream, event_queue, fused_input_data, buffer_data, num_elements, buffer_len, host_buffer);
```
alsrgv: `DoAllreduce`

horovod/common/ops/cuda_operations.cc
```diff
@@ -0,0 +1,274 @@
...
+
+    for (auto& e : entries) {
+      timeline.End(e.tensor_name, e.output);
+      e.callback(Status::OK());
+    }
+  });
+
+  finalizer_thread.detach();
+
+  return Status::Finalizing();
```
alsrgv: `s/Finalizing/InProgress/`

horovod/common/ops/mpi_operations.h
```diff
@@ -0,0 +1,105 @@
...
+
+  void AllocateSharedBuffer(int64_t window_size, int element_size, void* baseptr, Communicator comm) override;
+
+  void FreeSharedBuffer() override;
+
+  void QuerySharedBuffer(int rank, void* baseptr) override;
+
+  void GetTypeSize(DataType dtype, int* out) override;
+
+  inline std::string AllreduceActivity() const override {
```
alsrgv: `AllreduceActivityName`

horovod/common/ops/nccl_operations.cc
```diff
@@ -0,0 +1,293 @@
...
+void NCCLContext::ErrorCheck(std::string op_name, ncclResult_t nccl_result) {
+  if (nccl_result != ncclSuccess) {
+    throw std::logic_error(std::string(op_name) + " failed: " + ncclGetErrorString(nccl_result));
+  }
+}
+
+
+NCCLAllreduce::NCCLAllreduce(NCCLContext* nccl_context,
+                             CUDAContext* cuda_context,
+                             CommunicationContext* comm_context,
```
alsrgv: `Orchestration`

horovod/common/ops/mpi_operations.h
tgaddair: Removed and moved this behavior to framework-specific operations.

horovod/common/ops/cuda_operations.cc
tgaddair: AllreduceOp is now abstract, default is now MPIAllreduce.

horovod/common/ops/cuda_operations.cc
tgaddair: Changed to CUDAAllreduceAsync to better reflect the different behavior.

horovod/common/ops/nccl_operations.cc
tgaddair: Changed to Channel to more accurately reflect its role as a separate collective communication channel, and made its use limited to specific frameworks (NCCL).

horovod/common/message.cc
```diff
@@ -14,65 +14,73 @@
...
+    case HOROVOD_NULL:
+      static const std::string null_("null");
+      return null_;
+    default:
+      static const std::string unknown("<unknown>");
+      return unknown;
   }
 }
 
 const std::string& MPIRequest::RequestType_Name(RequestType value) {
```
alsrgv: This can be renamed from `MPIRequest` to simply Request or HorovodRequest.  There should not be any MPI specifics.

horovod/common/ops/mpi_operations.h
```diff
@@ -0,0 +1,193 @@
...
+protected:
+  void DoAllreduce(std::vector<TensorTableEntry>& entries,
+                   const void* fused_input_data, void* buffer_data,
+                   int64_t& num_elements, size_t& buffer_len) override;
+
+  MPIChannel* mpi_channel_;
+};
+
+#if HAVE_CUDA
+class MPI_CUDAAllreduce : public CUDAAlrreduce {
```
alsrgv: Typo

horovod/common/ops/mpi_operations.h
```diff
@@ -0,0 +1,193 @@
...
+namespace common {
+
+// Table for storing Tensor metadata on rank zero. This is used for error
+// checking, stall checking and size calculations, as well as determining
+// when a reduction is ready to be done (when all nodes are ready to do it).
+using MessageTable = std::unordered_map<
+    std::string,
+    std::tuple<std::vector<MPIRequest>, std::chrono::steady_clock::time_point>>;
+
+class MPIChannel : public Channel {
```
alsrgv: This [MPI]Channel concept feels very confusing - it seems to be combining some coordinator state, an abstraction over MPI for collective calls, and even abstraction for MPI ops.  Let's remove it for now and inline MPI everywhere to make this change simpler.

horovod/common/ops/collective_operations.cc
```diff
@@ -0,0 +1,401 @@
...
+    RecordEventStart(MEMCPY_IN_FUSION_BUFFER, entries);
+    int64_t offset = 0;
+    for (auto& e : entries) {
+      void *buffer_data_at_offset = (uint8_t *) buffer_data + offset;
+      MemcpyInFusionBuffer(buffer_data_at_offset, e, entries);
+      offset += e.tensor->size();
+    }
+
+    StreamSynchronize(entries);
+    RecordEventEnd(MEMCPY_IN_FUSION_BUFFER, entries);
```
alsrgv: I got vastly confused here - "why would we synchronize and then record an event?", later to realize that in some cases one of these methods is empty, and in others the other method is empty.  I believe it will be very hard to follow for folks who want to extend Horovod.  I think your original assessment that there is sufficient discrepancy between these methods is correct.

Can we try to refactor operations in a different way - by having a set of helper classes/functions, but allowing each strategy to fully implement the logic that it needs?  Ideally, we'd have just a single level of inheritance (from interface), hopefully at a cost of minimal code duplication.

horovod/common/ops/mpi_operations.h
tgaddair: I suppose the biggest downside of doing so is that we'd need to then couple the NCCL operations with MPI (as NCCL uses MPI ops at several points, e.g., hierarchical allreduce).  

In the current design the MPIChannel allows us to use composition in NCCL ops in place of inheritance, so we can sub out a different CPU channel as desired.  Do you feel it's preferable to have NCCL ops include MPI headers and have them call MPI operations directly?

horovod/common/ops/collective_operations.cc
tgaddair: I made a change to replace all this with `StartMemcpyInFusionBuffer` and `EndMemcpyInFusionBuffer`.  Let me know if that makes things more clear.

I'm thinking about how we could best implement the entire logic at the op level without duplicating a bunch of code.  The biggest issue is that all the fusion buffer stuff is completely generic, except for the actual implementation of the memcpy.  It's possible to expose functions that do fusion buffering in and out and passing in function pointers to do the work, but it could get equally messy.

Perhaps we should chat more in person.

horovod/common/message.cc
tgaddair: Done in #841.

horovod/common/ops/mpi_operations.h
tgaddair: Removed channel concept.

horovod/common/ops/mpi_operations.cc
```diff
@@ -0,0 +1,405 @@
...
+
+  if (entries.size() > 1) {
+    auto& buffer = global_state_->fusion_buffer.GetBuffer(
+        first_entry.device, first_entry.context->framework());
+    buffer_data = const_cast<void*>(buffer->AccessData(first_entry.context));
+
+    // Copy memory into the fusion buffer. Then the input data of each
+    // process is assumed to be in the area where that process would
+    // receive its own contribution to the receive buffer.
+    timeline.ActivityStartAll(entries, MEMCPY_IN_FUSION_BUFFER);
```
alsrgv: Should this logic be in abstract Allgather()?

horovod/common/ops/nccl_operations.cc
```diff
@@ -0,0 +1,294 @@
...
+
+    ncclUniqueId nccl_id;
+    if (nccl_rank == 0) {
+      nccl_context_->ErrorCheck("ncclGetUniqueId", ncclGetUniqueId(&nccl_id));
+    }
+
+    MPI_Broadcast(*mpi_context_, (void*) &nccl_id, sizeof(nccl_id), HOROVOD_BYTE, 0, nccl_id_bcast_comm);
+
+    ncclComm_t new_nccl_comm;
+    auto nccl_result = ncclCommInitRank(&new_nccl_comm, nccl_size, nccl_id, nccl_rank);
```
alsrgv: In mpi_operations, you created wrappers for MPI calls, should we align these two approaches?  (Either remove wrappers in MPI, or add here).

alsrgv:
cc @karakusc @romerojosh @apeforest @yuxihu, can you guys review this PR as well and provide feedback?  We want to make it easier to contribute enhancements to Horovod core operations.

horovod/common/ops/mpi_operations.cc
```diff
@@ -0,0 +1,405 @@
...
+                         (int) num_elements,
+                         ctx.GetMPIDataType(first_entry.tensor),
+                         first_entry.tensor->dtype() == HOROVOD_FLOAT16 ? ctx.mpi_float16_sum : MPI_SUM,
+                         ctx.GetMPICommunicator(comm));
+  if (op != MPI_SUCCESS) {
+    throw std::logic_error("MPI_Allreduce failed, see MPI output for details.");
+  }
+}
+
+void MPI_Allgatherv(MPIContext& ctx, const void* sendbuf, int sendcount, DataType sendtype,
```
alsrgv: Add to utilities files instead, instead of *_operations.cc/h

horovod/common/ops/mpi_operations.cc
```diff
@@ -0,0 +1,405 @@
...
+      return local_comm;
+    case CROSS:
+      return cross_comm;
+    default:
+      throw std::logic_error("Communicator " + CommunicatorName(comm) +
+                             " is not supported in MPI mode.");
+  }
+}
+
+void MPI_Allreduce(MPIContext& ctx, const void* buffer_data, int64_t num_elements,
```
romerojosh: Nit. This `MPI_Allreduce` wrapper is unlike the others in that it passes a `TensorTableEntry` as an argument to get the datatype, while all the other ones just pass the datatype directly. For consistency, should this do the same?

horovod/common/ops/mpi_operations.cc
romerojosh: On a related note, I find the MPI wrappers in this file to be a bit confusing generally. Are these wrappers intended to be used only in the files defining the various collective operation implementations, and not intended to be used to wrap MPI elsewhere (i.e. for coordination in `operations.cc`)?  Additionally, leaving the wrapper names unmodified from the existing unwrapped MPI API calls leads to some confusion as well. 

Could you clear up what the intent/expected use of these wrappers is?

horovod/common/operations.cc
```diff
@@ -272,14 +121,40 @@ const Status DUPLICATE_NAME_ERROR = Status::InvalidArgument(
...
-#define OP_ERROR(entries, error_message)                                       \
-  {                                                                            \
-    for (auto& e : (entries)) {                                                \
-      timeline.End(e.tensor_name, nullptr);                                    \
-      e.callback(Status::UnknownError(error_message));                         \
-    }                                                                          \
-    return;                                                                    \
-  }
+OperationManager* CreateOperationManager(HorovodGlobalState& state) {
+  // Order of these operations is very important. Operations will be checked sequentially from the first
```
karakusc: Would it make sense to have an explicit decision function to choose an implementation for each op dynamically? I feel like the sequential ordering makes the decision a bit implicit, and might make it harder to write more complex logic later, for example dynamically choosing an algorithm on a per-tensor basis.

horovod/common/ops/mpi_operations.h
```diff
@@ -0,0 +1,161 @@
...
+#include "collective_operations.h"
+
+#if HAVE_CUDA
+#include "cuda_operations.h"
+#endif
+
+namespace horovod {
+namespace common {
+
+struct MPIContext {
```
karakusc: I feel like `MPIContext` should be defined somewhere like `common.h`, since it is being used in the control plane as well. Might be more intuitive to have `mpi_operations.cc` and `mpi_operations.h` pertain only to the ops themselves. Also, the `Communicator` enum is defined there, so I think it would make more sense to have all the MPI communicator-related code in one place so that it is easier to try out custom communicators while developing.

horovod/common/ops/mpi_operations.cc
karakusc: I agree about the confusion related the the MPI wrappers. What is the motivation for the wrappers, is it to clean up the op implementation from the MPI error handling logic? Or is it simply to standardize access to the MPI communicators through `MPIContext`?

I think if the wrappers are used, the exposed API should be identical to the underlying core MPI API (with maybe a standard modification on it, something like `MPI_Bcast_(...)` instead of `MPI_Bcast(...)` ), to make things more consistent.

horovod/common/ops/nccl_operations.cc
tgaddair: Removed wrappers.

horovod/common/ops/mpi_operations.cc
tgaddair: I went ahead and removed the wrappers.

For context, the wrappers were originally added as an abstraction above MPI, in case we wanted to sub in another framework that supports similar collective primitives (for NCCL).  However, the design evolved so that NCCL became coupled to MPI for now, so the wrappers no longer have much use beyond the convenience of not having to convert an error op code into an exception.

Given that they provide little benefit in the current design, and that most devs working in this codebase are accustomed to using MPI functions directly, I went ahead and removed them.

horovod/common/ops/mpi_operations.cc
tgaddair: Went ahead and removed these wrappers (see above comment).

horovod/common/operations.cc
tgaddair: Personally, I think the sequencing approach reduces the complexity (but perhaps I'm biased from writing so many behavior trees over the years).

There's some complication that arises from using an explicit rule, which is that some of the ops available at runtime will be dynamic (based on what frameworks you have installed: NCCL, DDL, etc.), while others will be available given some parameter state, and still others (as you mention) will be available on a per-tensor basis (which is available to the Enabled function via the Response parameter).

Using an explicit rule, the operation manager becomes very complicated with preprocessor code, needing to couple itself to what frameworks are installed (because certain ops will not exist at compile time).  Using the sequential approach, all of that is isolated to a single function in `operations.cc`, which I think makes the separation between "what ops are available at compile time" and "what ops are available at runtime" more clear.

horovod/common/ops/mpi_operations.h
tgaddair: Good idea!  The MPIContext is a more generic object used in more places (like NCCL), so it makes sense to move it out of there.  I put it into a new file `mpi_context.h`.  In my mind, it makes sense to avoid putting it in `common.h`, since there is interest in decoupling Horovod from MPI in the future.  With this separation, we keep Horovod concepts independent of MPI.  What do you think?

horovod/common/ops/mpi_operations.cc
tgaddair: Good idea, done.

horovod/common/ops/cuda_operations.cc
```diff
@@ -0,0 +1,226 @@
...
+      e.callback(Status::OK());
+    }
+  });
+
+  finalizer_thread.detach();
+
+  return Status::InProgress();
+}
+
+void CUDAAllreduceAsync::StartMemcpyInFusionBuffer(std::vector<TensorTableEntry>& entries) {
```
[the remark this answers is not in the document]
tgaddair: Added comments in `cuda_operations.h` on the `event_queue_` member variable.  Let me know what you think.

horovod/common/ops/nccl_operations.cc
```diff
@@ -0,0 +1,374 @@
...
+
+  // Determine GPU IDs of the devices participating in this communicator.
+  std::vector<int32_t> nccl_device_map;
+  nccl_device_map.reserve(global_state_->local_comm_ranks.size());
+  for (int rank : global_state_->local_comm_ranks) {
+    nccl_device_map.push_back(devices[rank]);
+  }
+
+  InitCUDA(entries);
+  InitComm(entries, nccl_device_map);
```
alsrgv: InitNCCLComm?

horovod/common/operations.cc
karakusc: Makes sense. Would it be easy to override the decision logic implied by this list if needed? Is it sufficient to call `execute()` for the desired op in `OperationManager::ExecuteAllreduce`, or would it be more complicated?

horovod/common/ops/mpi_operations.h
karakusc: Agreed!

horovod/common/operations.cc
tgaddair: Yes, it should be a pretty simple change to enforce the use a certain operation under certain conditions.  It would just require giving the `OperationManager` a direct pointer to the operation or having a map (right now, we don't directly reference the operations, only store them in the list).

alsrgv:
🥳

PR #868, opened 2019-02-28
Address segfault when using MXNet MKLDNN backend

wuxun-zhang:
When I ran horovod with MXNet mkldnn backend, I got a segmentation fault. Then, I checked the file setup.py, there are no option found to specify the backend on which MXNet was built.  So, I added the option 'MXNET_USE_MKLDNN' when building horovod from source. At the same time, we should set environment variable 'MXNET_USE_MKLDNN=1'  before installing horovod. Now, the segfault is addressed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -277,8 +277,12 @@ def get_mx_flags(build_ext, cpp_flags):
     mx_libs = get_mx_libs(build_ext, mx_lib_dirs, cpp_flags)
 
     compile_flags = []
+    has_mkldnn = is_mx_mkldnn()
     for include_dir in mx_include_dirs:
         compile_flags.append('-I%s' % include_dir)
+        if has_mkldnn:
+            mkldnn_include = os.path.join(include_dir, 'mkldnn')
+            compile_flags.append('-I%s' % mkldnn_include)
 
     link_flags = []
     for lib_dir in mx_lib_dirs:
@@ -626,6 +630,34 @@ def parse_version(version_str):
     return version
 
 
+def is_mx_mkldnn():
+    try:
+        from mxnet import runtime
+        features = runtime.Features()
+        return features.is_enabled('MKLDNN')
+    except Exception:
+        msg = 'INFO: Cannot detect if MKLDNN is enabled in MXNet. Please \
+            set MXNET_USE_MKLDNN=1 if MKLDNN is enabled in your MXNet build.'
+        if 'linux' not in sys.platform:
+            # MKLDNN is only enabled by default in MXNet Linux build. Return 
+            # False by default for non-linux build but still allow users to 
+            # enable it by using MXNET_USE_MKLDNN env variable. 
+            print(msg)
+            return os.environ.get('MXNET_USE_MKLDNN', '0') == '1'
+        else:
+            try:
+                import mxnet as mx
+                mx_libs = mx.libinfo.find_lib_path()
+                for mx_lib in mx_libs:
+                    output = subprocess.check_output(['readelf', '-d', mx_lib])
+                    if 'mkldnn' in str(output):
+                        return True
+                    return False
+            except Exception:
+                print(msg)
+                return os.environ.get('MXNET_USE_MKLDNN', '0') == '1'
+
+
 def build_mx_extension(build_ext, options):
     check_mx_version()
     mx_compile_flags, mx_link_flags = get_mx_flags(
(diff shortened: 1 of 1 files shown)


wuxun-zhang:
please @pengzhao-intel @apeforest @yuxihu help review. Thanks

setup.py
```diff
@@ -636,6 +636,12 @@ def build_mx_extension(build_ext, options):
         mxnet_mpi_lib.define_macros += [('MSHADOW_USE_CUDA', '1')]
     else:
         mxnet_mpi_lib.define_macros += [('MSHADOW_USE_CUDA', '0')]
+    
+    if os.environ.get('MXNET_USE_MKLDNN'):
+        mxnet_mpi_lib.define_macros += [('MXNET_USE_MKLDNN', os.environ.get('MXNET_USE_MKLDNN'))]
+    else:
+        mxnet_mpi_lib.define_macros += [('MXNET_USE_MKLDNN', 1)]
```
yuxihu: Setting to 1 by default causes CI failure as we are using non-mkldnn mxnet package in the CI. I think we need to set it to 0 by default.

setup.py
```diff
@@ -636,6 +636,12 @@ def build_mx_extension(build_ext, options):
         mxnet_mpi_lib.define_macros += [('MSHADOW_USE_CUDA', '1')]
     else:
         mxnet_mpi_lib.define_macros += [('MSHADOW_USE_CUDA', '0')]
+    
+    if os.environ.get('MXNET_USE_MKLDNN'):
```
yuxihu: Can we automatically detect if mkldnn is enabled in MXNet? Using an env variable is not very good user experience. We added [runtime feature](https://github.com/apache/incubator-mxnet/commit/3a6fe22b660797f0b6f7d0530c604024b6f84bae) discovery API recently which might be able to used here.

yuxihu:
I went ahead to test it and made the following changes. The idea is to use the runtime discovery API to detect if MKLDNN is enabled in MXNet by default. If there is no runtime discovery API yet (e.g. 1.4.0 release), we will take MXNET_USE_MKLDNN env variable. I think this can be further automated by checking if libmkldnn is one of the dependent libraries of MXNet when the API is not there.

Adding a new function:
```
def is_mx_mkldnn():
    try:
        from mxnet import runtime
        features = runtime.Features()
        return features.is_enabled('MKLDNN')
    except ImportError:
        return os.environ.get('MXNET_USE_MKLDNN', '0') == '1'
```
In build_mx_extension():
```
    if is_mx_mkldnn():
        mxnet_mpi_lib.define_macros += [('MXNET_USE_MKLDNN', '1')]
    else:
        mxnet_mpi_lib.define_macros += [('MXNET_USE_MKLDNN', '0')]
```

yuxihu:
@wuxun-zhang Created a [PR](https://github.com/apache/incubator-mxnet/pull/14300) in MXNet to add MKLDNN headers to include dir, which is needed for this PR to work. Also, please refer to these [changes](https://github.com/yuxihu/horovod/commit/4924a67477bb032ab589a27331fd41a3dabeb872) to update the setup.py here. Thanks.

wuxun-zhang:
@yuxihu Many thanks for your quick suggestions. I'll update this PR soon.

setup.py
```diff
@@ -626,6 +630,34 @@ def parse_version(version_str):
     return version
 
 
+def is_mx_mkldnn():
+    try:
+        from mxnet import runtime
+        features = runtime.Features()
+        return features.is_enabled('MKLDNN')
+    except Exception:
+        msg = 'INFO: Cannot detect if MKLDNN is enabled in MXNet. Please \
```
apeforest: Do we still need this message if we are checking MKLDNN library using readelf?

setup.py
```diff
@@ -626,6 +630,34 @@ def parse_version(version_str):
...
+                import mxnet as mx
+                mx_libs = mx.libinfo.find_lib_path()
+                for mx_lib in mx_libs:
+                    output = subprocess.check_output(['readelf', '-d', mx_lib])
+                    if 'mkldnn' in str(output):
+                        return True
+                    return False
+            except Exception:
+                print(msg)
+                return os.environ.get('MXNET_USE_MKLDNN', '0') == '1'
```
apeforest: Should we just return false in this case?

yuxihu: we want to give users an option to set it if we fail to automatically detect it, just like for non-linux build.

apeforest: If user sets env variable to 1 but there is no libraries available, this will still cause the segmentation fault, right?

yuxihu: When we are here, it means that we failed to automatically detect if mkldnn is enabled in the mxnet library, we do not know if mkldnn is there or not. So we give users the option to set it by themselves.

apeforest: Discussed offline. This seems not an issue,  but leave option to user.

setup.py
```diff
@@ -626,6 +630,34 @@ def parse_version(version_str):
...
+            return os.environ.get('MXNET_USE_MKLDNN', '0') == '1'
+        else:
+            try:
+                import mxnet as mx
+                mx_libs = mx.libinfo.find_lib_path()
+                for mx_lib in mx_libs:
+                    output = subprocess.check_output(['readelf', '-d', mx_lib])
+                    if 'mkldnn' in str(output):
+                        return True
+                    return False
```
apeforest: Move this return outside of for block

PR #872, opened 2019-02-28
Update MXNet example in README.md

eric-haibin-lin:
No need to call `split_and_load `, which is meant for splitting data when the number of context is greater than 1. Calling `as_in_context` should be sufficient. Also I made the training loop simpler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,15 +258,12 @@ loss_fn = ...
 for epoch in range(num_epoch):
     train_data.reset()
     for nbatch, batch in enumerate(train_data, start=1):
-        data = gluon.utils.split_and_load(batch.data[0], ctx_list=[context],
-                                          batch_axis=0)
-        label = gluon.utils.split_and_load(batch.label[0], ctx_list=[context],
-                                           batch_axis=0)
+        data = batch.data[0].as_in_context(context)
+        label = batch.label[0].as_in_context(context)
         with autograd.record():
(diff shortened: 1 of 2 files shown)


alsrgv:
cc @yuxihu, @apeforest 

@eric-haibin-lin, thanks for the PR, could you update it with [DCO](https://github.com/horovod/horovod/pull/872/checks?check_run_id=70703648)?

README.md
```diff
@@ -258,15 +258,12 @@ loss_fn = ...
 for epoch in range(num_epoch):
     train_data.reset()
     for nbatch, batch in enumerate(train_data, start=1):
-        data = gluon.utils.split_and_load(batch.data[0], ctx_list=[context],
-                                          batch_axis=0)
-        label = gluon.utils.split_and_load(batch.label[0], ctx_list=[context],
-                                           batch_axis=0)
+        data = batch.data[0].as_in_context(context)
```
yuxihu: Could you please apply the same change to the [imagenet example](https://github.com/horovod/horovod/blob/master/examples/mxnet_imagenet_resnet50.py#L339)?

eric-haibin-lin: will do

alsrgv:
Thanks for the PR!  Looks like https://github.com/pytorch/vision/pull/768 broke our integration tests.  I'll see if they make a fix today, otherwise, I'll fix our integration tests and ask you to rebase.

PR #879, opened 2019-03-04
Fix detecting MKLDNN in MXNet

yuxihu:
Fix a bug in detecting MKLDNN in MXNet


yuxihu:
@apeforest please review

PR #886, opened 2019-03-05
Simplify mnnist example using Gluon API and update README

apeforest:
Current mxnet examples uses an old module API which has been gradually replaced by a more user-friendly Gluon API. Replacing the example and README using Gluon API for mxnet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,15 +214,10 @@ Horovod supports MXNet and regular TensorFlow in similar ways.
 
 See full training [MNIST](examples/mxnet_mnist.py) and [ImageNet](examples/mxnet_imagenet_resnet50.py) examples.
 
-**Note**: we recommend users to build MXNet from source following this [guide](https://mxnet.incubator.apache.org/install/build_from_source.html) when running Horovod with MXNet on a Linux OS with GCC version 5.X and above. The MXNet shared library distributed through MXNet pip package is currently built using GCC 4.8.4. If we build and install Horovod on a Linux OS with GCC 5.X+ with MXNet pip package, we will hit segmentation fault due to std::function definition change from GCC [4.X](https://github.com/gcc-mirror/gcc/blob/gcc-4_8_4-release/libstdc++-v3/include/std/functional#L2069) to GCC [5.X](https://github.com/gcc-mirror/gcc/blob/gcc-5_4_0-release/libstdc++-v3/include/std/functional#L1854).
-
-There are two ways to train a model using MXNet: [Gluon](http://mxnet.incubator.apache.org/api/python/gluon/gluon.html) API (preferred) and [Module](http://mxnet.incubator.apache.org/api/python/module/module.html) API. Here we provide the building block for each set of API to train a model using MXNet with Horovod.
-
-###### Gluon API
 ```python
-from mxnet import autograd, gluon
 import mxnet as mx
 import horovod.mxnet as hvd
+from mxnet import autograd, gluon
 
 # Initialize Horovod
 hvd.init()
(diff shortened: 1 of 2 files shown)


apeforest:
@yuxihu @eric-haibin-lin Please help to review. Thanks!

examples/mxnet_mnist.py
```diff
@@ -1,27 +1,29 @@
...
                     help='number of training epochs (default: 5)')
-parser.add_argument('--lr', type=float, default=0.05,
-                    help='learning rate (default: 0.05)')
-parser.add_argument('--momentum', type=float, default=0.5,
-                    help='SGD momentum (default: 0.5)')
+parser.add_argument('--lr', type=float, default=0.01,
+                    help='learning rate (default: 0.01)')
+parser.add_argument('--momentum', type=float, default=0.9,
+                    help='SGD momentum (default: 0.9)')
+parser.add_argument('--no-cuda', action='store_true', default=True,
```
yuxihu: how to set no_cuda to False?

apeforest: Good catch. Changed argument to -use-gpu

alsrgv:
How about using something like this: https://github.com/horovod/horovod/commit/a5fb7a215a8a62cd5b8862fa819f6371050321d9#diff-10729f4cf94906d5762883f7163f91dbR17 & https://github.com/horovod/horovod/commit/a5fb7a215a8a62cd5b8862fa819f6371050321d9#diff-10729f4cf94906d5762883f7163f91dbR27 ?

examples/mxnet_mnist.py
```diff
@@ -1,29 +1,36 @@
...
+import time
 
-import horovod.mxnet as hvd
 import mxnet as mx
+import horovod.mxnet as hvd
+from mxnet import autograd, gluon, nd
 from mxnet.test_utils import download
 
 # Training settings
 parser = argparse.ArgumentParser(description='MXNet MNIST Example')
```
eric-haibin-lin: This may be a general question for all horovod examples: 
There are a lot of examples for ResNet on ImageNet. This is great. But I do not find instructions on how to run those example (e.g. with x learning rate and y batch size on z GPUs this example converges to 75.5% accuracy). This will help people validate that everything is working when they try out the example.

README.md
```diff
@@ -214,15 +214,10 @@ Horovod supports MXNet and regular TensorFlow in similar ways.
...
-
-###### Gluon API
 ```python
-from mxnet import autograd, gluon
 import mxnet as mx
 import horovod.mxnet as hvd
+from mxnet import autograd, gluon
 
 # Initialize Horovod
 hvd.init()
```
eric-haibin-lin: Is the parameter hook for gluon already done? As we discussed last time, if the shape of a parameter has to be inferred after a batch of data is seen, the parameter will not be broadcast. Therefore, the correctness of the program depends on whether all workers are initialized with the same random seed. Maybe we should call it out as a known issue if it is not yet fixed.

yuxihu: Not yet. I will look into it soon.

examples/mxnet_mnist.py
apeforest: @eric-haibin-lin The default values are the ones we used to achieve 75.5% reported. Users can reproduce this using the defalt argument values.

README.md
eric-haibin-lin: I created an issue in case others it affects others: https://github.com/horovod/horovod/issues/895

yuxihu: @eric-haibin-lin I think you should mention Gluon in the title of the issue. It does not affect Module API.

PR #888, opened 2019-03-06
Fixed parameter manager to use the previously set value for Bayesian parameters when changing a free parameter to a constant

tgaddair:
Signed-off-by: Travis Addair

### DIFF
--- a/horovod/common/parameter_manager.cc
+++ b/horovod/common/parameter_manager.cc
@@ -404,6 +404,11 @@ ParameterManager::BayesianParameter::BayesianParameter(
 
 void ParameterManager::BayesianParameter::SetValue(BayesianVariable variable, double value, bool fixed) {
   if (fixed) {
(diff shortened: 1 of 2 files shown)


horovod/common/parameter_manager.cc
```diff
@@ -388,11 +398,17 @@ ParameterManager::BayesianParameter::BayesianParameter(
     test_points_(test_points),
     iteration_(0) {
   ResetBayes();
+  Reinitialize(FilterTestPoint(0));
   ResetState();
 }
 
 void ParameterManager::BayesianParameter::SetValue(BayesianVariable variable, double value, bool fixed) {
   if (fixed) {
+    int32_t index = index_[variable];
```
alsrgv: Scratching my head about what does this do - could you add comments?

tgaddair: Added comments, let me know if that helps clear it up.

PR #899, opened 2019-03-10
fix a mistake about HOROVOD_CYCLE_TIME

zrss:
@alsrgv , it seems a mistake about `-x HOROVOD_FUSION_THRESHOLD` in the doc


alsrgv:
@zrss, any reason for closing the PR?  Would be good to merge the fix.

PR #902, opened 2019-03-11
Enable faster worker coordination via response caching.

romerojosh:
This PR enables faster global coordination between workers via a response caching mechanism. It replaces the redundant MPI message passing of tensor information between workers at every step with a cache of responses and coordination via a cheap global intersection of bitvectors constructed from the worker cache states (carried out with `MPI_Allreduce` with `MPI_BAND`). We've seen major improvements to scaling performance at high node counts with this change. 

The cache capacity can be set at runtime via environment variable `HOROVOD_CACHE_CAPACITY`. By default, the cache capacity is set to 1024 responses. Furthermore, caching is limited to allreduce operations only in the current version.   

Thanks to @alsrgv for early feedback on this PR!


alsrgv:
Love it!  A bit swamped to review in details at the moment, planning to look into it next week during the GTC :-)

romerojosh:
Excellent! In the meantime, I am looking over failing TensorFlow tests. The results seem to indicate a couple outstanding issues we will need to discuss when time allows:

1. For graph-based operation, the cache state is being preserved between the tests, with reused tensor names causing problematic name collisions. I can resolve these tests by adding a cache reset each time `hvd.init()` is called. This begs the question though: Do you expect user scripts with graph based execution to result in these types of name collisions due to re-initialization of Tensorflow with a persistent Horovod instance?

2. For eager operation, there are even more issues with name collisions as TF doesn't assign names to any of the tensors in this mode (all tensors enqueued to allreduce are just named `HorovodAllreduce`). If I disambiguate the names with an index suffix, the tests pass. However, I think the correct solution here would be to disable caching if running in eager mode. I am not quite sure the best way to go about achieving this at the moment (perhaps introducing some kind of flag to eager allreduce requests to indicate they are  uncacheable?)

tgaddair:
Hey Josh,

   1. I do know of some users who call hvd.init() multiple times in the
   same script for such a reason.  So presumably there could also be users who
   are re-initializing TensorFlow without re-initializing Horovod as well.
   But I like the idea of clearing the cache in the init() function so at
   least we can solve those API issues if they arise.
   2. I'm curious as to how you're running eager execution.  I seem to
   recall that we got around this issue in our DistributedOptimizer by
   coercing Horovod operations into a graph.  I believe that should have
   enforced that all tensors have uniquely identifiable IDs (if not, then I'm
   not sure how the parallelism / tensor fusion is working in eager mode).
   Are calling allreduce directly without going through the
   DistributedOptimizer in eager mode?

Thanks, I'll continue going through the PR today.

romerojosh:
1. Okay, that is good info. I have a new commit with a `reset()` method for the `ResponseCache` that I can call in the `InitializeHorovodOnce` function. I'll hold off pushing it for now though.

2. I am just running the eager execution tests in the `test_tensorflow.py` script. In fact, at least on my machine, it seems like the tests in that file are *only* run in eager mode by default (unless all the related eager mode changes are commented out). If I inspect the names provided to `EnqueueTensorAllreduce`, I see that all tensors have the same names. Is tensor fusion supposed to be working with eager mode anyway? I assumed that with eager enabled, each all reduce op would block until completion, yielding no fusion anyway (also would explain why it can handle the lack of different tensor names).

tgaddair:
When running eagerly under normal conditions (not the unit tests), users make use of the `hvd.DistributedOptimizer` class, which defines a [subgraph](https://github.com/horovod/horovod/blob/master/horovod/tensorflow/__init__.py#L212) over all the gradients in a batch that need to be allreduced.  

This results in tensors being assigned names and being submitted asynchronously, such that tensor fusion can take place (this also improves performance by about about 6x ;)).

In the unit tests, we call `allreduce` functions directly, which is why you're seeing tensors with no names.  So, long story short, the caching should still work in eager mode, provided the user is using `DistributedOptimizer` and not calling `allreduce` directly.  

So my current thinking is that disambiguating the tensors with an index suffix as we do in [PyTorch](https://github.com/horovod/horovod/blob/master/horovod/torch/mpi_ops.cc#L39) is not a bad idea to address the issue in the unit tests.

alsrgv:
Could we support name reuse with a different shape by treating it as a cache miss scenario, which would raise cache miss bit and subsequently broadcast the MPIRequest with all the details?  As Travis mentioned, it's not a big production use case, but it would be good to support it nevertheless.

romerojosh:
@alsrgv, this would certainly be possible. The current issue is that `Responses`, unlike `Requests`, do not store any information on the tensor shape, which I think would be necessary to make this happen. Basically, when I check if a `Request` is cached with an existing name, I need to be able to compare the candidate request shape with the shape associated with the currently cached `Response`, however that information is currently unavailable. Similar requirements would apply to checking if a `Response` is cached as well. Are you suggesting this as an alternative to the index suffix disambiguation?

romerojosh:
To add to the conversation about this, I can think of a scenario in which name reuse could be problematic, even if we add additional logic to consider changes in tensor shape or device as cache misses. If you consider a variant of the `test_horovod_allreduce_multi_gpu` test with 2 ranks, the following could occur:
1. rank 0 reduces `tensor` on gpu 0, rank 1 reduces `tensor` on gpu 1
2. rank 0 reduces `tensor` on gpu 0, rank 1 reduces `tensor` on gpu 2

If the names of the reduce ops are not disambiguated and we rely on recaching based on changes in the device id, this would actually hang indefinitely. The reason for this is that for rank 0, the second allreduce would be cached, while for rank 1, the second allreduce would trigger a cache miss due to a change in device. This triggers a situation where rank 0 will wait forever for a matching cache hit on rank 1, while rank 1 will trigger the cache miss bit, ship off the request to be recorded in the global message table of rank 0 and wait to hear back from rank 0 about this "new" uncached response. This type of situation is remedied by the index suffix method, as each allreduce op, regardless of device, is treated separately.

alsrgv:
You're highlighting one interesting property of Horovod - you can't always know outcome based on your local information, and your local information may correspond to multiple global outcomes (e.g. from viewpoint of rank 0 both 1) and 2) are possible).

I'm still a believer that you could be reusing the same names, as long as you don't try to mix the names at exactly the same time.

One possibility could be to cache based on `Request` and allow multi-matching to few `Response`s.

romerojosh:
Caching based on `Request` would still fall victim to the scenario I posted earlier. In that case, rank 0 would see the same exact `Request` and believe it was cached, and rank 1 would see a unique `Request`, causing the same issue. The problem here is that by allowing multiple responses per tensor name, we lose cache coherence between workers.  

I have an idea on how to resolve this. Instead of having a name conflict trigger a cache miss, we can add an additional cache invalidate state (when a request has the same name but one or more of the  parameters doesn't match the existing cached `Response`). In the bitvector, we can add an additional status bit that indicates if any worker has cache entries the need to be invalidated. If this status is set, all workers proceed to do a second `MPI_Allreduce` on a bitvector with bits that need to be invalidated flipped. Then all workers erase entries associated with the invalidated bits and maintain a coherent cache. In my bad example above, this would cause rank 0 to realize it cannot use a cached response and will proceed to share information with rank 1 to update. 

We could alternatively append these additional bits to the first bitvector, but my preference would be to avoid bloating that initial message size for what I assume should be a rare situation. 

To enable this, I will modify the cache so that the entries are a pair/tuple of a `Response` and tensor params (taken from the `tensor_table`), but the cache will still remain keyed by tensor name only. 

Let me know what you think!

alsrgv:
I like the invalidation mechanism.  I wonder if params from the tensor table should be in the cache, or they could be taken directly from the tensor table?  (didn't have a chance to do pass over the code yet :-( )

romerojosh:
I think the tensor params need to be stored in the cache since the tensor table entries are erased during `PerformOperation`. 

No worries on not having looked through the code yet. I am close to committing changes with the cache invalidation in place which should pass all tests.

alsrgv:
I wonder if we should keep them there then, and use TensorTable as a part of the cache?

romerojosh:
That is possible, but you would also need to deal with name collisions in the TensorTable as well then. Keeping the essential tensor params in the cache isn't too bad actually. I have a small struct which stores only the dtype, the tensor shape, and the device.

alsrgv:
Makes sense.  I think looking at the code will be most helpful at this point :-)

romerojosh:
Changes will be up shortly.

horovod/common/response_cache.h
```diff
@@ -0,0 +1,84 @@
...
+
+  void erase_response(uint32_t cache_bit);
+
+  void update_cache_bits();
+
+private:
+  void put_(const Response& response, TensorParams& params);
+
+  uint32_t capacity_;
+  std::list<std::pair<Response, TensorParams>> cache_;
```
alsrgv: Can you put comments for these three structures describing their roles?  Maybe also rename more semantically?

horovod/common/operations.cc
```diff
@@ -445,6 +449,134 @@ int64_t TensorFusionThresholdBytes() {
   return proposed_fusion_threshold;
 }
 
+// Populates provided ResponseList with responses from deque.
+void PopulateResponseList(ResponseList& response_list,
+                             std::deque<Response>& responses,
```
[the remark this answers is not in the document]
romerojosh: Done. I limited the scope of the reformatting to lines of code I changed. Otherwise, there is a lot more code that would've been modified.

horovod/common/common.cc
```diff
@@ -109,5 +109,10 @@ int64_t TensorShape::num_elements() const {
   return result;
 }
 
+// TODO: Better name for this?
+const std::vector<int64_t>& TensorShape::shape() const {
```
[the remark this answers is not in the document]
romerojosh: Done.

horovod/common/operations.cc
```diff
@@ -794,6 +1052,9 @@ void BackgroundThreadLoop(HorovodGlobalState& state, MPIContext& ctx) {
     state.timeline.Initialize(std::string(horovod_timeline),
                               static_cast<unsigned int>(size));
   }
+  if (horovod_timeline != nullptr) {
+    state.timeline_enabled = true;
```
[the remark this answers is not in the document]
romerojosh: The issue here is that `state.timeline.Initialized()` is only true on rank 0, but for the timeline handling with the bitvector, all ranks need to be aware that the timeline is active. The `state.timeline_enabled` flag is true on all ranks if the timeline is enabled to handle this.

horovod/common/operations.cc
```diff
@@ -638,6 +770,132 @@ void set_int_from_env(const char* env, int& val) {
...
+void GetCommonCacheAndState(std::set<int>& cache_hits,
+                            std::set<int>& invalid_bits,
+                            bool& uncached_in_queue,
+                            bool& invalid_in_queue,
+                            bool& should_shut_down,
+                            HorovodGlobalState& state,
+                            MPIContext& ctx) {
+  // Resize and reset bit vector.
+  int nbits = state.response_cache.current_size() + NUM_STATUS_BITS;
+  int count = (nbits + sizeof(long long) * CHAR_BIT - 1) /
```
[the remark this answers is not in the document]
romerojosh: The formula here `(nbits + nbits_in_long_long - 1) / nbits_in_long_long` will provide exactly the right number of `long long` words required to cover `nbits`.

horovod/common/operations.cc
```diff
@@ -638,6 +770,132 @@ void set_int_from_env(const char* env, int& val) {
...
+    state.cache_mask[shift] |= (1ull <<
+                               (shifted_bit % (sizeof(long long) * CHAR_BIT)));
+    if (state.timeline_enabled) {
+      state.cache_mask[count + shift] ^= (1ull <<
+                                         (shifted_bit % (sizeof(long long) * CHAR_BIT)));
+    }
+  }
+
+  // Global MPI AND operation to get intersected bit array.
+  MPI_Allreduce(MPI_IN_PLACE, state.cache_mask.data(), fullcount,
```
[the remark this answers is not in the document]
romerojosh: We make calls that mutate the cache identically on all workers (i.e. the number of active bits in the cache should be consistent). Thus, `count`, and also `fullcount`, should be the same. Note that all workers will have `fullcount = 2 * count` in the case that the timeline is enabled (related to comment above about new timeline boolean).

horovod/common/operations.cc
```diff
@@ -638,6 +770,132 @@ void set_int_from_env(const char* env, int& val) {
...
+    // Erase cache entries associated with invalid bits.
+    for (auto bit : invalid_bits) {
+      state.response_cache.erase_response(bit);
+    }
+  }
+
+  if (state.timeline_enabled) {
+    // For timeline, check for entries with cache hits on *any* worker to
+    // mark start of negotiation phase.
+    for (int i = 0; i < count; ++i) {
```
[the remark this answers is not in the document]
romerojosh: Could you elaborate a bit more on what you would like the API to this utility function look like?

horovod/common/response_cache.h
romerojosh: Added comments and renamed the data structures. This definitely makes things more clear I think. Let me know if you think there are better names to use.

horovod/common/timeline.cc
```diff
@@ -186,6 +186,11 @@ void Timeline::NegotiateStart(const std::string& tensor_name,
   if (!initialized_) {
     return;
   }
+  // Note: Need to enable repeated calls to this routine during negotiate
+  // phase. First call takes precedence.
+  if (tensor_states_[tensor_name] == TimelineState::NEGOTIATING) {
```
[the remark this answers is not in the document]
romerojosh: Moved the guard ahead and fleshed out the comment. Basically, this is required to handle the timeline management when using the cache. In this case, `NegotiateStart` gets called every time the extra bit indicated that _some_ workers hit the cache for this response. The negotiation phase finishes when either all workers get a cache hit for the response, or if the response proceeds to be handled by the non-bypass path.

horovod/common/operations.cc
```diff
@@ -638,6 +770,132 @@ void set_int_from_env(const char* env, int& val) {
...
+                            HorovodGlobalState& state,
+                            MPIContext& ctx) {
+  // Resize and reset bit vector.
+  int nbits = state.response_cache.current_size() + NUM_STATUS_BITS;
+  int count = (nbits + sizeof(long long) * CHAR_BIT - 1) /
+              (sizeof(long long) * CHAR_BIT);
+
+  // Allocate additional bits for timeline states if required.
+  int fullcount = count;
+  if (state.timeline_enabled) fullcount *= 2;
```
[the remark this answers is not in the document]
romerojosh: Added braces around all ifs.

horovod/common/operations.cc
```diff
@@ -638,6 +770,132 @@ void set_int_from_env(const char* env, int& val) {
...
+
+  state.cache_mask.resize(fullcount);
+
+  std::memset(&state.cache_mask[0], 0, count * sizeof(long long));
+  if (state.timeline_enabled) {
+    std::memset(&state.cache_mask[count], -1, count * sizeof(long long));
+  }
+
+  // Set reserved bits for additional states.
+  if (!should_shut_down) state.cache_mask[0] |= (1ull);
```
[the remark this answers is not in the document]
romerojosh: Added a `StatusBit` enum for these three state bits to `common.h`.

horovod/common/operations.cc
```diff
@@ -819,6 +1080,15 @@ void BackgroundThreadLoop(HorovodGlobalState& state, MPIContext& ctx) {
                                        true);
   }
 
+  // Override response cache capacity, if it's set.
+  // TODO: Put in param manager?
```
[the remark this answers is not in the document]
romerojosh: So I made a first pass in adding this to param manager, but it would be nice if @tgaddair could review and make sure things look alright. 

Before this though, I think there may be a bug in the current master branch with the autotuner. I get `std::bad_alloc` exceptions thrown when the autotuner is active after a few hundred steps (seems to occur when the parameters are updated). This was with a TF RN50 case with 2 processes. Can one of you verify that the autotuner is still in a working state?

horovod/common/operations.cc
```diff
@@ -1316,6 +1628,19 @@ bool RunLoopOnce(HorovodGlobalState& state, MPIContext& ctx, bool is_coordinator
       }
     }
 
+    // Add supported responses to cache.
+    if (state.response_cache.capacity() > 0) {
+      {
```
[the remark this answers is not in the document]
romerojosh: Removed extra braces.

horovod/common/response_cache.cc
```diff
@@ -0,0 +1,196 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =============================================================================
+
+#include "response_cache.h"
+#include <iostream>
```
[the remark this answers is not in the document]
romerojosh: `<iostream>` wasn't even needed so I just removed it.

horovod/common/response_cache.cc
```diff
@@ -0,0 +1,196 @@
...
+
+uint32_t ResponseCache::capacity() const {return capacity_;}
+
+size_t ResponseCache::current_size() const {return cache_.size();}
+
+ResponseCache::CacheState ResponseCache::cached(const Request& message) const {
+  auto it = table_.find(message.tensor_name());
+  if (it != table_.end()) {
+    // If entry associated with this request already exists in cache, check
+    // if tensor parameters match. If not, mark entry to be invalidated.
```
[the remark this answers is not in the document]
romerojosh: Very true. Updated the comment to be more precise.

horovod/common/response_cache.h
```diff
@@ -0,0 +1,84 @@
...
+  DataType dtype;
+  std::vector<int64_t> shape;
+  int32_t device;
+};
+
+// LRU cache of Responses
+class ResponseCache {
+public:
+  enum CacheState {
+    MISS = 0, HIT = 1, INVALIDATE = 2
```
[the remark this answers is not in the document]
romerojosh: Done.

horovod/common/operations.cc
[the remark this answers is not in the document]
alsrgv: Ah, makes sense.

horovod/common/operations.cc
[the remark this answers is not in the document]
alsrgv: I see, so it is rounding up :-)

horovod/common/timeline.cc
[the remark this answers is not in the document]
alsrgv: I see, makes sense.  Can you add this to the comment?

horovod/common/operations.cc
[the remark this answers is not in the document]
romerojosh: If it's useful, reverting commit aa605d6de1e07d1184e8793b6d4f65cee9d4efb3 seems to fix the exception issue.

horovod/common/timeline.cc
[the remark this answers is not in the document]
romerojosh: Not sure why this isn't showing up as outdated. Is the longer comment in the most recent committed code sufficient?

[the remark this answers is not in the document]
alsrgv: LGTM, thanks!

horovod/common/operations.cc
```diff
@@ -1000,32 +1338,118 @@ bool RunLoopOnce(HorovodGlobalState& state, MPIContext& ctx, bool is_coordinator
...
 
   // Collect all tensors that are ready to be reduced. Record them in the
   // tensor count table (rank zero) or send them to rank zero to be
   // recorded (everyone else).
   std::vector<std::string> ready_to_reduce;
   if (is_coordinator) {
     while (!message_queue.empty()) {
+      // Skip cached messages.
+      if (state.response_cache.capacity() > 0 &&
+          state.response_cache.cached(message_queue.front()) ==
```
alsrgv: If we're just skipping these, can we skip them on line 1415 and not have this code?

horovod/common/operations.cc
```diff
@@ -1282,6 +1604,12 @@ bool RunLoopOnce(HorovodGlobalState& state, MPIContext& ctx, bool is_coordinator
     RequestList message_list;
     message_list.set_shutdown(should_shut_down);
     while (!message_queue.empty()) {
+      // Skip cached messages.
+      if (state.response_cache.cached(message_queue.front()) ==
```
alsrgv: Same as other other question.

horovod/common/response_cache.cc
```diff
@@ -0,0 +1,223 @@
...
+  bits_outdated_ = true;
+}
+
+void ResponseCache::put(const Response& response, TensorTable& tensor_table) {
+  if (capacity_ == 0) {
+    return;
+  }
+
+  // If response is fused, split back into individual responses
+  if (response.tensor_names().size() > 0) {
```
alsrgv: CLion suggests to replace all `.size() > 0` with `!.empty()`

horovod/common/response_cache.cc
```diff
@@ -0,0 +1,223 @@
...
+    cache_.push_front(std::make_pair(response, std::move(params)));
+  }
+
+  cache_iters_[cache_bit] = cache_.begin();
+  tensor_name_to_bit_[response.tensor_names()[0]] = cache_bit;
+
+  bits_outdated_ = true;
+}
+
+void ResponseCache::put(const Response& response, TensorTable& tensor_table) {
```
alsrgv: make tensor_table const?

horovod/common/operations.cc
```diff
@@ -1325,6 +1664,11 @@ bool RunLoopOnce(HorovodGlobalState& state, MPIContext& ctx, bool is_coordinator
       LOG(TRACE, state.rank) << "Finished performing " << response.tensor_names_string();
     }
 
+    if (state.response_cache.capacity() > 0) {
+      // Reassign cache bits based on current cache state.
+      state.response_cache.update_cache_bits();
```
alsrgv: Should we move this to line 1655, i.e. in prior if statement?

horovod/common/response_cache.cc
```diff
@@ -0,0 +1,223 @@
...
+  return std::get<0>(*cache_iters_[cache_bit]);
+}
+
+const Response& ResponseCache::peek_response(uint32_t cache_bit) const {
+  assert(cache_bit < cache_iters_.size());
+  return std::get<0>(*cache_iters_[cache_bit]);
+}
+
+uint32_t ResponseCache::peek_cache_bit(const Request& message) const {
+  assert(this->cached(message));
```
alsrgv: Add assert that bits are not outdated?

horovod/common/response_cache.cc
```diff
@@ -0,0 +1,223 @@
...
+  cache_.push_front(std::move(*it));
+  cache_.erase(it);
+  cache_iters_[cache_bit] = cache_.begin();
+  bits_outdated_ = true;
+  return cache_.front().first;
+}
+
+const Response& ResponseCache::peek_response(const Request& message) const {
+  assert(this->cached(message));
+  uint32_t cache_bit = tensor_name_to_bit_.at(message.tensor_name());
```
alsrgv: Add assert that bits are not outdated?

horovod/common/response_cache.cc
```diff
@@ -0,0 +1,223 @@
...
+            cache_params.dtype == params.dtype &&
+            cache_params.shape == params.shape)
+               ? CacheState::HIT
+               : CacheState::INVALID;
+  } else {
+    return CacheState::MISS;
+  }
+}
+
+void ResponseCache::put_(const Response& response, TensorParams& params) {
```
alsrgv: Seems that there's a lot of logic spent maintaining consistency, but yet there's still `update_cache_bits()` required. 

What if we make this more batch-oriented, i.e. collect all responses this cycle, and once we call `update()`, new `cache_` will be created that is `new responses` + `responses from old cache_ that don't have same tensors as new responses`.  Then we don't need iterators and the process is super simple, right?

horovod/common/operations.cc
```diff
@@ -638,6 +765,146 @@ void set_int_from_env(const char* env, int& val) {
...
+  int count = (nbits + sizeof(long long) * CHAR_BIT - 1) /
+              (sizeof(long long) * CHAR_BIT);
+
+  // Allocate additional bits for timeline states if required.
+  int fullcount = count;
+  if (state.timeline_enabled) {
+    fullcount *= 2;
+  }
+
+  state.cache_mask.resize(fullcount);
```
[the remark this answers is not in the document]
romerojosh: I originally did this to save on reallocation. Will be resolved with bitvector class abstraction.

horovod/common/operations.cc
```diff
@@ -638,6 +765,146 @@ void set_int_from_env(const char* env, int& val) {
   }
 }
 
+// Routine to intersect cache hits across workers. Also determines global
+// shutdown state and whether uncached requests exist on any worker.
+void GetCommonCacheAndState(std::set<int>& cache_hits,
```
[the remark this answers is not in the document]
romerojosh: Done.

horovod/common/operations.cc
```diff
@@ -1000,32 +1338,118 @@ bool RunLoopOnce(HorovodGlobalState& state, MPIContext& ctx, bool is_coordinator
...
+    {
+      // Get lock in order to safely replace messages to global queue
+      std::lock_guard<std::mutex> guard(state.mutex);
+
+      // Remove uncommon cached tensors from queue and replace to state
+      // queue for next cycle.
+      size_t num_messages = message_queue.size();
+      for (size_t i = 0; i < num_messages; ++i) {
+        auto message = message_queue.front();
+        if (!(state.response_cache.cached(message) ==
```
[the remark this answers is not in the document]
romerojosh: Done, but modified to also include the "Skip cached message" logic which was removed.

horovod/common/operations.cc
```diff
@@ -1000,32 +1338,118 @@ bool RunLoopOnce(HorovodGlobalState& state, MPIContext& ctx, bool is_coordinator
...
     }
   }
 
   if (!message_queue.empty()) {
     LOG(DEBUG, state.rank) << "Sent " << message_queue.size() << " messages";
   }
 
-  // Flag indicating that the background thread should shut down.
-  bool should_shut_down = state.shut_down;
+  if (state.response_cache.capacity() > 0) {
```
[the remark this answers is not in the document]
romerojosh: Done.

horovod/common/operations.cc
```diff
@@ -445,6 +448,130 @@ int64_t TensorFusionThresholdBytes() {
   return proposed_fusion_threshold;
 }
 
+// Populates provided ResponseList with responses from deque.
+void PopulateResponseList(ResponseList& response_list,
```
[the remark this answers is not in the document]
romerojosh: Done.

horovod/common/operations.cc
```diff
@@ -954,6 +1233,65 @@ void BackgroundThreadLoop(HorovodGlobalState& state, MPIContext& ctx) {
   }
 }
 
+// If all messages in queue have responses in cache, use fast path with
+// no additional MPI coordination.
+void RunBypass(std::queue<Request>& message_queue, std::set<int>& cache_hits,
+               HorovodGlobalState& state, MPIContext& ctx) {
+
+  // Clear message queue.
+  message_queue = std::queue<Request>();
```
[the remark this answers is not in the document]
romerojosh: Since we skip adding cached messages to the `message_queue`, this clear is no longer necessary. Removed.

horovod/common/operations.cc
romerojosh: Done.

horovod/common/operations.cc
romerojosh: Done.

horovod/common/response_cache.cc
romerojosh: The original statement here was actually incorrect. It should've been `response.tensor_names().size() > 1`. Corrected it.

horovod/common/response_cache.cc
romerojosh: Done.

horovod/common/operations.cc
romerojosh: Done.

horovod/common/response_cache.cc
romerojosh: An assert on `bits_outdated_` would not be correct here. This is because in the case that invalid elements are erased and `erase_response` is called, this flag is set to be true. However, we still require extracting response information on the bits we collected before erasing. See https://github.com/horovod/horovod/blob/11a5d6d365960efd90374144cffb83d558e259e1/horovod/common/operations.cc#L1411.

horovod/common/response_cache.cc
romerojosh: Similar to other comment. Responses need to be visible here: https://github.com/horovod/horovod/blob/11a5d6d365960efd90374144cffb83d558e259e1/horovod/common/operations.cc#L1517

horovod/common/operations.cc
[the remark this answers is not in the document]
romerojosh: Would hiding all this cache bitvector logic into a separate class, with methods to set, intersect, and get cache/timeline/invalid bits as well as states work for you? I'd aim to have the exposure in the `operations.cc` level just be manipulating and working with set of bits.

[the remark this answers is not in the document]
alsrgv: I think that'd be great!

horovod/common/operations.cc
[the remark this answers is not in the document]
tgaddair: The changes look good to me.  I haven't seen the autotuner issue you're describing before, but I'll play around with a few examples and see if I can reproduce it.

horovod/common/response_cache.cc
```diff
@@ -0,0 +1,223 @@
...
+    TensorParams params;
+    params.device = tensor_entry.device;
+    params.dtype = tensor_entry.tensor->dtype();
+    params.shape = tensor_entry.tensor->shape().to_vector();
+
+    this->put_(response, params);
+  }
+}
+
+const Response& ResponseCache::get_response(const Request& message) {
```
alsrgv: CLion is telling me this method is not used, remove?

horovod/common/response_cache.cc
```diff
@@ -0,0 +1,223 @@
...
+  assert(cache_bit < cache_iters_.size());
+  auto it = cache_iters_[cache_bit];
+  cache_.push_front(std::move(*it));
+  cache_.erase(it);
+  cache_iters_[cache_bit] = cache_.begin();
+  bits_outdated_ = true;
+  return cache_.front().first;
+}
+
+const Response& ResponseCache::peek_response(const Request& message) const {
```
alsrgv: CLion is telling me this method is not used, remove?

horovod/common/response_cache.cc
```diff
@@ -0,0 +1,223 @@
...
+    this->clear();
+  }
+
+  capacity_ = capacity;
+  cache_iters_.reserve(capacity);
+}
+
+uint32_t ResponseCache::capacity() const { return capacity_; }
+
+size_t ResponseCache::current_size() const { return cache_.size(); }
```
alsrgv: CLion is telling me this method is not used, remove?

horovod/common/response_cache.cc
```diff
@@ -0,0 +1,223 @@
...
+  uint32_t cache_bit = tensor_name_to_bit_[message.tensor_name()];
+  auto it = cache_iters_[cache_bit];
+  cache_.push_front(std::move(*it));
+  cache_.erase(it);
+  cache_iters_[cache_bit] = cache_.begin();
+  bits_outdated_ = true;
+  return cache_.front().first;
+}
+
+const Response& ResponseCache::get_response(uint32_t cache_bit) {
```
alsrgv: Let's add more comments, e.g. how this method pulls the cached response to the front of the cache?

horovod/common/operations.cc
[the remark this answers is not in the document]
romerojosh: I added a `CacheCoordinator` class which encapsulates all the bit vector communication into a `sync()` method. This class also cleans up some of the various bit set and state flag handling in `operations.cc`. Let me know what you think.

horovod/common/response_cache.cc
romerojosh: Removed.

horovod/common/response_cache.cc
romerojosh: Removed.

horovod/common/response_cache.cc
romerojosh: Removed.

horovod/common/response_cache.cc
romerojosh: I added some more comments to the various methods in `response_cache.cc`. Please take a look and let me know if anything jumps out at you that we can simplify.

horovod/common/operations.cc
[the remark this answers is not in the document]
romerojosh: @tgaddair I am able to produce the error using the `tensorflow_synthetic_benchmark.py` file with Horovod 0.16.1 installed with pip. Running with the command `mpirun -np 2 -x HOROVOD_AUTOTUNE=1 -x HOROVOD_AUTOTUNE_LOG=at.log python tensorflow_synthetic_benchmark.py --num-iters 300` generates the exception to after reporting iteration 23. This doesn't occur with Horovod 0.16.0.

horovod/common/operations.cc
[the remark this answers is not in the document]
alsrgv: Looks great!

horovod/common/operations.cc
```diff
@@ -954,6 +1126,61 @@ void BackgroundThreadLoop(HorovodGlobalState& state, MPIContext& ctx) {
...
+  // recently used responses get priority.
+  std::deque<Response> responses;
+  for (auto bit : cache_coordinator.cache_hits()) {
+    responses.push_back(state.response_cache.get_response(bit));
+  }
+
+  // Fuse responses as normal.
+  auto response_list = FuseResponses(responses, state, ctx);
+
+  if (!response_list.responses().empty()) {
```
alsrgv: Seems like lines 1141-1181 are currently copy-pasted three times, can we extract them as a helper functions?

horovod/common/response_cache.h
alsrgv: Looks good, thanks!

horovod/common/operations.cc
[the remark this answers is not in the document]
tgaddair: Let me see if I can reproduce and get back to you.

alsrgv:
Left a couple of comments, otherwise LGTM!

alsrgv:
@romerojosh, can you rebase on the latest master for the CI to work?

horovod/common/operations.cc
romerojosh: Alright, I tried to consolidate some of the repeated code. I wrote a small helper function to get the tensor names and size data for the autotuner. I also merged the coordinator and worker rank logic after the response list broadcast to finally remove some code repetition that has existed in the code for a while. The remaining repeated code is just a few function calls now and logging code.

alsrgv: Maybe we want to merge it as well, but perhaps another time :-)

romerojosh:
Hey Alex, addressed your comment about repeated code and tried to consolidate some of it. I also added a bit more commentary around the cache usage in `operations.cc` to try and be a bit more clear about where the cache order is mutated.

Any updates on the autotuner @tgaddair? I'd like to be able to test that functionality before this is merged, but am still blocked by the error I'm seeing with master.

tgaddair:
> Hey Alex, addressed your comment about repeated code and tried to consolidate some of it. I also added a bit more commentary around the cache usage in `operations.cc` to try and be a bit more clear about where the cache order is mutated.
> 
> Any updates on the autotuner @tgaddair? I'd like to be able to test that functionality before this is merged, but am still blocked by the error I'm seeing with master.

Apologies for the delay @romerojosh.  Fix in #995.

tgaddair:
@romerojosh #995 has been landed to master.

romerojosh:
Thanks for the update @tgaddair! I rebased with the change and the autotuner seems to be working with the cache. Are there are ways to limit the choices for caching to something a bit more discrete though? I don't think the cache capacity value needs to be as precise as it currently is now (since the biggest difference in impact will be due to two factors (1. Is the cache on or off? and 2. Can all my responses fit into the cache?) Also, I want to make sure the autotuner can fairly test the case where the cache is disabled (i.e. `HOROVOD_CACHE_CAPACITY=0`) since that may be a better choice if there are responses that cannot be cached. Any ideas?

@alsrgv, I do have one last thought on this PR before merge concerning stall messaging. As it stands now, if all responses are cached, there will be no stall messages as the cached messages are not handled via the `message_table`. However, the cache path does not by default generate the kind of global data used in the current stall messaging . Rather than have no stall messaging at all, maybe we should at least add a generic stall message if no cycles have proceeded to perform an operation in the stall check interval?

alsrgv:
@romerojosh, we could add a categorical variable for caching (0 vs non-zero), but every categorical variable makes tuning process X times longer (X - cardinality of the variable).  It may be worth doing for this use case.  Do you want to give it a shot?

As for stalls - that's a great call out and we definitely need stall logic.  One idea I have is to run a full coordination loop if the stall is detected and call out slow nodes that way.  Do we / can we keep enough information to do that?

romerojosh:
Concerning the autotuner, I can try adding the categorical variable for cache on/off. If it saves optimization time, we can also just fix the cache capacity to a large value (1024 is the current default, but we can make it larger), and have the autotuner focus *only* on whether the cache is on or off and not the capacity. This might miss cases that could benefit from a larger cache capacity, but if we set the default large enough, that case should be pretty rare. What do you think?

For stall messaging, forcing the full coordination path is a good idea. I should be able to do this with a couple of additional status bits in the bit vector and an additional timer that tracks the time between when any worker reports a cache hit and calls to `PerformOperation`. If this time exceeds a threshold, we set a status bit that triggers a cache reset on all workers, forcing the full coordination path and the normal stall message handling. Should be straightforward to implement.

alsrgv:
We could start with that (0 or 1024) and see how it goes.  We could also emit a one-time warning once cache capacity is reached with the recommendation of how it can be increased.

Your proposal of resetting cache in case of a stall sounds good, let's do it :-)

romerojosh:
Okay, just pushed a few commits implementing changes to the autotuner treatment of cache capacity as well as stall message handling for cached messages. Please take a look when you can.

horovod/common/operations.cc
```diff
@@ -1282,14 +1294,29 @@ bool RunLoopOnce(HorovodGlobalState& state, MPIContext& ctx, bool is_coordinator
...
       should_shut_down |= CheckForStalledTensors(state, ctx);
       state.last_stall_check = std::chrono::steady_clock::now();
     }
+
+    // Check if response cache should be cleared to trigger stall warnings.
+    if (state.response_cache.capacity() > 0) {
+        if (!state.cached_messages_pending) {
+          // If no outstanding cached messages, reset perform operation clock.
+          state.last_perform_operation = std::chrono::steady_clock::now();
+        } else if (std::chrono::steady_clock::now() - state.last_perform_operation >
```
alsrgv: Is there any way for us to track that there's a tensor that's been sitting in the queue for more than X time rather than tracking perform operations, add that as a bit flag and OR it?  It's possible that there has been no tensors to reduce for a while (running a large validation), or (less likely): there are two application threads, and one thread keeps reducing something, while other thread is stalled.

alsrgv:
Thanks! Autotuner change LGTM, have a comment about stall change.

horovod/common/operations.cc
romerojosh: The `cached_messages_pending` is set from an additional bit status flag I added to the cache coordinator (`cached_in_queue`) that indicates if any worker has a cached tensor pending (see line here: https://github.com/horovod/horovod/pull/902/files#diff-98391bc0c5b3bf87519e745b4ec3b9fdR796). So in the case that no tensors are reduced for a while, this flag is false and the stall message logic isn't triggered (and the `last_perform_operation` time point is updated so that once another cache messaged shows up, the stall warning isn't immediately triggered).

I considered keep track of how long the `cached_messages_pending` flag remains true and if that exceeds a threshold, trigger the stall message. The issue with this though is that this flag does cannot discern stale tensors, so in the case that you have 60 seconds of continuous cycles with tensors to reduce, it would view that as a stall condition. This is why I based the stall timing on the time since the last perform operation.

alsrgv: Do individual workers see that they have pending tensors?  Maybe we could change `cached_messages_pending` to mean that the tensor is pending for 60 seconds?

romerojosh: Hmm these are good points. I can keep track of individual cached tensors when they are replaced to the state queue here: https://github.com/horovod/horovod/pull/902/files#diff-98391bc0c5b3bf87519e745b4ec3b9fdR1345
I think in this way, I should be able to check if a specific tensor has been pending for >60 seconds on any worker and if so, trigger an invalidation of that tensor entry to get the stall message rather than a full cache flush. I'll try that instead.

romerojosh: Okay, just pushed an alternative version of the stall message handling for cached messages. I just keep a map of pending cached tensors on every worker and the initial time they are queued. If a worker sees a tensor listed in that map for greater than 60 seconds, it triggers an invalidation of that cache entry. Entries are removed from the map when they are forwarded to be processed either by RunBypass or the normal path.

horovod/common/operations.cc
```diff
@@ -768,6 +768,23 @@ bool CheckForStalledTensors(HorovodGlobalState& state, MPIContext& ctx) {
   return should_shut_down;
 }
 
+// Check for cached tensors that have been pending for a long time.
+void CheckForStalledCachedTensors(HorovodGlobalState& state,
```
alsrgv: Rename to `InvalidateStalledCachedTensors`?

alsrgv:
LGTM, added one nit comment.

horovod/common/operations.cc
romerojosh: Done!

huoliquankai7:
Hey, @romerojosh @alsrgv  I have a problem now: when tensorflow goes through multiple steps, one of the workers adds a join request (missing state),but  other workers' tensor will hit the cache,(not join request)。After entering the sync you mentioned, all the tensors of other workers' hits are put back into the message_queue, waiting for the next loop, so that the program has been stuck, how to solve this situation?

romerojosh:
@huoloquankai7 You can try disabling the response cache by setting the environment variable `HOROVOD_CACHE_CAPACITY=0`, however, I don't think the issue you described is due to the response cache. It sounds like you are having a single worker submit an operation (join) that is uncommon to all other workers, which I do not think is currently supported.

huoliquankai7:
@romerojosh Yes, this is a single worker submit an operation (join) that is uncommon to all other workers. This design is for the recent horovod data imbalance problem. The current pr only supports pytorch, I am trying to support tensorflow. So, in the process of encountering the cache this situation. In that case, I will temporarily disabling the response cache.

tgaddair:
@huoliquankai7 that sounds like a bug in the join op.  Can you post about the issue in #1058 so they can address it?

huoliquankai7:
@tgaddair Ok, the background to this problem is based on the join operation that I wrote to support tensorflow. The support for pytorch mentioned in pr is no problem. According to my understanding, is pytorch not using cache for the sake of suffix? (maybe my understanding is wrong), so pytorch will not have this problem, and tensorflow appears.

Richie-yan:
Hi,@romerojosh 
According to my understanding, using mpi_allreduce instead of the negotiation of rank0, is this approach a decentralized process?

romerojosh:
@Richie-yan Yes, when all requests in a cycle are in cache, the workers coordinate via a MPI_Allreduce instead of the centralized coordinator path. The coordinator path is still used in cases for uncached requests.

PR #869, opened 2019-02-28
Added horovodrun.

abditag2:
now, we can run horovod scripts using:
```
horovodrun -np 2 -H worker-0:1,worker-1:1 model.py
```

PR #916, opened 2019-03-15
Use os.setsid() instead of os.setpgid()

alsrgv:
We've encountered an issue with launching ssh inside safe_shell_exec.
OpenSSH makes use of tcsetattr() to set terminal properties, which gets
propagated to whole process group.  setsid() provides better isolation
of newly spawned process.

Additional reading: https://en.wikipedia.org/wiki/Process_group

### DIFF
--- a/horovod/spark/util/safe_shell_exec.py
+++ b/horovod/spark/util/safe_shell_exec.py
@@ -78,7 +78,7 @@ def execute(command, env=None, stdout=None, stderr=None):
     if middleman_pid == 0:
         # Close unused file descriptors to enforce PIPE behavior.
         os.close(w)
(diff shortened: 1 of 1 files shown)


alsrgv:
Prior negative behavior:
```bash
python -c "import horovod.spark.util.safe_shell_exec as s; s.execute('ssh <ip which asks password>')"
<get stuck>
```

After the fix:
```
python -c "import horovod.spark.util.safe_shell_exec as s; s.execute('ssh <ip which asks password>')"
Permission denied, please try again.
Permission denied, please try again.
Permission denied (publickey,password).
```

@abditag2, can you include this fix in #869 as well?

PR #920, opened 2019-03-16
Shutdown horovod if some rank stalls for too long

thuningxu:
1. Changed `CheckForStalledTensors()` to return bool to indicate whether some ranks has lags exceeding `HOROVOD_STALL_SHUTDOWN_TIME_SECONDS`.
2. Reformulated the log printout about stalled ranks and tensors
3. In `RunLoopOnce()`, shutdown if `CheckForStalledTensors()` returns true.

Tested offline with a script. Example: https://gist.github.com/thuningxu/6311503ec04b3f38ee70d7a99b9bfa4c


horovod/common/operations.cc
```diff
@@ -108,6 +110,10 @@ std::unique_ptr<OperationManager> op_manager;
 // Stall-check warning time
 #define STALL_WARNING_TIME std::chrono::seconds(60)
 
+// Stall-check shutdown time. If any rank is stalled for longer than this
+// horovod will shutdown
+#define STALL_SHUTDOWN_TIME std::chrono::seconds(180)
```
alsrgv: This value is too small - we should make both warning time & shutdown time configurable through environment variables.  Default behavior for stall check should be to never timeout (for now).  Recoverable stall can happen if one of the ranks is doing full validation.  While not desirable, it should not be a fatal situation.

alsrgv: That way, we should also be able to make an integration test with a small timeout value.

horovod/common/operations.cc
```diff
@@ -1235,7 +1266,9 @@ bool RunLoopOnce(HorovodGlobalState& state, MPIContext& ctx, bool is_coordinator
     if (state.perform_stall_check &&
         std::chrono::steady_clock::now() - state.last_stall_check >
             STALL_WARNING_TIME) {
-      CheckForStalledTensors(state, ctx);
+      // I would prefer to do "should_shut_down |= CheckForStalledTensors(state, ctx);"
+      // but for some reason it leaves the program hanging instead of quitting
```
alsrgv: Can we move this before `response_list.set_shutdown(should_shut_down);` above?  Then the original `should_shut_down |= CheckForStalledTensors(state, ctx);` will work w/o getting stuck.

horovod/common/operations.cc
thuningxu: In new diff this and stall warning time are configurable from environment variables.

horovod/common/operations.cc
thuningxu: done

horovod/common/operations.cc
```diff
@@ -541,53 +540,97 @@ void PerformOperation(TensorTable& tensor_table, Response response) {
...
+      LOG(ERROR) << message.str();
+    } else {
       LOG(WARNING) << message.str();
     }
   }
+
+  return should_shut_down;
+}
+
+void set_bool_from_env(const char* env, bool& val, bool value_if_set) {
```
alsrgv: Nice!

horovod/common/operations.cc
```diff
@@ -541,53 +540,97 @@ void PerformOperation(TensorTable& tensor_table, Response response) {
...
+        message << ", " << *it;
+      }
+
       message << "]";
+    }
+
+    if (should_shut_down) {
+      message << std::endl
+              << "One or more rank (marked by \"!\") is stalled for longer than "
+              << stall_shutdown_time.count() << " seconds. Will shutdown.";
```
alsrgv: Can you add an example of how this message looks like to the PR description?

thuningxu: Would you like example in code? Here is what it looks like:

https://gist.github.com/thuningxu/6311503ec04b3f38ee70d7a99b9bfa4c

horovod/common/operations.cc
```diff
@@ -541,53 +540,97 @@ void PerformOperation(TensorTable& tensor_table, Response response) {
...
+    for (auto& kv: missing_ranks) {
+      message << std::endl << kv.first;
+      if (shutdown_ranks.find(kv.first) != shutdown_ranks.end()) {
+        message << "!";
+      }
+
+      message << ": [";
+      auto it = kv.second.begin();
+      message << *it;
+      while (++it != kv.second.end()) {
```
alsrgv: One of the complaints from the user was that sometimes these messages are too long - we could probably just "..." after the first 5 tensors?

horovod/common/operations.cc
alsrgv: Looks great!

horovod/common/operations.cc
thuningxu: done

test/test_stall.py
```diff
@@ -0,0 +1,31 @@
...
+import os
+import signal
+from common import env
+
+def test():
+  signal.alarm(45)
+  with env(HOROVOD_STALL_CHECK_TIME_SECONDS="2",
+    HOROVOD_STALL_SHUTDOWN_TIME_SECONDS="5"):
+    config = tf.ConfigProto()
+    tf.enable_eager_execution(config=config)
```
alsrgv: Let's add `config.gpu_options.allow_growth = True`, otherwise each process will try to use up all the GPU memory.

alsrgv: Or, since we don't require TensorFlow specific functionality, we could make this a PyTorch test, and model it after `test_torch.py`, which uses `unittest.TestCase`

thuningxu: Changed to using torch.

Issue #893, opened 2019-03-07
Horovod hangs -- include/shm.h:42 NCCL WARN Cuda failure 'invalid argument'


maxhgerlach:
Hi,

I am trying to get Horovod trainings running on a cluster built from Nvidia RTX 2080 ti GPUs and Infiniband interconnects.

Installed software:

- Ubuntu 16.04.5 LTS
- Cuda 10.0.130-1
- NCCL 2.4.2-1+cuda10
- Open MPI 3.1.2
- Mellanox OFED 4.4-2.0.7.0
- Tensorflow 1.13.1, custom built without XLA support
- Horovod 0.16

This is the device topology (PCI express dual root):

```
$ nvidia-smi topo -m
	GPU0	GPU1	GPU2	GPU3	GPU4	GPU5	GPU6	GPU7	mlx4_0	CPU Affinity
GPU0	 X 	PIX	PIX	PIX	SYS	SYS	SYS	SYS	SYS	0-7,16-23
GPU1	PIX	 X 	PIX	PIX	SYS	SYS	SYS	SYS	SYS	0-7,16-23
GPU2	PIX	PIX	 X 	PIX	SYS	SYS	SYS	SYS	SYS	0-7,16-23
GPU3	PIX	PIX	PIX	 X 	SYS	SYS	SYS	SYS	SYS	0-7,16-23
GPU4	SYS	SYS	SYS	SYS	 X 	PIX	PIX	PIX	PIX	8-15,24-31
GPU5	SYS	SYS	SYS	SYS	PIX	 X 	PIX	PIX	PIX	8-15,24-31
GPU6	SYS	SYS	SYS	SYS	PIX	PIX	 X 	PIX	PIX	8-15,24-31
GPU7	SYS	SYS	SYS	SYS	PIX	PIX	PIX	 X 	PIX	8-15,24-31
mlx4_0	SYS	SYS	SYS	SYS	PIX	PIX	PIX	PIX	 X

Legend:

  X    = Self
  SYS  = Connection traversing PCIe as well as the SMP interconnect between NUMA nodes (e.g., QPI/UPI)
  NODE = Connection traversing PCIe as well as the interconnect between PCIe Host Bridges within a NUMA node
  PHB  = Connection traversing PCIe as well as a PCIe Host Bridge (typically the CPU)
  PXB  = Connection traversing multiple PCIe switches (without traversing the PCIe Host Bridge)
  PIX  = Connection traversing a single PCIe switch
[...]

PR #924, opened 2019-03-18
Adopt horovodrun

alsrgv:
Rolled back NCCL due to #893


abditag2:
should we also add something about caching in horovodrun and --disable-cache flag?

alsrgv:
@abditag2, I want to see what kind of issues people experience with it, and put it into troubleshooting at that point.

PR #931, opened 2019-03-19
horovodrun: use os.execve() for mpirun

alsrgv:
safe_shell_exec uses line buffering which does not play very well with
Keras and tqdm -- libraries that make use of progress bars.


alsrgv:
Will merge to unblock the release, minor change.

PR #932, opened 2019-03-19
add nccl comm destroy

Sergei-Lebedev:
Horovod doesn't free NCCL communicator

### DIFF
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -848,7 +848,7 @@ void BackgroundThreadLoop(HorovodGlobalState& state, MPIContext& ctx) {
   // Signal that shutdown has been requested.
   state.shut_down = true;
 
-#if HAVE_NCCL && HOROVOD_GPU_ALLREDUCE == 'N'
+#if HAVE_NCCL
   nccl_context.ShutDown();
 #endif
   // TODO: init.cu:645 WARN Cuda failure 'driver shutting down'
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -848,6 +848,9 @@ void BackgroundThreadLoop(HorovodGlobalState& state, MPIContext& ctx) {
   // Signal that shutdown has been requested.
   state.shut_down = true;
 
+#if HAVE_NCCL && HOROVOD_GPU_ALLREDUCE == 'N'
+  nccl_context.ShutDown();
+#endif
   // TODO: init.cu:645 WARN Cuda failure 'driver shutting down'
   //#if HAVE_NCCL
   //  for (auto it = horovod_global.streams.begin();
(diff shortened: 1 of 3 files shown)


horovod/common/operations.cc
```diff
@@ -848,6 +848,9 @@ void BackgroundThreadLoop(HorovodGlobalState& state, MPIContext& ctx) {
   // Signal that shutdown has been requested.
   state.shut_down = true;
 
+#if HAVE_NCCL && HOROVOD_GPU_ALLREDUCE == 'N'
```
tgaddair: Since we create the `NCCLContext` regardless of `HOROVOD_GPU_ALLREDUCE` value, I think this should just be:

```
if HAVE_NCCL`
```

horovod/common/ops/nccl_operations.cc
```diff
@@ -43,6 +43,12 @@ void NCCLContext::ErrorCheck(std::string op_name, ncclResult_t nccl_result) {
   }
 }
 
+void NCCLContext::ShutDown(){
+  for(auto it = nccl_comms.begin(); it != nccl_comms.end(); ++it){
+    ncclCommDestroy(it->second);
+  }
```
tgaddair: Should we also clear the map?  `NCCLAllreduce::InitNCCLComm()` will only initialize a communicator if it is not in the map.

Sergei-Lebedev:
Thanks for your comments, I've updated the PR. Previous NCCL communicator free code raises cuda driver error and I can't reproduce it with the patch. Could you please check that's not the case anymore?

tgaddair:
Hey @Sergei-Lebedev, do you have an example script that raises the CUDA driver error (pre-patch), or did this happen for every Horovod script you ran?

Sergei-Lebedev:
I mean that this part failed in previous versions of Horovod  for any script that used nccl  
https://github.com/horovod/horovod/blob/4f7319e265591c3231ee87ae9dfd91e9c5f9b53e/horovod/common/operations.cc#L894-L904

horovod/common/ops/nccl_operations.cc
```diff
@@ -43,6 +43,13 @@ void NCCLContext::ErrorCheck(std::string op_name, ncclResult_t nccl_result) {
   }
 }
 
+void NCCLContext::ShutDown(){
```
alsrgv: @Sergei-Lebedev, can you add `cudaStreamSynchronize()` before shutting down NCCL communicators?

The reason for this is that there could be outstanding work enqueued to NCCL, and I'm not sure if it's safe to destroy NCCL communicators while there's enqueued work.

After you make this change, can you check if `init.cu:645 WARN Cuda failure 'driver shutting down'` appears?

Sergei-Lebedev: @alsrgv, I think there is no need to explicitly wait for operations to be completed as long as NCCL do it internally. According to nccl documentation ncclCommDestroy waits for any uncompleted operations before destroying communicator. 
https://docs.nvidia.com/deeplearning/sdk/nccl-developer-guide/docs/api/comms.html#ncclcommdestroy

alsrgv: Good find!  In this case, can you remove the comment in the operations.cc?

PR #943, opened 2019-03-21
MXNet: add DistributedTrainer for Gluon

yuxihu:
DistributedTrainer is a subclass of Gluon Trainer. We override the `_allreduce_grads` function in Trainer to use Horovod `allreduce_` instead of kvstore pull/push methods. This makes distributed training using MXNet Gluon API with PS and Horovod have the same internal workflow.

Thanks @ptrendx and @romerojosh for proposing this [idea](https://github.com/horovod/horovod/pull/915#issuecomment-474022074).

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ See full training [MNIST](examples/mxnet_mnist.py) and [ImageNet](examples/mxnet
 ```python
 import mxnet as mx
 import horovod.mxnet as hvd
-from mxnet import autograd, gluon
+from mxnet import autograd
 
 # Initialize Horovod
 hvd.init()
@@ -220,12 +220,9 @@ num_workers = hvd.size()
 model = ...
 model.hybridize()
 
-# Define hyper parameters
+# Create optimizer
 optimizer_params = ...
-
-# Add Horovod Distributed Optimizer
 opt = mx.optimizer.create('sgd', **optimizer_params)
-opt = hvd.DistributedOptimizer(opt)
 
 # Initialize parameters
 model.initialize(initializer, ctx=context)
@@ -235,8 +232,10 @@ params = model.collect_params()
 if params is not None:
     hvd.broadcast_parameters(params, root_rank=0)
 
-# Create trainer and loss function
-trainer = gluon.Trainer(params, opt, kvstore=None)
+# Create DistributedTrainer, a subclass of gluon.Trainer
+trainer = hvd.DistributedTrainer(params, opt)
+
+# Create loss function
 loss_fn = ...
 
 # Train model
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ See a full training [example](examples/tensorflow_mnist_estimator.py).
 
 Horovod supports MXNet and regular TensorFlow in similar ways.
 
-See full training [MNIST](examples/mxnet_mnist.py) and [ImageNet](examples/mxnet_imagenet_resnet50.py) examples.
+See full training [MNIST](examples/mxnet_mnist.py) and [ImageNet](examples/mxnet_imagenet_resnet50.py) examples. The script below provides a simple skeleton of code block based on MXNet Gluon API.
 
 ```python
 import mxnet as mx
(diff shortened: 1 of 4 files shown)


yuxihu:
@alsrgv @apeforest @eric-haibin-lin @ptrendx @romerojosh Please help review.

horovod/mxnet/__init__.py
```diff
@@ -68,6 +68,17 @@ def set_wd_mult(self, args_wd_mult):
         self._optimizer.set_wd_mult(args_wd_mult)
 
 
+class DistributedTrainer(mx.gluon.Trainer):
+    def __init__(self, params, optimizer, optimizer_params=None):
+        super(DistributedTrainer, self).__init__(
```
romerojosh: Do you think it would be a good idea to add something here to deal with the case where a user erroneously passes a `DistributedOptimizer` into the `DistributedTrainer`? You could just automatically unwrap the `DistributedOptimizer` for them (and maybe print a warning message).

yuxihu: Make sense. Will add the check.

yuxihu:
@alsrgv Looks like [Keras test](https://travis-ci.org/horovod/horovod/jobs/509631730#L9825) is failing.

README.md
```diff
@@ -235,8 +232,10 @@ params = model.collect_params()
 if params is not None:
     hvd.broadcast_parameters(params, root_rank=0)
 
-# Create trainer and loss function
-trainer = gluon.Trainer(params, opt, kvstore=None)
+# Create DistributedTrainer, a subclass of gluon.Trainer
```
apeforest: Gluon was not mentioned in any context.

alsrgv:
@yuxihu, the tests are broken due to https://github.com/tensorflow/tensorflow/issues/27067.  BTW, is it possible to switch all these https://github.com/apache/incubator-mxnet/blob/791c5d5e8091c8c16d1a30df1b0d53e4af60d245/include/mxnet/engine.h#L122 to function pointers?  Then we could avoid gcc4/gcc5 incompatibility, too.

horovod/mxnet/__init__.py
```diff
@@ -68,6 +69,22 @@ def set_wd_mult(self, args_wd_mult):
         self._optimizer.set_wd_mult(args_wd_mult)
 
 
+class DistributedTrainer(mx.gluon.Trainer):
```
eric-haibin-lin: Since this implements the Trainer API, are all APIs like save_state/load_state validated and tested?

yuxihu:
@alsrgv I see. It looks like the change which broke the CI will not be reverted. What's the plan for Horovod CI?

It won't work by simply changing `std:function` to function poiner in the MXNet Engine. Lambda function with capture cannot be passed to a function pointer interface. There are more work to be done besides changing the signature which I am working with priority.

README.md
yuxihu: Done

horovod/mxnet/__init__.py
yuxihu: I manually tested save_states and load_states and it worked. The DistributedTrainer is a subclass of Trainer which we only overrided one function `_allreduce_grads`.  This function has been covered by the MNIST example which we run in Horovod CI. For other APIs, the Python inheritance rules should govern and are covered by our MXNet unit tests.

alsrgv:
@yuxihu, could you rebase this PR on latest master as well?

alsrgv:
@apeforest @eric-haibin-lin @romerojosh, good to merge?

romerojosh:
LGTM!

horovod/mxnet/__init__.py
```diff
@@ -68,6 +69,22 @@ def set_wd_mult(self, args_wd_mult):
...
+    def __init__(self, params, optimizer, optimizer_params=None):
```
eric-haibin-lin: It is worthwhile to document the difference between `DistributedTrainer` and `Trainer`, that `DistributedTrainer` performs allreduce(summation) and average, while `Trainer` only performs allreduce(summation)? That will help existing Trainer + kvstore users understand how gradient is calculated

yuxihu: Added

yuxihu:
@apeforest @eric-haibin-lin Ready for final review. Please take a look.

PR #946, opened 2019-03-24
Fix typo in DDL

alsrgv:
Signed-off-by: Alex Sergeev <alsrgv@users.noreply.github.com>

cc @denfromufa


denfromufa:
thank you, confirmed the fix!

PR #954, opened 2019-03-25
Updated README link for Kubeflow MPI Job

tlkh:
In `README.md`, the "Kubeflow" link is broken (goes to a non-existing `kubeflow/openmpi` folder). I've updated the link to point to `kubeflow/mpi-job` instead, which I believe is the updated name of the old component.


alsrgv:
LGTM, thanks!

PR #955, opened 2019-03-25
Pin GCC to 4.8

alsrgv:
See https://github.com/tensorflow/tensorflow/issues/27067

Updates symlinks to gcc-4.8/g++-4.8 before Horovod installation, and restores them afterward.

NOTE: if Python was built with a more recent version of GCC, it may have incompatible flags in `/usr/lib/python${version}/.../_sysconfigdata*.py`.  We patch the Python installation to remove those flags.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ before_install:
   - export CONTAINER=$(docker ps -q | head -n 1)
   - docker exec ${CONTAINER} /bin/sh -c "apt-get update -qq"
   # install necessary network tools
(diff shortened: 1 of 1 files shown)


yuxihu:
Any specific reason that we use Ubuntu in the CI? CentOS 7 comes with gcc 4, which might be an option.

alsrgv:
@yuxihu, I guess just historical reasons - internally we use Debian, and we used to use it until it became too annoying. At that point, we switched to Ubuntu :-)

alsrgv:
GCC 4.8 does not work with pybind11.  This issue seems related: https://github.com/pybind/pybind11/issues/1032

alsrgv:
Superseded by #958

PR #958, opened 2019-03-27
Pin GCC 4.9 in CI

alsrgv:
See tensorflow/tensorflow#27067

Updates symlinks to gcc-4.9/g++-4.9 before Horovod installation, and restores them afterward.

gcc-4.8 would work for TensorFlow as well, but does not interoperate well with pybind11 used by PyTorch (https://github.com/pybind/pybind11/issues/1032)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,7 @@ services:
   - docker
 
 before_install:
-  # force latest Ubuntu for Python 3.6 and nightly TensorFlow which requires new glibc
-  - |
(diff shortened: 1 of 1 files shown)


yuxihu:
We may not need to, but do you think it is worth doing the same in [Dockerfile](https://github.com/horovod/horovod/blob/master/Dockerfile)? In this way, we can use MXNet pip packages.

alsrgv:
@yuxihu, let's run it for a while, and if nothing breaks, we can switch Dockerfile as well.

PR #965, opened 2019-03-28
MXNet: check CUDA runtime availability when building Horovod

yuxihu:
With this PR, we perform the following two checks:

1.  If HOROVOD_GPU_ALLREDUCE/BROADCAST/ALLGATHER is set but MXNet is not built with CUDA runtime, Horovod installation will error out
2. If HOROVOD_GPU_ALLREDUCE/BROADCAST/ALLGATHER is not set but MXNet is built with CUDA runtime, Horovod will be built with CUDA runtime.

Along with enabling new build options, we fix the issues for performing allreduce and allgather on CPU when HAVE_CUDA is true and HOROVOD_GPU_ALLREDUCE/ALLGATHER is not set. We tested the following three installation methods where all unit tests and MNIST example passed.

1. `pip install --no-cache-dir ~/horovod` with MXNet CPU package
2. `pip install --no-cache-dir ~/horovod` with MXNet GPU package
3. `HOROVOD_GPU_ALLREDUCE=NCCL pip install --no-cache-dir ~/horovod` with MXNet GPU package


yuxihu:
@apeforest @ctcyang @alsrgv Please help review.

yuxihu:
PyTorch test is [failing](https://travis-ci.org/horovod/horovod/jobs/512749699#L8945). All MXNet tests passed.

setup.py
```diff
@@ -668,11 +668,50 @@ def is_mx_mkldnn():
...
+            'Horovod build with GPU support was requested, but this MXNet '
+            'installation does not support CUDA.')
+
+    # Update HAVE_CUDA to mean that MXNet supports CUDA. Internally, we will be checking
+    # HOROVOD_GPU_(ALLREDUCE|ALLGATHER|BROADCAST) to decide whether we should use GPU
+    # version or transfer tensors to CPU memory for those operations.
+    if mx_have_cuda and not macro_have_cuda:
+        cuda_include_dirs, cuda_lib_dirs = get_cuda_dirs(build_ext, options['COMPILE_FLAGS'])
+        options['MACROS'] += [('HAVE_CUDA', '1')]
+        options['INCLUDES'] += cuda_include_dirs
```
alsrgv: Do we need anything other than MACRO update here?  AFAIK other frameworks don't have this logic.  You can also use `updated_macros = set_macro(options['MACROS'], 'HAVE_CUDA', str(int(mx_have_cuda)))` for safety.

test/test_mxnet.py
```diff
@@ -221,32 +221,6 @@ def test_horovod_allreduce_type_error(self):
         except (MXNetError, RuntimeError):
             pass
 
-    @unittest.skipUnless(has_gpu, "no gpu detected")
-    def test_horovod_allreduce_cpu_gpu_error(self):
```
alsrgv: Can we keep this test, but enable it only if Horovod was built with HORVOD_GPU_ALLREDUCE?  Maybe it's a good time to add functionality to `horovod.common` to expose flags it was built with.

setup.py
yuxihu: I did what you mentioned in my very first [commit](input) in this PR. However, it cannot compile if I do not set HOROVOD_GPU_ALLREDUCE/BROADCAST/ALLGATHER.

I do not know how other frameworks worked. But for MXNet, I think we need to call `get_cuda_dirs` here when HOROVOD_GPU_ALLREDUCE/BROADCAST/ALLGATHER is not set. Right now, it is [called](https://github.com/horovod/horovod/blob/master/setup.py#L504) in `get_comm_options`. If those options are not set, it will be skipped here. Therefore, if HAVE_CUDA is not set in MACROS and we detect CUDA is supported in the MXNet build, we need to set the include/lib/macro etc. for CUDA.

test/test_mxnet.py
yuxihu: I thought about it but there seems no way for me to decide if it needs to run. I can keep it but skip it for both CPU/GPU tests. Later on we can enable it when we have a way to know the flags.

setup.py
alsrgv: What's the error message if you just set the macro?

test/test_mxnet.py
alsrgv: Sounds good, let's just disable them, and add a TODO comment to fix it later.

setup.py
yuxihu: This was the error:
```
    In file included from horovod/common/operations.cc:45:0:
    horovod/common/ops/cuda_operations.h:23:26: fatal error: cuda_runtime.h: No such file or directory
    compilation terminated.
```

test/test_mxnet.py
yuxihu: Done.

setup.py
alsrgv: Oh, I see. Fair enough :-)

Issue #936, opened 2019-03-20
broadcast optimizer state dict or not in PyTorch?


hubutui:
Hey, I notice there is a minor difference in https://github.com/horovod/horovod/blob/master/examples/pytorch_synthetic_benchmark.py#L64, https://github.com/horovod/horovod/blob/master/examples/pytorch_imagenet_resnet50.py#L155, and https://github.com/horovod/horovod/blob/master/examples/pytorch_mnist.py#L94. Do I need to broadcast optimizer state dict or not?
Also, there's no API reference document?

alsrgv:
@hubutui, good catch!  `hvd.broadcast_optimizer_state()` is useful when restoring from checkpoint, as it can broadcast optimizer state, such as momentums, from first rank to everyone else.  While it's not strictly necessary in either synthetic benchmark or mnist example, it's good to add.

alsrgv:
We have a plan to build a doc website, but didn't get to it yet.  It will have an API reference.

PR #967, opened 2019-03-29
Add broadcast_optimizer_state to pytorch_mnist.py

alsrgv:
See #936

### DIFF
--- a/examples/pytorch_mnist.py
+++ b/examples/pytorch_mnist.py
@@ -90,8 +90,9 @@ if args.cuda:
     # Move model to GPU.
     model.cuda()
 
-# Horovod: broadcast parameters.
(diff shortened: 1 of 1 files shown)


examples/pytorch_mnist.py
```diff
@@ -90,8 +90,9 @@ def forward(self, x):
     # Move model to GPU.
     model.cuda()
 
-# Horovod: broadcast parameters.
+# Horovod: broadcast parameters & optimizer state.
 hvd.broadcast_parameters(model.state_dict(), root_rank=0)
+hvd.broadcast_optimizer_state(optimizer, root_rank=0)
```
tgaddair: Optimizer is defined below, which is causing unit tests to fail.

PR #976, opened 2019-04-02
MXNet: Normalize rescale_grad in optimizer by Horovod size

yuxihu:
We currently perform average on gradients using allreduce methods in DistributedOptimizer and DistributedTrainer. This is equivalent to normalizing optimizer.rescale_grad by Horovod size(), which has better performance. 

This PR replaces allreduce average with normalized rescale_grad. Our tests show 2~3% performance boost when training ResNet50 with 8 GPUs.

### DIFF
--- a/examples/mxnet_imagenet_resnet50.py
+++ b/examples/mxnet_imagenet_resnet50.py
@@ -279,15 +279,6 @@ net.cast(args.dtype)
 initializer = mx.init.Xavier(rnd_type='gaussian', factor_type="in",
                              magnitude=2)
 
-# Create optimizer
-optimizer_params = {'wd': args.wd,
-                    'momentum': args.momentum,
-                    'rescale_grad': 1.0 / batch_size,
-                    'lr_scheduler': lr_sched}
-if args.dtype == 'float16':
-    optimizer_params['multi_precision'] = True
-opt = mx.optimizer.create('sgd', **optimizer_params)
-
 
 def train_gluon():
     def evaluate(epoch):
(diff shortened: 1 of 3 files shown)


yuxihu:
@romerojosh @eric-haibin-lin @apeforest Please help review.

yuxihu:
The test [failures](https://travis-ci.org/horovod/horovod/jobs/514525027) are not related to MXNet.

examples/mxnet_imagenet_resnet50.py
```diff
@@ -317,6 +308,14 @@ def evaluate(epoch):
     if params is not None:
         hvd.broadcast_parameters(params, root_rank=0)
 
+    # Create optimizer
+    optimizer_params = {'wd': args.wd,
+                        'momentum': args.momentum,
+                        'lr_scheduler': lr_sched}
```
romerojosh: Does `rescale_grad` default to `1.0 / batch_size` for gluon but not module? Maybe it would be good to set it here anyway for consistency, or document the difference somehow.

yuxihu: This is a good question. Short answer is yes. For Gluon, we take care of it in [Trainer.step()](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/gluon/trainer.py#L309) function so there is no need to set it in optimizer_params. But for Module, because we create optimizer first and wrap it with DistributedOptimizer, we need to specify [rescale_grad](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/optimizer/optimizer.py#L51) in optimizer_params when creating optimizer.

I can add a comment here to document the difference.

apeforest:
Can you also post performance numbers in this PR for future reference?

alsrgv:
Love the removal of `rescale_grad` from the training script.  Restarted failed integration tests.

yuxihu:
@alsrgv Looks like the tensorflow nightly [tests](https://travis-ci.org/horovod/horovod/jobs/514792270#L9894) failed.

alsrgv:
We're fixing that separately, looks like TensorFlow changes the library name but did not update `tf.sysconfig.get_link_flags()`. 

@romerojosh @eric-haibin-lin @apeforest, can you stamp, so we can merge?

PR #985, opened 2019-04-04
MXNet: use MXEnginePushAsync C API to push horovod operations

### LINKED ISSUE #884: Segmentation fault: running MXNet with Horovod on Ubuntu Linux 16.04
**Environment:**
1. Framework: MXNet
2. Framework version: 1.4.0
3. Horovod version: 0.16.0
4. MPI version: 3.1.1
5. CUDA version: 9.2
6. NCCL version: 2.2.13
7. Python version: 3.5
8. OS and version: Linux Ubuntu 16.04

**Checklist:**
Yes
2. If your question is about hang, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/running.md)?
3. If your question is about docker, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/docker.md)?

**Bug report:**
When running Horovod with MXNet on Linux Ubuntu 16.04, there will be a segmentation fault.

yuxihu:
Fixes #884 

This PR changes PushAsync to MXEnginePushAsync C API to avoid std::function incompatibility between GCC [4.x](https://github.com/gcc-mirror/gcc/blob/gcc-4_8_4-release/libstdc++-v3/include/std/functional#L2069) and GCC [5.x](https://github.com/gcc-mirror/gcc/blob/gcc-5_4_0-release/libstdc++-v3/include/std/functional#L1854).

The dependent [PR](https://github.com/apache/incubator-mxnet/pull/14615) in MXNet has been merged.


yuxihu:
@eric-haibin-lin @ctcyang @apeforest Please help review

horovod/mxnet/mpi_ops.h
```diff
@@ -32,6 +32,46 @@ typedef ::mxnet::NDArray NDArray;
 typedef ::mxnet::engine::CallbackOnComplete CallbackOnComplete;
 typedef std::shared_ptr<MXTemporaryBuffer<NDArray>> MXTempBufferShared;
 
+struct MpiOpsParam {
+  std::shared_ptr<NDArray*> input;
```
alsrgv: Since you're passing around a pointer to this structure, do fields of the structure need to be pointers, or could they be underlying objects?

yuxihu: These fields also need to be shared_ptr to prevent them from being destroyed prematurely. I tried not to wrap them with shared_ptr but hit segfault.

alsrgv: That's odd.  What kind of segfault were you getting?  The only thing that can be destroyed is `std::string`, but I imagine it gets copied around as needed.

yuxihu: Yes, you are right. I confused myself with another error. Will change it.

alsrgv:
@yuxihu, please note that there is another tf-nightly build break that I just pushed workaround for in the master, so please rebase.

yuxihu:
@alsrgv Thanks for letting me know. I will do it later once the dependent PR is merged into MXNet. I will need to update the MXNet package in order to use the new API so it will fail anyway right now.

yuxihu:
@apeforest @alsrgv The PR is ready for review.

horovod/mxnet/mpi_ops.cc
```diff
@@ -49,206 +51,154 @@ inline void InvokeCompleteCallback(CallbackOnComplete on_complete, const Status&
...
   // Not in-place
-  if (input->var() != output->var()) {
-    Engine::Get()->PushAsync(exec_fn, MX_EXEC_CTX,
-                             {input->var()}, {output->var()},
-                             FnProperty::kCPUPrioritized, priority,
-                             op_type_name.c_str());
+  auto input_var = input->var();
+  auto output_var = output->var();
+  if (input_var != output_var) {
+    MXEnginePushAsync(DoHorovodOperation, ops_param, DeleteMpiOpsParam,
```
eric-haibin-lin: What happens if an user installed mxnet-1.4 instead of mxnet nightly? Does hovorod plan to drop mxnet 1.4 support? 
(I'm fine with mxnet nightly for my project, but I cannot speak for others)

yuxihu: Users can still use Horovod 0.16.0 and 0.16.1 releases with MXNet 1.4.0 release. Given that GCC incompatibility issue we had in earlier releases, I would encourage users to upgrade to latest Horovod and MXNet releases. Maintaining two internal implementations here is cumbersome.

docker-compose.test.yml
```diff
@@ -105,7 +105,8 @@ services:
         KERAS_PACKAGE: git+https://github.com/keras-team/keras.git
         PYTORCH_PACKAGE: torch-nightly
         PYSPARK_PACKAGE: pyspark==2.4.0
-  test-cpu-openmpi-py3_6-tfhead-kerashead-torchhead-mxnet1_4_0-pyspark2_4_0:
+        MXNET_PACKAGE: mxnet --pre
```
apeforest: why not use the same 1.5.0b20190412 as above here?

docker-compose.test.yml
```diff
@@ -115,7 +116,8 @@ services:
         KERAS_PACKAGE: git+https://github.com/keras-team/keras.git
         PYTORCH_PACKAGE: torch-nightly
         PYSPARK_PACKAGE: pyspark==2.4.0
-  test-cpu-mpich-py2_7-tf1_12_0-keras2_2_2-torch1_0_0-mxnet1_4_0-pyspark2_4_0:
+        MXNET_PACKAGE: mxnet --pre
```
apeforest: why not use the same 1.5.0b20190412 as above here?

docker-compose.test.yml
```diff
@@ -190,9 +192,9 @@ services:
         TENSORFLOW_PACKAGE: tf-nightly-gpu==1.14.1-dev20190408
         KERAS_PACKAGE: git+https://github.com/keras-team/keras.git
         PYTORCH_PACKAGE: torch-nightly
-        MXNET_PACKAGE: mxnet-cu100==1.4.0.post0
+        MXNET_PACKAGE: mxnet-cu100 --pre
```
apeforest: why not use the same 1.5.0b20190412 as above here?

docker-compose.test.yml
```diff
@@ -204,9 +206,9 @@ services:
...
```
apeforest: why not use the same 1.5.0b20190412 as above here?

horovod/mxnet/mpi_ops.cc
```diff
@@ -49,206 +51,154 @@ inline void InvokeCompleteCallback(CallbackOnComplete on_complete, const Status&
...
+    case OperationType::BROADCAST:
+      return BROADCAST_OP_TYPE_NAME;
+    default:
+      throw std::logic_error("Unsupported Horovod operation type.");
+  }
 }
 
-void DoAllgather(NDArray* tensor, NDArray* output, std::string& name,
-                 CallbackOnComplete on_complete) {
+void DoHorovodOperation(void*, void* on_complete_ptr, void* param) {
```
apeforest: Do we still need the first argument in the function?

horovod/mxnet/mpi_ops.cc
```diff
@@ -49,206 +51,154 @@ inline void InvokeCompleteCallback(CallbackOnComplete on_complete, const Status&
...
+    MXEnginePushAsync(DoHorovodOperation, ops_param, DeleteMpiOpsParam,
+                      &MX_EXEC_CTX, nullptr, 0, &output_var, 1,
+                      &MX_FUNC_PROP, priority, op_type_name);
   }
 }
 
 #if HAVE_CUDA
-void DoAllreduceCudaOnCPU(MXTempBufferShared& hvd_cpu_buffer, std::string& name,
-                          CallbackOnComplete on_complete) {
+void DoHorovodOperationCudaOnCPU(void*, void* on_complete_ptr, void* param) {
```
apeforest: Do we still need the first argument in the function?

horovod/mxnet/mpi_ops.h
```diff
@@ -22,17 +22,53 @@
...
+
+struct MpiOpsParam {
+  NDArray* input;
+  NDArray* output;
+  NDArray* cpu_tensor;
+  OperationType op_type;
+  std::string op_name;
+  int root_rank;
+
+  MpiOpsParam(NDArray* input, NDArray* output, NDArray* cpu_tensor,
```
apeforest: Add `const` to immutable arguments

docker-compose.test.yml
yuxihu: This is the nightly test. We want to test our nightly package instead of a specific release.

docker-compose.test.yml
yuxihu: See above answer.

docker-compose.test.yml
yuxihu: See above answer.

horovod/mxnet/mpi_ops.cc
yuxihu: We need it to comply with the [function signature](https://github.com/apache/incubator-mxnet/blob/master/include/mxnet/c_api.h#L106). The first argument is RunContext, which we do not use here but might be used in other cases. That's why the function signature has it to make it more generic.

docker-compose.test.yml
yuxihu: See above answer.

horovod/mxnet/mpi_ops.cc
yuxihu: Answered below.

horovod/mxnet/mpi_ops.h
yuxihu: Actually adding `const` to these pointers requires cascading changes downstream where `const` does not apply (e.g. TensorUtil::Copy()). We have to keep it like this.

horovod/mxnet/mpi_ops.cc
eric-haibin-lin: ok

yuxihu:
@alsrgv Could you please take a look?

A side QQ: when will be the next Horovod release?

PR #986, opened 2019-04-04
MXNet: push Horovod operations to engine using CPU context

yuxihu:
@romerojosh noticed that small but noticeable time is spent pushing the Horovod ops to the engine, which has the effect of holding back GPU kernel launches if these ops are scheduled on the GPU context.

These small stalls are removed if these operations are put into the CPU context explicitly.
Placing these operations into the CPU context also makes some sense, since the underlying Horovod operation that is pushed to the MXNet engine itself doesn't mutate the GPU data, but just enqueues work to Horovod.

Since now all operations are pushed to CPU context, we also use kCPUPrioritized when pushing such that we can leverage the default multiple [MXNET_CPU_PRIORITY_NTHREADS](http://mxnet.incubator.apache.org/versions/master/faq/env_var.html) to schedule CPU jobs. This also aligns with the way of [pushing operations](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/gluon/trainer.py#L354) into kvstore.

### DIFF
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -51,7 +51,8 @@ class DistributedOptimizer(mx.optimizer.Optimizer):
     def _do_allreduce(self, index, grad):
         if isinstance(index, (tuple, list)):
             for i in range(len(index)):
-                allreduce_(grad[i], average=False, name=str(index[i]))
+                allreduce_(grad[i], average=False,
+                           name=str(index[i]), priority=-i)
         else:
             allreduce_(grad, average=False, name=str(index))
 
@@ -97,7 +98,8 @@ class DistributedTrainer(mx.gluon.Trainer):
     def _allreduce_grads(self):
         for i, param in enumerate(self._params):
             if param.grad_req != 'null':
-                allreduce_(param.list_grad()[0], average=False, name=str(i))
+                allreduce_(param.list_grad()[0], average=False,
+                           name=str(i), priority=-i)
 
 
 # Wrapper to inject Horovod broadcast after parameter initialization
--- a/horovod/mxnet/mpi_ops.cc
+++ b/horovod/mxnet/mpi_ops.cc
@@ -28,7 +28,7 @@ namespace {
 
 std::atomic_int op_count;
 
-std::string GetOpName(std::string prefix, char* name) {
+std::string GetOpName(const std::string& prefix, const char* name) {
   if (name != nullptr) {
     return prefix + "." + std::string(name);
   }
@@ -37,6 +37,8 @@ std::string GetOpName(std::string prefix, char* name) {
   return prefix + ".noname." + std::to_string(op_count);
 }
 } // namespace
+
+const auto EXEC_CTX = Context::CPU();
 
 inline void InvokeCompleteCallback(CallbackOnComplete on_complete, const Status& status) {
   if (status.ok()) {
@@ -162,7 +164,8 @@ void DoBroadcastCudaOnCPU(MXTempBufferShared& hvd_cpu_buffer, int root_rank,
 #endif
 
 extern "C" int horovod_mxnet_allreduce_async(NDArray* input, NDArray* output,
-                                             char* name, bool average) {
+                                             const char* name, bool average,
+                                             int priority) {
   MX_API_BEGIN();
 
   std::string op_name = GetOpName("allreduce", name);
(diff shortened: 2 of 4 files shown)


yuxihu:
@romerojosh @eric-haibin-lin @apeforest Please help review.

alsrgv:
@yuxihu, could you rebase?  Let's try out our new amazing CI :-)

yuxihu:
@alsrgv Just rebased. How can I track the test status?

alsrgv:
@yuxihu, sorry - didn't have checkbox to allow forked repos selected, could you try again?

alsrgv:
@yuxihu, a question - does the fact that we use CPU for enqueueing operations have any implications on whether we need to wait for previous kernels to finish before Horovod operations can start?

yuxihu:
> @yuxihu, a question - does the fact that we use CPU for enqueueing operations have any implications on whether we need to wait for previous kernels to finish before Horovod operations can start?

We are not changing anything (except that we are using kCPUPrioritized now, which should be more optimal) if the tensor context is already CPU. When the context is GPU, there may be contention for GPU kernels between pushing Horovod operation and performing the actual computation. Changing pushing to CPU context should alleviate the contention.

alsrgv:
Got it. SGTM, @romerojosh @eric-haibin-lin @apeforest could you stamp?

romerojosh:
LGTM. :+1:

horovod/mxnet/mpi_ops.py
```diff
@@ -61,6 +61,8 @@ def allreduce(tensor, average=True, name=None):
         average: A flag indicating whether to compute average or summation,
                  defaults to average.
         name: A name of the reduction operation.
+        priority: The priority of this operation. Higher priority operations are
+                  likely to be executed before other actions.
```
apeforest: nit: change the word "actions" to "operations" to be consistent

horovod/mxnet/mpi_ops.py
```diff
@@ -95,6 +99,8 @@ def allreduce_(tensor, average=True, name=None):
...
```
apeforest: nit: change the word "actions" to "operations" to be consistent

horovod/mxnet/mpi_ops.py
```diff
@@ -127,6 +135,8 @@ def allgather(tensor, name=None):
     Arguments:
         tensor: A tensor to allgather.
         name: A name of the allgather operation.
+        priority: The priority of this operation. Higher priority operations are
+                  likely to be executed before other actions.
```
apeforest: nit: change the word "actions" to "operations" to be consistent

horovod/mxnet/mpi_ops.py
```diff
@@ -166,6 +176,8 @@ def broadcast(tensor, root_rank, name=None):
         tensor: A tensor to broadcast.
         root_rank: The rank to broadcast the value from.
         name: A name of the broadcast operation.
+        priority: The priority of this operation. Higher priority operations are
+                  likely to be executed before other actions.
```
apeforest: nit: change the word "actions" to "operations" to be consistent

horovod/mxnet/mpi_ops.py
```diff
@@ -198,6 +212,8 @@ def broadcast_(tensor, root_rank, name=None):
...
```
apeforest: nit: change the word "actions" to "operations" to be consistent

horovod/mxnet/__init__.py
```diff
@@ -51,7 +51,8 @@ def create_state_multi_precision(self, index, weight):
     def _do_allreduce(self, index, grad):
         if isinstance(index, (tuple, list)):
             for i in range(len(index)):
-                allreduce_(grad[i], average=False, name=str(index[i]))
+                allreduce_(grad[i], average=False,
+                           name=str(index[i]), priority=-i)
```
apeforest: why the priority is decreasing with index? can't we just all use 0?

yuxihu: If all is 0, there is no priority anymore. Followed the way of [pushing](https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/gluon/trainer.py#L354) to kvstore.

horovod/mxnet/mpi_ops.py
yuxihu: Done

horovod/mxnet/mpi_ops.py
yuxihu: Done

horovod/mxnet/mpi_ops.py
yuxihu: Done

horovod/mxnet/mpi_ops.py
yuxihu: Done

horovod/mxnet/mpi_ops.py
yuxihu: Done

horovod/mxnet/mpi_ops.cc
```diff
@@ -38,6 +38,8 @@ std::string GetOpName(std::string prefix, char* name) {
 }
 } // namespace
 
+const auto EXEC_CTX = Context::CPU();
```
apeforest: nit: use static const since this variable is only for MXNet; maybe also name it MX_EXEC_CTX?

yuxihu: MX_EXEC_CTX sounds good. Let's do it in my next PR if this is the only concern.

apeforest: SGTM.

PR #993, opened 2019-04-05
MXNet: perform copy to CPU when needed

yuxihu:
Currently, if HAVE_CUDA is set and HOROVOD_GPU_ALLREDUCE/ALLGATHER/BROADCAST is not set, we copy the tensor to CPU, perform Horovod operations and copy it back from CPU regardless of the tensor context. However, copy to CPU is needed only when the tensor is not in CPU context.

This PR checks the context of the tensors to avoid unnecessary copy to CPU. We also refactor the code to make this change cleaner.

### DIFF
--- a/horovod/mxnet/mpi_ops.cc
+++ b/horovod/mxnet/mpi_ops.cc
@@ -38,7 +38,7 @@ std::string GetOpName(const std::string& prefix, const char* name) {
 }
 } // namespace
 
-const auto EXEC_CTX = Context::CPU();
+static const auto MX_EXEC_CTX = Context::CPU();
 
 inline void InvokeCompleteCallback(CallbackOnComplete on_complete, const Status& status) {
   if (status.ok()) {
@@ -143,13 +143,13 @@ inline void PushHorovodOperation(OperationType op_type, NDArray* input,
 
   // Not in-place
   if (input->var() != output->var()) {
-    Engine::Get()->PushAsync(exec_fn, EXEC_CTX,
+    Engine::Get()->PushAsync(exec_fn, MX_EXEC_CTX,
                              {input->var()}, {output->var()},
                              FnProperty::kCPUPrioritized, priority,
                              op_type_name.c_str());
   // In-place
   } else {
-    Engine::Get()->PushAsync(exec_fn, EXEC_CTX,
+    Engine::Get()->PushAsync(exec_fn, MX_EXEC_CTX,
                              {}, {output->var()},
                              FnProperty::kCPUPrioritized, priority,
                              op_type_name.c_str());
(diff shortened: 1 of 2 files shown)


yuxihu:
@apeforest @ctcyang @alsrgv Please help review.

horovod/mxnet/mpi_ops.cc
```diff
@@ -277,49 +307,19 @@ extern "C" int horovod_mxnet_broadcast_async(NDArray* input, NDArray* output,
...
-
-  // Make async copy of CPU tensor to output tensor.
-  TensorUtil::AsyncCopyCPUToCuda(cpu_tensor, output);
-#else
-  auto broadcast_async_fn = [input, output, op_name,
-                             root_rank](RunContext rctx,
-                                        CallbackOnComplete on_complete) mutable {
-    DoBroadcast(input, output, root_rank, op_name, on_complete);
-  };
+  if (input->ctx().dev_mask() == cpu::kDevMask &&
```
alsrgv: Seems like this logic is now duplicated and could be abstracted away as well?

yuxihu: This is actually the check I wanted to add here. If both input and output are in CPU context, there is no need to perform copy from and to CPU.

alsrgv: Nevermind, found few differences between blocks upon closer inspection.

yuxihu:
@alsrgv Looks like tensorflow [breaks](https://buildkite.com/horovod/horovod/builds/53#fa5f8d55-5294-4e5b-a449-7851687085b9/202-416)?

alsrgv:
@yuxihu, just fixed (#999) - could you rebase and try again?

yuxihu:
@alsrgv It passed now. Thanks.

@apeforest Could you please review?

horovod/mxnet/mpi_ops.cc
```diff
@@ -93,35 +74,16 @@ void DoAllgather(NDArray* tensor, NDArray* output, std::string& name,
...
       [on_complete](const Status& status) {
         InvokeCompleteCallback(on_complete, status);
       });
   ThrowIfError(enqueue_result);
 }
-#endif
 
-void DoBroadcast(NDArray* tensor, NDArray* output, int root_rank,
-                 std::string& name, CallbackOnComplete on_complete) {
+void DoBroadcast(NDArray* tensor, NDArray* output, std::string& name,
```
apeforest: const std::string&name?

yuxihu: Let's not worry about this. This function will be changed in the other PR solving the GCC compatibility issue.

yuxihu:
@alsrgv I think we are ready to merge this one.

PR #1011, opened 2019-04-13
Unpin tf-nightly & pin MXNet

alsrgv:
Unpin tf-nightly once tensorflow/tensorflow#27697 is resolved.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -101,7 +101,7 @@ services:
       args:
         MPI_KIND: OpenMPI
         PYTHON_VERSION: 2.7
-        TENSORFLOW_PACKAGE: tf-nightly==1.14.1-dev20190408
+        TENSORFLOW_PACKAGE: tf-nightly
         KERAS_PACKAGE: git+https://github.com/keras-team/keras.git
         PYTORCH_PACKAGE: torch-nightly
         PYSPARK_PACKAGE: pyspark==2.4.0
@@ -112,7 +112,7 @@ services:
       args:
         MPI_KIND: OpenMPI
         PYTHON_VERSION: 3.6
-        TENSORFLOW_PACKAGE: tf-nightly==1.14.1-dev20190408
+        TENSORFLOW_PACKAGE: tf-nightly
         KERAS_PACKAGE: git+https://github.com/keras-team/keras.git
         PYTORCH_PACKAGE: torch-nightly
         PYSPARK_PACKAGE: pyspark==2.4.0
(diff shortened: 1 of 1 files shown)


alsrgv:
It's still broken. Will hold back until it's fixed.

alsrgv:
@apeforest, @yuxihu, looks like latest MXNet introduced a [breaking change](https://github.com/apache/incubator-mxnet/commit/3f3ba92ae1468d08de088d2291ca14e2d5dc5515#diff-726d19e2c39201681813130c8e2ac154R406), could you make a fix?

docker-compose.test.yml
```diff
@@ -101,22 +101,22 @@ services:
       args:
         MPI_KIND: OpenMPI
         PYTHON_VERSION: 2.7
-        TENSORFLOW_PACKAGE: tf-nightly==1.14.1-dev20190408
+        TENSORFLOW_PACKAGE: tf-nightly
         KERAS_PACKAGE: git+https://github.com/keras-team/keras.git
         PYTORCH_PACKAGE: torch-nightly
         PYSPARK_PACKAGE: pyspark==2.4.0
-        MXNET_PACKAGE: mxnet --pre
+        MXNET_PACKAGE: mxnet==1.5.0b20190412 --pre
```
yuxihu: --pre can be removed. Same for other places.

yuxihu:
> @apeforest, @yuxihu, looks like latest MXNet introduced a [breaking change](https://github.com/apache/incubator-mxnet/commit/3f3ba92ae1468d08de088d2291ca14e2d5dc5515#diff-726d19e2c39201681813130c8e2ac154R406), could you make a fix?

Sure, I will look into it.

docker-compose.test.yml
alsrgv: I wanted to keep it to not forget it once we remove specific version tag.

yuxihu: sounds good.

alsrgv:
@yuxihu, looks like there's no published wheel before the breaking change and after the `MXEnginePushAsync` addition.  Since MXNet 1.5.0 is required anyway, I guess we could just make a fix [here](https://github.com/horovod/horovod/blob/master/horovod/mxnet/tensor_util.cc#L138)?  I'll give it a shot.

yuxihu:
> @yuxihu, looks like there's no published wheel before the breaking change and after the `MXEnginePushAsync` addition. Since MXNet 1.5.0 is required anyway, I guess we could just make a fix [here](https://github.com/horovod/horovod/blob/master/horovod/mxnet/tensor_util.cc#L138)? I'll give it a shot.

We are releasing MXNet 1.4.1 where I cherry-picked the GCC fix. If we change to the new TShape API, it will not be backward compatible. I am actually looking into using this [C API](https://github.com/apache/incubator-mxnet/blob/master/include/mxnet/c_api.h#L752).

Can we just use this build `1.5.0b20190412` in the CI for now?

alsrgv:
@yuxihu, unfortunately, `1.5.0b20190412` doesn't have `MXEnginePushAsync` and I haven't seen other Linux wheels on PyPI.  Another option is to make TShape API backward compatible, but I don't have a strong preference.

yuxihu:
@alsrgv I just checked that `MXEnginePushAsync` is in `mxnet==1.5.0b20190412` and `mxnet-cu90==1.5.0b20190412`. For CUDA 10, we can use `mxnet-cu100==1.5.0b20190413`.

alsrgv:
@yuxihu, thanks!  Let me try that.

alsrgv:
That worked.  I'll merge this in.

PR #1052, opened 2019-05-06
Add a ParallelNCCLHierarchicalAllreduce

amazingyyc:
Rough idea:
The NCCLHierarchicalAllreduce can be split to 3 steps:
a: in 1 host, copy tensor to fusion buffer,  reducescatter/reduce, copy fusion data to CPU memory
b: use MPI to do allreduce cross host
c: copy data back to fusion buffer, allgather/bcast, copy data to output tensor
So support have n step, the whole process can be:
`a1, b1, c1, ...., an, bn, cn`

Because all the operation is serially so the total time is:
sum(a1...n) + sum(b1...n) + sum(c1...n)

use 3 thread to do the HierarchicalAllreduce
step a in main thread (Horovod background thread), step b and step c in another 2 thread. Than it become to:
```
a1, a2, ...., an
     b1, b2, ..., bn
           c1, c2, ..., cn
```

The cost time can be: max(sum(a1...n, bn, cn), sum(a1, b1...n, cn), sum(a1, b1, c1...n))

deadlock attention:
because parallel allreduce will do nccl collection operation in multi-thread,
so it maybe cause deadlock But if the GPU have enough resource it can be avoid
ref:https://github.com/NVIDIA/nccl/issues/208


amazingyyc:
@alsrgv How do you thinks about this idea, thanks

alsrgv:
@amazingyyc, I think it's a great idea and something we were thinking about before.  Were you able to measure performance improvement coming out of this PR?  NCCL 2.4.x has built-in support for hierarchical allreduce, so we should compare those two against each other.

We'd need to make sure there are no deadlocks.  I reviewed the NCCL issue, it sounds like they are suggesting to use a single NCCL stream.  You mentioned you had difficulties with that, can you elaborate?

cc @romerojosh

amazingyyc:
Thanks @alsrgv  I have not test the NCCL 2.4.x hierarchical for now, I just test the ParallelNCCLHierarchicalAllreduce with NCCLHierarchicalAllreduce, I have a simple result using tcp with modified Bert model (https://github.com/google-research/bert), it can save about 20% -30% time, I can measure performance improvement much detailed later.

let me explain the deadlock first.
suppose a scenario:2 process all NCCL operation in main thread serial like NCCLHierarchicalAllreduce do, than
```
process-0: op(0), op(1), ..., op(n)
process-1: op(0), op(1), ..., op(n)
```
So it will never get deadlock because op(i) in every process is serial. in process-0 the op(i) can always communicate with process1's op(i).

suppose another scenario:
```
process-0: op(0), op(1)
process-1: op(1), op(0)
```
This will be get deadlock because when process-0 do op(0) it will wait process-1's op(0), but the same time process-1 is doing op(1) and waiting process-0's op(1).

Suppose another scenario have 2 process, every process has 2 thread, every thread bind with a unique nccl_comm and cuda_stream and the 2 thread operate the same GPU. the thread(i) the every process always communicate the i-thread in another process.
```
process-0:
thread-0: op(0), op(1),..., op(n)
thread-1: op(0), op(1),..., op(n)

process-1:
thread-0: op(0), op(1),..., op(n)
thread-1: op(0), op(1),..., op(n)
```

Because every thread bind a unique cuda_stream so it can not guarantee the op's order in different thread. So in process-0 the op order called by GPU can be
process-0:
`thread-0-op(0), thread-1-op(0) .... `
But in process-2 it can be:
`thread-1-op(0), thread-0-op(0)`

If the GPU have enough resource (can do the stream parallel) it works well, If not like some old GPU only can call one stream's kernel at one time (ref:https://stackoverflow.com/questions/3565793/is-there-a-maximum-number-of-streams-in-cuda) or the op(0) occupy all GPU resource than it may be get deadlock. Ref the wiki of nccl, the default thread count is 256 is much smaller than modern GPU.

And as I test in GPU P100, I never get deadlock.

amazingyyc:
About the deadlock explain is based on my experience and some wiki on web. Than I do not have much experience about nvidia GPU hardware architecture, So maybe someone have enough knowledge can explain it clearly.

romerojosh:
Using _multiple_ CUDA streams with _one_ NCCL communicator is a recipe for deadlock. This is because there is no guaranteed order of execution between kernels launched between two streams, _regardless of the order they were queued_ into the streams. So one worker might launch allreduce A before allreduce B, while another worker might launch allreduce B before allreduce A. Since there is only 1 NCCL communicator to serve them, it will deadlock.

Having a distinct NCCL communicator per stream alleviates this issue somewhat, but deadlocks can still occur if the GPU is resource constrained. For example, if you have too many concurrent NCCL kernels in flight, you may run into a scheduling situations where deadlock occurs because some critical allreduce kernels do not have the resources to be scheduled on some workers. 

I have been experimenting with enabling multiple streams for NCCL operations recently (each with their own distinct NCCL communicators, fusion buffers, etc) and it does work well. Perhaps just enabling that alone will provide a good performance boost without needing the hierarchical allreduce?

amazingyyc:
> Using _multiple_ CUDA streams with _one_ NCCL communicator is a recipe for deadlock. This is because there is no guaranteed order of execution between kernels launched between two streams, _regardless of the order they were queued_ into the streams. So one worker might launch allreduce A before allreduce B, while another worker might launch allreduce B before allreduce A. Since there is only 1 NCCL communicator to serve them, it will deadlock.
> 
> Having a distinct NCCL communicator per stream alleviates this issue somewhat, but deadlocks can still occur if the GPU is resource constrained. For example, if you have too many concurrent NCCL kernels in flight, you may run into a scheduling situations where deadlock occurs because some critical allreduce kernels do not have the resources to be scheduled on some workers.
> 
> I have been experimenting with enabling multiple streams for NCCL operations recently (each with their own distinct NCCL communicators, fusion buffers, etc) and it does work well. Perhaps just enabling that alone will provide a good performance boost without needing the hierarchical allreduce?

The ParallelNCCLHierarchicalAllreduce use 3 thread to do the AllReduce, and 2 of them contain the NCCL  operation. As you say. the deadlock maybe not a critical problem. 
I think the hierarchical allreduce still need in some special scenario, like Microsoft Azure. when training on Azure VM. The RDMA does not support the GPU in Azure vm, it only support the intel mpi, So in this scenario hierarchical allreduce have a better performance than NCCL.

amazingyyc:
Attached the test result:
scenario:
machine: Azure VM, nccl 2.4.2, cuda9.0, tensorflow 1.12.0, GPU P100
Model: modified Bert base, batch size 32, 
FABRICS: shm:tcp

script demo , the -np and -ppn will be change in different test:
```
NCCL, NCCL_MIN_NRINGS=12 script demo:

mpirun -np 4 -ppn 2 -hosts item0-0,item1-0 \
-env HOROVOD_FUSION_THRESHOLD=134217728 \
-env I_MPI_DAPL_TRANSLATION_CACHE=0 \
-env I_MPI_DYNAMIC_CONNECTION=0 \
-env I_MPI_FABRICS=shm:tcp \
-env I_MPI_DAPL_PROVIDER=ofa-v2-ib0 \
-env I_MPI_SHM_LMT=shm \
-env I_MPI_SSHM_BUFFER_SIZE=10737418240 \
-env NCCL_SOCKET_IFNAME=eth0 \
-env NCCL_IB_DISABLE=1 \
-env NCCL_MIN_NRINGS=12 \
-genvall \

HOROVOD_HIERARCHICAL_ALLREDUCE:

mpirun -np 3 -ppn 2 -hosts item0-0,item1-0 \
-env HOROVOD_HIERARCHICAL_ALLREDUCE=1 \
-env HOROVOD_FUSION_THRESHOLD=134217728 \
-env I_MPI_DAPL_TRANSLATION_CACHE=0 \
-env I_MPI_DYNAMIC_CONNECTION=0 \
-env I_MPI_FABRICS=shm:tcp \
-env I_MPI_DAPL_PROVIDER=ofa-v2-ib0 \
-env I_MPI_SHM_LMT=shm \
-env I_MPI_SSHM_BUFFER_SIZE=10737418240 \
-env NCCL_SOCKET_IFNAME=eth0 \
-env NCCL_IB_DISABLE=1 \
-genvall \

HOROVOD_PARALLEL_HIERARCHICAL_ALLREDUCE:

mpirun -np 3 -ppn 2 -hosts item0-0,item1-0 \
-env HOROVOD_PARALLEL_HIERARCHICAL_ALLREDUCE=1 \
-env HOROVOD_FUSION_THRESHOLD=134217728 \
-env I_MPI_DAPL_TRANSLATION_CACHE=0 \
-env I_MPI_DYNAMIC_CONNECTION=0 \
-env I_MPI_FABRICS=shm:tcp \
-env I_MPI_DAPL_PROVIDER=ofa-v2-ib0 \
-env I_MPI_SHM_LMT=shm \
-env I_MPI_SSHM_BUFFER_SIZE=10737418240 \
-env NCCL_SOCKET_IFNAME=eth0 \
-env NCCL_IB_DISABLE=1 \
-genvall \
```

result:

scenario\10 steps time |  NCCL, NCCL_MIN_NRINGS=12 | HOROVOD_HIERARCHICAL_ALLREDUCE | HOROVOD_PARALLEL_HIERARCHICAL_ALLREDUCE
:-----------------------:|:--------------------------------:|:-------------------:|:----------------------------------------:
2 host, each has 1 GPU | 14s | 69.2s | 66.8s
2 host, each has 2 GPU | 21s | 44.5s | 35.5s
2 host, one has 1 GPU, another has 2| 18.3s | 73.7s | 64.7s

Unfortunately, I do not have machine that RDMA support the MPI, so I just test the result in tcp connection. As I experience HIERARCHICAL_ALLREDUCE will get a better perf than NCCL-alone.

-env NCCL_MIN_NRINGS=12 set the ring count of NCCL will get error in HIERARCHICAL_ALLREDUCE and PARALLEL_HIERARCHICAL_ALLREDUCE

amazingyyc:
From the result the PARALLEL will get a better perf than current solution in tcp connection. But both of them is much slower than NCCL alone.

alsrgv:
@amazingyyc, we need to find a use case where `PARALLEL_HIERARCHICAL_ALLREDUCE` will have the best performance for this to be merge-able.  On the subject of deadlocks, it looks like your code is already using two comms with corresponding two streams, is that right?

@romerojosh, would be interesting to hear more about your experiments with concurrent NCCL operations.  Maybe let's start a PR and talk there? :-)

cc @karakusc FYI

amazingyyc:
> @amazingyyc, we need to find a use case where `PARALLEL_HIERARCHICAL_ALLREDUCE` will have the best performance for this to be merge-able.

The case that used with `PARALLEL_HIERARCHICAL_ALLREDUCE` is same with `HIERARCHICAL_ALLREDUCE`, `PARALLEL` is the improvement if `HIERARCHICAL_ALLREDUCE`. It will get better performance in Microsoft Azure VM scenario that the VM's RDMA does not support the NCCL.

>On the subject of deadlocks, it looks like your code is already using two comms with corresponding two streams, is that right?

You are right, I have used 2 nccl_comm for 2 cuda_stream in 2 sub-thread.

alsrgv:
@amazingyyc, I'm not sure I fully understand.  Is there an example of a workload that has better performance with `PARALLEL_HIERARCHICAL_ALLREDUCE` than recent NCCL supporting tree algos?  NCCL support both RDMA and TCP networks.

amazingyyc:
> @amazingyyc, I'm not sure I fully understand. Is there an example of a workload that has better performance with `PARALLEL_HIERARCHICAL_ALLREDUCE` than recent NCCL supporting tree algos? NCCL support both RDMA and TCP networks.

Yes, NCCL support both RDMA and TCP networks. But in some machine like Microsoft Azure VM, the RDMA does not support the NCCL, the RDMA in azure only support intel mpi. In this scenario use HIERARCHICAL_ALLREDUCE will get a better perf than only-NCCL.

alsrgv:
@amazingyyc, ah, I see!  Do you have any benchmark that demonstrates this?

amazingyyc:
> @amazingyyc, ah, I see! Do you have any benchmark that demonstrates this?

@alsrgv I only have some old data to prove the HOROVOD_HIERARCHICAL_ALLREDUCE is better than NCCL-only, see below:

Scenario | examples/sec   Disable HOROVOD_HIERARCHICAL,   only NCCL, NCCL_MIN_NRINGS=6 | examples/sec   HOROVOD_HIERARCHICAL with RDMA
-- | -- | --
1 host, each 1 GPU | 3.825 | 3.82
1 host, each 4 GPU | 11.68 | 13.6
4 host, each 1 GPU | 8.06 | 7.6
4 host, each 4 GPU | 27.84 | 37.12

Unfortunately, I do not have Azure VM machine that support the RDMA anymore, So I have no chance to test the perf of HOROVOD_PARALLEL_ HIERARCHICAL_ALLREDUCE.

alsrgv:
@amazingyyc, thanks!  Any idea why HOROVOD_HIERARCHICAL with RDMA is better than NCCL on a single host?  How many times did you run the benchmarks to ensure the stability of results?  Any chance to get Azure VM in future?

amazingyyc:
> Any idea why HOROVOD_HIERARCHICAL with RDMA is better than NCCL on a single host? 
Actually I cannot explain why In 1 host with multi-GPU the NCCL-only is slower than HOROVOD_HIERARCHICAL, As I know 2 differents of NCCL-only and HOROVOD_HIERARCHICAL
1: the NCCL-only use ncclAllReduce to do the task, but HOROVOD_HIERARCHICAL combine the ncclScatter with ncclAllGather
2: The NCCL_MIN_NRINGS can affect the final perf, I only test it use NCCL_MIN_NRINGS=6.

> How many times did you run the benchmarks to ensure the stability of results? 
About 200-300 steps, when it get stable.

> Any chance to get Azure VM in future?
I'm not sure. It does not depend on me actually....... If I have the chance to get the machine I will update the perf test as soon as possible.

alsrgv:
I think it'd be great to rerun these benchmarks multiple times and get confidence intervals.

amazingyyc:
> I think it'd be great to rerun these benchmarks multiple times and get confidence intervals.

Agreed
I can test it If I can get the resource. For now, I do not have appropriate machine to do this. 
Let me close it firstly.

Issue #832, opened 2019-02-15
Is it possible run horovod with unequal steps on each worker?


yayamao-ponyai:
Background, I prefer specify epochs instead of steps, but if the data sizes are not equal on each worker, then the steps could be different.

e.g: for batch size 10, epoch 20,
worker0, data size: 100, total steps: 100 / 10 * 20 = 200
worker1, data size: 110, total steps: 110 / 10 * 20 = 220

But horovod will raise error if steps are unequal on each worker due to some workers may exist before other workers. The error is:

```console
UnknownError (see above for traceback): Horovod has been shut down. 
This was caused by an exception on one of the ranks or an attempt to allreduce, allgather or broadcast a tensor after one of the ranks finished execution.
If the shutdown was caused by an exception, you should see the exception in the log before the first shutdown message.
```

My question is:
Is there a way to let early completed worker to wait other workers? something like `hvd.join()`?

alsrgv:
@yayamao-ponyai, would you want the other ranks that are still training to average gradients with the missing node(s) contributing zero gradients in that case?

LiweiPeng:
A horovod design that can handle this case should also help my original case in $694. For my NMT application, my desired result is that if one worker finishes early, all other rank workers should quit gracefully. It doesn't make much sense to average the remaining N-1 ranks for my case.

alsrgv:
@LiweiPeng, it wouldn't average `N-1`, but rather it would sum all gradients and still divide by N to account for constant LR.

LiweiPeng:
@alsrgv Because I am not sure how this will affect the overall training for NMT, can you make it optional? Option1: sum all the gradients and still divide by N; Option2: just stop training (or discard remaining data)

yayamao-ponyai:
@alsrgv Another solution I can come up is introduce a `join` method, these method does:
1. If a worker finished, tell other workers, so the average become `N - 1`, and the finished worker become waiting mode.
2. After all worker finished, broadcast the final weights to every worker, so make sure every worker has identical weights.

This is because if the worker who is responsible for checkpoint finished earlier can still get a chance to checkpoint after join.

alsrgv:
@LiweiPeng, as we discussed in the other issue, we don't have a good way to nicely abort the training gracefully.  However, typically these data imbalance outliers would be small and it should be OK to actively involve waiting nodes in the finalization process.

[...]

PR #1058, opened 2019-05-07
Add Join op (with only support for AllReduce and PyTorch for now)

kit1980:
This is very earlier work in progress PR to add hvd.join() as described in https://github.com/horovod/horovod/issues/832

This is mostly boilerplate code, it compiles and I can call hvd.join() from the pytorch test (the Join op goes to the queue but does nothing).

The purpose of this PR for me is to understand if I'm going in the right direction.
Some other questions I have (in no particular order):

1. Should Join op have name?
2. Join op doesn't have an input, and should have single int output (last joined rank), right?
3. Does the Join op need to be implemented separately for cuda, nccl, in cuda_operations.cc,  nccl_operations.cc, etc. ?
4. In operation manager, should it be a vector of available Join ops or just one Join op?
5. The logic for join op should track already joined ranks, this should be similar to the logic in AllReduce and AllGather ops? Any code pointers/suggestions?
6. AllReduce and AllGather ops need to know already joined ranks. Is it going to be stored in global state? Any code pointers/suggestions?
7. Any suggestions how to make adding Join op more granular (several smaller PRs)?

### DIFF
--- a/horovod/common/common.h
+++ b/horovod/common/common.h
@@ -214,6 +214,8 @@ public:
                      std::shared_ptr<PersistentBuffer>* tensor) = 0;
   virtual Status AllocateOutput(TensorShape shape,
                                 std::shared_ptr<Tensor>* tensor) = 0;
+  virtual Status AllocateZeros(int64_t num_elements, DataType dtype,
+                                std::shared_ptr<Tensor>* tensor) = 0;
   virtual Framework framework() const = 0;
   virtual ~OpContext() = default;
 };
--- a/horovod/common/common.h
+++ b/horovod/common/common.h
@@ -95,6 +95,9 @@ namespace common {
 // Device ID used for CPU.
 #define CPU_DEVICE_ID (-1)
 
+// Temporary tensor name for ranks that did Join().
+#define JOIN_TENSOR_NAME "join.noname"
+
 // List of supported frameworks.
 enum Framework { TENSORFLOW, PYTORCH, MXNET };
 
--- a/horovod/common/controller.cc
+++ b/horovod/common/controller.cc
@@ -278,6 +278,7 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down,
         responses.push_back(std::move(response));
       }
       if (state.joined_size == size_) {
+        // All ranks did Join(). Send the response, reset joined size.
         Response join_response;
         join_response.set_response_type(Response::JOIN);
         join_response.add_tensor_name(JOIN_TENSOR_NAME);
@@ -613,14 +614,9 @@ ResponseList Controller::FuseResponses(std::deque<Response>& responses) {
 
     if (response.response_type() == Response::ResponseType::ALLREDUCE) {
       // Attempt to add more responses to this fused response.
-      try {
-        const auto& entry =
-            tensor_queue_.GetTensorEntry(response.tensor_names()[0]);
-        tensor_size = entry.tensor->size();
-        dtype = entry.tensor->dtype();
-      } catch (std::out_of_range& e) {
-        // Ranks that did Join don't have the tensor entry.
-      }
+
+      // found_tensor can be false for ranks that did Join.
+      bool found_tensor = tensor_queue_.GetTensorSizeAndType(response.tensor_names()[0], tensor_size, dtype);
 
       std::deque<Response> skipped_responses;
       int64_t skipped_size = 0;
@@ -631,7 +627,7 @@ ResponseList Controller::FuseResponses(std::deque<Response>& responses) {
             tensor_queue_.GetTensorEntry(new_response.tensor_names()[0]);
         int64_t new_tensor_size = new_entry.tensor->size();
 
-        if (tensor_size > 0 &&
+        if (found_tensor &&
             response.response_type() == new_response.response_type() &&
             response.devices() == new_response.devices() &&
             dtype == new_entry.tensor->dtype() &&
--- a/horovod/common/controller.cc
+++ b/horovod/common/controller.cc
@@ -51,7 +51,8 @@ Controller::Controller(ResponseCache& response_cache, TensorQueue& tensor_queue,
       timeline_(timeline), response_cache_(response_cache),
       parameter_manager_(parameter_manager) {}
 
-ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down) {
+ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down,
+                                             HorovodGlobalState& state) {
   // Update cache capacity if autotuning is active.
   if (parameter_manager_.IsAutoTuning()) {
     response_cache_.set_capacity((int)parameter_manager_.CacheEnabled() *
@@ -68,6 +69,11 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down) {
   std::deque<Request> message_queue_tmp;
   tensor_queue_.PopMessagesFromQueue(message_queue_tmp);
   for (auto& message : message_queue_tmp) {
+    if (message.request_type() == Request::JOIN) {
+      state.joined = true;
+      state.join_device = message.device();
+    }
+
     // Keep track of cache hits
     if (response_cache_.capacity() > 0) {
       auto cache_ = response_cache_.cached(message);
@@ -193,7 +199,12 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down) {
         Request message = message_queue_tmp.front();
         message_queue_tmp.pop_front();
 
-        bool reduce = IncrementTensorCount(message);
+        if (message.request_type() == Request::JOIN) {
+          state.joined_size++;
+          continue;
+        }
+
+        bool reduce = IncrementTensorCount(message, state.joined_size);
         stall_inspector_.RecordUncachedTensorStart(
             message.tensor_name(), message.request_rank(), size_);
         if (reduce) {
@@ -211,7 +222,13 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down) {
         auto received_message_list = ready_list[i];
         for (auto& received_message : received_message_list.requests()) {
           auto& received_name = received_message.tensor_name();
-          bool reduce = IncrementTensorCount(received_message);
+
+          if (received_message.request_type() == Request::JOIN) {
+            state.joined_size++;
+            continue;
+          }
+
+          bool reduce = IncrementTensorCount(received_message, state.joined_size);
           stall_inspector_.RecordUncachedTensorStart(
               received_message.tensor_name(), received_message.request_rank(),
               size_);
@@ -222,6 +239,19 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down) {
         if (received_message_list.shutdown()) {
           // Received SHUTDOWN request from one of the workers.
           should_shut_down = true;
+        }
+      }
+
+      // Check if tensors from previous ticks are ready to reduce after Joins.
+      if (state.joined_size > 0) {
+        for (auto& table_iter : message_table_) {
+          int count = (int)table_iter.second.size();
+          if (count == (size_ - state.joined_size) &&
+              std::find(ready_to_reduce.begin(), ready_to_reduce.end(),
+                        table_iter.first) == ready_to_reduce.end()) {
+            state.timeline.NegotiateEnd(table_iter.first);
+            ready_to_reduce.push_back(table_iter.first);
+          }
         }
       }
 
@@ -244,10 +274,16 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down) {
       }
 
       for (auto& tensor_name : ready_to_reduce) {
-        Response response = ConstructResponse(tensor_name);
+        Response response = ConstructResponse(tensor_name, state.joined_size);
         responses.push_back(std::move(response));
       }
-
+      if (state.joined_size == size_) {
+        Response join_response;
+        join_response.set_response_type(Response::JOIN);
+        join_response.add_tensor_name(JOIN_TENSOR_NAME);
+        responses.push_back(std::move(join_response));
+        state.joined_size = 0;
+      }
       response_list = FuseResponses(responses);
       response_list.set_shutdown(should_shut_down);
 
@@ -317,7 +353,7 @@ int64_t Controller::TensorFusionThresholdBytes() {
   return proposed_fusion_threshold;
 }
 
-Response Controller::ConstructResponse(std::string& name) {
+Response Controller::ConstructResponse(std::string& name, int joined_size) {
   bool error = false;
   auto it = message_table_.find(name);
   assert(it != message_table_.end());
@@ -390,11 +426,12 @@ Response Controller::ConstructResponse(std::string& name) {
     }
   }
 
+  std::vector<int64_t> tensor_sizes;
   // If we are doing an allgather, make sure all but the first dimension are
   // the same. The first dimension may be different and the output tensor is
   // the sum of the first dimension. Collect the sizes by rank.
-  std::vector<int64_t> tensor_sizes(requests.size());
   if (message_type == Request::ALLGATHER) {
+    tensor_sizes.resize(requests.size());
     TensorShape tensor_shape;
     for (auto dim : requests[0].tensor_shape()) {
       tensor_shape.AddDim(dim);
@@ -449,6 +486,16 @@ Response Controller::ConstructResponse(std::string& name) {
 
       tensor_sizes[requests[i].request_rank()] = request_shape.dim_size(0);
     }
+  }
+
+  // If there is at least one rank that requested Join, communicate tensor sizes
+  // in the response, because joined ranks don't have this info.
+  if (joined_size > 0 && message_type == Request::ALLREDUCE) {
+    TensorShape tensor_shape;
+    for (auto dim : requests[0].tensor_shape()) {
+      tensor_shape.AddDim(dim);
+    }
+    tensor_sizes.push_back(tensor_shape.num_elements());
   }
 
   // If we are doing a broadcast, check that all root ranks are identical.
@@ -508,6 +555,12 @@ Response Controller::ConstructResponse(std::string& name) {
     }
   } else if (message_type == Request::ALLREDUCE) {
     response.set_response_type(Response::ALLREDUCE);
+    if (joined_size > 0) {
+      for (auto dim : tensor_sizes) {
+        response.add_tensor_size(dim);
+      }
+      response.set_tensor_type(data_type);
+    }
   } else if (message_type == Request::BROADCAST) {
     response.set_response_type(Response::BROADCAST);
   }
@@ -556,11 +609,18 @@ ResponseList Controller::FuseResponses(std::deque<Response>& responses) {
     assert(response.tensor_names().size() == 1);
     responses.pop_front();
     int64_t tensor_size = 0;
+    DataType dtype;
+
     if (response.response_type() == Response::ResponseType::ALLREDUCE) {
       // Attempt to add more responses to this fused response.
-      const auto& entry =
-          tensor_queue_.GetTensorEntry(response.tensor_names()[0]);
-      tensor_size = entry.tensor->size();
+      try {
+        const auto& entry =
+            tensor_queue_.GetTensorEntry(response.tensor_names()[0]);
+        tensor_size = entry.tensor->size();
+        dtype = entry.tensor->dtype();
+      } catch (std::out_of_range& e) {
+        // Ranks that did Join don't have the tensor entry.
+      }
 
       std::deque<Response> skipped_responses;
       int64_t skipped_size = 0;
@@ -571,9 +631,10 @@ ResponseList Controller::FuseResponses(std::deque<Response>& responses) {
             tensor_queue_.GetTensorEntry(new_response.tensor_names()[0]);
         int64_t new_tensor_size = new_entry.tensor->size();
 
-        if (response.response_type() == new_response.response_type() &&
+        if (tensor_size > 0 &&
+            response.response_type() == new_response.response_type() &&
             response.devices() == new_response.devices() &&
-            entry.tensor->dtype() == new_entry.tensor->dtype() &&
+            dtype == new_entry.tensor->dtype() &&
             tensor_size + new_tensor_size <= TensorFusionThresholdBytes()) {
           // These tensors will fuse together well.
           tensor_size += new_tensor_size;
@@ -697,7 +758,7 @@ int Controller::GetLocalSizeAtCrossRank(int i) {
   return local_sizes_for_cross_rank_[i];
 }
 
-bool Controller::IncrementTensorCount(const Request& msg) {
+bool Controller::IncrementTensorCount(const Request& msg, int joined_size) {
   auto& name = msg.tensor_name();
   auto table_iter = message_table_.find(name);
   if (table_iter == message_table_.end()) {
@@ -715,7 +776,7 @@ bool Controller::IncrementTensorCount(const Request& msg) {
 
   std::vector<Request>& messages = table_iter->second;
   int count = (int)messages.size();
-  bool ready_to_reduce = count == size_;
+  bool ready_to_reduce = count == (size_ - joined_size);
   if (ready_to_reduce) {
     timeline_.NegotiateEnd(name);
   }
--- a/horovod/common/controller.cc
+++ b/horovod/common/controller.cc
@@ -428,10 +428,16 @@ Response Controller::ConstructResponse(std::string& name, int joined_size) {
   }
 
   std::vector<int64_t> tensor_sizes;
-  // If we are doing an allgather, make sure all but the first dimension are
-  // the same. The first dimension may be different and the output tensor is
-  // the sum of the first dimension. Collect the sizes by rank.
   if (message_type == Request::ALLGATHER) {
+    if (joined_size > 0) {
+      error = true;
+      error_message_stream << "Allgather is not supported with Join at this time. "
+                           << "Specify sparse_to_dense=True if using DistributedOptimizer";
+    }
+
+    // If we are doing an allgather, make sure all but the first dimension are
+    // the same. The first dimension may be different and the output tensor is
+    // the sum of the first dimension. Collect the sizes by rank.
     tensor_sizes.resize(requests.size());
     TensorShape tensor_shape;
     for (auto dim : requests[0].tensor_shape()) {
@@ -499,8 +505,13 @@ Response Controller::ConstructResponse(std::string& name, int joined_size) {
     tensor_sizes.push_back(tensor_shape.num_elements());
   }
 
-  // If we are doing a broadcast, check that all root ranks are identical.
   if (message_type == Request::BROADCAST) {
+    if (joined_size > 0) {
+      error = true;
+      error_message_stream << "Broadcast is not supported with Join at this time.";
+    }
+
+    // If we are doing a broadcast, check that all root ranks are identical.
     int first_root_rank = requests[0].root_rank();
     for (unsigned int i = 1; i < requests.size(); ++i) {
       if (error) {
--- a/horovod/common/global_state.h
+++ b/horovod/common/global_state.h
@@ -102,9 +102,10 @@ struct HorovodGlobalState {
 
   // Number of ranks that did Join()
   int joined_size = 0;
+  // If a rank is Joined, AllReduce uses temporary 0 tensors for it.
   bool joined = false;
   // ID of the device to create temporary tensors while Joined
-  int join_device = -1;
+  int join_device = CPU_DEVICE_ID;
 
   ~HorovodGlobalState() {
     // Make sure that the destructor of the background thread is safe to
--- a/horovod/common/global_state.h
+++ b/horovod/common/global_state.h
@@ -100,6 +100,12 @@ struct HorovodGlobalState {
   // operations.
   LibType control_operation;
 
+  // Number of ranks that did Join()
+  int joined_size = 0;
+  bool joined = false;
+  // ID of the device to create temporary tensors while Joined
+  int join_device = -1;
+
   ~HorovodGlobalState() {
     // Make sure that the destructor of the background thread is safe to
     // call. If a thread is still joinable (not detached or complete) its
--- a/horovod/common/message.cc
+++ b/horovod/common/message.cc
@@ -72,6 +72,9 @@ const std::string& Request::RequestType_Name(RequestType value) {
     case RequestType::BROADCAST:
       static const std::string broadcast("BROADCAST");
       return broadcast;
+    case RequestType::JOIN:
+      static const std::string join("JOIN");
+      return join;
     default:
       static const std::string unknown("<unknown>");
       return unknown;
@@ -236,6 +239,9 @@ const std::string& Response::ResponseType_Name(ResponseType value) {
     case ResponseType::BROADCAST:
       static const std::string broadcast("BROADCAST");
       return broadcast;
+    case ResponseType::JOIN:
+      static const std::string join("JOIN");
+      return join;
     case ResponseType::ERROR:
       static const std::string error("ERROR");
       return error;
@@ -256,6 +262,10 @@ void Response::set_response_type(ResponseType value) {
 const std::vector<std::string>& Response::tensor_names() const {
   return tensor_names_;
 }
+
+DataType Response::tensor_type() const { return tensor_type_; }
+
+void Response::set_tensor_type(DataType value) { tensor_type_ = value; }
 
 const std::string Response::tensor_names_string() const {
   std::string result;
@@ -321,6 +331,7 @@ void Response_ParseFromWire(Response& response,
   for (const auto& tensor_name_obj : *obj->tensor_names()) {
     response.add_tensor_name(tensor_name_obj->str());
   }
+  response.set_tensor_type((DataType) obj->tensor_type());
   response.set_error_message(obj->error_message()->str());
   response.set_devices(
       std::vector<int32_t>(obj->devices()->begin(), obj->devices()->end()));
@@ -347,6 +358,8 @@ void Response_SerializeToWire(const Response& response,
   response_builder.add_response_type(
       (wire::ResponseType) response.response_type());
   response_builder.add_tensor_names(tensor_names_wire);
+  response_builder.add_tensor_type(
+      (wire::DataType) response.tensor_type());
   response_builder.add_error_message(error_message_wire);
   response_builder.add_devices(devices_wire);
   response_builder.add_tensor_sizes(tensor_sizes_wire);
--- a/horovod/common/message.h
+++ b/horovod/common/message.h
@@ -45,7 +45,7 @@ const std::string& DataType_Name(DataType value);
 class Request {
 public:
   enum RequestType {
-    ALLREDUCE = 0, ALLGATHER = 1, BROADCAST = 2
+    ALLREDUCE = 0, ALLGATHER = 1, BROADCAST = 2, JOIN = 3
   };
 
   static const std::string& RequestType_Name(RequestType value);
@@ -130,7 +130,7 @@ private:
 class Response {
 public:
   enum ResponseType {
-    ALLREDUCE = 0, ALLGATHER = 1, BROADCAST = 2, ERROR = 3
+    ALLREDUCE = 0, ALLGATHER = 1, BROADCAST = 2, JOIN = 3, ERROR = 4
   };
 
   static const std::string& ResponseType_Name(ResponseType value);
@@ -141,6 +141,10 @@ public:
 
   // Empty if the type is DONE or SHUTDOWN.
   const std::vector<std::string>& tensor_names() const;
+
+  DataType tensor_type() const;
+
+  void set_tensor_type(DataType value);
 
   const std::string tensor_names_string() const;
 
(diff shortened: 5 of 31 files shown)


kit1980:
8. For PyTorch, are both mpi_lib_v2 and mpi_lib implementations will be needed eventually or only v2?

.cache/v/cache/lastfailed
```diff
@@ -0,0 +1,8 @@
+{
```
alsrgv: Let's remove this file from the PR

horovod/common/message.cc
```diff
@@ -237,6 +240,9 @@ const std::string& Response::ResponseType_Name(ResponseType value) {
     case ResponseType::ERROR:
       static const std::string error("ERROR");
       return error;
+    case ResponseType::JOIN:
```
alsrgv: Nit: move JOIN before ERROR

horovod/common/message.h
```diff
@@ -130,7 +130,7 @@ class RequestList {
 class Response {
 public:
   enum ResponseType {
-    ALLREDUCE = 0, ALLGATHER = 1, BROADCAST = 2, ERROR = 3
+    ALLREDUCE = 0, ALLGATHER = 1, BROADCAST = 2, ERROR = 3, JOIN = 4
```
alsrgv: Nit: move JOIN before ERROR, reorder #s

horovod/common/operations.cc
```diff
@@ -1772,5 +1773,39 @@ Status EnqueueTensorBroadcast(std::shared_ptr<OpContext> context,
...
 }
 
+// MPI must be initialized and the background thread must be running before
+// this function is called.
+Status EnqueueJoin(std::shared_ptr<OpContext> context,
+                              std::shared_ptr<ReadyEvent> ready_event,
+                              const std::string name, const int device,
+                              StatusCallback callback) {
+
+  printf("EnqueueJoin start\n");
```
alsrgv: Remove debug printf

horovod/common/operations.cc
```diff
@@ -1772,5 +1773,39 @@ Status EnqueueTensorBroadcast(std::shared_ptr<OpContext> context,
...
+    return SHUT_DOWN_ERROR;
+  }
+  if (horovod_global.tensor_table.find(name) !=
+      horovod_global.tensor_table.end()) {
+    return DUPLICATE_NAME_ERROR;
+  }
+  horovod_global.tensor_table.emplace(name, std::move(e));
+  horovod_global.message_queue.push(message);
+  LOG(TRACE, horovod_global.rank) << "Enqueued " << name;
+  printf("EnqueueJoin end\n");
```
alsrgv: Remove debug printfs

horovod/common/operations.cc
```diff
@@ -1772,5 +1773,39 @@ Status EnqueueTensorBroadcast(std::shared_ptr<OpContext> context,
   return Status::OK();
 }
 
+// MPI must be initialized and the background thread must be running before
+// this function is called.
+Status EnqueueJoin(std::shared_ptr<OpContext> context,
+                              std::shared_ptr<ReadyEvent> ready_event,
```
alsrgv: Nit: use `clang-format` to format code w/o unaligned spaces

horovod/common/ops/collective_operations.h
```diff
@@ -116,6 +116,19 @@ class ErrorOp : public HorovodOp {
   virtual Status Execute(std::vector<TensorTableEntry>& entries, const Response& response);
 };
 
+class JoinOp : public HorovodOp {
```
alsrgv: Nit: move above Error?

horovod/common/ops/collective_operations.h
```diff
@@ -116,6 +116,19 @@ class ErrorOp : public HorovodOp {
...
+public:
+  JoinOp(HorovodGlobalState* global_state);
+
+  virtual ~JoinOp() = default;
+
+  virtual Status Execute(std::vector<TensorTableEntry>& entries, const Response& response);
+
+  // virtual bool Enabled(const ParameterManager& param_manager,
```
alsrgv: Remove?

horovod/common/ops/operation_manager.cc
```diff
@@ -59,6 +61,15 @@ Status OperationManager::ExecuteBroadcast(std::vector<TensorTableEntry>& entries
   throw std::logic_error("No Broadcast operation enabled");
 }
 
+Status OperationManager::ExecuteJoin(std::vector<TensorTableEntry>& entries,
+                                          const Response& response) const {
+  // if (join_op_->Enabled(*param_manager_, entries, response)) {
```
alsrgv: Remove?

horovod/common/ops/collective_operations.cc
```diff
@@ -175,5 +175,13 @@ Status ErrorOp::Execute(std::vector<TensorTableEntry>& entries, const Response&
   return Status::PreconditionError(response.error_message());
 }
 
+// Join
+JoinOp::JoinOp(HorovodGlobalState* global_state) : HorovodOp(global_state) {}
```
alsrgv: Move up before Error, add TODO?

.cache/v/cache/lastfailed
kit1980: Done.

horovod/common/message.cc
kit1980: Done.

horovod/common/message.h
kit1980: Done.

horovod/common/operations.cc
kit1980: Done.

horovod/common/operations.cc
kit1980: Done.

horovod/common/ops/collective_operations.h
kit1980: Done.

horovod/common/ops/operation_manager.cc
kit1980: Done.

horovod/common/ops/collective_operations.cc
kit1980: Done.

kit1980:
Rebased on master, addressed style comments, fixed "Segmentation fault" because of trying to cache Join op.

horovod/common/ops/collective_operations.h
kit1980: Done.

horovod/common/operations.cc
kit1980: Done.

horovod/common/message.cc
```diff
@@ -70,6 +70,9 @@ const std::string& Request::RequestType_Name(RequestType value) {
     case RequestType::BROADCAST:
       static const std::string broadcast("BROADCAST");
       return broadcast;
+    case RequestType::JOIN:
+      static const std::string join("JOIN");
+      return broadcast;
```
alsrgv: Bug, should be `return join`

horovod/common/message.h
```diff
@@ -130,7 +130,7 @@ class RequestList {
 class Response {
 public:
   enum ResponseType {
-    ALLREDUCE = 0, ALLGATHER = 1, BROADCAST = 2, ERROR = 3
+    ALLREDUCE = 0, ALLGATHER = 1, BROADCAST = 2, JOIN = 4, ERROR = 3
```
alsrgv: Can you sort the numbers, i.e. JOIN = 3, ERROR = 4.  There's no need for backward compatibility in this layer as all workers must have the same Horovod version.

horovod/common/wire/message.fbs
```diff
@@ -73,7 +74,8 @@ enum ResponseType:byte {
     ALLREDUCE = 0,
     ALLGATHER = 1,
     BROADCAST = 2,
-    ERROR = 3
+    ERROR = 3,
+    JOIN = 4
```
alsrgv: Need to regenerate this file

horovod/common/wire/message_generated.h
```diff
@@ -1,18 +1,3 @@
-// Copyright 2019 Uber Technologies, Inc. All Rights Reserved.
```
alsrgv: Please manually put back the copyright header

horovod/common/wire/message_generated.h
```diff
@@ -125,16 +113,18 @@ enum ResponseType {
   ResponseType_ALLGATHER = 1,
   ResponseType_BROADCAST = 2,
   ResponseType_ERROR = 3,
+  ResponseType_JOIN = 4,
```
alsrgv: Need to regenerate this file

horovod/torch/mpi_ops.py
```diff
@@ -436,3 +436,9 @@ def synchronize(handle):
     mpi_lib.horovod_torch_wait_and_clear(handle)
     _, output = _handle_map.pop(handle)
     return output
+
+
+def join(name=None):
+    """
+    """
+    return mpi_lib.horovod_torch_join()
```
alsrgv: Please add comment and end-of-file newline

test/test_torch.py
```diff
@@ -1293,5 +1293,28 @@ def test_synchronize_step_warning(self):
...
+            dtypes += [torch.cuda.ByteTensor, torch.cuda.CharTensor, torch.cuda.ShortTensor,
+                       torch.cuda.IntTensor, torch.cuda.LongTensor,
+                       torch.cuda.FloatTensor, torch.cuda.DoubleTensor]
+            if _fp16_supported:
+                dtypes += [torch.cuda.HalfTensor]
+        dims = [1, 2, 3]
+        for dtype, dim in itertools.product(dtypes, dims):
+            tensor = torch.FloatTensor(*([17] * dim)).fill_(1).mul_(rank)
+            tensor = self.cast_and_place(tensor, dtype)
+            ret = hvd.join()
```
alsrgv: The join test should test the actual Join behavior - e.g. you can allreduce on rank 0, join on rank 1, and rank 1 will allreduce with zero dummy tensor.

kit1980:
Hi @alsrgv,

I've coded some basic logic for the JoinOp (the code is not in this PR currently yet): global state stores number of ranks already joined on the coordinator, and each rank knows about itself if it already joined. Then `IncrementTensorCount` function takes into account number of ranks already joined, and sets `ready_to_reduce` if the number of ranks reducing the tensor equals to `mpi_size - joined_count`. And each rank checks it's own `joined` bit to decide if it needs to actually perform the reduce operation after getting reduce response.

The problems is that the ranks that actually perform the reduce (not yet joined) use MPI or NCCL reduce, that expects all ranks to participate. So one way to solve it is to make every rank know the list of already joined ranks, and create new MPI communicator that only uses subset of all nodes; this seems not really feasible... Another idea is that all joined ranks still participate in MPI reduce (Join op Execute calls MPI reduce), but with zero tensors; but in this case the joined nodes need to somehow know the shape of the reduced tensor...

Any suggestions?

kit1980:
An update after my last comment.
Now I'm trying to implement this design:

The coordinator tracks how many ranks sent Join request, and each rank knows if it Joined.
Then, when sending Reduce response, if at least one rank Joined, the coordinator sends Join response with the information about the reduced tensors - type and size (there will be many Join responses after one Join request). The Join response will be processed by the joined ranks only, and these ranks will use the information from the response to fill their tensor_table (the actual content of the tensors will be 0).
Then all the ranks will execute Reduce normally.

@alsrgv, could you review if it makes sense?

kit1980:
Another problem I have is how to keep Python thread alive after one rank Joins earlier. The wait must be in the Python thread, not in the background thread, and I'm not sure how to communicate back that the wait is over (all ranks sent Join requests).

kit1980:
So I've implemented this design (similar to my previous comment, but with some modifications):

The coordinator tracks how many ranks already sent Join requests, and each rank knows if it Joined.
Running join() from Python blocks Python thread for that rank until Join response, which is sent only when all ranks Joined.
When sending Reduce response, if at least one rank Joined, the coordinator includes tensor type and size information in the Reduce response. This information is used by the Joined ranks to construct temporary 0-filled data for MPI AllReduce, while non-Joined ranks execute normal AllReduce.

This seems to work OK with 1 small test I have. If you think this approach makes sense, I'll fix all TODOs and test/fix more scenarios (Fused responses, all data types, cache...)

alsrgv:
@kit1980, sorry for going dark on you.  Your proposed approach sounds good!

alsrgv:
Note that it should work with AllGather, too (send empty slices from joined nodes).

horovod/common/operations.cc
```diff
@@ -205,17 +205,17 @@ OperationManager* CreateOperationManager(HorovodGlobalState& state) {
...
-                              allgather_ops, broadcast_ops, error_op);
+  return new OperationManager(&state.param_manager, allreduce_ops, allgather_ops, broadcast_ops, join_op, error_op);
 }
 
 // Store the Request for a name, and return whether the total count of
 // Requests for that tensor is now equal to the MPI size (and thus we are
 // ready to reduce the tensor).
 bool IncrementTensorCount(std::unique_ptr<MessageTable>& message_table,
-                          const Request& msg, int mpi_size) {
+                          const Request& msg, int mpi_size, int joined_size = 0) {
```
thiagocrepaldi: NIT: should it be uint32_t instead of int?

kit1980: This should be the same type as `HorovodGlobalState.size`, which is `int`. I'll change `HorovodGlobalState.joined_size` to be `int`.
BTW, this PR is not exactly ready for review at the moment...

kit1980:
Join with AllReduce now passes tests for all data types for MPI and NCCL.
Please review.

DEKHTIARJonathan:
I have a simple question how different it is from this:

```
from mpi4py import MPI
MPI.COMM_WORLD.Barrier()  # Waiting for all MPI processes to sync
```

kit1980:
@DEKHTIARJonathan, this is for the case when data is unevenly distributed between workers, so different workers do different _number of steps_. MPI barrier can't help in this case - the workers that still have data will forever wait in `optimizer.step()` for the workers that already processed all their data.
Please see this issue for more details and discussion: #832

DEKHTIARJonathan:
Oh I see. Thanks for the pointer. Never had this usecase

kit1980:
@alsrgv Could you take a look at this?

horovod/common/operations.cc
```diff
@@ -235,7 +235,7 @@ bool IncrementTensorCount(std::unique_ptr<MessageTable>& message_table,
 
   std::vector<Request>& messages = std::get<0>(table_iter->second);
   int count = (int)messages.size();
-  bool ready_to_reduce = count == mpi_size;
+  bool ready_to_reduce = count == (mpi_size - joined_size);
```
alsrgv: What happens with allgather & broadcast?  Could we support them as well?

horovod/common/operations.cc
```diff
@@ -648,28 +665,126 @@ int64_t GetTensorDataForAutotuner(const ResponseList& response_list,
...
 // Process a Response by doing a reduction, a gather, a broadcast, or
 // raising an error.
-void PerformOperation(TensorTable& tensor_table, Response response) {
+void PerformOperation(TensorTable& tensor_table, Response response,
+                      HorovodGlobalState& state) {
+  if (response.response_type() == Response::JOIN) {
+    if (!state.joined) {
+      return;
+    } else {
+      auto iter = tensor_table.find("join.noname");
```
alsrgv: Turn `join.noname` into a constant in a macro

horovod/common/operations.cc
```diff
@@ -648,28 +665,126 @@ int64_t GetTensorDataForAutotuner(const ResponseList& response_list,
...
+          iter->second.device = state.join_device;
+        }
+
+        entries.push_back(iter->second);
+        // Clear the tensor table of this tensor and its callbacks; the rest of
+        // this function takes care of it.
+        tensor_table.erase(iter);
+      } else {
+        TensorTableEntry entry;
+        switch (response.tensor_type()) {
```
alsrgv: Extract this switch statement into a helper function, so switch above and here can be deduplicated?

horovod/common/global_state.h
```diff
@@ -146,6 +146,12 @@ struct HorovodGlobalState {
   // A string indicating what framework we are using to perform CPU operations.
   std::string cpu_operation;
 
+  // Number of ranks that did Join()
+  int joined_size = 0;
+  bool joined = false;
```
alsrgv: Can you add comment clarifying what does `joined` do and what does it mean if it's true or false?

kit1980:
Rebased on master (which was painful because of recent large code moves with simultaneous logic and formatting changes).

horovod/common/operations.cc
kit1980: I haven't added support for allgather & broadcast yet, but it should be relatively easy after allreduce.
I hope to get the allreduce PR in and then add support for allgather & broadcast & non-PyTorch frameworks in subsequent PRs.

horovod/common/operations.cc
kit1980: Removed the duplication.

horovod/common/operations.cc
```diff
@@ -249,7 +249,7 @@ bool IncrementTensorCount(std::unique_ptr<MessageTable>& message_table,
 //
 // Constructing the Response, thus, requires a whole lot of error checking.
 Response ConstructResponse(std::unique_ptr<MessageTable>& message_table,
-                           std::string name) {
+                           std::string name, uint32_t joined_size) {
```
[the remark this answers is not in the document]
kit1980: Done.

horovod/common/operations.cc
huoliquankai7: Hello, please, if you want to support tensorflow, is it necessary to implement a join op.
E.g:
Class HorovodJoinOp : public AsyncOpKernel{}

kit1980: @huoliquankai7 In the end TensorFlow will be supported, this particular PR supports only PyTorch.

horovod/common/operations.cc
```diff
@@ -200,17 +200,28 @@ OperationManager* CreateOperationManager(HorovodGlobalState& state) {
...
   return new OperationManager(&state.parameter_manager, allreduce_ops,
-                              allgather_ops, broadcast_ops, error_op);
+                              allgather_ops, broadcast_ops, join_op, error_op);
 }
 
 // Process a Response by doing a reduction, a gather, a broadcast, or
 // raising an error.
-void PerformOperation(Response response) {
+void PerformOperation(Response response, HorovodGlobalState& state) {
+  if (response.response_type() == Response::JOIN) {
```
tgaddair: Making this into a special case seems strange since you already have the `JoinOp`, which currently does nothing (throws `PreconditionError`).  Can we just move this logic into the `JoinOp` to make the code more general?

horovod/common/operations.cc
```diff
@@ -231,7 +242,9 @@ void PerformOperation(Response response) {
     if (!status.ok()) {
       for (auto& e : entries) {
         timeline.End(e.tensor_name, nullptr);
-        e.callback(status);
+        if (!state.joined) {
```
tgaddair: In `cuda_operations.cc` you just check that the callback is not null.  Maybe we can do the same here to make it more general?

horovod/common/operations.cc
```diff
@@ -273,7 +286,9 @@ void PerformOperation(Response response) {
   if (!status.in_progress()) {
     for (auto& e : entries) {
       timeline.End(e.tensor_name, status.ok() ? e.output : nullptr);
-      e.callback(status);
+      if (!state.joined) {
+        e.callback(status);
```
tgaddair: Same as above.

horovod/torch/mpi_ops.cc
```diff
@@ -216,6 +216,38 @@ int DoBroadcastCudaOnCPU(TC* tensor, TC* output, int root_rank, char* name) {
...
+
+int DoJoin(int device) {
+  ThrowIfError(common::CheckInitialized());
+
+  auto handle = handle_manager.AllocateHandle();
+  auto ready_event = RecordReadyEvent(device);
+
+  auto hvd_cpu_buffer =
+      std::make_shared<TorchTemporaryBuffer<DataType::HOROVOD_INT64, DeviceType::CPU, THLongTensor>>(
+          CPU_DEVICE_ID);
```
tgaddair: Why `CPU_DEVICE_ID`?  If we use the `device` passed in to the function, we can allocate Tensors on the correct device through the framework itself instead of needing to do it ourselves.

horovod/common/controller.cc
```diff
@@ -225,6 +242,19 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down) {
         }
       }
 
+      // Check if tensors from previous ticks are ready to reduce after Joins.
+      if (state.joined_size > 0) {
+        for (auto& table_iter : message_table_) {
+          int count = (int)table_iter.second.size();
+          if (count == (size_ - state.joined_size) &&
+              std::find(ready_to_reduce.begin(), ready_to_reduce.end(),
```
tgaddair: Can you explain why this scan through the list is necessary?  Under what circumstances would the tensor already be in the `ready_to_reduce` list?

horovod/common/controller.cc
```diff
@@ -244,10 +274,16 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down) {
       }
 
       for (auto& tensor_name : ready_to_reduce) {
-        Response response = ConstructResponse(tensor_name);
+        Response response = ConstructResponse(tensor_name, state.joined_size);
         responses.push_back(std::move(response));
       }
-
+      if (state.joined_size == size_) {
```
tgaddair: Please add a comment here.

horovod/common/controller.cc
```diff
@@ -451,6 +488,16 @@ Response Controller::ConstructResponse(std::string& name) {
     }
   }
 
+  // If there is at least one rank that requested Join, communicate tensor sizes
+  // in the response, because joined ranks don't have this info.
+  if (joined_size > 0 && message_type == Request::ALLREDUCE) {
```
tgaddair: What about Allgather?

huoliquankai7:
@kit1980 Hey, I have a problem now: when tensorflow goes through multiple steps, one of the workers adds a join request (missing state),but other workers' tensor will hit the cache,(not join request)。After entering the sync function, all the tensors of other workers' hits are put back into the message_queue, waiting for the next loop, so that the program has been stuck, how to solve this situation?
the background to this problem is based on the join operation that I wrote to support tensorflow. The support for pytorch mentioned in pr is no problem. According to my understanding, is pytorch not using cache for the sake of suffix? (maybe my understanding is wrong), so pytorch will not have this problem, and tensorflow appears.

horovod/common/ops/cuda_operations.cc
```diff
@@ -159,7 +159,10 @@ Status CUDAAllreduce::FinalizeCUDAQueue(const std::vector<TensorTableEntry>& ent
   // Claim a std::shared_ptr to the fusion buffer to prevent its memory from being reclaimed
   // during finalization.
   auto fusion_buffer = global_state_->fusion_buffer.GetBuffer(
-      first_entry.device, first_entry.context->framework(), global_state_->current_nccl_stream);
+      first_entry.device,
+      // JoinTensor doesn't have context or framework
+      first_entry.context ? first_entry.context->framework() : PYTORCH,
```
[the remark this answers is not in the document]
kit1980: Here PYTORCH isn't actually used afterwards, it can be any valid framework constant.
But I agree with your suggestion about delegating Tensor allocation to the framework, I'll try to do that.

horovod/common/controller.cc
kit1980: My intention is to make this PR about AllReduce only, and add AllGather in the next PR.

horovod/common/tensor_queue.cc
```diff
@@ -69,26 +69,80 @@ TensorQueue::GetTensorDataForAutotuner(const ResponseList& response_list,
...
+        case HOROVOD_INT32:
+          entry.tensor = std::make_shared<JoinTensor<HOROVOD_INT32, int32_t>>(
+              join_device, response.tensor_sizes()[0]);
+          break;
+        case HOROVOD_INT64:
+          entry.tensor = std::make_shared<JoinTensor<HOROVOD_INT64, int64_t>>(
+              join_device, response.tensor_sizes()[0]);
+          break;
+        case HOROVOD_FLOAT16:
+          entry.tensor = std::make_shared<JoinTensor<HOROVOD_FLOAT16, float>>(
```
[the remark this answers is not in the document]
kit1980: @tgaddair Should it be eventually a switch statement on framework, then calling framework-specific allocator? This doesn't seem right, because now Join op code needs to know about every framework.
If `float` doesn't work for fp16 (it works in my tests, but maybe I need better tests), maybe some int variant will work? It only needs to work for value 0...

horovod/common/controller.cc
kit1980: The code scans for tensors that are currently **not** in the `ready_to_reduce` list, but have enough `table_iter.second.size()` that now, after new rank joined, they need to be in `ready_to_reduce`.

horovod/common/tensor_queue.cc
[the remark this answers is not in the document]
tgaddair: There's no need to use a switch statement here.  If you implement the allocation of the correct tensor within [OpContext](https://github.com/horovod/horovod/blob/master/horovod/common/common.h#L206), then polymorphism will take care the rest.

horovod/common/global_state.h
```diff
@@ -100,6 +100,12 @@ struct HorovodGlobalState {
   // operations.
   LibType control_operation;
 
+  // Number of ranks that did Join()
+  int joined_size = 0;
+  bool joined = false;
```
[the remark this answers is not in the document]
kit1980: My terminology idea was that a rank "does join" or "joins" when it sends Join request.
I can change this, and all comments when I say "did Join()" or similar to "sent Join request", if you think it's more clear.

kit1980:
Pushed intermediate update. I think I've addressed all the comments from the last review except creation of temporary tensors using the underlying frameworks.

horovod/common/controller.cc
tgaddair: If we exclude Allgather and Broadcast from this commit, then we should to be careful about handling the case where the user submits an Allgather or Broadcast request during a join.  Ideally, we should throw an error saying "Allgather is not supported with Join at this time. Specify sparse_to_dense=True if using DistributedOptimizer".  Something like that.

horovod/common/controller.cc
tgaddair: I understand that, but my question is: why is it necessary to check if the tensor is in `ready_to_reduce` before doing the insertion?  Why not just insert it without the check?  I am not sure how it's possible that a tensor could already be in the list at the time this condition is met.  Can multiple `table_iter` objects meet the `count == (size_ - state.joined_size` condition for the same `table_iter.first` object?

kit1980: I see. I think you're right, no need to check. I'll fix it.

kit1980: I've tested this, and the check is actually needed because about 40 lines before this the coordinator can add to  `ready_to_reduce` based in its own messages.

horovod/common/tensor_queue.cc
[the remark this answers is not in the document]
kit1980: Hi @tgaddair,
So I'm planning to implement it this way. Any suggestions?

In `GetTensorEntriesFromResponse` to create temporary zero tensor, call `AllocatePersistent` of the `JOIN_TENSOR_NAME` context. `AllocatePersistent` will automatically call framework-specific implementation.

`AllocatePersistent` needs size in bytes, that can be computed from size in elements using  `GetTypeSize()` https://github.com/horovod/horovod/blob/c5b53eca9686ba296ba3a8b3a7e25dda25065eea/horovod/common/gloo/gloo_controller.cc#L94

The newly allocated buffer needs to be filled with zeros. Not really sure how to do it best, especially for fp16.

Or do you think reusing `AllocatePersistent` for this is not a good idea?

kit1980:
@huoliquankai7 Hi, I'm not sure I understand the TensorFlow issue now. Let me first complete this PyTorch-only part.

tgaddair:
Hey @kit1980 just wanted to checkin and see how the refactor is going.  Let me know if you have any questions about the OpContext interface.

kit1980:
> Hey @kit1980 just wanted to checkin and see how the refactor is going.  Let me know if you have any questions about the OpContext interface.
Can you comment on this? https://github.com/horovod/horovod/pull/1058#discussion_r330836127

horovod/common/tensor_queue.cc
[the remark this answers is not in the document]
tgaddair: Hey @kit1980, there are two limitations to `AllocatePersistent`, one is (as you mentioned) it returns an empty tensor, not a zeroed-tensor, and the other is that it allocates a tensor of data type int8.  What I think would be better would be to have a new method alongside `AllocatePersistent` like `AllocateZeros` that creates a zeroed-tensor of a given data type.  

So:

```
AllocateZeros(int device, int64_t size, horovod::common::DataType dtype)
```

TensorFlow and PyTorch both have C++ APIs for creating zero tensors and/or zeroing out tensors that can be used.

huoliquankai7:
Hi,@alsrgv @kit1980 @tgaddair  I have written a join function code that supports tensorflow(The horovod framework code for the join function is based on @kit1980 ). Can I provide pr?

kit1980:
> Hey @kit1980 just wanted to checkin and see how the refactor is going. Let me know if you have any questions about the OpContext interface.

Hi @tgaddair, I refactored (not pushed yet) as you suggested, and everything seems to work OK for PyTorch 1.0 and later (adapter_v2.cc, mpi_ops_v2.cc, etc.). But for older PyTorch (adapter.cc, mpi_ops.cc, etc.) it looks much more cumbersome and requires to call cuda API directly anyway.

Do you think it's OK to not support Join op for PyTorch before version 1.0? (I'll have stub implementations and skip all Join tests for older versions).

tgaddair:
Hey @kit1980, I think it's fine to not support the Join Op prior to v1.0.  The one thing we should be careful of is to throw a good error if the user is using an old version of PyTorch and calls `join` (in the Python API) or `AllocateZeroTensor` (in the C++ API).

tgaddair:
> Hi,@alsrgv @kit1980 @tgaddair I have written a join function code that supports tensorflow(The horovod framework code for the join function is based on @kit1980 ). Can I provide pr?

Hey @huoliquankai7, would be great to see a PR.  Is it based on this one or an entirely new implementation?

huoliquankai7:
> > Hi,@alsrgv @kit1980 @tgaddair I have written a join function code that supports tensorflow(The horovod framework code for the join function is based on @kit1980 ). Can I provide pr?
> 
> Hey @huoliquankai7, would be great to see a PR. Is it based on this one or an entirely new implementation?

Hey @tgaddair ,it is not an entirely new implementation, it is based on @kit1980 . I just added a join_op that supports tensorflow.

tgaddair:
Nice!  Sounds good @huoliquankai7.  Feel free to create a PR now if you'd like and I can review it.  We'll just be sure to rebase/land your change after this lands.

kit1980:
Hi @tgaddair,
I believe I addressed all your comments, could you have another look?

huoliquankai7:
Hi @kit1980 I have a question: Does horovod's cache support the join function of pytorch? Because once there is a rank join, according to the cache, the tensor of other rank hits will be put back into tensor_queue_, so that the current horovod is always in an infinite loop state.

kit1980:
> Hi @kit1980 I have a question: Does horovod's cache support the join function of pytorch? Because once there is a rank join, according to the cache, the tensor of other rank hits will be put back into tensor_queue_, so that the current horovod is always in an infinite loop state.

Can you provide a test that exhibits this behavior? I don't understand why this would happen "the tensor of other rank hits will be put back into tensor_queue_". I have test_horovod_join_allreduce for PyTorch that passes OK.

horovod/common/common.h
```diff
@@ -211,6 +214,8 @@ class OpContext {
                      std::shared_ptr<PersistentBuffer>* tensor) = 0;
   virtual Status AllocateOutput(TensorShape shape,
                                 std::shared_ptr<Tensor>* tensor) = 0;
+  virtual Status AllocateZeros(int64_t num_elements, DataType dtype,
```
tgaddair: Can you add stub implementations of this method to the other `OpContext` children, including `TFOpContext` in `tensorflow/mpi_ops.cc` and `MXOpContext` in `mxnet/adapter.{h,cc}`?  These implementations should raise an error similar to the one for PyTorch <1.0.

horovod/common/tensor_queue.h
```diff
@@ -22,6 +22,10 @@
 
 #include "common.h"
 
+#if HAVE_CUDA
```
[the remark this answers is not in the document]
kit1980: Not used anymore, removed.

horovod/torch/adapter_v2.cc
```diff
@@ -101,6 +124,20 @@ Status TorchOpContext::AllocateOutput(TensorShape shape,
   return Status::OK();
 }
 
+Status TorchOpContext::AllocateZeros(int64_t num_elements, DataType dtype,
+                                      std::shared_ptr<Tensor>* tensor) {
+  with_device device_context(device_);
+  ::torch::Tensor zero_tensor;
+  auto torch_data_type = GetTorchDataType(dtype);
+  if (device_ == CPU_DEVICE_ID) {
```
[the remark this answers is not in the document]
kit1980: done

test/test_torch.py
```diff
@@ -1378,5 +1379,122 @@ def forward(self, x):
...
+                ret = hvd.join()
+        else:
+            try:
+                hvd.allgather(tensor)
+                assert False, 'hvd.allgather did not throw error'
+            except (torch.FatalError, RuntimeError):
+                pass
+
+            if torch.cuda.is_available():
+                ret = hvd.join(hvd.local_rank())
```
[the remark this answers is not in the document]
kit1980: Yes, not necessary, removed.

horovod/torch/mpi_ops.cc
```diff
@@ -216,6 +216,11 @@ int DoBroadcastCudaOnCPU(TC* tensor, TC* output, int root_rank, char* name) {
 }
 #endif
 
+int DoJoin(int device) {
+  throw std::runtime_error("Join Op is not supported for PyTorch < 1.0");
+  return 0;
```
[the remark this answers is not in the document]
kit1980: done

tgaddair:
@kit1980 looks like TensorFlow just switched nightly over to 2.0.  I'll make a quick update to our test script and then your tests should pass.

kit1980:
> @kit1980 looks like TensorFlow just switched nightly over to 2.0.  I'll make a quick update to our test script and then your tests should pass.

Thanks, I was trying to understand what's going on - the previous nightly build passed.

horovod/common/common.h
kit1980: Done for TF and MXNet.

kit1980:
All the tests passed.

kit1980:
> Hi @kit1980 I have a question: Does horovod's cache support the join function of pytorch? Because once there is a rank join, according to the cache, the tensor of other rank hits will be put back into tensor_queue_, so that the current horovod is always in an infinite loop state.
> 
> Can you provide a test that exhibits this behavior? I don't understand why this would happen "the tensor of other rank hits will be put back into tensor_queue_". I have test_horovod_join_allreduce for PyTorch that passes OK.
> 
> Hi， @kit1980 I don't have a pytorch test here, but if horovod's cache supports pytorch's join function, then the following situation will occur:
> For ranks without join:
> Train to step 1000: Since tensor1 tensor2 is cached  the previous step, then it will hit
> For the rank of join: there is no tensor1 tensor2
> Then after the cheap mpi of the cache , because of the rank join, so for tensor1 and tensor2, not all ranks are ready. In this case, for the rank without join will put tensor1 tensor2 back Tensor_queue
> That caused the program to loop
> Of course, the assumption that the above situation is that pytorch will take the logic of the cache.
> So my question is:
> 1.pytorch whether to execute the logic of the cache？

Hi @huoliquankai7,
I think I see what you mean now in my tests.
I'm working on fixing this.

tgaddair:
Hey @huoliquankai7, any update on the TensorFlow PR you mentioned?  Now that PyTorch has landed, would be great to get it for TF as well!

tgaddair:
Hey @kit1980, what is your plan for adding Allgather, Broadcast, TensorFlow, and MXNet support for this operation?  No worries if there are no immediate plans, we can add some of this work to the backlog, but just wanted to make sure we don't duplicate our efforts.

kit1980:
> Hey @kit1980, what is your plan for adding Allgather, Broadcast, TensorFlow, and MXNet support for this operation?  No worries if there are no immediate plans, we can add some of this work to the backlog, but just wanted to make sure we don't duplicate our efforts.

Hi,
I've found couple of issues with Join and fusion and caching while testing with MNIST, should be fixed very soon.
After that I'll add support for Allgather and Broadcast, should be easy.
For TensorFlow and MXNet I'd like someone's help, because I don't really use these.

huoliquankai7:
> Hey @huoliquankai7, any update on the TensorFlow PR you mentioned? Now that PyTorch has landed, would be great to get it for TF as well!

Hi， @tgaddair ,Because @kit1980's join and cache have a problem, so wait for him to solve it, and then I sort it out, it should be put forward tensorflow pr soon.

tgaddair:
Hey @kit1980, and update on the cache issue?

kit1980:
> Hey @kit1980, and update on the cache issue?

Hi @tgaddair
I added 1 more special bit to the cache bit vector to decide that communication is needed if one of the ranks did join this tick.
It mostly works, still need to fix couple of issues I found after adding more extensive tests.

tgaddair:
Thanks for the update, @kit1980.  When the PR is ready, we should also get @romerojosh to take a look, as he's the creator of the bit cache.

PR #1092, opened 2019-05-21
don't average the "size" metric of keras

noamwies:
when using tensorboard callback with update_freq != 'epoch'  the current behavior create incorrect plots (especially when you want to compare runs with different number of hosts)


alsrgv:
@noamwies, thanks for the PR.  I understand the idea, but I'm worried that if the callback called `MetricAverageCallback` will start summing things up, people will get confused.

@tgaddair, @abditag2, WDYT?

alsrgv:
Issue #1095 also touches upon batch size.  Perhaps we'll have to make `size` a special element.

@noamwies, while reviewing the code, I noticed that we only do average at the end of the epoch.  If we want to support intra-epoch updates, we should make MetricAverageCallback do average on a batch basis.  Considering that it's just averaging a few tiny elements, we may be able to just do averaging on every batch, but it should be something to test.

tgaddair:
> @noamwies, thanks for the PR. I understand the idea, but I'm worried that if the callback called `MetricAverageCallback` will start summing things up, people will get confused.
> 
> @tgaddair, @abditag2, WDYT?

Seems like the right approach would be to call the averaging with the same cadence as the update frequency, or if it's not too much of a performance penalty (which I doubt it would be) to just do it once per batch as you suggest.

noamwies:
I don't have the resources to run [this benchmarks](https://github.com/horovod/horovod/blob/master/docs/benchmarks.rst).
You have any suggestion  for alternative (smaller) benchmark ?

PR #1093, opened 2019-05-21
Initialize correctly IBM Spectrum MPI provided with IBM PowerAI DDL

nvcastet:
IBM Spectrum MPI is provided with IBM PowerAI DDL. Spectrum MPI needs to be initialized with the proper license. This PR let DDL initialize MPI when using the DDL backend.

### DIFF
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -883,7 +883,13 @@ void BackgroundThreadLoop(HorovodGlobalState& state, MPIContext& ctx) {
                       "likely cause a segmentation fault.";
     }
   } else {
+#if HAVE_DDL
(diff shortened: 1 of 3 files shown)


horovod/common/operations.cc
```diff
@@ -883,7 +883,13 @@ void BackgroundThreadLoop(HorovodGlobalState& state, MPIContext& ctx) {
                       "likely cause a segmentation fault.";
     }
   } else {
+#if HAVE_DDL
+    // DDL comes with IBM Spectrum MPI
+    // and needs to inialize MPI with the proper license.
```
alsrgv: Nit: `initialize`

PR #1088, opened 2019-05-19
Docs migration to Sphinx

IgorWilbert:
The bloated README really needs to be divided, but right now we have links on it that don't work because they point to non `.rst` files. To fix it, I'll update them to the `.rst` syntax.

## Roadmap

- [X] Concepts.md

- [x] Benchmarks.md

- [x] Docker.md

- [x] Gpus.md

- [x] Inference.md

- [x] Running.md

- [x] Spark.md

- [x] Tensor-fusion.md

- [x] Timeline.md

- [x] Troubleshooting.md

PR #1111, opened 2019-05-31
Changes subtitle emphasis 

IgorWilbert:
As mentioned in the comments [here](https://github.com/horovod/horovod/pull/1088), the emphasis of the titles for subsections has to be changed so that they will become links in the `index`.

### DIFF
--- a/docs/benchmarks.rst
+++ b/docs/benchmarks.rst
@@ -43,8 +43,8 @@ To reproduce the benchmarks:
     total images/sec: 1656.82
 
 
-**Real data benchmarks**
-
+Real data benchmarks
+~~~~~~~~~~~~~~~~~~~~
 The benchmark instructions above are for the synthetic data benchmark.
 
 To run the benchmark on a real data, you need to download the `ImageNet dataset <http://image-net.org/download-images>`__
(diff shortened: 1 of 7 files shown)


IgorWilbert:
@alsrgv I think I got them all. Let me know if you want to include other changes in this PR. BTW, the documentation has changed very fast, good job!

alsrgv:
@IgorWilbert, thanks for the PR!  I agree, docs are getting better and better :-)

Can you try subsections (`~~~`) instead of sections (`---`)?  The sections font feels a bit too big for what used to be subsections before.

IgorWilbert:
Hi @alsrgv ! I did the changes you asked, but it looks identical to me (I'm using Firefox). Also, I think some of the titles could be shorter, like `transport/p2p.cu:431 WARN failed to open CUDA IPC handle : 30 unknown error` seems way too long. What do you think?

alsrgv:
I guess what I wanted to do is not possible with titles: Rather than imposing a fixed number and order of section title adornment styles, the order enforced will be the order as encountered. The first style encountered will be an outermost title (like HTML H1), the second style will be a subtitle, the third will be a subsubtitle, and so on.

Let's keep longer FAQ subtitles - they contain important part of the error, it should make them more discoverable with better search engine matches.

PR #1128, opened 2019-06-06
Adding support for multiple CUDA streams for NCCL operations.

romerojosh:
Feature adding support for multiple CUDA streams for NCCL operations. Number of steams is controllable via env var `HOROVOD_NUM_STREAMS`. Using an additional stream of NCCL operations can improve performance in mixed-precision cases where there may be multiple allreduces of different dtypes scheduled in close proximity. 

As a usage note, each stream will allocate its own resources (tensor fusion buffer, NCCL communicator and related resources) which should be considered when increasing the number of streams. Also, as discussed in #1052 and NVIDIA/nccl#208, the number of concurrent NCCL operations should be limited to avoid hangs due to GPU resource constraints.


alsrgv:
@romerojosh, looks pretty much commit-ready, thanks!  Any benchmark data you can share?

romerojosh:
In our internal benchmarking efforts for MLPerf with MXNet, we saw that using two NCCL streams result in a 1-2% end-to-end performance increase when training RN50 at large scale with small per GPU batch sizes. In this small local batch configuration, each iteration is relatively quick and the overheads from the last pair allreduces in a step that are typically not overlapped with much computation are important to consider (recall, MXNet does a mix of  FP32 and FP16 reductions). Using a second NCCL stream enables these final FP32 and FP16 reductions to occur concurrently and reduces the overhead. 

Cases with larger local batch sizes would be less sensitive to this latency, so using a second NCCL stream would provide no benefit in those scenarios.

horovod/common/operations.cc
```diff
@@ -1092,6 +1093,21 @@ void BackgroundThreadLoop(HorovodGlobalState& state, MPIContext& ctx) {
            "allgather and hierarchical allreduce.";
   }
 
+#if HAVE_CUDA
+  // Set number of CUDA streams to use
+  auto horovod_num_streams =
+      std::getenv("HOROVOD_NUM_STREAMS");
```
alsrgv: Could you make a constant in operations.h, and rename to `HOROVOD_NUM_NCCL_STREAMS` (rename variable as well)?

romerojosh: Done. I also changed the names of the new variables in the global state struct from `num_streams` -> `num_nccl_streams` and `current_stream` to `current_nccl_stream`.

alsrgv:
@romerojosh, can you rebase on master? I bet you're being bitten by one of the recent breaks.

Richie-yan:
Hi, @romerojosh 
When I was profiling horovod, a long QUEUE time was displayed on the timeline. My understanding of QUEUE is: the time to wait for the nccl stream to end.
So I used HOROVOD_NUM_NCCL_STREAMS = 4 and found that the final training speed is almost the same as when HOROVOD_NUM_NCCL_STREAMS = 1, but there is almost no QUEUE on the timeline, so I am confused about this phenomenon.
I compared timeline information using one NCCL stream and multiple NCCL streams:
Although multiple NCCL streams are opened, there is almost no QUEUE time on the timeline, but the NCCL_ALLREDUCE time has increased relatively, which is almost equal to the QUEUE time + NCCL_ALLREDUCE time for opening one NCCL stream.
When there are many QUEUE times, adding NCCL streams and improving the overlap of computing and communication has not improved the performance?

romerojosh:
@Richie-yan Increasing the number of NCCL streams allows multiple NCCL allreduces to run concurrently, which does remove the queue time. However, if the individual NCCL allreduces are already saturating network resources, running multiple in parallel will not improve performance, as each would be splitting the available resources. Multiple NCCL streams are useful in cases where the framework may schedule some small allreduces that don't saturate resources along with larger ones  (for example, in MXNet mixed precision training where there are a few FP32 allreduces mixed in with mostly FP16 ones). 

It sounds like you may see better performance in your case instead by either increasing the `HOROVOD_FUSION_THRESHOLD` and/or decreasing `HOROVOD_CYCLE_TIME`.

NanYan1119:
Hi, I would like to known why does the use of multiple streams improve performance in the case of mixed precision of fp16 and fp32? If only have the fp32 allreduce, multiple CUDA stream for NCCL can improve performance?

firejq:
@romerojosh Hi, in my case, I use multiple NCCL streams and it does make NCCLAllreduces run concurrently. 
However, the wall duration of each NCCLAllreduce increased several times, which make me puzzled. Could you please give any suggestions?

PR #1130, opened 2019-06-07
Add grouped allreduce feature

romerojosh:
This is a preliminary PR for a new feature that enables explicit assignment of allreduce operations into manually defined groups. This enables users to exert more direct control over how/when allreduces are scheduled within Horovod.

Current implementation is for TensorFlow only and exposed via additional `num_groups` argument to `DistributedOptimizer` wrapper. Right now, the wrapper will evenly split and assign tensors to requested number of groups. Grouping is enabled via new env var `HOROVOD_GROUPED_ALLREDUCES`. 

A few benefits of this feature:
1. Ability to disable all overlap of allreduces (`num_groups = 1`) and computation which can be useful.
2. Explicit group assignment can be used to impose minimum fused allreduce size. This enables users to reduce `HOROVOD_CYCLE_TIME` for reduced latency *without* the side effect of generating many small allreduces as is the case with the fully dynamic strategy.
3. (not implemented yet) Can be used to enforce deterministic allreduces. Fully dynamic strategy results in non-deterministic allreduce results due to dynamic fusion buffer packing. If multiple groups are prohibited from fusing, and the fusion order is explicitly defined within groups, the allreduces will be deterministic.

### DIFF
--- a/horovod/tensorflow/__init__.py
+++ b/horovod/tensorflow/__init__.py
@@ -231,7 +231,7 @@ if _SessionRunHook is not None and _get_default_graph is not None:
 
 @_cache
 def _make_allreduce_grads_fn(name, device_dense, device_sparse,
-                             compression, sparse_as_dense, op, num_groups):
+                             compression, sparse_as_dense, op, num_groups=0):
     def allreduce_grads(grads):
         with tf.name_scope(name + "_Allreduce"):
             if sparse_as_dense:
--- a/horovod/tensorflow/__init__.py
+++ b/horovod/tensorflow/__init__.py
@@ -34,7 +34,7 @@ from horovod.tensorflow.mpi_ops import nccl_built, ddl_built, mlsl_built
 from horovod.tensorflow.mpi_ops import Average, Sum, Adasum
 from horovod.tensorflow.mpi_ops import _check_has_gpu
 from horovod.tensorflow.mpi_ops import handle_average_backwards_compatibility, check_num_rank_power_of_2
-
+from horovod.tensorflow.mpi_ops import register_group
 from horovod.tensorflow.util import _executing_eagerly, _make_subgraph, _cache
 
 import tensorflow as tf
@@ -42,7 +42,7 @@ import warnings
 
 has_gpu = gpu_available('tensorflow')
 def allreduce(tensor, average=None, device_dense='', device_sparse='',
-              compression=Compression.none, op=None):
+              compression=Compression.none, op=None, group_id=-1):
     """Perform an allreduce on a tf.Tensor or tf.IndexedSlices.
 
     This function performs a bandwidth-optimal ring allreduce on the input
@@ -63,6 +63,7 @@ def allreduce(tensor, average=None, device_dense='', device_sparse='',
                      using compression.
         op: The reduction operation to combine tensors across different ranks.
             Defaults to Average if None is given.
+        group_id: Optional integer id for explicitly grouping allreduce ops.
 
     Returns:
         A tensor of the same shape and type as `tensor`, summed across all
@@ -94,6 +95,7 @@ def allreduce(tensor, average=None, device_dense='', device_sparse='',
             horovod_size = tf.cast(size(), dtype=tensor.dtype)
             tensor_compressed, ctx = compression.compress(tensor)
             summed_tensor_compressed = _allreduce(tensor_compressed, op=true_op)
+            summed_tensor_compressed = _allreduce(tensor_compressed, op=true_op, group_id)
             summed_tensor = compression.decompress(summed_tensor_compressed, ctx)
             if op == Adasum:
                 if ('CPU' not in tensor.device and has_gpu):
@@ -229,7 +231,7 @@ if _SessionRunHook is not None and _get_default_graph is not None:
 
 @_cache
 def _make_allreduce_grads_fn(name, device_dense, device_sparse,
-                             compression, sparse_as_dense, op):
+                             compression, sparse_as_dense, op, num_groups):
     def allreduce_grads(grads):
         with tf.name_scope(name + "_Allreduce"):
             if sparse_as_dense:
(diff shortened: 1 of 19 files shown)


alsrgv:
cc @duncanriach

alsrgv:
@romerojosh, did you have better performance with `HOROVOD_GROUPED_ALLREDUCES=1` or `>1`?

romerojosh:
I think generally it would be case dependent. In once case, we achieved good results at very large scale with `num_groups = 1` combined with a small `HOROVOD_FUSION_THRESHOLD` set to 4 MiB which still broke up the allreduces into smaller chunks. In a different case, we achieved good results by setting num_groups to a small fixed value (16) and turning down `HOROVOD_CYCLE_TIME` to a very small value (0.2). By the way, `HOROVOD_GROUPED_ALLREDUCES` is just a switch to enable or disable the feature.

If you recall, much of the performance benefit from the original grouping implementation (i.e. use of the bitvector for readiness) was factored out into the more general #902. I think the main benefit of this PR is not necessarily a big performance boost, but that it exposes a mechanism for greater control over the way Horovod performs allreduces. This _could_ present opportunities for higher performance than the default if the user is knowledgeable about their network/system and tunes for that.

horovod/tensorflow/mpi_ops.py
```diff
@@ -76,7 +77,10 @@ def _allreduce(tensor, name=None):
     """
     if name is None and not _executing_eagerly():
         name = 'HorovodAllreduce_%s' % _normalize_name(tensor.name)
-    return MPI_LIB.horovod_allreduce(tensor, name=name)
+    if (group_id is not None):
```
tgaddair: Nit: remove parens

horovod/tensorflow/__init__.py
```diff
@@ -75,7 +77,10 @@ def allreduce(tensor, average=True, device_dense='', device_sparse='',
         with tf.device(device_dense):
             horovod_size = tf.cast(size(), dtype=tensor.dtype)
             tensor_compressed, ctx = compression.compress(tensor)
-            summed_tensor_compressed = _allreduce(tensor_compressed)
+            if (group_id is not None):
```
tgaddair: What if we just set the default to -1?  Then we could simplify these four lines to:

```
summed_tensor_compressed = _allreduce(tensor_compressed, group_id)
```

Personally, I think that would make the code simpler to parse (fewer special cases to consider).

horovod/tensorflow/__init__.py
```diff
@@ -144,7 +149,7 @@ class DistributedOptimizer(tf.train.Optimizer):
 
     def __init__(self, optimizer, name=None, use_locking=False, device_dense='',
                  device_sparse='', compression=Compression.none,
-                 sparse_as_dense=False):
+                 sparse_as_dense=False, num_groups=None):
```
tgaddair: Can num_groups default 0?

horovod/tensorflow/__init__.py
```diff
@@ -191,12 +200,41 @@ def allreduce_grads(grads):
...
+                    num_grads_per_group = (len(grads) + self._num_groups - 1) // self._num_groups
+                    group_ids = [register_group(num_grads_per_group,
+                                                "%s:%s:%d" % (grads[0].name, grads[-1].name, i))
+                                 for i in range(len(grads) // num_grads_per_group)]
+                    if len(grads) % num_grads_per_group != 0:
+                        group_ids.append(register_group(len(grads) % num_grads_per_group,
+                                                        "%s:%s:%d" % (grads[0].name, grads[-1].name,
+                                                        len(grads) // num_grads_per_group + 1)))
+
+                    return [allreduce(grad,
```
tgaddair: Perhaps we can simplify this to remove the else:

```
return [allreduce(grad,
                             device_dense=self._device_dense,
                             device_sparse=self._device_sparse,
                             compression=self._compression,
                             group_id=group_ids[idx // num_grads_per_group] if group_ids else -1)
            if grad is not None else grad
            for idx, grad in enumerate(grads)]
```

Then we declare `group_ids = None` above outside the if statement.

horovod/tensorflow/mpi_ops.py
romerojosh: Removed parens and updated with new default value. I left the if/else statment here to ensure the default `group_id` value is set properly in the C++ backend (as defined by `NULL_GROUP_ID` in `message.h`)

horovod/tensorflow/__init__.py
romerojosh: Sounds good to me. Changed default to -1 and removed extra cases.

horovod/tensorflow/__init__.py
romerojosh: Yep, should be fine. Done.

horovod/tensorflow/__init__.py
romerojosh: Thanks for the suggestion! Implemented.

eric-haibin-lin:
will this support mxnet?

romerojosh:
@eric-haibin-lin I am planning to revise this PR to be more framework agnostic, with the group handling being taken care of completely in the C++ backend. The goal is to have the feature support all the frameworks with little to no need for user script modification (or modification of existing Horovod APIs) for greater ease of use.

duncanriach:
FYI, @romerojosh and I had a one-on-one and discussed this current PR. I'm now fully, and finally, up-to-speed on it. :-)

duncanriach:
No specific update, @bryandeng. I haven't yet got to working on this myself much, but there have been resourcing discussions. It's on my radar. Sorry for making you wait.

chandana1332:
> No specific update, @bryandeng. I haven't yet got to working on this myself much, but there have been resourcing discussions. It's on my radar. Sorry for making you wait.

Hello,

Just wanted to pop-in here and voice my team's needs for this feature as well. We need our training runs to be deterministic and are currently noticing significant variance in model performance. We use MXNet with Horovod.

romerojosh:
Hi @chandana1332, glad you will find this useful and apologies that it has not been updated in quite a while, but given the recent comments by you and others, I will try prioritizing it.

In the meantime, you can get deterministic behavior from Horovod in MXNet (and other frameworks as well) by setting `HOROVOD_FUSION_THRESHOLD=0`. If this variable is set and you are still getting non-deterministic results, ensure that the other framework operations you are using are deterministic.

EiffL:
Hi @romerojosh, thanks for initiating this PR! I'm looking into using horovod as the backend for collective operations in  https://github.com/tensorflow/mesh, and grouped communications are necessary for doing model parallelism over multidimensional GPU meshes.
I took a look at the code, and resolved some conflicts with respect to the current state of master on my fork over here: https://github.com/DifferentiableUniverseInitiative/horovod/tree/grouping 
I'm not sure my merge is working properly though, or maybe it's just that I haven't used the API properly, but it looks like when the grouping mechanism is enabled, the computation gets stuck in the allreduce, waiting for the group to be complete:
https://github.com/romerojosh/horovod/blob/cf8d91d5c78d073a1499dc88ab38f65480e88657/horovod/common/controller.cc#L107
but the group never seems to fill-up :-/ . Maybe I'm not using the API correctly, or maybe there have been other changes in the controller mechanism since your PR that are messing with the logic. 

Do you have a minimal working example that I can use to test if things are working properly?  

And let me know if you would be interested in me opening a PR to add these updates to your `romerojosh:grouping_pr`? I've also started looking  separately on adding group_id to the other collective ops (all2all, allgather, etc.).

EiffL:
And as I write this comment, it suddently strikes me that by grouped allreduce, you probably didn't mean in terms of GPU groups/ different communicators, but grouping communications so that they occur at the same time ^^'. I'm confused now.
So, here is an additional question, is it possible to use/adapt this grouping mechanism to perform collective operations only over a subset of GPUs?

romerojosh:
Hi @EiffL, thanks for taking a look at this. As you mention in your second comment, yes, this grouping is referring to forcing Horovod to perform a specific set of allreduce communications at the same time (with tensor fusion). I actually have a different version of this PR locally that changes the API up a bit to make it more obvious that this is the case, via introducing an operation like `hvd.grouped_allreduce([tensor1, tensor2])`, without the need to manually register and add `group_id` arguments to calls, which is a bit of a hassle. It's been on my TODO list for a while to post the PR and will post soon.

For your additional question, I'm not sure that this particular grouping mechanism in this PR (or in the updated version I will post) can be adapted easily to control multiple subcommunicators. We do have an issue open (#2139) about this exact topic and I would be happy to discuss some ideas there if you are interested.

EiffL:
Great, thanks for the confirmation of what this PR is doing. So at any rate, I've updated this code to the current master, happy to make a PR to your PR if useful :-)
And many thanks for the pointer to this this issue, that's exactly what I need, and I'll comment more over there.

romerojosh:
Superseded by #2453

PR #1132, opened 2019-06-10
Replace .step(synchronize=False) with optimizer.skip_synchronize()

alsrgv:
NVIDIA AMP does not support passing additional flags to `optimizer.step()`, such as `optimizer.step(synchronize=False)`.

This PR switches API to use context manager:
```python
optimizer.synchronize()
with optimizer.skip_synchronize():
    optimizer.step()
```

### DIFF
--- a/horovod/torch/__init__.py
+++ b/horovod/torch/__init__.py
@@ -160,7 +160,7 @@ class _DistributedOptimizer(torch.optim.Optimizer):
         self._synchronized = True
 
     @contextmanager
-    def already_synchronized(self):
+    def skip_synchronize(self):
         """
         A context manager used to specify that optimizer.step() should
         not perform synchronization.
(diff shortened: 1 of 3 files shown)


horovod/torch/__init__.py
```diff
@@ -157,13 +159,31 @@ def synchronize(self):
...
 
-    def step(self, closure=None, synchronize=True):
-        if synchronize:
+    @contextmanager
+    def already_synchronized(self):
+        """
+        A context manager used to specify that optimizer.step() should
+        not perform synchronization.
+
+        It's typically used in a following pattern:
```
abditag2: nit: It's typically used in a following pattern: >> It's typically used in the following pattern:

abditag2:
New interface LGTM. Shall we keep the old interface for backward compatibility and raise a warning and remove it in a later release?

alsrgv:
@abditag2, I'm debating that, since the API was released a couple of days ago and it does not work for the only user who asked for it. Seems less confusing to just hard-migrate to a new API at this point.

abditag2:
Sounds good then!

PR #1155, opened 2019-06-18
fix imagenet example with synthetic input

apeforest:
Fix the example of imagenet when users runs with synthetic data. Before this fix, it would fail with error 'batch_fn' not defined.

### DIFF
--- a/examples/mxnet_imagenet_resnet50.py
+++ b/examples/mxnet_imagenet_resnet50.py
@@ -137,8 +137,6 @@ else:
     raise ValueError('Invalid lr mode')
 
 
-
-
 # Function for reading data from record file
 # For more details about data loading in MXNet, please refer to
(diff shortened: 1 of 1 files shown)


apeforest:
@yuxihu @eric-haibin-lin please help review. thanks

examples/mxnet_imagenet_resnet50.py
```diff
@@ -136,6 +136,9 @@
 else:
     raise ValueError('Invalid lr mode')
 
+
```
alsrgv: Less whitespace?

yuxihu: +1 for less empty lines

apeforest: removed. only 2 blank lines now to satisfy pep8

PR #1156, opened 2019-06-19
MLSL: move mlsl_init before mpi_init, add mlsl_finalize call

mshiryaev:
Signed-off-by: Mikhail Shiryaev <mikhail.shiryaev@intel.com>


mshiryaev:
@alsrgv - Intel MPI worked in fallback mode i.e. functionally correct but without performant configuration.

PR #1158, opened 2019-06-19
Restructure Horovod doc landing page

sblotner:
- Add Get Started accordion to top of the page
- Use Helvetica font
- Add basic pages for Keras, MXNet, PyTorch, and TensorFlow
- Rename some pages and restructure page order for better left navigation flow
- Shorten the README a bit


sblotner:
Ok, finally fixed those merge conflicts! 

Here's a preview of the new landing page: 
Screen Shot 2019-06-19 at 1 58 49 PM

docs/index.rst
```diff
@@ -2,26 +2,106 @@ Horovod documentation
...
+         <ol>
+            <li><a href="https://www.open-mpi.org/faq/?category=building#easy-build">Install Open MPI 3.1.2 or 4.0.0</a>, or another MPI implementation. </li>
+            <li>Install the Horovod pip package: <code>pip install horovod</code></li>
+            <li>Read <a href="https://horovod.readthedocs.io/en/latest/tensorflow.html">Horovod with TensorFlow</a> for best practices and examples. </li>
+         </ol>
+         Or, use <a href="https://horovod.readthedocs.io/en/latest/gpus_include.html">Horovod on GPUs</a>, in <a href="https://horovod.readthedocs.io/en/latest/spark_include.html">Spark</a>, <a href="https://horovod.readthedocs.io/en/latest/docker_include.html">Docker</a>, <a href="https://github.com/sylabs/examples/tree/master/machinelearning/horovod">Singularity</a>, or Kubernetes (<a href="https://github.com/kubeflow/kubeflow/tree/master/kubeflow/mpi-job">Kubeflow</a>, <a href="https://github.com/kubeflow/mpi-operator/">MPI Operator</a>, <a href="https://github.com/helm/charts/tree/master/stable/horovod">Helm Chart</a>, and <a href="https://github.com/IBM/FfDL/tree/master/etc/examples/horovod/">FfDL</a>).
```
alsrgv: Is it possible to use include to reduce duplication, like this one?

sblotner: Using .. include:: would be ideal, but it will get ignored because .. raw:: will parse everything as HTML

I think I could do it with JavaScript, but would have to put the shared content in a separate file and reference it. Would you prefer to make edits in 1 place in a separate .js file, or make them 4+ times in the same .rst file? 

Another option is to change it from an accordion to buttons for each, and include the install steps in the individual Keras/PyTorch .rst files and use include to share the steps there

alsrgv: I see, maybe let's keep it as-is for now.  Separate JavaScript file sounds a bit too magical :-)

sblotner: Sounds good :) You have good tech writing instincts!

PR #1163, opened 2019-06-21
Update "Horovod with Keras" documentation page

IgorWilbert:
Hello! I was talking to @sblotner about the next steps for the documentation, and she mentioned some pages were missing content. In this one, the idea is to give tips and describe special situations that can be useful when using `Horovod` with `Keras`. I still don't have much experience with the project, so for now I just pasted the description of `Horovod with Tensorflow` together with an example using `Keras`. If @alsrgv or someone else could give me high-level inputs of what is important to write I can improve the text accordingly. Thanks!

### DIFF
--- a/docs/keras.rst
+++ b/docs/keras.rst
@@ -18,16 +18,17 @@ Horovod supports Keras and regular TensorFlow in similar ways. To use Horovod wi
 
    The distributed optimizer delegates gradient computation to the original optimizer, averages gradients using *allreduce* or *allgather*, and then applies those averaged gradients.
 
-5. Add ``hvd.BroadcastGlobalVariablesHook(0)`` to broadcast initial variable states from rank 0 to all other processes.
+5. Add ``hvd.callbacks.BroadcastGlobalVariablesCallback(0)`` to broadcast initial variable states from rank 0 to all other processes.
 
    This is necessary to ensure consistent initialization of all workers when training is started with random weights or restored from a checkpoint.
-   Alternatively, if you're not using ``MonitoredTrainingSession``, you can execute the ``hvd.broadcast_global_variables`` op after global variables have been initialized.
 
 6. Modify your code to save checkpoints only on worker 0 to prevent other workers from corrupting them.
 
-   Accomplish this by passing ``checkpoint_dir=None`` to ``tf.train.MonitoredTrainingSession`` if ``hvd.rank() != 0``.
+   Accomplish this by guarding model checkpointing code with ``hvd.rank() != 0``.
 
-.. NOTE:: Keras 2.0.9 has a `known issue <https://github.com/fchollet/keras/issues/8353>`_ that makes each worker allocate all GPUs on the server, instead of the GPU assigned by the *local rank*. If you have multiple GPUs per server, upgrade to Keras 2.1.2 or downgrade to Keras 2.0.8.
+.. NOTE:: - Keras 2.0.9 has a `known issue <https://github.com/fchollet/keras/issues/8353>`_ that makes each worker allocate all GPUs on the server, instead of the GPU assigned by the *local rank*. If you have multiple GPUs per server, upgrade to Keras 2.1.2 or downgrade to Keras 2.0.8.
+
+          - To use ``tensorflow`` with ``keras`` you must use ``from tensorflow import keras`` instead of ``import keras`` and ``import horovod.tensorflow.keras as hvd`` instead of ``import horovod.keras as hvd`` in the import statements.
 
 See full training `simple <https://github.com/horovod/horovod/blob/master/examples/keras_mnist.py>`_ (shown below) and `advanced <https://github.com/horovod/horovod/blob/master/examples/keras_mnist_advanced.py>`_ examples.
 
--- a/docs/keras.rst
+++ b/docs/keras.rst
@@ -1,7 +1,131 @@
 Horovod with Keras
 ==================
-Horovod supports Keras and regular TensorFlow in similar ways.
+Horovod supports Keras and regular TensorFlow in similar ways. To use Horovod with Keras, make the following additions to your program. 
 
-See full training `simple <https://github.com/horovod/horovod/blob/master/examples/keras_mnist.py>`_ and `advanced <https://github.com/horovod/horovod/blob/master/examples/keras_mnist_advanced.py>`_ examples.
+1. Run ``hvd.init()``.
+
+2. Pin a server GPU to be used by this process using ``config.gpu_options.visible_device_list``.
+
+   With the typical setup of one GPU per process, you can set this to *local rank*. In that case, the first process on
+   the server will be allocated the first GPU, the second process will be allocated the second GPU, and so forth.
+
+3. Scale the learning rate by the number of workers.
+
+   Effective batch size in synchronous distributed training is scaled by the number of workers.
+   An increase in learning rate compensates for the increased batch size.
+
+4. Wrap the optimizer in ``hvd.DistributedOptimizer``.
+
+   The distributed optimizer delegates gradient computation to the original optimizer, averages gradients using *allreduce* or *allgather*, and then applies those averaged gradients.
+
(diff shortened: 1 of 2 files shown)


docs/keras.rst
```diff
@@ -1,7 +1,131 @@
...
+3. Scale the learning rate by the number of workers.
+
+   Effective batch size in synchronous distributed training is scaled by the number of workers.
+   An increase in learning rate compensates for the increased batch size.
+
+4. Wrap the optimizer in ``hvd.DistributedOptimizer``.
+
+   The distributed optimizer delegates gradient computation to the original optimizer, averages gradients using *allreduce* or *allgather*, and then applies those averaged gradients.
+
+5. Add ``hvd.BroadcastGlobalVariablesHook(0)`` to broadcast initial variable states from rank 0 to all other processes.
```
alsrgv: Replace with `hvd.callbacks.BroadcastGlobalVariablesCallback`

docs/keras.rst
```diff
@@ -1,7 +1,131 @@
...
+
+   This is necessary to ensure consistent initialization of all workers when training is started with random weights or restored from a checkpoint.
+   Alternatively, if you're not using ``MonitoredTrainingSession``, you can execute the ``hvd.broadcast_global_variables`` op after global variables have been initialized.
```
alsrgv: Remove the "alternatively..." part.

docs/keras.rst
```diff
@@ -1,7 +1,131 @@
...
+
+6. Modify your code to save checkpoints only on worker 0 to prevent other workers from corrupting them.
+
+   Accomplish this by passing ``checkpoint_dir=None`` to ``tf.train.MonitoredTrainingSession`` if ``hvd.rank() != 0``.
```
alsrgv: Accomplish this by guarding model checkpointing code with `hvd.rank() != 0`.

IgorWilbert:
@alsrgv done! I did not fix DCO before because I thought there would be many changes. If there is anything else you would like to change, let me know!

docs/keras.rst
```diff
@@ -1,7 +1,130 @@
...
+    from __future__ import print_function
+    import keras
+    from keras.datasets import mnist
+    from keras.models import Sequential
+    from keras.layers import Dense, Dropout, Flatten
+    from keras.layers import Conv2D, MaxPooling2D
+    from keras import backend as K
+    import math
+    import tensorflow as tf
+    import horovod.keras as hvd
```
alsrgv: Maybe let's also document that to use TF-Keras, user should do `from tensorflow import keras` instead of `import keras` and `import horovod.tensorflow.keras as hvd` instead of `import horovod.keras as hvd`?

Or should we make a separate section for this?

IgorWilbert: Good idea, since this would make the imports more clear. I don't think it is necessary to make a separate section for something so simple, so if you agree I'll change this example. :)

alsrgv: What are you thinking w.r.t. change, add TF-Keras as an alternative commented out?  If so, I think we should also add **NOTE** before/after to draw additional attention to those changes to be made if switching to TF-Keras.

IgorWilbert:
@alsrgv sorry for the late reply, things were a bit messy here, but now everything is fine. 

- I decided to add a bullet point instead of making a new **NOTE** because it seemed unnecessarily large (image below). Let me know if you want me to change it.

example

- I did not put an additional example exclusively for `TF-Keras` because I think the existing example is pretty large. So, I guess we could take three different paths: making the examples smaller and having two of them in `keras.rst`; having only one example and explaining all the things that should be changed for `TF-Keras` in `keras.rst` (similar to what it is now); having a separate page for `TF-Keras` with its own example and explanation; Let me know which one you prefer.
- Paragraph breaking looks pretty weird (image below). Same thing can be seen in `Horovod with Tensorflow` page. I believe there should be spacing only from one numbered paragraph to the other, not in the middle of the same paragraph. Let me know if this should be fixed.

weird

docs/keras.rst
```diff
@@ -1,7 +1,132 @@
...
+
+   This is necessary to ensure consistent initialization of all workers when training is started with random weights or restored from a checkpoint.
+
+6. Modify your code to save checkpoints only on worker 0 to prevent other workers from corrupting them.
+
+   Accomplish this by guarding model checkpointing code with ``hvd.rank() != 0``.
+
+.. NOTE:: - Keras 2.0.9 has a `known issue <https://github.com/fchollet/keras/issues/8353>`_ that makes each worker allocate all GPUs on the server, instead of the GPU assigned by the *local rank*. If you have multiple GPUs per server, upgrade to Keras 2.1.2 or downgrade to Keras 2.0.8.
+
+          - To use ``tensorflow`` with ``keras`` you must use ``from tensorflow import keras`` instead of ``import keras`` and ``import horovod.tensorflow.keras as hvd`` instead of ``import horovod.keras as hvd`` in the import statements.
```
alsrgv: Looks good, I'd rephrased "use ``tensorflow`` with ``keras``" as "use Keras bundled with TensorFlow"

sblotner:
> I think a bullet point is fine. Line breaks are indeed un-appealing. Can you take a look at how to fix them?

You may be able to force a line break with: 

`
|
`

IgorWilbert:
@alsrgv I changed the phrase you asked and fixed the line breaks on both pages. `|` worked, thank you @sblotner! Right now it looks like this. If you don't want any other changes, please let me know so that I will **squash** the commits before merging. Thanks! 

fixed

alsrgv:
The paragraph breaks are super-huge, is it possible to reduce them by 2x?

IgorWilbert:
@alsrgv I noticed this too. Unfortunately, I already used only one `|`, so from what I looked into it this is the minimum space you can insert. The space above it is smaller because it is not really a space inserted, but actually a hierarchy made by Sphinx. A better alternative might be not using a new line at all (image below). If you did not like any of these, I can search more to find a different way of inserting space.
other

alsrgv:
I wonder if we could just add a line break (not a new paragraph) after the first sentence of each bullet and make the rest italic / smaller font / something to highlight that it's an explanation of the key idea?

IgorWilbert:
It would look great, the problem is actually this "add a line break (not a new paragraph) after the first sentence of each bullet". If I do just that, it would have the same problem I noticed at first (below), because don't know how to add a space just as small between the numbered items. To avoid this, I need to find a way how to insert a space that is smaller than a `|`, but is not related to the hierarchy.
issue

alsrgv:
@sblotner, do you have any idea how to insert `<br/>` in RST?

sblotner:
> @sblotner, do you have any idea how to insert `<br/>` in RST?

This is kind of clunky, but you could try to use the .. raw:: directive like this: 

`.. raw:: html

   <br>`

(Another idea is to update the custom CSS file I added https://github.com/horovod/horovod/blob/master/docs/_static/custom.css to override the paragraph spacing behavior of Alabaster theme https://github.com/bitprophet/alabaster/blob/master/alabaster/static/alabaster.css_t)

IgorWilbert:
Thanks @alsrgv and @sblotner ! Using
```
.. raw:: html

<br>
```
the space reduces by about 1/3, but I think it is still a bit too large. 
changes
An empty `<p/>` tag achieves better results
c2
So, would you like me to use `<p/>`  or look for how to do it with custom CSS?

sblotner:
@IgorWilbert , I think the <p> looks pretty good

alsrgv:
`<p/>` looks great!

Can we still have another space break between step's "header" like `Pin a server GPU to be used by this process using ``config.gpu_options.visible_device_list``.` and step's "description"?

IgorWilbert:
Will do!

IgorWilbert:
@alsrgv sorry for the late reply, it actually took much less time than I thought. Right now it looks like below. Would you like to change anything else? If not, I'll rebase it and then we can merge it. Thanks! 
maybe

alsrgv:
Can you factor these into https://horovod.readthedocs.io/en/latest/tensorflow.html as well?

IgorWilbert:
I'm sorry @alsrgv, I should have checked this. Of course, I will send a PR to fix it ASAP, probably later today. Thank you!

Issue #1176, opened 2019-06-27
Add `horovod.run()` to make horovod notebook friendly


mengxr:
**Is your feature request related to a problem? Please describe.**

`horovod.spark.run` makes it easy for users to run Horovod on Spark. The interface that takes a train method is quite convenient and fits notebook users well:) However, it only supports Spark. Sometimes, users might want to run distributed training on a multi-GPU node inside a notebook environment, or users might want to trigger an arbitrary job with worker IPs. Right now, they still need to create a Python script and triggered via `horovodrun` command-line.

**Describe the solution you'd like**

I want to have a Python method works similarly to `horovodrun` but takes a python method instead. It will pickle the method to disk and trigger the MPI job using `horovodrun`.

~~~python
import horovod
import horovod.tensorflow as hvd

def train_hvd():
  hvd.init()
  # training code

horovod.run(train_hvd, args, kwargs, num_proc, hosts, ...)
~~~

Users should be able to quickly modify `train_hvd` method and re-run `horoovd.run`. The simplification is that users do not need to manually save the code to disk and call `horovodrun`

@WeichenXu123 will submit a prototype.

cc: @alsrgv

alsrgv:
LGTM, looking forward to the prototype!

denfromufa:
I like this approach, we were just thinking about this last week.

tgaddair:
Fixed by #1192.

PR #1192, opened 2019-07-03
Add `horovod.run.run_func` to make horovod notebook friendly

WeichenXu123:
## Proposed API:

Exported `run_fun` method in package `horovod.run`

```
def run_func(
        fn,
        num_proc,
        host=None,
        ssh_port=None,
        disable_cache=False,
        start_timeout=None,
        env=None,
        verbose=False):
    """
    Run horovod inside python program.
    Run `num_proc` processes executing `fn` on specified hosts.

    :param fn: Function to run.
    :param num_proc: Number of Horovod processes.
    :param host: To specify the list of host names as well as the
                 number of available slots on each host for
                 training processes using the following format:
                 <hostname>:<number of slots>,... .
                 E.g., host1:2,host2:4,host3:1 indicates that 2 processes can run on
                 host1, 4 processes on host2, and 1 process on host3.
                 If None, launch all mpi processes on local node.
    :param ssh_port: SSH port on all the hosts.
                     If None, use default SSH port.
    :param disable_cache: If the flag is not set, horovodrun will perform
                          the initialization checks only once every 60
                          minutes -- if the checks successfully pass.
                          Otherwise, all the checks will run every time
                          horovodrun is called."
    :param start_timeout: Horovodrun has to perform all the checks and
                          start the processes before the specified
                          timeout. The default value is 30 seconds.
                          Alternatively, The environment variable
                          HOROVOD_START_TIMEOUT can also be used to
                          specify the initialization timeout.
    :param env: Environment dictionary to use in Horovod run.
                If None, use `os.environ`.
    :param verbose: If this flag is set, extra messages will be printed.

    :return: A list contains returned result from each MPI process. The ith value of the list
             corresponds to the returned result of rank i process.
    """
```

## Example:

```
import horovod

def hvd_train1():
    # define your training function, with no args.
    # train code
    ...
    if hvd.rank == 0:
        return result

# result_list will be [result_from_rank_0_process, None]
result_list = horovod.run.run_func(hvd_train1, num_proc=2, host="10.95.255.92:1,10.95.232.120:1")

```

## Test

To be added. This is a version for first review.

### DIFF
--- a/horovod/run/run_func.py
+++ b/horovod/run/run_func.py
@@ -115,7 +115,7 @@ def run_func(
         rank=${{OMPI_COMM_WORLD_RANK:-0}}
         {exec} -c "import cloudpickle; cloudpickle.load(open('{fn_file}', 'rb'))(rank=$rank)"
         if [[ "$rank" -eq 0 ]]; then
-        scp -q {scp_port_opt} -o StrictHostKeyChecking=no \
+        scp -v -q {scp_port_opt} -o StrictHostKeyChecking=no \
         {wdir}/{local_result} {local_ip}:{wdir}/{result}
         fi
         """.format(wdir=wdir,
--- a/horovod/run/run_func.py
+++ b/horovod/run/run_func.py
@@ -19,7 +19,7 @@ import tempfile
 import os
 import sys
 import six
-import subprocess
+import shutil
 import textwrap
 
 from horovod.run.common.util import safe_shell_exec
@@ -46,7 +46,10 @@ def run_func(
     Run horovod inside python program.
     Run `num_proc` processes executing `fn` on specified hosts.
 
-    :param fn: Function to run.
+    :param fn: Function to run. The function take no arguments and its return value will be
+               collected and return by `run_func`.
+               We can use a lambda function to wrap functions with arguments, such as:
+               `run_func(fn=lambda: fn1(arg1, arg2, ..), num_proc=2, ...)`
     :param num_proc: Number of Horovod processes.
     :param host: To specify the list of host names as well as the
                  number of available slots on each host for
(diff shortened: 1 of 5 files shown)


mengxr:
This PR addresses #1176 .

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+import six
+import subprocess
+import collections
+import textwrap
+import time
+
+from horovod.run.util import network
+from horovod.run.common.util import safe_shell_exec
+
+_TAIL_LINES_TO_KEEP = 100
```
mengxr: User can see all log output already, no need to keep.

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+_PICKLED_PROC_FN_FILENAME = "proc_fn.pkl"
+_PROC_LAUNCHER_FILENAME = "launch.sh"
+
+
+def run_func(
+        fn,
+        args=None,
+        kwargs=None,
+        num_proc=None,
+        host=None,
```
mengxr: `host` or `hosts`?

mengxr: `host` or `hosts`?

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+        ssh_port=None,
+        disable_cache=False,
+        start_timeout=None,
+        env=None,
+        verbose=False,
+        driver_host=None):
```
mengxr: this is only relevant for spark.

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+
+    :param fn: Function to run.
+    :param args: Arguments to pass to `fn`.
+    :param kwargs: Keyword arguments to pass to `fn`.
+    :param num_proc: Number of Horovod processes. Default to be number of all host slots.
+    :param host: To specify the list of host names as well as the
+                 number of available slots on each host for
+                 training processes using the following format:
+                 <hostname>:<number of slots>,... .
+                 E.g., host1:2,host2:4,host3:1 indicates that 2 processes can run on
```
mengxr: We might need to use sequential mapper to deterministically tell which host runs rank 0.

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+                 host1, 4 processes on host2, and 1 process on host3.
+    :param ssh_port: SSH port on all the hosts.
```
mengxr: should document what the default value `None` means

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+    :param disable_cache: The same argument with horovodrun command.
```
mengxr: Explain what the param does first.

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+    :param start_timeout: Horovodrun has to perform all the checks and
+                          start the processes before the specified
+                          timeout. The default value is 30 seconds.
+                          Alternatively, The environment variable
+                          HOROVOD_START_TIMEOUT can also be used to
+                          specify the initialization timeout.
+    :param env: Environment dictionary to use in Horovod run.  Defaults to `os.environ`.
+    :param verbose: If this flag is set, extra messages will be printed.
+    :param driver_host: The host of node `run_func` running on.
+    :return: The returned value of rank 0 process.
```
mengxr: @alsrgv Do you prefer collecting all results or only from rank #0?

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+
+    wdir = tempfile.mkdtemp()
+    wdir_par = os.path.dirname(wdir)
+    if verbose:
+        print("mpirun working dir is " + wdir)
```
mengxr: `mpirun` -> `horovod.run_func`

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def wrapped_proc_fn(rank=0):
+        return_value = fn(*args, **kwargs)
+        if rank == 0:
+            with open(_LOCAL_PICKLED_RESULT_FILENAME, 'wb') as f:
+                try:
+                    cloudpickle.dump(return_value, f)
+                except Exception as e:
+                    raise RuntimeError("Caught an excpetion while pickling "
+                                       "return value: {}".format(repr(e)))
```
mengxr: Does it hide the stacktrace from `e`? You can wrap it under `RuntimeError`.

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+    launcher_path = os.path.join(wdir, _PROC_LAUNCHER_FILENAME)
+    with open(launcher_path, 'w') as f:
+        f.write(textwrap.dedent("""
+        set -e
+        cd {wdir}
+        rank=${{OMPI_COMM_WORLD_RANK:-0}}
+        {exec} -c "import cloudpickle; cloudpickle.load(open('{fn_file}', 'rb'))(rank=$rank)"
+        if [[ "$rank" -eq 0 ]]; then
+        scp -q {scp_port_opt} -o StrictHostKeyChecking=no \
+        {wdir}/{local_result} {driver_host}:{wdir}/{result}
```
mengxr: Cannot assume that worker can ssh back. Should `scp` from mpirun node.

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+        fi
+        """.format(wdir=wdir,
+                   exec=sys.executable,
+                   scp_port_opt=scp_port_opt,
+                   fn_file=_PICKLED_PROC_FN_FILENAME,
+                   driver_host=driver_host,
+                   local_result=_LOCAL_PICKLED_RESULT_FILENAME,
+                   result=_PICKLED_RESULT_FILENAME)))
+
+    os.chmod(launcher_path, 0o777)
```
mengxr: owner read access is sufficient

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+_TAIL_LINES_TO_KEEP = 100
+
+_LOCAL_PICKLED_RESULT_FILENAME = "local_result.pkl"
+_PICKLED_RESULT_FILENAME = "result.pkl"
+_PICKLED_PROC_FN_FILENAME = "proc_fn.pkl"
+_PROC_LAUNCHER_FILENAME = "launch.sh"
+
+
+def run_func(
+        fn,
```
mengxr: an alternative is just take `fn` without `args`, `kwargs`, user can do `lambda: fn(args, kwargs)`. @alsrgv your preference?

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+        host = "localhost:1"
+    elif host is None:
+        host = "localhost:{np}".format(np=num_proc)
+    elif num_proc is None:
+        num_proc = sum(int(x.split(':')[1]) for x in host.split(','))
+
+    all_host_names = list(set(x.split(':')[0] for x in host.split(',')))
+    remote_host_names = network.filter_local_addresses(all_host_names)
+
+    for remote_host in remote_host_names:
```
mengxr: maybe leave a TODO here to `scp` in parallel

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+        print("Run command: " + " ".join(cmd))
+
+    if env is None:
+        env = os.environ
+    task = subprocess.Popen(cmd,
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.STDOUT,
+                            stdin=subprocess.PIPE,
+                            env=env)
+    task.stdin.close()
```
mengxr: make a comment here why we close it

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+        cmd += ["--verbose"]
+
+    cmd += ["bash", launcher_path]
+
+    if verbose:
+        print("Run command: " + " ".join(cmd))
+
+    if env is None:
+        env = os.environ
+    task = subprocess.Popen(cmd,
```
mengxr: Should be the same as

~~~python
subprocess.check_call([...], stdin=subprocess.DEVNULL)
~~~

horovod/run/run_func.py
```diff
@@ -0,0 +1,220 @@
...
+    if host is None and num_proc is None:
+        num_proc = 1
+        host = "localhost:1"
+    elif host is None:
+        host = "localhost:{np}".format(np=num_proc)
+    elif num_proc is None:
+        num_proc = sum(int(x.split(':')[1]) for x in host.split(','))
+
+    all_host_names = list(set(x.split(':')[0] for x in host.split(',')))
+    remote_host_names = network.filter_local_addresses(all_host_names)
```
mengxr: might be just use a different path but do a simple distinct on all hosts

test/test_run_func.py
```diff
@@ -0,0 +1,51 @@
...
+import torch
+import traceback
+import unittest
+import warnings
+
+import horovod.torch as hvd
+from horovod.run import run_func
+
+
+class RunFuncTests(unittest.TestCase):
```
mengxr: need more tests:)

mengxr:
I suggest using sequential mapper so it is easy to determine the ranks. Not sure whether it is compatible with all MPI implementations though.

horovod/run/run_func.py
WeichenXu123: keep the same with `horovodrun` arg `host`

horovod/run/run_func.py
WeichenXu123: Collect all processes results.

horovod/run/run_func.py
WeichenXu123: Update to be collecting all processes results.

horovod/run/run_func.py
WeichenXu123: Update to be `fn` without args/kwargs

horovod/run/run_func.py
```diff
@@ -0,0 +1,199 @@
...
+    cmd += ["bash", os.path.join(wdir, _PROC_LAUNCHER_FILENAME)]
+
+    if verbose:
+        print("Run command: " + " ".join(cmd))
+
+    if env is None:
+        env = os.environ
+
+    # close stdin for sub-process, avoid sub-process to read data from stdin.
+    subprocess.check_call(cmd, stdin=subprocess.DEVNULL, env=env)
```
mengxr: Sorry, I'm wrong about check_call would capture stdout/stderr in notebook env. We still need:

~~~
p = subprocess.Popen(cmd, stdin=DEVNULL, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, env=env)
for line in p.stdout:
  sys.stdout.write(line)
p.wait()
~~~

horovod/run/run_func.py
```diff
@@ -0,0 +1,199 @@
...
+                               "Error message is {msg}."
+                               .format(hostname=hostname, msg=output_msg))
+
+    results = [None] * num_proc
+
+    for i in range(num_proc):
+        result_path = "{wdir}/result/{prefix}.{rank}".format(
+            wdir=wdir, prefix=_PICKLED_RESULT_PREFIX, rank=i)
+        with open(result_path, 'rb') as f:
+            result_bytes = f.read()
```
mengxr: `cloudpickle.load(f)`

horovod/run/run_func.py
```diff
@@ -0,0 +1,199 @@
...
+        results[i] = cloudpickle.loads(result_bytes)
+
```
mengxr: ensure clean-up of `wdir`

horovod/run/run_func.py
```diff
@@ -0,0 +1,199 @@
...
+             corresponds to the returned result of rank i process.
+    """
+
+    wdir = tempfile.mkdtemp()
+    if verbose:
+        print("horovod.run_func working dir is " + wdir)
+
+    if ssh_port:
+        ssh_port_opt = "-p " + str(ssh_port)
+        scp_port_opt = "-P " + str(ssh_port)
```
mengxr: no need to define `scp_port_opt`, it is the same as SSH port

mengxr:
@alsrgv Do you have time to make a pass? Need your input on API and compatibility with different MPI impls.

horovod/run/run_func.py
WeichenXu123: oh, they're different. scp should only use "-P", ssh on some system only support "-p", some system support both "-p" and "-P". So for safety ssh I use "-p"

WeichenXu123:
@alsrgv Seemingly some test env do not configure `scp` correctly (scp file to/from 127.0.0.1 failed). My minimal unit-test require this. Who can help fix this ? Thanks!

horovod/run/run_func.py
```diff
@@ -0,0 +1,199 @@
...
+    local_pickled_proc_fn_path = os.path.join(wdir, _LOCAL_PICKLED_PROC_FN_FILENAME)
+    with open(local_pickled_proc_fn_path, 'wb') as f:
+        f.write(pickled_proc_fn_str)
+
+    local_launcher_path = os.path.join(wdir, _LOCAL_PROC_LAUNCHER_FILENAME)
+    with open(local_launcher_path, 'w') as f:
+        f.write(textwrap.dedent("""
+        set -e
+        cd {wdir}
+        rank=${{OMPI_COMM_WORLD_RANK:-0}}
```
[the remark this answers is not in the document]
WeichenXu123: I can remove this. We can get the rank id in mpi process code.

alsrgv:
@WeichenXu123 @mengxr thanks for the PR!

Before going into too many details, I wanted to discuss the high-level design.  Here's my thinking:

1. `horovod.run()` API should align as much as possible with `horovod.spark.run()`. Including `args`, `kwargs`. I think we can keep the name `horovod.run()` by exporting `run` method in `horovod/__init__.py` (there are other ways, similar to what TensorFlow does, but we don't have to go there yet).  We can rename the current `horovod.run` package into `horovod._run` to make it work.

2. I think `horovod.run` should integrate with `horovodrun` code, rather than execute it.

3. We should not scp any code. Instead, we can do what `horovod.spark.run` does - run a thread in `horovod.run` that the executors can call back to and determine which code to run.  @zsh-thu is working on another task (#1181) that will add server thread in `horovodrun` that will perform registry role, so perhaps this server thread can be used for sharing the function as well.

What do you think?

horovod/run/run_func.py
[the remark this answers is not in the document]
WeichenXu123: Thought again. Remove this is troublesome. It requires the arg `fn` return "rank", because I don't know which framework in `fn` user will use, e.g., If tensorflow, get rank by `horovod.tensorflow.rank()`, if pytorch, get rank by `horovod.pytorch.rank()`.

horovod/run/run_func.py
```diff
@@ -0,0 +1,199 @@
...
+        {exec_path} -c "import cloudpickle; cloudpickle.load(open('{fn_file}', 'rb'))(rank=$rank)"
+        """.format(wdir=wdir,
+                   exec_path=sys.executable,
+                   fn_file=_PICKLED_PROC_FN_FILENAME)))
+
+    if host is None:
+        host = "localhost:{np}".format(np=num_proc)
```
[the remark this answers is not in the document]
WeichenXu123: Will remove this line code.
then if host is none, will launch `horovod` without `-H` arg, so that mpi will automatically use local node.

horovod/run/run_func.py
```diff
@@ -0,0 +1,199 @@
...
+        start_timeout=None,
+        env=None,
+        verbose=False):
+    """
+    Run horovod inside python program.
+    Run `num_proc` processes executing `fn` on specified hosts.
+
+    :param fn: Function to run.
+    :param num_proc: Number of Horovod processes.
+    :param host: To specify the list of host names as well as the
```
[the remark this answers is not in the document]
WeichenXu123: It require a single string in current code.
I prefer a single string split by "," rather then a list. It keep consistent with `horovodrun` command.

horovod/run/run_func.py
```diff
@@ -0,0 +1,199 @@
...
+
+    all_host_names = set(x.split(':')[0] for x in host.split(','))
+
+    # copy pickled function and launcher script to all hosts via scp
+    # TODO: parallelize the scp copy
+    for hostname in all_host_names:
+        scp_cmd = textwrap.dedent(
+            """
+            ssh {ssh_port_opt} -o StrictHostKeyChecking=no {hostname} mkdir -p {wdir} && \
+            scp -r -q {scp_port_opt} -o StrictHostKeyChecking=no \
```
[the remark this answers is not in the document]
WeichenXu123: Current impl will change file name when copy to dest location (remove "local_" prefix). This is to avoid error when scp copy to local machine.
But the scp copy multiple file syntax `scp "/path/fileA /path/fileB" remote:/path/` could not change dest file name.
Another way is to use `scp -r` to copy the whole dir to dest but it generate more dirs.

horovod/run/run_func.py
```diff
@@ -0,0 +1,199 @@
...
+            {wdir}/{local_launcher} {hostname}:{wdir}/{launcher}
+            """).format(ssh_port_opt=ssh_port_opt, scp_port_opt=scp_port_opt,
+                        wdir=wdir, hostname=hostname,
+                        local_fn_file=_LOCAL_PICKLED_PROC_FN_FILENAME,
+                        fn_file=_PICKLED_PROC_FN_FILENAME,
+                        local_launcher=_LOCAL_PROC_LAUNCHER_FILENAME,
+                        launcher=_PROC_LAUNCHER_FILENAME)
+
+        output = six.StringIO()
+        exit_code = safe_shell_exec.execute(scp_cmd, stdout=output, stderr=output)
```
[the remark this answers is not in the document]
WeichenXu123: I read `safe_shell_exec` code. It is a wrapper for `subprocess.Popen`. But it add two features:
* when parent process exit, force child process exit.
* forward child process stdout/stderr to user-specific stream. We can set it to redirect to parent process stdout/stderr.

Btw, I add a arg `disable_stdin` for `safe_shell_exec`.

So that I suggest to use `safe_shell_exec` in every places.

horovod/run/run_func.py
WeichenXu123: See my comments above.

WeichenXu123:
@alsrgv Good suggestion! I will discuss with @mengxr and update then.

alsrgv:
@WeichenXu123 any updates?  Note, you will need to rebase on #1181 once it lands as it does a few changes to `horovodrun`.

WeichenXu123:
@alsrgv I will rebase #1181 and update code in a few days. Thanks!

WeichenXu123:
@alsrgv 
Current code `RendezvousServer` only launched in gloo. If we want to support `run_func` in both gloo and openmpi, then should we launch a separate http server for sharing the executor function ? Or do you have some recommends ? Thanks!

zsh-thu:
> @alsrgv
> Current code `RendezvousServer` only launched in gloo. If we want to support `run_func` in both gloo and openmpi, then should we launch a separate http server for sharing the executor function ?
> 
> My proposal is in `run_func`, I launch a specific http server for sharing executor functions. This make code clear.
> 
> Thanks!

@WeichenXu123  Thanks for the pr!
I think the `RendezvousServer`  is only an underlying implementation. You can just add a line to parse the `--gloo` option for `horovodrun` at [here](https://github.com/horovod/horovod/pull/1192/files#diff-6f3d5bb650876f00c1c37873c99a105dR162) and run the command through `safe_shell_exec.execute`. The rest should be taken cared for by `horovodrun`.

WeichenXu123:
@zsh-thu 
Oh, sorry I do not explain my issue clearly. I know the `--gloo` option, but here I want to discuss the issue @alsrgv mentioned in earlier comment:
> We should not scp any code. Instead, we can do what horovod.spark.run() does - run a thread in horovod.run() that the executors can call back to and determine which code to run. @zsh-thu is working on another task (#1181) that will add server thread in horovodrun that will perform registry role, so perhaps this server thread can be used for sharing the function as well.

So I want to launch a separate http server used for broadcasting the "executor func". (so do not reuse the http server in gloo impl)

WeichenXu123:
See new impl here
https://github.com/horovod/horovod/pull/1307

PR #1209, opened 2019-07-10
Separately build gloo for each plugins

zsh-thu:
This pr is an amendment to the pr that Alex created for autodetecting matching gcc compiler. Since different plugins may be compiled using different compilers, the gloo needs to be compiled with the same compiler, which means gloo should be compiled each time a plugin is compiled.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ import subprocess
 import sys
 import textwrap
 import traceback
+from copy import deepcopy
 from distutils.errors import CompileError, DistutilsError, \
     DistutilsPlatformError, LinkError
 from distutils.sysconfig import customize_compiler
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -826,6 +828,9 @@ def build_tf_extension(build_ext, options):
     try:
         with env(CC=cc_compiler, CXX=cxx_compiler, CFLAGS=cflags, CPPFLAGS=cppflags,
                  LDSHARED=ldshared):
+
```
alsrgv: Remove empty line?

setup.py
```diff
@@ -899,7 +904,17 @@ def is_mx_cuda():
     return False
 
 
-def build_mx_extension(build_ext, options):
+def build_mx_extension(build_ext, global_options):
+
```
alsrgv: Style: remove empty lines in the beginning of the block

setup.py
```diff
@@ -899,7 +904,17 @@ def is_mx_cuda():
...
+    # Backup the options, preventing other plugins access libs that
+    # compiled with compiler of this plugin
+    options = deepcopy(global_options)
+
+    # First build gloo
+
```
alsrgv: Remove empty line

setup.py
```diff
@@ -1036,13 +1051,22 @@ def __exit__(self, type, value, traceback):
             os.rename(file + '.protected', file)
 
 
-def build_torch_extension(build_ext, options, torch_version):
+def build_torch_extension(build_ext, global_options, torch_version):
+
```
alsrgv: Remove empty line

PR #1223, opened 2019-07-15
Do not return random ints from init and shutdown

orivej:
C API defines horovod_init, horovod_init_comm, and horovod_shutdown as returning
void.  When ctypes interprets their return values as ints, it returns garbage.


alsrgv:
@orivej, thanks for the PR. LGTM, but can you rebase on master so that integration tests pass?

orivej:
I have force pushed, but everything failed because currently registry-1.docker.io returns 500 Internal Server Error.

alsrgv:
@orivej, thanks!  I think the public docker registry had an outage, I'll restart the build.

PR #1241, opened 2019-07-21
Add abi related compiler args when build cmake extension.

zsh-thu:
Due to ABI compatibility with GCC after 5.1, different ABI compiler setting can lead to "undefined symbol" errors for submodule Gloo during runtime for Pytorch plugin. So we must use the same ABI compiler flag that PyTorch uses to compile Gloo. By setting `CMAKE_CXX_FLAGS_[DEBUG/RELEASE]` (whichever mode we are using), cmake can append these flags to CMAKE_CXX_FLAGS when compiling files. 
This pr also added `cmake` as a required package for test docker, so that integration tests can build Gloo for each plugins.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ import subprocess
 import sys
 import textwrap
 import traceback
+import pipes
 from copy import deepcopy
 from distutils.errors import CompileError, DistutilsError, \
     DistutilsPlatformError, LinkError
@@ -772,7 +773,7 @@ def remove_offensive_gcc_compiler_options(compiler_version):
 def filter_compile_macros(compile_flags):
     res = []
     for flag in compile_flags:
-        if flag[:2] == '-D':
+        if flag.startswith('-D'):
             res.append(flag)
     return res
 
@@ -786,7 +787,7 @@ def build_tf_extension(build_ext, global_options):
     tf_compile_flags, tf_link_flags = get_tf_flags(
         build_ext, options['COMPILE_FLAGS'])
 
-    gloo_compile_flag = filter_compile_macros(tf_compile_flags)
+    gloo_compile_macros = filter_compile_macros(tf_compile_flags)
 
     tensorflow_mpi_lib.define_macros = options['MACROS']
     tensorflow_mpi_lib.include_dirs = options['INCLUDES']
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -1240,20 +1254,25 @@ def build_cmake(build_ext, ext, prefix, plugin_ext=None, options=None):
...
-        '-DUSE_MPI=ON',
-        '-DCMAKE_BUILD_TYPE=' + config,
-        '-DCMAKE_LIBRARY_OUTPUT_DIRECTORY_{}={}'.format(config.upper(), extdir),
-        '-DCMAKE_ARCHIVE_OUTPUT_DIRECTORY_{}={}'.format(config.upper(),
-                                                        lib_output_dir),
-    ]
+
+    # Pass additional compiler flags by setting CMAKE_CXX_FLAGS_[DEBUG/RELEASE]
+    # so that cmake will append these flags to CMAKE_CXX_FLAGS
+    additional_cxx_flags = ' '.join(additional_flags)
```
alsrgv: Do you need to pipes.quote that, to ensure it's in " "s?

setup.py
```diff
@@ -1240,20 +1254,25 @@ def build_cmake(build_ext, ext, prefix, plugin_ext=None, options=None):
...
+    cmake_args = ['-DUSE_MPI=ON',
+                  '-DCMAKE_BUILD_TYPE=' + config,
+                  '-DCMAKE_CXX_FLAGS_'+config.upper()+':STRING='+additional_cxx_flags,
```
alsrgv: Style: keep spaces between + or maybe use string formatting?

setup.py
```diff
@@ -767,6 +767,16 @@ def remove_offensive_gcc_compiler_options(compiler_version):
     return None, None, None
 
 
+# Filter out all the compiler macros (starts with -D)
+# that need to be passed to compiler
+def filter_compile_macros(compile_flags):
+    res = []
+    for flag in compile_flags:
+        if flag[:2] == '-D':
```
alsrgv: .startswith() is better

setup.py
```diff
@@ -776,6 +786,8 @@ def build_tf_extension(build_ext, global_options):
     tf_compile_flags, tf_link_flags = get_tf_flags(
         build_ext, options['COMPILE_FLAGS'])
 
+    gloo_compile_flag = filter_compile_macros(tf_compile_flags)
```
alsrgv: gloo_compile_flag  -> gloo_compile_macros

setup.py
zsh-thu: Yes, that's much more secure.

zsh-thu: But I use a string prefix in the bash args, I assume they are of the same functionality?

setup.py
zsh-thu: Yes.

setup.py
zsh-thu: changed.

setup.py
zsh-thu: Fixed.

Issue #1240, opened 2019-07-21
Feature Request: Sum/Average over Batch


DEKHTIARJonathan:
Hi,

As you may have seen, TensorFlow made default to aggregate gradients use sum across workers:

> Set default loss reduction as AUTO for improving reliability of loss scaling with distribution strategy and custom training loops. AUTO indicates that the reduction option will be determined by the usage context. For almost all cases this defaults to SUM_OVER_BATCH_SIZE. When used in distribution strategy scope, outside of built-in training loops such as tf.keras compile and fit, we expect reduction value to be 'None' or 'SUM'. Using other values will raise an error. 

Source: https://github.com/tensorflow/tensorflow/releases/tag/v1.14.0

That would be nice to have a flag switching for one to another. If I'm right the only way the change the default behavior for HVD is to re-install the lib with `HOROVOD_GPU_ALLREDUCE="NCCL"` or `HOROVOD_GPU_ALLGATHER="NCCL"`

Is this a feature that may be possible ?

PR #1257, opened 2019-07-28
[TF] New Feature: Sum/Average over Rank

DEKHTIARJonathan:
Hi,

As stated in #1240, as you may have seen, TensorFlow made default to aggregate gradients use sum across workers:

> Set default loss reduction as AUTO for improving reliability of loss scaling with distribution strategy and custom training loops. AUTO indicates that the reduction option will be determined by the usage context. For almost all cases this defaults to SUM_OVER_BATCH_SIZE. When used in distribution strategy scope, outside of built-in training loops such as tf.keras compile and fit, we expect reduction value to be 'None' or 'SUM'. Using other values will raise an error. 

Source: https://github.com/tensorflow/tensorflow/releases/tag/v1.14.0

That would be nice to have a flag switching for one to another. If I'm right the only way the change the default behavior for HVD is to re-install the lib with `HOROVOD_GPU_ALLREDUCE="NCCL"` or `HOROVOD_GPU_ALLGATHER="NCCL"`

I drafted something quick to be able to test this feature without using having to use TF Distributed, let me know what u think about it

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,16 @@
+# Checklist before submitting (remove before creating the PR)
+
+- [ ] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)?
+- [ ] Did you update the docs?
+- [ ] Did you write any tests to validate this change?  
+- [ ] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users?
+
+## Description
+
+Fixes # (issue).
+
+## Review process to land 
+
+1. All tests and other checks must succeed.
+2. At least one member of the [technical steering committee](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md) must review and approve.
+3. If any member of the technical steering committee requests changes, they must be addressed.


horovod/tensorflow/__init__.py
```diff
@@ -282,14 +286,17 @@ def DistributedOptimizer(optimizer, name=None, use_locking=False, device_dense='
         Treat all sparse gradients as dense tensors.  This can help improve
         performance and memory utilization if the original sparse gradient
         has high density.  Defaults to false.
+      sum_over_ranks: If True, computes the sum of the gradients
```
alsrgv: Please add new line before `If True`, to align with previous args.

DEKHTIARJonathan: Fixed in ffda60499308d659f7502b762e3e2371a4c23750

DEKHTIARJonathan:
Honestly with the few attempts I did, the results were catastrophic. My attempt was to implement the feature to allow anyone to use the same setup as TF Distributed (might be useful in some cases).

I don't really understand what is the motivation behind making this default in Tensorflow, nonetheless I believe we should allow the functionality (even if we don't recommend it) for reproducibility and test objectives.

alsrgv:
I don't know why TensorFlow made this change, but I think we need to find a use case where it's useful and clearly document it, before adding an option.  We should not add options that don't have known useful use cases.

DEKHTIARJonathan:
I would partly agree. Just for a result reproducibility or test objective, it would make sense to actually add the feature. And if Google made it default there are most certainly numerous use-cases.

I'm going to investigate a little. And try to understand why they went that direction

DEKHTIARJonathan:
As explained in the Tensorflow issue: https://www.tensorflow.org/beta/tutorials/distribute/training_loops

> Normally, on a single machine with 1 GPU/CPU, loss is divided by the number of 
examples in the batch of input. So, how should the loss be calculated when using 
a tf.distribute.Strategy?For an example, let's say you have 4 GPU's and a batch 
size of 64.One batch of input is distributed across the replicas (4 GPUs), each 
replica getting an input of size 16.The model on each replica does a forward pass 
with its respective input and calculates the loss. Now, instead of dividing the loss
by the number of examples in its respective input (BATCH_SIZE_PER_REPLICA = 16),
the loss should be divided by the GLOBAL_BATCH_SIZE (64).
Why do this?
This needs to be done because after the gradients are calculated on each replica, 
they are synced across the replicas by summing them.
How to do this in TensorFlow? If you're writing a custom training loop, as in this 
tutorial, you should sum the per example losses and divide the sum by the 
GLOBAL_BATCH_SIZE: scale_loss = tf.reduce_sum(loss) * (1. / GLOBAL_BATCH_SIZE) 
or you can use tf.nn.compute_average_loss which takes the per example loss, optional 
sample weights, and GLOBAL_BATCH_SIZE as arguments and returns the scaled loss.

One point is interesting:

How does Horovod handle different batch size per GPU ?
- Case A: Average gradients without taking local batch size in consideration
- Case B: Average gradients using the global batch size

If Horovod follows case A, then doing sum of gradients actually makes sense. You can do `local_loss = local_loss / global_batch_size` and then sum the gradients instead of averaging

alsrgv:
Ok, I think the reason for this is that TensorFlow actually uses global batch size in their estimator framework to make decisions.

In Horovod, the concept is about per-GPU batch size, and scaling it up with Horovod ranks & LR adjustment.

It's just a different thought model.

I do have a draft code for HorovodDistributionStrategy which will be compatible with their philosophy. Sneak peek [here](https://gist.github.com/alsrgv/148866f4747f4b9a77085c7a5311c944).  Part focused on averaging is [here](https://gist.github.com/alsrgv/148866f4747f4b9a77085c7a5311c944#file-hvd_dist-py-L70).  Since it does not use DistributedOptimizer, I don't think we need this change.

PR #1265, opened 2019-07-31
TF 2.x BroadcastGlobalVariablesHook Fix

DEKHTIARJonathan:
Hi,

I found that `BroadcastGlobalVariablesHook` is not working with Tensorflow 2.x.

Issue:

```python
  File "cnn_session.py", line 169, in <module>
    training_hooks.append(hvd.BroadcastGlobalVariablesHook(0))
AttributeError: module 'horovod.tensorflow' has no attribute 'BroadcastGlobalVariablesHook'
```
Same issue with: `broadcast_global_variables`
```python
  File "/usr/local/lib/python3.6/dist-packages/horovod/tensorflow/__init__.py", line 167, in begin
    self.bcast_op = broadcast_global_variables(self.root_rank)
NameError: name 'broadcast_global_variables' is not defined
```

Same issue with: `get_default_graph`
```python
  File "/usr/local/lib/python3.6/dist-packages/horovod/tensorflow/__init__.py", line 172, in begin
    if not self.bcast_op or self.bcast_op.graph != tf.get_default_graph():
AttributeError: module 'tensorflow' has no attribute 'get_default_graph'
```
-------------

Basically, the hook is not being defined. This PR implements fixes for these issues.

Tested on MaskRCNN and it works as expected


alsrgv:
@DEKHTIARJonathan, the reason these were excluded is due to the fact that when I examined `tf.compat.v1.global_variables()` I found that it was always empty. So, having this hook gives the user a false impression that variable synchronization is happening, while in reality, it does not.

Do you see `tf.compat.v1.global_variables()` returning non-empty list in your case?

alsrgv:
This is more TF 2.0 way of broadcasting model & optimizer weights: https://github.com/horovod/horovod/blob/master/examples/tensorflow2_mnist.py#L75

DEKHTIARJonathan:
In tensorflow 2 you have the possibility to deactivate eager execution. And if done Gradient Tape is not the way to go. And `tf.compat.v1.global_variables()` returns everything you expect

alsrgv:
I see.  But it seems to defeat the purpose of migrating to TensorFlow 2.0.  Why not keep using TensorFlow 1.14?

I'm worried that bringing back these hooks will be very misleading.  Unless eager execution is disabled, they will do nothing w/o complaining.

A middle ground could be throwing an error with TensorFlow 2.0 is eager mode was not disabled.

DEKHTIARJonathan:
Going this direction will enforce basically block everyone using Horovod to migrate to TF 2.x without completely rewriting the project for eager execution. In addition, it's still unclear if eager execution has any performance cost.

Tensorflow provide a script `tf_upgrade_v2` which basically convert most of the API to `tf.compat.v1.****`. Enforcing directly "if you use TF2.x you must use eager execution" without any technical limitation will block numerous people from using Horovod in TF2.x or from upgrading to TF 2.x at all.

I agree it can be misleading/dangerous, hence we should issue a warning. However, I'm not convinced blocking it completely is good decision.

alsrgv:
That's fair. Can we throw an error if the hook/functions are used outside of the graph mode?

DEKHTIARJonathan:
I think it's doable. Let me look into it ;)

DEKHTIARJonathan:
@alsrgv Seems like yes => https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/eager/context.py#L648

alsrgv:
Sounds good.  We should add tests for these scenarios.

DEKHTIARJonathan:
@alsrgv I've looked into it. And I don't know how to add unittest for this. It's impossible to use a TF Hook in Eager mode ... At the least to the best of my knowledge

horovod/tensorflow/__init__.py
```diff
@@ -158,7 +178,16 @@ def __init__(self, root_rank, device=''):
             self.device = device
 
         def begin(self):
-            if not self.bcast_op or self.bcast_op.graph != tf.get_default_graph():
+            if LooseVersion(tf.__version__) >= LooseVersion('1.7.0'):  # Eager Mode has been introduced in TF 1.7.0
```
alsrgv: Can you use `_executing_eagerly()` from util.py instead?

horovod/tensorflow/__init__.py
```diff
@@ -158,7 +178,16 @@ def __init__(self, root_rank, device=''):
...
+                from tensorflow.python.eager import context
+                if context._context is not None and context.executing_eagerly():
+                    raise RuntimeError(
+                        "Eager Execution is not supported by `hvd.BroadcastGlobalVariablesHook`\n"
+                        "We recommend using `hvd.DistributedGradientTape` instead"
+                    )
+
+            if not self.bcast_op or self.bcast_op.graph != _get_default_graph():
+
```
alsrgv: Nit: remove space

horovod/tensorflow/__init__.py
```diff
@@ -121,14 +130,25 @@ def broadcast_global_variables(root_rank):
...
                        to all other processes.
         """
-        return broadcast_variables(tf.global_variables(), root_rank)
-
-
-if hasattr(tf, 'train') and hasattr(tf.train, 'SessionRunHook'):
-    if hasattr(tf, 'estimator') and hasattr(tf.estimator, 'SessionRunHook'):
-        _SessionRunHook = tf.estimator.SessionRunHook
-    else:
+        return broadcast_variables(_global_variables(), root_rank)
```
alsrgv: We should fail this call in the eager mode.

horovod/tensorflow/__init__.py
```diff
@@ -158,7 +178,16 @@ def __init__(self, root_rank, device=''):
             self.device = device
 
         def begin(self):
-            if not self.bcast_op or self.bcast_op.graph != tf.get_default_graph():
+            if LooseVersion(tf.__version__) >= LooseVersion('1.7.0'):  # Eager Mode has been introduced in TF 1.7.0
+                from tensorflow.python.eager import context
+                if context._context is not None and context.executing_eagerly():
+                    raise RuntimeError(
```
alsrgv: Let's move this check to broadcast_global_variables()

DEKHTIARJonathan:
@alsrgv You can't use the session in eager mode. That's the whole point actually

alsrgv:
I see, that's good.  What about estimators, can you use estimator in eager mode?

DEKHTIARJonathan:
If you use estimator, graph mode is enforced

https://github.com/tensorflow/estimator/blob/r2.0/tensorflow_estimator/python/estimator/estimator.py#L34
Actually this check about eager exec is useless...

alsrgv:
OK, gotcha. That's good then. Let's just make sure `broadcast_global_variables()` throws an exception in eager mode, and we should be good to go.

DEKHTIARJonathan:
I'll apply the change tomorrow and you can merge if its good with you

DEKHTIARJonathan:
@alsrgv I have done the followings:
- updated `util._executing_eagerly` to a non deprecated version of the API
=> https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/eager/context.py#L1615

- applied the cnages requested

horovod/tensorflow/util.py
```diff
@@ -26,7 +26,7 @@
 
 def _executing_eagerly():
     """Returns true if eager execution is supported and enabled."""
-    return _has_eager and context.in_eager_mode()
+    return _has_eager and context.executing_eagerly()()
```
alsrgv: Too many parenthesis?

horovod/tensorflow/__init__.py
```diff
@@ -121,14 +130,31 @@ def broadcast_global_variables(root_rank):
...
                        to all other processes.
         """
-        return broadcast_variables(tf.global_variables(), root_rank)
-
-
-if hasattr(tf, 'train') and hasattr(tf.train, 'SessionRunHook'):
-    if hasattr(tf, 'estimator') and hasattr(tf.estimator, 'SessionRunHook'):
-        _SessionRunHook = tf.estimator.SessionRunHook
-    else:
+        if _executing_eagerly():
```
alsrgv: Great!  Can you add a test for this functionality (verify that it fails in eager mode)?

horovod/tensorflow/util.py
DEKHTIARJonathan: Oupsi. Sent a patch

alsrgv:
@DEKHTIARJonathan, bump. We're planning to release Horovod 0.17.0 next week and it would be great to include this fix.  We just need to add a test which validates exception, [like this](https://github.com/horovod/horovod/blob/master/test/test_tensorflow.py#L328) :-)

DEKHTIARJonathan:
I'll do my best to send smthg tomorrow. I really hope it will make it to 0.17.0

DEKHTIARJonathan:
@alsrgv : Done, if it's good with you. Please merge. Otherwise feel free to directly modify

DEKHTIARJonathan:
@alsrgv : Done, if it's good with you. Please merge. Otherwise feel free to directly modify.
Took me a few iterations ;)

horovod/tensorflow/__init__.py
```diff
@@ -19,6 +19,7 @@
 from __future__ import division
 from __future__ import print_function
 
+from distutils.version import LooseVersion
```
alsrgv: Remove?

test/test_tensorflow.py
```diff
@@ -975,6 +975,38 @@ def test_horovod_broadcast_grad_gpu(self):
                             "gradient %s differs from expected %s, "
                             "error: %s" % (grad_out, expected, str(err)))
 
+    def test_horovod_broadcast_eager_mode_error(self):
+        """Test that tries to broadcast tensorflow global variables
+        in eager execution mode. This call should raise a RuntimeError."""
+
+        if not hvd.util._has_eager:
```
alsrgv: These tests are already executing eagerly and in graph mode (can be checked via `executing_eagerly()`), maybe we can just check for that and skip the test if there is a mode mismatch?

test/test_tensorflow.py
```diff
@@ -975,6 +975,38 @@ def test_horovod_broadcast_grad_gpu(self):
...
+            return
+
+        hvd.init()
+
+        from tensorflow.python.eager import context
+        with context.eager_mode():
+
```
alsrgv: Style: avoid empty line after `:`

alsrgv:
@DEKHTIARJonathan, just few comments

test/test_tensorflow.py
DEKHTIARJonathan: Oh that explain why I had so much trouble allowing/disabling eager exec. I'll update my code :)

horovod/tensorflow/__init__.py
DEKHTIARJonathan: Indeed, unused

DEKHTIARJonathan:
@alsrgv requested changes applied

DEKHTIARJonathan:
Awesome ! Let''s merge and make sure it makes it for 0.17.0 release :)

PR #1270, opened 2019-08-01
Added g++-7 in docker file for the latest tf-nighly

zsh-thu:
Signed-off-by: Sihan Zeng <zsh@uber.com>

### DIFF
--- a/Dockerfile.test.cpu
+++ b/Dockerfile.test.cpu
@@ -27,6 +27,13 @@ RUN apt-get install -y --no-install-recommends \
         build-essential \
         g++-4.8
 
+# install g++-7
+RUN apt-get install -y --no-install-recommends software-properties-common
+RUN apt-get update
+RUN add-apt-repository ppa:ubuntu-toolchain-r/test
+RUN apt-get update
+RUN apt-get install -y --no-install-recommends gcc-7 g++-7
+
 # Install Python.
 RUN apt-get install -y python${PYTHON_VERSION} python${PYTHON_VERSION}-dev
 RUN if [[ "${PYTHON_VERSION}" == "3.6" ]]; then \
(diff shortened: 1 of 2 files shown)


Dockerfile.test.cpu
```diff
@@ -27,6 +27,13 @@ RUN apt-get install -y --no-install-recommends \
         build-essential \
         g++-4.8
 
+# install g++-7
+RUN apt-get install -y --no-install-recommends software-properties-common
+RUN apt-get update
```
alsrgv: I don't think this line is necessary.

Dockerfile.test.cpu
```diff
@@ -27,6 +27,13 @@ RUN apt-get install -y --no-install-recommends \
...
+RUN add-apt-repository ppa:ubuntu-toolchain-r/test
+RUN apt-get update
```
alsrgv: Can you add `-qq` to not spam the output?

Dockerfile.test.cpu
```diff
@@ -27,6 +27,13 @@ RUN apt-get install -y --no-install-recommends \
...
+RUN apt-get install -y --no-install-recommends gcc-7 g++-7
```
alsrgv: We can just say `g++-7`, it will bring `gcc-7`.

Dockerfile.test.cpu
zsh-thu: These lines install the `add-apt-repository` command. Otherwise, we cannot add the test repo.

Dockerfile.test.cpu
zsh-thu: Yes.

Dockerfile.test.cpu
zsh-thu: Ok.

PR #1279, opened 2019-08-05
Update reference to horovod/common/ddl_mpi_context_manager.cc

bethune-bryant:
#1154 Broke the build for PowerAI DDL.

The location of "horovod/common/ddl_mpi_context_manager.cc" is incorrect in setup.py

https://github.com/horovod/horovod/commit/4451dd0df5243fd9941022f72b2ff17cb6833222#diff-2eeaed663bd0d25b7e608891384b7298R676

@nvcastet


bethune-bryant:
@alsrgv Does this look okay?

PR #1288, opened 2019-08-10
Changed CUDA operations to claim a copy of the fusion buffer shared pointer to prevent deallocation during finalization

### LINKED ISSUE #1282: Autotuner causes unstable training behavior at startup due to possible race condition
**Environment:**
1. Framework: Tensorflow
2. Framework version: 1.14.0
3. Horovod version: 0.16.4
4. MPI version: OpenMPI 3.1.4
5. CUDA version: 10.1
6. NCCL version: 2.4.8
7. Python version: 3.6.8
8. OS and version: Ubuntu 18.04.2
9. GCC version: 7.4.0

**Bug report:**
Recently, one of our developers has been running cases with TensorFlow + AMP and noticed some very unstable training behavior at startup while the autotuner is active, indicating some type of data corruption occurring with the gradients. The instability does not occur with the autotuner disabled.

We ran a lot of experiments and found that the issue appears to be related to modifications to `HOROVOD_FUSION_THRESHOLD`, as the training with autotuning stabilizes _if_ we set that parameter to a fixed value. 

I dug around in the code and I think it is possible that there may be a race condition which occurs when the fusion buffer size is updated by the autotuner. Since NCCL/CUDA operations are completed asynchronously w.r.t to the main Horovod thread, there seems to be a possibility that a subsequent op can trigger a reallocation of the fusion buffer *before* the previous NCCL/CUDA operation completes.

To test this hypothesis, I changed the `detach()` of the finalizer thread to `join()` to force the main thread to wait until NCCL/CUDA operations complete before continuing (see  romerojosh/horovod@5983aa6115bdc2cc039c249546270d8ee7fd0160) With this change, the training runs stably with the autotuner enabled and a variable fusion threshold. I don't think this is a proper fix as it introduces unnecessary serialization, however it does suggest that some type of sync is necessary when modifying the fusion threshold during training.

cc @DEKHTIARJonathan

tgaddair:
Fixes #1282.

### DIFF
--- a/horovod/common/fusion_buffer_manager.cc
+++ b/horovod/common/fusion_buffer_manager.cc
@@ -45,7 +45,7 @@ Status FusionBufferManager::InitializeBuffer(int64_t threshold, int device, std:
   return Status::OK();
 }
 
-std::shared_ptr<PersistentBuffer>& FusionBufferManager::GetBuffer(int device, Framework framework, int stream_id) {
+std::shared_ptr<PersistentBuffer> FusionBufferManager::GetBuffer(int device, Framework framework, int stream_id) {
   return tensor_fusion_buffers_[std::make_tuple(device, framework, stream_id)].first;
 }
 
--- a/horovod/common/fusion_buffer_manager.h
+++ b/horovod/common/fusion_buffer_manager.h
@@ -44,7 +44,7 @@ public:
                           std::function<void()> on_end_init);
 
   // Returns the buffer associated with the given device and framework, or null.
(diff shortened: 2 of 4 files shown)


DEKHTIARJonathan:
@romerojosh @tgaddair : I have launched a stability test to reproduce the issue with the commit: 54e2a9d1c14b715d653889e2ec296ab9fb5e4165.
I'll give you an update in 3-4 hours.

DEKHTIARJonathan:
@tgaddair : This PR fixes the problem.
@romerojosh : if it's good with you, let's merge 👍

romerojosh:
LGTM. Nice fix @tgaddair!

PR #1297, opened 2019-08-13
Fixed progress bars for TF Keras with Gloo controller, added and updated Keras examples

tgaddair:


tgaddair:
> Overall LGTM, can we rename new example as `tensorflow_keras_mnist.py` for consistency?

Done.

PR #1299, opened 2019-08-13
Fix optimizer broadcast bug for Adam

gkanwar:
Resolves an issue with creating an Adam optimizer after setting default tensor type to cuda, which resulted in error "can't convert CUDA tensor to numpy".

### DIFF
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -99,7 +99,7 @@ run_test() {
   echo "      pull-retries: 3"
   echo "  - ecr#v1.2.0:"
   echo "      login: true"
-  echo "  timeout_in_minutes: 7"
+  echo "  timeout_in_minutes: 5"
   echo "  retry:"
   echo "    automatic: true"
   echo "  agents:"
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -99,7 +99,7 @@ run_test() {
   echo "      pull-retries: 3"
   echo "  - ecr#v1.2.0:"
   echo "      login: true"
-  echo "  timeout_in_minutes: 5"
+  echo "  timeout_in_minutes: 7"
   echo "  retry:"
   echo "    automatic: true"
   echo "  agents:"
--- a/horovod/torch/__init__.py
+++ b/horovod/torch/__init__.py
@@ -349,12 +349,12 @@ def broadcast_optimizer_state(optimizer, root_rank):
     # new unwrapped scalar value via a callback.
     def _create_callback(pid, name, t, p):
         def _from_tensor():
-            state_dict['state'][pid][name] = t(p.numpy()[0])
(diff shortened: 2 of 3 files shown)


alsrgv:
@gkanwar, thanks! Could you add a test case that covers this scenario?

gkanwar:
@alsrgv done!

alsrgv:
@gkanwar, looks like the test is getting stuck. Could you take a look?

gkanwar:
@alsrgv sorry to say I have no idea why this would hang. It simply delegates to the other test test_broadcast_state after setting the default torch Tensor type. Any ideas?

gkanwar:
@alsrgv, following up on this before I lose track, do you have a suggestion for how to investigate this hanging test? I'm pretty unfamiliar with your test setup.

alsrgv:
@gkanwar, sorry for the delay.  I tried this on my machine, and it worked - but perhaps a bit slower. GPU PyTests are on the verge of 5-minute timeout threshold, so perhaps this test pushes them over the boundary just enough.

Can you simplify the test definition back to:
```
    def test_broadcast_state_gpu(self):
        # Only do this test if there are GPUs available.
        if not torch.cuda.is_available():
            return
        # Set default tensor type, ensuring optimizer tensor-wrapping is robust
        # to this setting.
        try:
            torch.set_default_tensor_type(torch.cuda.FloatTensor)
            self.test_broadcast_state()
        finally:
            torch.set_default_tensor_type(torch.FloatTensor)
```

And set the timeout in [run_test](https://github.com/horovod/horovod/blob/master/.buildkite/gen-pipeline.sh#L102) to 7 minutes?

alsrgv:
@gkanwar, unfortunately, this tests deadlocks K80s that we use for the integration test.

Can you add skip pragma to the unit test, like this:
```
    # TODO: investigate why this hangs on K80s
    @unittest.skip
    def test_broadcast_state_gpu(self):
        # Only do this test if there are GPUs available.
        if not torch.cuda.is_available():
            return
        # Set default tensor type, ensuring optimizer tensor-wrapping is robust
        # to this setting.
        try:
            torch.set_default_tensor_type(torch.cuda.FloatTensor)
            self.test_broadcast_state()
        finally:
            torch.set_default_tensor_type(torch.FloatTensor)
```

and revert the 7 minute change?  Then we can land & call it a day.

alsrgv:
@gkanwar, I've added commits to resolve my comments, so we can land this and make a release #1318.

gkanwar:
@alsrgv, thanks for taking care of it. Landing this soon sounds good. Any idea what is the cause of the deadlock?

alsrgv:
@gkanwar, I have a feeling there's some conflict between PyTorch's CUDA threads & NCCL CUDA threads.  K80 may have old enough scheduling infrastructure to not be able to resolve those.

alsrgv:
Unit test failure is known, merging.

PR #1302, opened 2019-08-14
Fixed Gloo control plane to work when world size is 1

tgaddair:


horovod/common/gloo/gloo_context.cc
```diff
@@ -29,6 +33,38 @@
 namespace horovod {
 namespace common {
 
+// Horovod Gloo rendezvous knobs.
```
alsrgv: Any reason to move to .cc? AFAIK other vars are in .h files.

tgaddair: I noticed those variables were only used within the CC.  Moving them there helps reduce clutter in the namespace when importing the header, so I think it's generally a good idea.  Though maybe we could move the environment variables into common.h for consistency (perhaps in followup).

PR #1303, opened 2019-08-14
Added horovodrun action --check-build to print available frameworks and libraries

tgaddair:


tgaddair:
> Amazing!
> 
> Can you add a test?

Done.

PR #1304, opened 2019-08-14
(MXNet) Fixed: unknown device mask issue

karan6181:
This issue arises `[1,0]<stderr>:mxnet.base.MXNetError: [05:07:07] src/ndarray/ndarray.cc:1282: unknown device mask` when running some specific MXNet model script when horovod installed without NCCL.

The issue is because we had two Async nested calls, for example, `A` and `B` and it is called as `B(A)`. The temporary CPU tensor is created in Async call method `A` and it's deleted when it's goes out of scope, so the Async method `B`doesn't aware of it and tries to copy the data from CPU tensor(Which was created in `A`) to GPU which results in above failure. This block of code is only executed when horovod compiled without NCCL. The fix is to create a shared pointer outside of the `B` scope and pass it in `B(A)` and once the tensor usage has finished, it will be deleted from the memory automatically.

@apeforest @yuxihu @eric-haibin-lin

### DIFF
--- a/horovod/mxnet/mpi_ops.h
+++ b/horovod/mxnet/mpi_ops.h
@@ -33,18 +33,18 @@ using namespace horovod::common;
 typedef ::mxnet::NDArray NDArray;
 typedef ::mxnet::Engine::CallbackOnComplete CallbackOnComplete;
 typedef Request::RequestType OperationType;
-typedef std::shared_ptr<MXTensor<NDArray>> MXTensor_sh_ptr;
+typedef std::shared_ptr<MXTensor<NDArray>> MXTensorSharedPtr;
 
 struct MpiOpsParam {
   NDArray* input;
   NDArray* output;
-  MXTensor_sh_ptr cpu_tensor;
+  MXTensorSharedPtr cpu_tensor;
   OperationType op_type;
   std::string op_name;
   int root_rank;
 
   MpiOpsParam(NDArray* input, NDArray* output,
-              MXTensor_sh_ptr cpu_tensor,
+              MXTensorSharedPtr cpu_tensor,
               const OperationType& op_type, const std::string& op_name,
               int root_rank)
       : input(input),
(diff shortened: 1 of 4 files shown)


horovod/mxnet/mpi_ops.h
```diff
@@ -33,16 +33,18 @@ using namespace horovod::common;
 typedef ::mxnet::NDArray NDArray;
 typedef ::mxnet::Engine::CallbackOnComplete CallbackOnComplete;
 typedef Request::RequestType OperationType;
+typedef std::shared_ptr<MXTensor<NDArray>> MXTensor_sh_ptr;
 
 struct MpiOpsParam {
   NDArray* input;
   NDArray* output;
-  NDArray* cpu_tensor;
+  MXTensor_sh_ptr cpu_tensor;
```
yuxihu: The name MXTensor_sh_ptr looks weired. Can you rename it, e.g. MXTensorSharedPtr?

karan6181: Sure. Thanks for the input.

karan6181: addressed in [ca9400a](https://github.com/horovod/horovod/pull/1304/commits/ca9400a9587cb0f8b2efe22f4a7016de06b74be2).

horovod/mxnet/mpi_ops.h
```diff
@@ -33,16 +33,18 @@ using namespace horovod::common;
 typedef ::mxnet::NDArray NDArray;
 typedef ::mxnet::Engine::CallbackOnComplete CallbackOnComplete;
 typedef Request::RequestType OperationType;
+typedef std::shared_ptr<MXTensor<NDArray>> MXTensorSharedPtr;
```
yuxihu: hvd_cpu_buffer is [created](https://github.com/horovod/horovod/pull/1304/files#diff-79b78539e15db1d75d73029490eed708R190) as a shared ptr of MXTemporaryBuffer. Here we define a type of shared ptr for MXTensor. Do they need to be consistent? Can we use either MXTensor or MXTemporaryBuffer?

karan6181:
@apeforest Addressed the comments and also resolved the DCO. Thanks!

horovod/mxnet/mpi_ops.h
karan6181: I used `MXTemporaryBuffer` in `PushHorovodOperationCudaOnCPU()` because it has the destructor method to delete the tensor when not in use. In the above line, I have used `MXTensor` to make it more generaliz(can be use in future scenarios).

PR #1307, opened 2019-08-15
Add `horovod.run.run` to make horovod notebook friendly (new impl)

WeichenXu123:
## Proposed API:

```
def run(
        func,
        args=(),
        kwargs={},
        np=1,
        hosts=None,
        hostfile=None,
        start_timeout=None,
        ssh_port=None,
        disable_cache=None,
        output_filename=None,
        verbose=None,
        use_gloo=None,
        use_mpi=None)
```
I launch a http server in driver side, so that:
* remote processes can read the exec function from the server
* remote process can put return value into the server.

I integrate the code into current `run` code.


horovod/run/run.py
```diff
@@ -484,12 +494,32 @@ def run():
...
+        raise RuntimeError(
+            'Cannot find an IPv4 address of driver.')
+
+    if exec_func:
+        run_func_server = KVStoreServer(verbose=settings.verbose)
+        run_func_server_port = run_func_server.start_server()
+        pickled_exec_func = cloudpickle.dumps(exec_func)
+        put_data_into_kvstore(driver_ip, run_func_server_port,
+                              'runfunc', 'func', pickled_exec_func)
+        settings.command = ['horovod_launch_exec_func', str(driver_ip), str(run_func_server_port)]
```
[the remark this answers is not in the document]
WeichenXu123: Good idea.

horovod/run/run.py
```diff
@@ -503,6 +533,45 @@ def run():
...
+        args=(),
+        kwargs={},
+        np=2,
+        hosts=None,
+        hostfile=None,
+        start_timeout=None,
+        ssh_port=None,
+        disable_cache=None,
+        verbose=None,
+        use_gloo=None):
```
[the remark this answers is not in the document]
WeichenXu123: Yes. This is troublesome.
I am considering two approach: 
1) keep this API simple and only provide basic arguments.
2) Providing a builder pattern API, like:
```
FuncRunnerBuilder().set(...).set(...).run_fun()
```
What do you think ?

WeichenXu123:
I update PR for addressing some comments.
I will add test later, but it is ready for first pass review.

WeichenXu123:
I update PR for addressing some comments.
I will add test later, but it is ready for first pass review.

horovod/run/run.py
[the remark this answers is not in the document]
tgaddair: My preference would be to have a structure to encapsulate these parameters.  That would make it easy to reuse them between function calls.

tgaddair:
Since the rendezvous server has been made more generic, I would move it out of the `horovod.run.rendezvous` package into either `horovod.run` or `horovod.run.http`.

WeichenXu123:
@tgaddair I add a structure `HorovodArgs` to encapsulate all the configs.

WeichenXu123:
@tgaddair Seemingly the buildkite system has some issue. Sometimes building failed.

tgaddair:
> @tgaddair Seemingly the buildkite system has some issue. Sometimes building failed.

Hey @WeichenXu123, can you rebase onto master?  There was a change made to `pytorch-nightly`.

WeichenXu123:
@alsrgv @tgaddair 
This PR is almost ready. Could you take a review ?
If the API is OK, then I will add doc and expose it to package `horovod.run` ?

Btw, we need to add `yaml` into the `requirement.txt` in the doc building machine. Otherwise building doc may raise error. Because in `horovod.run` package `yaml` is imported.

WeichenXu123:
@alsrgv There is one issue in test:
mpirun do not support recursive call, but current tests are all launched by
```
mpirun pytest /path/to/test_xxx.py
```
So that the `InteractiveRunTests` will raise error like "mpirun does not support recursive calls"

Maybe we need to modify the test script. For `InteractiveRunTests` we should launch it specially by
```
pytest /path/to/test_interactive_run.py
```

What do you think ?

WeichenXu123:
Ready for review.

I add test `test_interactiverun.py` and update the test script to:
* run `test_interactiverun.py` separately by command `pytest ... test_interactiverun.py`
* remove test against `mpich`. horovod do not support `mpich`.

About the API, should we add all the config args into the `run` function ?

@alsrgv @tgaddair Thanks!

WeichenXu123:
Gently ping @alsrgv @tgaddair
Ready for final review, thanks!

WeichenXu123:
Gently ping @alsrgv
Any updates ? Thanks !!

WeichenXu123:
Gently ping @alsrgv Thanks!

WeichenXu123:
Gently ping @karakusc @tgaddair @alsrgv 
Thanks!

tgaddair:
Hey @WeichenXu123, sorry for the very late reply.  I'll be taking a look this weekend and back to you by Monday at the latest.  Thanks for your patience.

.buildkite/gen-pipeline.sh
```diff
@@ -28,7 +28,6 @@ tests=( \
        test-cpu-openmpi-py3_6-tf2_0_0-keras2_2_4-torch1_2_0-mxnet1_5_0-pyspark2_4_0 \
        test-cpu-openmpi-py2_7-tfhead-kerashead-torchhead-mxnethead-pyspark2_4_0 \
        test-cpu-openmpi-py3_6-tfhead-kerashead-torchhead-mxnethead-pyspark2_4_0 \
-       test-cpu-mpich-py3_6-tf1_14_0-keras2_2_4-torch1_1_0-mxnet1_5_0-pyspark2_4_0 \
```
tgaddair: Please rebase as we've made some updates to `gen-pipeline.sh`.

docs/requirements.txt
```diff
@@ -2,3 +2,4 @@ git+https://github.com/sphinx-doc/sphinx@2.0
 sphinxcontrib-napoleon
 alabaster
 nbsphinx
+pyyaml
```
tgaddair: How is this being used?

horovod/run/common/util/settings.py
```diff
@@ -48,4 +47,9 @@ def __init__(self, verbose=0, ssh_port=None, key=None, timeout=None,
         self.num_proc = num_proc
         self.hosts = hosts
         self.output_filename = output_filename
+        self.command = None
+        self.run_func_mode = None
+
+    def set_command(self, command, run_func_mode):
```
tgaddair: I think it'd be cleaner to just decouple the command from the settings and pass it in as a separate parameter.  The `run_func_mode` property can be set in the constructor because you know its value from `args.run_func`.

horovod/run/http/http_client.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =============================================================================
+import sys
+import base64
+if sys.version < '3':
```
tgaddair: Replace with:

```
from distutils.version import LooseVersion
if LooseVersion(sys.version) < LooseVersion('3.0.0')
```

horovod/run/http/http_server.py
```diff
@@ -199,6 +201,53 @@ def listen_loop(self):
...
+
+        # Cache that provides the store
+        self.cache_lock = threading.Lock()
+        self.cache = {}
+
+        self.verbose = verbose
+
+
+class KVStoreServer:
+
```
tgaddair: Remove blank line.

horovod/run/run.py
```diff
@@ -709,14 +782,43 @@ def run():
...
+    driver_ip = _get_driver_ip(common_intfs)
+
+    if args.run_func:
+        run_func_server = KVStoreServer(verbose=settings.verbose)
+        run_func_server_port = run_func_server.start_server()
+        pickled_exec_func = cloudpickle.dumps(args.run_func)
+        put_data_into_kvstore(driver_ip, run_func_server_port,
+                              'runfunc', 'func', pickled_exec_func)
+
+        command = [sys.executable, "-m", "horovod.run.run_task", str(driver_ip), str(run_func_server_port)]
```
tgaddair: Nit: single quotes for consistency.

horovod/run/run_task.py
```diff
@@ -0,0 +1,36 @@
...
+
+def main(addr, port):
+    pickled_func = read_data_from_kvstore(addr, port, 'runfunc', 'func')
+    func = cloudpickle.loads(pickled_func)
+    try:
+        ret_val = func()
+    except BaseException as e:
+        sys.stderr.write("User function raise error: {error}".format(error=str(e)))
+        raise e
+    if ret_val:
```
tgaddair: There are a few issues here we should think about.

This will resolve to `False` for valid return values like `0`.  You could check `ret_val is None` instead, but in some cases `None` may also be a valid return value.  It's also possible that multiple ranks can return something, in which case this approach will result in all but one of them being thrown out.

What I suggest we do instead is to fetch the rank from the environment (you can see an example of how to do this in `test_tensorflow.py::test_horovod_rank`).  Then we can have every rank publish its `ret_val` (regardless of whether it is `None` or not) to the KV store with a key unique to its rank.  Back in driver, we can then iterate through the ranks and fetch every return value, adding them to a list we return to the caller.

What do you think?

test/test_interactiverun.py
```diff
@@ -0,0 +1,76 @@
...
+    def test_failed_run(self):
+
+        def fn():
+            hvd.init()
+            rank = hvd.rank()
+            if rank == 1:
+                raise RuntimeError()
+
+        try:
+            run(fn, np=2, use_gloo=True)
```
tgaddair: You can use:

```
with pytest.raises(RuntimeError, match='Gloo job detected that one or more processes exited'):
    run(fn, np=2, use_gloo=True)
```

WeichenXu123:
@tgaddair Thanks! will update PR soon.

horovod/run/run_task.py
WeichenXu123: Yes, I considered this way. But there is an issue: the `rank()` method should import `horovod.tensorflow` or ``horovod.pytorch` or some other backends. Here I don't know which backend user use. How to address this ? @tgaddair

docs/requirements.txt
WeichenXu123: In `horovod/run/run.py` there is `import yaml`
So before building doc we should install this dependency first, otherwise doc building will fail.

WeichenXu123:
@tgaddair I address most comments from you, only leave the one "ret value", we need to address the issue first which I comment above https://github.com/horovod/horovod/pull/1307#discussion_r335983591

horovod/run/run_task.py
tgaddair: You can pull the rank directly from environment variables without needing to import a framework.  I would suggest adding this method to `horovod/run/common/util/env.py`:

```
def get_env_rank_and_size():
    rank_env = ['HOROVOD_RANK', 'OMPI_COMM_WORLD_RANK', 'PMI_RANK']
    size_env = ['HOROVOD_SIZE', 'OMPI_COMM_WORLD_SIZE', 'PMI_SIZE']

    for rank_var, size_var in zip(rank_env, size_env):
        rank = os.environ.get(rank_var)
        size = os.environ.get(size_var)
        if rank is not None and size is not None:
            return int(rank), int(size)

    # Default to rank zero and size one if there are no environment variables
    return 0, 1
```

You can use that to get the rank and size directly from the environment.

.buildkite/gen-pipeline.sh
tgaddair: Why was this test removed?

WeichenXu123: @tgaddair 
Looks like horovod only support openmpi, do not support mpich (it cause test fail).
Also, env `OMPI_COMM_WORLD_RANK` is only supported in openmpi.

WeichenXu123: Oh, Spectrum MPI was added yesterday https://github.com/horovod/horovod/pull/1441
I doubt whether this version MPI can support `OMPI_COMM_WORLD_RANK`.

horovod/run/run_task.py
WeichenXu123: @tgaddair Spectrum MPI was added yesterday #1441
I doubt whether this version MPI can support OMPI_COMM_WORLD_RANK.

.buildkite/gen-pipeline.sh
WeichenXu123: One option is `run_func` only support opemMPI, what do you think ? @tgaddair

horovod/run/run_task.py
tgaddair: According to the [Spectrum MPI user guide](https://hpc.llnl.gov/sites/default/files/MPI-SpectrumUserGuide.pdf), you can access the rank using `OMPI_COMM_WORLD_RANK`, same as for Open MPI.  

@nvcastet can you confirm that this is correct (and that `OMPI_COMM_WORLD_SIZE` is the correct env variable for world size)?

.buildkite/gen-pipeline.sh
tgaddair: @WeichenXu123, unit tests using mpich are passing for HEAD.  If they're failing with your change, we should figure out why and fix them (some Horovod users use mpich).  

I don't see any reason why `run_func` should only support OpenMPI.  Using the function I provided above, it should work for OpenMPI, Spectrum, and mpich.  Please see my comment above regarding Spectrum support.

tgaddair: Looks like you may need to rebase to get the change for Spectrum MPI.

horovod/run/run_task.py
nvcastet: @tgaddair I confirm that `OMPI_COMM_WORLD_SIZE`, `OMPI_COMM_WORLD_LOCAL_SIZE `, `OMPI_COMM_WORLD_RANK `, and `OMPI_COMM_WORLD_LOCAL_RANK ` work with SMPI.

WeichenXu123:
@tgaddair I add back the mpich test, then when running `test_interactiverun.py`, it raise error like:
```
test_interactiverun.py:68:
  | _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  | /usr/local/lib/python3.6/dist-packages/horovod/run/run.py:922: in run
  | return _run(hargs)
  | /usr/local/lib/python3.6/dist-packages/horovod/run/run.py:805: in _run
  | _launch_job(args, remote_host_names, settings, common_intfs, command)
  | /usr/local/lib/python3.6/dist-packages/horovod/run/run.py:832: in _launch_job
  | mpi_run(settings, common_intfs, env, command)
  | _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  |  
  | settings = <horovod.run.common.util.settings.Settings object at 0x7f7b47467860>
  | common_intfs = {'lo'}
  | env = {'HOME': '/root', 'HOROVOD_CACHE_CAPACITY': '(None,)', 'HOROVOD_CYCLE_TIME': '(None,)', 'HOSTNAME': 'be96015d6083', ...}
  | command = ['/usr/bin/python', '-m', 'horovod.run.run_task', '127.0.0.1', '50171']
  |  
  | def mpi_run(settings, common_intfs, env, command):
  | mpi_impl_flags = _get_mpi_implementation_flags()
  | if mpi_impl_flags is None:
  | raise Exception(
  | >               'horovodrun convenience script does not find an installed OpenMPI.\n\n'
  | 'Choose one of:\n'
  | '1. Install Open MPI 4.0.0+ or IBM Spectrum MPI and re-install Horovod '
  | '(use --no-cache-dir pip option).\n'
  | '2. Run distributed '
  | 'training script using the standard way provided by your'
  | ' MPI distribution (usually mpirun, srun, or jsrun).\n'
  | '3. Use built-in gloo option (horovodrun --gloo ...).')
  | E           Exception: horovodrun convenience script does not find an installed OpenMPI.
  | E
  | E           Choose one of:
  | E           1. Install Open MPI 4.0.0+ or IBM Spectrum MPI and re-install Horovod (use --no-cache-dir pip option).
  | E           2. Run distributed training script using the standard way provided by your MPI distribution (usually mpirun, srun, or jsrun).
  | E           3. Use built-in gloo option (horovodrun --gloo ...).
  |  
  | /usr/local/lib/python3.6/dist-packages/horovod/run/mpi_run.py:65: Exception
  | ______________________ InteractiveRunTests.test_happy_run ______________________
```

I think the code here need to update
https://github.com/horovod/horovod/blob/36a98fe3b6ed2085011b4598de0433f982f6a176/horovod/run/mpi_run.py#L62

Current master only support OpenMPI and IBM spectrum MPI, and current master test do not cover the code path here.

WeichenXu123:
@tgaddair See my another PR test for MPICH (print log version), we can see MPICH do not support `--allow-run-as-root`
https://github.com/horovod/horovod/pull/1455
```
E               RuntimeError: mpirun failed with exit code 255, stdout
  | E
  | E               , stderr
  | E               [mpiexec@e099377ab812] match_arg (utils/args/args.c:159): unrecognized argument allow-run-as-root
  | E               [mpiexec@e099377ab812] HYDU_parse_array (utils/args/args.c:174): argument matching returned error
  | E               [mpiexec@e099377ab812] parse_args (ui/mpich/utils.c:1597): error parsing input array
  | E               [mpiexec@e099377ab812] HYD_uii_mpx_get_parameters (ui/mpich/utils.c:1649): unable to parse user arguments
  | E               [mpiexec@e099377ab812] main (ui/mpich/mpiexec.c:149): error parsing parameters
  |  
  | /usr/local/lib/python3.6/dist-packages/horovod/run/mpi_run.py:134: RuntimeError
 ```

**Comment by tgaddair** [comment]
Good catch @WeichenXu123.  Until we've done more testing with MPICH, how about we skip the interactive run test unless we're using OpenMPI similar to how we do for Spark?

```
# Seems that spark tests depend on MPI, do not test those when mpi is not available
  local exclude_spark_if_needed=""
  if [[ ${test} != *"mpi"* ]]; then
    exclude_spark_if_needed="| sed 's/[a-z_]*spark[a-z_.]*//g'"
  fi
```

**Comment by WeichenXu123** [comment]
Ready.
@tgaddair Before merging, let's wait @mengxr to take a final look.

File: horovod/run/run.py [hunk 11]
```diff
@@ -714,28 +789,140 @@ def run():
...
+def run_commandline():
+    args = parse_args()
+    args.run_func = None
+    _run(args)
+
+
+def run(
+        func,
+        args=(),
+        kwargs={},
```
mengxr: Do not use `{}` as the default value because it is mutable. Use `None` instead.

~~~python
>>> def test(a={}):
...   a["b"] = 1
...   print(a)

>>> test()

>>> help(test)

test(a={'b': 1})
~~~

horovod/run/run.py
```diff
@@ -714,28 +789,140 @@ def run():
...
+    :param func: The function to be run in Horovod job processes. The function return value will
+                 be collected as the corresponding Horovod process return value.
+    :param args: Arguments to pass to `func`.
+    :param kwargs: Keyword arguments to pass to `func`.
+    :param np: Number of Horovod processes.
+    :param hosts: List of host names and the number of available slots
+                  for running processes on each, of the form: <hostname>:<slots>
+                  (e.g.: host1:2,host2:4,host3:1 indicating 2 processes can run on host1,
+                  4 on host2, and 1 on host3). If not specified, defaults to using localhost:<np>
+    :param hostfile: Path to a host file containing the list of host names and the number of
```
mengxr: Do we allow both `hosts` and `hostfile`? If both are provided, what is the behavior?

horovod/run/run.py
```diff
@@ -714,28 +789,140 @@ def run():
...
+        ssh_port=None,
+        disable_cache=None,
+        output_filename=None,
+        verbose=None,
+        use_gloo=None,
+        use_mpi=None):
+    """
+    Launch a Horovod job to run the specified process function and get the return value.
+
+    :param func: The function to be run in Horovod job processes. The function return value will
```
mengxr: Could you mention that we assume `func` must be compatible with pickle?

horovod/run/run.py
```diff
@@ -591,6 +608,66 @@ def parse_args():
     return args
 
 
+class HorovodArgs(object):
```
mengxr: Shall we use `namedtuple`?

horovod/run/run.py
```diff
@@ -714,28 +789,140 @@ def run():
...
+    :return: Return a list which contains values return by all Horovod processes.
+             The index of the list corresponds to the rank of each Horovod process.
+    """
+
+    def wrapped_func():
+        return func(*args, **kwargs)
+
+    hargs = HorovodArgs()
+
+    hargs.np = np
```
mengxr: simpler if we use `namedtuple` to define `HorovodArgs`.

horovod/run/common/util/env.py
```diff
@@ -23,3 +24,18 @@
...
+
+
+def get_env_rank_and_size():
+    rank_env = ['HOROVOD_RANK', 'OMPI_COMM_WORLD_RANK', 'PMI_RANK']
+    size_env = ['HOROVOD_SIZE', 'OMPI_COMM_WORLD_SIZE', 'PMI_SIZE']
+
+    for rank_var, size_var in zip(rank_env, size_env):
+        rank = os.environ.get(rank_var)
+        size = os.environ.get(size_var)
+        if rank is not None and size is not None:
```
mengxr: if only one is defined, shall we throw an error?

horovod/run/http/http_client.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def read_data_from_kvstore(addr, port, scope, key):
+    try:
+        url = "http://{addr}:{port}/{scope}/{key}".format(
+            addr=addr, port=str(port), scope=scope, key=key
+        )
+        req = Request(url)
+        resp = urlopen(req)
+        return base64.b64decode(resp.read())
```
mengxr: base64 is not efficient. do we want to leave a TODO here to use binary stream?

horovod/run/run.py
```diff
@@ -265,6 +268,20 @@ def _driver_fn(all_host_names, local_host_names, settings):
         driver.shutdown()
 
 
+def _get_driver_ip(common_intfs):
```
mengxr: Could you document `common_intfs`?

horovod/run/run.py
```diff
@@ -265,6 +268,20 @@ def _driver_fn(all_host_names, local_host_names, settings):
...
+    iface = list(common_intfs)[0]
+    server_ip = None
```
mengxr: The method is called `_get_driver_ip` while the variable name is `server_ip` here. I think we should just call the method `get_local_ip` since that is what it does.

horovod/run/run.py
```diff
@@ -714,28 +789,140 @@ def run():
...
+        put_data_into_kvstore(driver_ip, run_func_server_port,
+                              'runfunc', 'func', pickled_exec_func)
+
+        command = [sys.executable, '-m', 'horovod.run.run_task', str(driver_ip), str(run_func_server_port)]
+
+        try:
+            _launch_job(args, remote_host_names, settings, common_intfs, command)
+            results = [None] * args.np
+            for i in range(args.np):
+                pickled_result = read_data_from_kvstore(driver_ip, run_func_server_port,
```
mengxr: Could you leave a TODO to make it parallel?

horovod/run/run.py
WeichenXu123: Do not allowed providing both. See `parse_args` method there's exclude rule checking.
But here I can add additional checking logic

horovod/run/run.py
WeichenXu123: Here there're many args I defined default value None in class `HorovodArgs` constructor. And in `run` we only set few arguments. If use  `namedtuple`, we need to copy the argments list twice completely, like:

```
HorovodArgs = namedtuple('arg1', 'arg2', ..., 'arg100')
hargs = HorovodArgs(arg1=XXX, arg2=XXX, ..., arg100=XXX)

```
but most of the args we only need to keep default value `None`.

horovod/run/common/util/env.py
WeichenXu123: Not sure. CC @tgaddair

horovod/run/run.py
WeichenXu123: This is not only local ip.
A machine could have multiple local ip, but not all local ip could be used here.
This method will get such ip on driver machine, which could make sure all worker could connect to this ip. So call it "driver ip" is more accurate.

horovod/run/run.py
tgaddair: Unfortunately, in the current design we cannot use `namedtuple` because the downstream `_run` function modifies some of the object's values (therefore, we cannot assume immutability).  We cannot change the behavior to use `namedtuple._replace` as it is currently designed because there command line entrypoint uses the same code path, but passes in parsed arguments object (here @WeichenXu123 is relying on Python's duck typing to get around this interface overloading).

Longterm, I think we should consolidate these two code paths (functional vs command line) to use the same object to represent materialized argument values.  But I don't think it's necessary to do that in this PR since it's an implementation detail not exposed to the end user.

PR #1311, opened 2019-08-17
Move allreduce to opt.apply_gradients when possible

alsrgv:
Motivation: new TensorFlow-Keras 2.0 [doesn't use](https://github.com/tensorflow/tensorflow/blob/498e815097e74aff7fefdbbae69ba9daf6e9c023/tensorflow/python/keras/engine/training_eager.py#L272) `Optimizer.get_gradients()` in eager mode.  However, it started using `Optimizer.apply_gradients()`, which was not the case with legacy Keras.

The proposal is to align gradient aggregation to happen during apply, similar to MXNet.  This way we will be able to deprecate `hvd.DistributedGradientTape()` and align user experience between Graph and Eager mode further.


alsrgv:
cc @kuroko1t

PR #1314, opened 2019-08-19
Limit # of CPU threads for PyTorch

alsrgv:
Default thread settings sometimes cause PyTorch to overuse CPU resources available on the server.


alsrgv:
Ignore unrelated test failures.

PR #1319, opened 2019-08-19
Add links to Horovod mailing lists

alsrgv:


sblotner:
I think it’s up to date instead of up-to-date in this case, but otherwise looks good

alsrgv:
Docs build is passing, merging.

PR #1320, opened 2019-08-20
Update PyTorch to 1.2.0

alsrgv:

PR #1323, opened 2019-08-20
Pin stricter versions for PyTorch 1.2.0 upgrade

alsrgv:
Previous upgrade #1320 did not take into account that PyTorch https://download.pytorch.org/whl/torch_stable.html is not usable for latest cuda10 releases, and that PyTorch silently ignores CUDA mismatch errors instead of loudly failing.


alsrgv:
See https://github.com/pytorch/pytorch/issues/23656 for more details.

Issue #1125, opened 2019-06-05
Overlapping NCCL and MPI


eric-haibin-lin:
When using hierarchical allreduce, I see that NCCL operations and MPI operations are not overlapped, leading to underutilization of network bandwidth. Is this by design? 

![Screen Shot 2019-04-11 at 11 49 58 PM](https://user-images.githubusercontent.com/5545640/58937201-9502b380-8726-11e9-811d-5ca22b76a0f8.png)

alsrgv:
@eric-haibin-lin, it is by design since MPI allreduce typically happens on CPU and requires memcopy before the data can be allreduced by NCCL.

eric-haibin-lin:
Yes, I meant overlapping cross tensors :)

alsrgv:
That would be very interesting to have :-)

alsrgv:
@eric-haibin-lin, any chance you could submit a PR for this?

eric-haibin-lin:
Unfortunately I'm currently busy with a few conference events and haven't had a chance to look into this yet. Could you give more pointers on which components in horovod are related to this?

alsrgv:
@eric-haibin-lin, you need to study up coordinator ([operations.cc](https://github.com/horovod/horovod/blob/master/horovod/common/operations.cc)) and [operation implementations](https://github.com/horovod/horovod/tree/master/horovod/common/ops).

@tgaddair recently published [Contributor Guide](https://horovod.readthedocs.io/en/latest/contributors_include.html) that should help to get started.

PR #1326, opened 2019-08-20
[WIP] Enable pipelining NCCL and MPI operations in hierarchical allreduce path

romerojosh:
This work in progress PR adds a feature requested in issue #1125. 

The current implementation allows the user to enable pipelining of the `MPI_Allreduce` and `NCCLAllGather` phases of the hierarchical allreduce scheme in fixed sized chunks. The chunk size used for the pipelining is user configurable via a new environment variable `HOROVOD_HIERARCHICAL_ALLREDUCE_CHUNK_SIZE_KB`, which is the desired chunk size in kilobytes. This variable defaults to 0, which disables the pipelining feature. We may need to explore alternative options to control the chunk sizing if performance requires it. 

To enable this PR, the original `NCCLAllGather` in the scheme has been replaced with a grouped set of `NCCLBroadcast` calls which emulate the functionality of an `allgatherv` type operation. This is required for the pipeling since the reduced chunks need to be gathered into displaced locations in the result buffer. Additionally, changing to this allgatherv like operation enables simplifying the handling of "remainder" portions of the buffer, as now the last local rank `local_size-1` can just process the remainder portion together with its non-remainder share of the data buffer without the need for a separate `NCCLBroadcast`. 

PR is currently a work in progress pending some performance testing.


romerojosh:
@eric-haibin-lin, would you be able to try this PR out and see if it improves your performance? I'd suggest trying a pipeline chunk size of 1024 KB to start (by setting `HOROVOD_HIERARCHICAL_ALLREDUCE_CHUNK_SIZE_KB=1024`).

PR #1345, opened 2019-08-23
Added command line arguments for Horovod knob environment variables, config file, and new knobs for autotuning

tgaddair:


DEKHTIARJonathan:
Some description for documentation with what it does, possible values and why it would be interesting to use these flags would be big plus IMHO

horovod/run/run.py
```diff
@@ -369,7 +450,15 @@ def parse_args():
...
 
-    return parser.parse_args()
+    args = parser.parse_args()
+
+    if args.param_file:
+        with open(args.param_file, 'r') as f:
+            config = yaml.load(f, Loader=yaml.FullLoader)
+        config_parser.set_args_from_config(args, config)
+    config_parser.validate_config_args(args)
+
```
abditag2: I think it makes more sense that the values passed in the command line overwrite the values in the config file. If we want that to be the case, we should parse the config file before parsing any of the command line args.

alsrgv: +1, let's change CLI args to take precedence.

tgaddair: Yes, I agree.  I was debating whether to go that route, or make those CLI args mutually exclusive with the config file.  I like the idea of CLI args taking precedence, but it's a bit tricky because the order of evaluation of args is dependent on the order the users provides them, but I'll keep exploring it.

Issue #1346, opened 2019-08-23
Support gradient clipping before & after allreduce


alsrgv:
**Is your feature request related to a problem? Please describe.**
Currently, each framework allows one (not always consistent) way to do gradient clipping. Generally, one could do gradient clipping locally (before allreduce) or globally (after allreduce).

In PyTorch, gradient clipping is done globally (after allreduce).

In MXNet and TensorFlow (since Horovod 0.17.0), gradient clipping is done locally (before allreduce).

**Describe the solution you'd like**
We should allow the user to specify gradient clipping and other gradient transformations both before and after allreduce.  This can easily be done with either a couple of Lambdas passed to `hvd.DistributedOptimizer()`, or by implementing a callback mechanism with the list of callbacks similarly passed to `hvd.DistributedOptimizer()`.

The latter approach has the benefit of being more generalizable and extensible in the future.

cc @DEKHTIARJonathan

PR #1316, opened 2019-08-19
Add *args to opt.apply_gradients()

alsrgv:
cc @DEKHTIARJonathan 
Fixes #1315

PR #1347, opened 2019-08-24
Revert hvd.DistributedOptimizer to allreduce in compute_gradients(), require graph mode

alsrgv:
Since landing #1311 and #1316 we have identified few issues with that proposal:
1. Inconsistency between PyTorch and TensorFlow in terms of handling of gradient clipping (see https://github.com/horovod/horovod/issues/1346).
2. Incompatibility with loss scaling in NVIDIA AMP and native TensorFlow loss scaling optimizer.

The idea is to revert the change and instead require graph mode in TF2-Keras.  Since this is another behavior change, we'll need to release 0.18.0.

cc @DEKHTIARJonathan

### DIFF
--- a/horovod/_keras/__init__.py
+++ b/horovod/_keras/__init__.py
@@ -40,6 +40,7 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
             In DistributedOptimizer, get_gradients() is overriden to also
             allreduce the gradients before returning them.
             """
+            self._fail_if_executing_eagerly()
             gradients = super(self.__class__, self).get_gradients(loss, params)
             if hvd.size() > 1:
                 averaged_gradients = []
@@ -59,6 +60,15 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
                     return averaged_gradients
             else:
                 return gradients
+
+        def apply_gradients(self, *args, **kwargs):
+            self._fail_if_executing_eagerly()
+            return super(self.__class__, self).apply_gradients(*args, **kwargs)
+
+        def _fail_if_executing_eagerly(self):
+            if hvd._executing_eagerly():
+                  raise Exception('hvd.DistributedOptimizer() does not support eager '
+                                'execution. Please specify `run_eagerly=False` in `fit()`.')
 
     # We dynamically create a new class that inherits from the optimizer that was passed in.
     # The goal is to override get_gradients() method with an allreduce implementation.
--- a/horovod/_keras/__init__.py
+++ b/horovod/_keras/__init__.py
@@ -29,6 +29,7 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
             self._device_sparse = device_sparse
             self._compression = compression
             self._sparse_as_dense = sparse_as_dense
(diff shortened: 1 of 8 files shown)


horovod/_keras/__init__.py
```diff
@@ -60,22 +40,42 @@ def get_gradients(self, loss, params):
...
+                                                     compression=self._compression)
+                            averaged_gradients.append(avg_grad)
+                        else:
+                            averaged_gradients.append(None)
+                    return averaged_gradients
             else:
                 return gradients
 
+        def apply_gradients(self, *args, **kwargs):
+            self._fail_if_executing_eagerly()
```
romerojosh: Is it possible to just have the eager failure occur during `__init__` instead of in both `apply_gradients` and `compute_gradients`?

horovod/tensorflow/__init__.py
```diff
@@ -243,23 +242,27 @@ def __init__(self, optimizer, name=None, use_locking=False, device_dense='',
...
 
-            See Optimizer.apply_gradients() for more info.
+            See Optimizer.compute_gradients() for more info.
 
-            In DistributedOptimizer, apply_gradients() is overriden to also
-            allreduce the gradients before applying them.
+            In DistributedOptimizer, compute_gradients() is overriden to also
+            allreduce the gradients before returning them.
             """
+            self._fail_if_executing_eagerly()
```
romerojosh: Same here. Is it possible to just have the eager failure occur during `__init__` instead of in both `apply_gradients` and `compute_gradients`?

horovod/_keras/__init__.py
```diff
@@ -60,22 +40,42 @@ def get_gradients(self, loss, params):
...
+            return super(self.__class__, self).apply_gradients(*args, **kwargs)
+
+        def _fail_if_executing_eagerly(self):
+            if hvd._executing_eagerly():
```
tgaddair: Does this behavior differ between TF 2.0 and older versions?

horovod/_keras/__init__.py
tgaddair: My understanding is that at the time the optimizer is created it's fine to be running eagerly, only during fit you need to be in graph mode.

alsrgv: Yeah, that's exactly right.

horovod/_keras/__init__.py
alsrgv: In order versions you need to opt in to eager mode. @romerojosh has a good point that eager mode is slower than graph mode, so if somebody is throwing a lot of resources at a problem, they should switch to graph mode first.

horovod/_keras/__init__.py
romerojosh: I figured that might be the case but wanted to check. Thanks!

alsrgv:
@romerojosh, @tgaddair, I had to make a fix. `run_eagerly` was insufficient because it was off by default.  TF 2.0 uses `tf.function` to run `fit()` which runs the actual training in graph mode, but it's still using GradientTape to compute the gradient.  I've modified the example to disable that behavior and the optimizer to detect if get_gradient() was never called.

PR #1376, opened 2019-09-03
Make gloo job raise error when any task failed

WeichenXu123:
Gloo job should raise error if any task failed. But current master code if some task failed, gloo job will still exit normally.

I add checking code and raise error if needed.

(This keep similar behavior with `mpirun`, which will exit with non-zero value if any task failed)

### Test
We could manually test this by:
```
horovodrun -np 1 --gloo python -c "raise RuntimeError('error1')"
echo $?
```

**Before**
`horovodrun` exit with 0

**After**
`horovodrun` exit with 1, and raise Runtime error "Run gloo job failed."

### DIFF
--- a/horovod/run/gloo_run.py
+++ b/horovod/run/gloo_run.py
@@ -298,3 +298,4 @@ def gloo_run(settings, remote_host_names, common_intfs, env):
 
     _launch_jobs(settings, env, host_alloc_plan, remote_host_names, run_command)
     return
+
--- a/horovod/run/gloo_run.py
+++ b/horovod/run/gloo_run.py
@@ -298,4 +298,3 @@ def gloo_run(settings, remote_host_names, common_intfs, env):
 
     _launch_jobs(settings, env, host_alloc_plan, remote_host_names, run_command)
     return
-
--- a/horovod/run/gloo_run.py
+++ b/horovod/run/gloo_run.py
@@ -250,9 +250,12 @@ def _launch_jobs(settings, env, host_alloc_plan, remote_host_names, _run_command
     res = threads.execute_function_multithreaded(_exec_command,
                                                  args_list,
                                                  block_until_all_done=True)
-    for k in res:
-        if res[k] != 0:
-            raise RuntimeError("Run gloo job failed.")
+    for name in res:
+        if res[name] != 0:
+            raise RuntimeError("Gloo job detected that one or more processes exited with non-zero "
+                               "status, thus causing the job to be terminated. The first process "
+                               "to do so was:\nProcess name: {name}\nExit code: {code}\n"
+                               .format(name=name, code=res[name]))
 
 
 def gloo_run(settings, remote_host_names, common_intfs, env):
--- a/horovod/run/gloo_run.py
+++ b/horovod/run/gloo_run.py
@@ -247,9 +247,12 @@ def _launch_jobs(settings, env, host_alloc_plan, remote_host_names, _run_command
     # a SIGINT, the event will be set and the spawned threads will kill their
     # corresponding middleman processes and thus the jobs will be killed as
     # well.
-    threads.execute_function_multithreaded(_exec_command,
-                                           args_list,
-                                           block_until_all_done=True)
+    res = threads.execute_function_multithreaded(_exec_command,
+                                                 args_list,
+                                                 block_until_all_done=True)
+    for k in res:
+        if res[k] != 0:
+            raise RuntimeError("Run gloo job failed.")
 
 
 def gloo_run(settings, remote_host_names, common_intfs, env):
--- a/horovod/run/gloo_run.py
+++ b/horovod/run/gloo_run.py
@@ -20,6 +20,7 @@ import os
 import signal
 import sys
 import threading
+import time
 
 from psutil import net_if_addrs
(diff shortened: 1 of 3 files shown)


WeichenXu123:
Gently ping @tgaddair

test/test_common.py
```diff
@@ -43,7 +43,7 @@ def test_gloo_built(self):
         """Test that Gloo has been built if env is set."""
         gloo_rank = int(os.getenv('HOROVOD_RANK', -1))
         if gloo_rank >= 0:
-            self.assertEqual(gloo_built())
+            self.assertTrue(gloo_built())
```
WeichenXu123: The old test is invalid:
`self.assertEqual` has 3 arguments so it will raise error, but launch gloo job will always return zero so the old test is always passed.

horovod/run/gloo_run.py
```diff
@@ -190,9 +192,12 @@ def set_event_on_sigterm(signum, frame):
...
     # well.
-    threads.execute_function_multithreaded(_exec_command,
-                                           args_list,
-                                           block_until_all_done=True)
+    res = threads.execute_function_multithreaded(_exec_command,
+                                                 args_list,
+                                                 block_until_all_done=True)
+    for k in res:
+        if res[k] != 0:
+            raise RuntimeError("Run gloo job failed.")
```
abditag2: Can we print out the number of ranks that failed?

tgaddair: In most cases all the ranks will fail, as one failure leads to a cascading failure of the whole system.  To be consistent with MPI, we should print the first rank to fail.  Example from `mpirun`:

```
--------------------------------------------------------------------------
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[38080,1],1]
  Exit code:    1
--------------------------------------------------------------------------
```

I don't think there's an elegant way to do this will `threads.execute_function_multithreaded`, however, so that will require a larger refactor.  I think this solution is a good stopgap.

horovod/run/gloo_run.py
```diff
@@ -136,9 +136,11 @@ def _exec_command(_command, _index, event_):
             exit_code = safe_shell_exec.execute(_command, index=_index, event=event_)
             if exit_code != 0:
                 print('Process {idx} exit with status code {ec}.'.format(idx=_index, ec=exit_code))
-        except Exception as e:
+                return 1
```
tgaddair: Let's return the failing exist code here.

horovod/run/gloo_run.py
tgaddair: I do think it would be good to print the error code, however.

horovod/run/gloo_run.py
```diff
@@ -190,9 +192,15 @@ def set_event_on_sigterm(signum, frame):
...
     # well.
-    threads.execute_function_multithreaded(_exec_command,
-                                           args_list,
-                                           block_until_all_done=True)
+    res = threads.execute_function_multithreaded(_exec_command,
+                                                 args_list,
+                                                 block_until_all_done=True)
+    for name in res:
+        if res[name] != 0:
+            raise RuntimeError("Gloo job detected that one or more processes exited with non-zero "
```
tgaddair: Because `res` is an unordered dict, we cannot know the order that the processes terminated, so it's not accurate to say that the first process to exit was <name>.  So, here's what I'd suggest doing instead:

In `_exec_command`, instead of returning a single integer exit code, try returning a tuple: `(exit_code, time.time())`.  That will give you a floating-point timestamp, which you can then use to sort the dictionary like so:

```
def get_timestamp(item):
    name, value = item
    exit_code, timestamp = value
    return timestamp

for name, value in sorted(res.items(), key=get_timestamp):
    exit_code, timestamp = value
    ...
```

horovod/run/gloo_run.py
```diff
@@ -190,9 +192,15 @@ def set_event_on_sigterm(signum, frame):
...
+                               "status, thus causing the job to be terminated. The first process "
```
tgaddair: Please replace double-quotes with single-quotes for consistency.

horovod/run/gloo_run.py
```diff
@@ -136,9 +136,11 @@ def _exec_command(_command, _index, event_):
             exit_code = safe_shell_exec.execute(_command, index=_index, event=event_)
             if exit_code != 0:
                 print('Process {idx} exit with status code {ec}.'.format(idx=_index, ec=exit_code))
-        except Exception as e:
+                return exit_code
```
tgaddair: Following my suggestion from below, we can simplify this to just return once:

```
try:
            exit_code = safe_shell_exec.execute(command, index=index, event=event,
                                                stdout=output_file, stderr=output_file)
            if exit_code != 0:
                print('Process {idx} exit with status code {ec}.'.format(idx=index, ec=exit_code))
except BaseException as e:
            print('Exception happened during safe_shell_exec, exception '
                  'message: {message}'.format(message=e))
            exit_code = 1
return (exit_code, time.time())
```

WeichenXu123:
@tgaddair I merge master and addressed comments. Looks like test failed on some unrelated cases. Does master code change something ?

PR #1391, opened 2019-09-09
Fix error check when initializing Horovod with existing mpi4py communicator

romerojosh:
This error check in `init()` when attempting to initialize Horovod with an existing mpi4py communicator breaks this functionality completely. The issue is that the flag indicating that MPI is enabled is set in within `horovod_init_comm/InitializeHorovodOnce`, the function being blocked by the error check.

### DIFF
--- a/horovod/common/basics.py
+++ b/horovod/common/basics.py
@@ -40,10 +40,11 @@ class HorovodBasics(object):
         atexit.register(self.shutdown)
 
         if not isinstance(comm, list):
-            mpi_enabled = self.MPI_LIB_CTYPES.horovod_mpi_enabled()
-            if not bool(mpi_enabled):
+            mpi_built = self.MPI_LIB_CTYPES.horovod_mpi_built()
+            if not bool(mpi_built):
                 raise ValueError(
-                    'Horovod MPI is not enabled; Please make sure it\'s installed and enabled.')
+                    "Horovod has not been built with MPI support. Ensure MPI is installed and "
+                    "reinstall Horovod with HOROVOD_WITH_MPI=1 to debug the build error.")
 
             from mpi4py import MPI
             if MPI._sizeof(MPI.Comm) == ctypes.sizeof(ctypes.c_int):


tgaddair:
Good catch @romerojosh!  Instead of removing this guard completely, what about changing it to `horovod_mpi_built()`, which only checks that Horovod was installed with MPI support?  Then we could change the error message to something like:

```
Horovod has not been built with MPI support.  Ensure MPI is installed and reinstall Horovod with HOROVOD_WITH_MPI=1 to debug the build error.
```

I think this would be useful diagnostic info since we can now fallback to using Gloo if MPI failed to install correctly.

romerojosh:
Makes sense. Do you think a similar change should also be made for the error check in `mpi_threads_supported`?

tgaddair:
That's a good question.  The problem with changing it to check `mpi_built()` is that `mpi_threads_supported()` will blindly perform a dynamic cast on the controller:

```
auto mpiController =
      std::dynamic_pointer_cast<MPIController>(horovod_global.controller);
  return mpiController->IsMpiThreadsSupported() ? 1 : 0;
```

... which can fail if MPI is installed, but Gloo is being used as the controller.  

Checking `mpi_context.Enabled()` takes care of most of the problem, but doesn't account for one fringe edge case: where you're using MPI for CPU operations (thus `mpi_context` is enabled), but Gloo for the controller.  I don't think anyone is likely to do this in practice (other than benchmarking purposes), but since it's possible, I'll make a fix for that.

But I think in general, checking `mpi_enabled()` is correct for `mpi_threads_supported()` for that reason.

romerojosh:
Sounds good to me @tgaddair! Pushed a change modifying rather than removing the existing error check.

romerojosh:
Looks like there are some failing tests unrelated to this change.

tgaddair:
Seems to be related to a change in Keras nightly (#1392).

tgaddair:
@romerojosh Landed #1392, can you rebase and push again?

Issue #1409, opened 2019-09-24
SegFault when using horovod.mxnet.allgather


yjxiong:
**Environment:**
1. Framework: MXNet
2. Framework version: 1.5.0
3. Horovod version: 0.18.1
4. MPI version: 4.0.1
5. CUDA version: 10.0
6. NCCL version: 2.3.7
7. Python version: 3.6.8
8. OS and version: Linux 4.14.138-89.102.amzn1.x86_64 (EC2 instance on AWS)
9. GCC version: 4.8.5

**Checklist:**
1. Did you search issues to find if somebody asked this question before? Yes
2. If your question is about hang, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/running.rst)? N/A
3. If your question is about docker, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/docker.rst)? N/A
4. Did you check if you question is answered in the [troubleshooting guide](https://github.com/horovod/horovod/blob/master/docs/troubleshooting.rst)? N/A

**Bug report:**
Please describe errorneous behavior you're observing and steps to reproduce it.

The function `horovod.mxnet.allgather` will lead to `Segmentation fault: 11` when I use more than 1 worker. The minimal code block to reproduce the error is as below:
```
import horovod.mxnet as hvd
import mxnet as mx

hvd.init()

ctx = mx.gpu(hvd.local_rank())

v = mx.nd.ones((100, 50), ctx=ctx)

r = hvd.allgather(v)

print(r.shape)
print(r)
```

Running it with the command `horovodrun -np 2 -H localhost:2 python3 test_allgather.py`  will result in stack traces pasted below:

```
[1,1]<stderr>:
[1,1]<stderr>:Segmentation fault: 11
[1,1]<stderr>:
[1,0]<stderr>:
[1,0]<stderr>:Segmentation fault: 11
[1,0]<stderr>:
[1,1]<stderr>:Stack trace:
[1,1]<stderr>:  [bt] (0) /usr/local/lib/python3.6/site-packages/mxnet/libmxnet.so(+0x2e6b160) [0x7f5c59089160]
[1,1]<stderr>:  [bt] (1) /lib64/libc.so.6(+0x362f0) [0x7f5d3593b2f0]
[...]

PR #1424, opened 2019-09-28
Fix segment fault in MXNet integration

yjxiong:
This PR solves the segment fault in: https://github.com/horovod/horovod/issues/1409

Basically, AllGather is a function that will allocate more space in the output than the size of input. The previous implementation falsely used the same address space for input and output and did not allocate enough space for output. If we want to use IN_PLACE operation, the MPI standard requires that the send data block to be sitting in the same location as if it is in the AllGather result. 

Currenlty there is a blocking call at the end of the MXNet API calls. Otherwise there will be data corruption. The previous implementation of Allreduce also has this risk when HOROVOD_GPU_REDUCE=MPI.

### DIFF
--- a/horovod/common/ops/mpi_operations.cc
+++ b/horovod/common/ops/mpi_operations.cc
@@ -15,6 +15,7 @@
 // =============================================================================
 
 #include "mpi_operations.h"
+#include "../logging.h"
 
 namespace horovod {
 namespace common {
@@ -28,6 +29,8 @@ Status MPIAllreduce::Execute(std::vector<TensorTableEntry>& entries, const Respo
   void* buffer_data;
   size_t buffer_len;
   int64_t num_elements = NumElements(entries);
+
+  
 
   // Copy memory into the fusion buffer.
   auto& timeline = global_state_->timeline;
@@ -54,6 +57,7 @@ Status MPIAllreduce::Execute(std::vector<TensorTableEntry>& entries, const Respo
     throw std::runtime_error("MPI_Allreduce failed, see MPI output for details.");
   }
   timeline.ActivityEndAll(entries);
+  
 
   // Copy memory out of the fusion buffer.
   if (entries.size() > 1) {
@@ -110,12 +114,12 @@ Status MPIAllgather::Execute(std::vector<TensorTableEntry>& entries, const Respo
 
   SetDisplacements(recvcounts, displcmnts);
   SetEntryComponentOffsets(entries, entry_component_sizes, recvcounts, entry_component_offsets);
-
   int element_size = mpi_context_->GetMPITypeSize(first_entry.tensor->dtype());
 
   const void* sendbuf = nullptr;
   void* buffer_data;
   int64_t total_num_elements = NumElements(entries);
+  LOG(TRACE)<<"total send amount: "<<total_num_elements;
 
   if (entries.size() > 1) {
     timeline.ActivityStartAll(entries, MEMCPY_IN_FUSION_BUFFER);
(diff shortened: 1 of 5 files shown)


horovod/common/ops/mpi_operations.cc
```diff
@@ -29,6 +30,8 @@ Status MPIAllreduce::Execute(std::vector<TensorTableEntry>& entries, const Respo
   size_t buffer_len;
   int64_t num_elements = NumElements(entries);
 
+
```
apeforest: please remove unnecessary blank lines

horovod/mxnet/mpi_ops.cc
```diff
@@ -138,6 +140,9 @@ inline void PushHorovodOperation(OperationType op_type, NDArray* input,
                       &MX_EXEC_CTX, nullptr, 0, &output_var, 1,
                       &MX_FUNC_PROP, priority, op_type_name);
   }
+
+  // Preventing exiting the API call prematurely and causing shape issue.
+  output->WaitToRead();
```
apeforest: The async actions are handled by MXNet engine. No need to synchronize here.

horovod/mxnet/mpi_ops.cc
```diff
@@ -185,24 +192,49 @@ void DoHorovodOperationCudaOnCPU(void*, void* on_complete_ptr, void* param) {
...
+                    &MX_FUNC_PROP, priority, op_type_name, true);
 
   // Make async copy of CPU tensor to output tensor.
-  TensorUtil::AsyncCopyCPUToCuda(hvd_cpu_buffer->tensor(), output);
+  if (op_type == OperationType::ALLGATHER){
+    hvd_output_buffer->tensor()->WaitToRead();
+    // Preventing exiting the API call prematurely and causing shape issue.
+    TensorUtil::AsyncCopyCPUToCuda(hvd_output_buffer->tensor(), output);
+  }else{
+    hvd_cpu_buffer->tensor()->WaitToRead();
```
apeforest: The async actions are handled by MXNet engine. No need to synchronize here.

horovod/mxnet/tensor_util.cc
```diff
@@ -137,13 +138,14 @@ void TensorUtil::Free(NDArray* tensor) { delete tensor; }
 
 // Resize tensor to nDimension with length size[i] in dimension i
 void TensorUtil::ResizeNd(NDArray* tensor, int nDimension, int64_t* size) {
-  void* temp_out;
-  MXNDArrayReshape64(tensor, nDimension, size, false, &temp_out);
-  tensor = static_cast<NDArray*>(temp_out);
+  mxnet::Tuple<dim_t> shape(size, size+nDimension);
+  tensor->ReshapeAndAlloc(shape);
```
apeforest: Can we still use the C API MXNDArrayReshape64 here instead of calling the lower level method inside NDArray?

yjxiong: After discussion, `MXNDArrayReshape64` is not a fit here. We should find a correct C API function for reshaping with a different size.

horovod/mxnet/mpi_ops.cc
yjxiong: We may need to add a blocking wait for the Horovod operation in the pushed MXNet operator to remove this wait.

yjxiong:
Hi @eric-haibin-lin 

I did not manage to make the AllGather without blocking wait work so I am not able to tell the difference in performance. If the performance is a concern with other Ops, we can limit that block to only the Allgather op.

yjxiong:
Some updates here:

- I have verified that for GPU training the blocking seems unnecessary in allreduce for gradients when NCCL is used. 

- The allgather is used in the forward for my training, where if the wait is removed it will segfault for either NCCL or MPI based allgather. 

So for now we only need to have the blocking wait for the Allgather op. It does not affect the performance for training jobs which are not using Allgather. I will update the CR soon.

yuxihu:
> So for now we only need to have the blocking wait for the Allgather op. It does not affect the performance for training jobs which are not using Allgather

why not adding the wait_to_read in your python code whenever you use allgather? It is not needed for allreduce.

yjxiong:
Because the segfault happens inside the allgather call. This is a BUG FIX. Not something I can cop with in application code.

yjxiong:
Close with https://github.com/horovod/horovod/pull/2092

PR #1429, opened 2019-10-04
Pin MXNet nightly

yuxihu:
Recently we compressed the MXNet pip package to meet PyPI package size limit. This causes undefined symbol errors. Pin MXNet nightly to a specific version while we are resolving the issue.


alsrgv:
LGTM, thanks!

Issue #1164, opened 2019-06-21
Why only openmpi is supported with horovodrun?


denfromufa:
Why only openmpi is supported with horovodrun? For example why not also support Spectrum MPI which is based on openmpi and should be mostly compatible.

```
(powerai.1.6.1) [@apps]$ horovodrun -np 4 python keras_mnist.py
Open MPI not found in output of mpirun --version.
Traceback (most recent call last):
  File "/data/gpfs/Users/j05/anaconda3/envs/powerai.1.6.1/bin/horovodrun", line 21, in <module>
    run.run()
  File "/data/gpfs/Users/j05/anaconda3/envs/powerai.1.6.1/lib/python3.6/site-packages/horovod/run/run.py", line 448, in run
    'horovodrun convenience script currently only supports '
Exception: horovodrun convenience script currently only supports Open MPI.

Choose one of:
1. Install Open MPI 4.0.0+ and re-install Horovod (use --no-cache-dir pip option).
2. Run distributed training script using the standard way provided by your MPI distribution (usually mpirun, srun, or jsrun).

```

alsrgv:
@denfromufa, would you be willing to send a PR for and maintain Spectrum MPI support?

denfromufa:
@alsrgv I read this document below and although smpi is based on openmpi and emulates the core APIs, still I think there are some differences in arguments used by Horovod. For example, the way openib & btl are specified.  Anyway I would probably be NOT the right person to support this. Maybe @nvcastet who merged the DDL support to Horovod :)

https://github.com/horovod/horovod/blob/master/docs/mpirun.rst

The use case is for benchmarks and for switching between openmpi and smpi on various machines or environments, without worrying about the differences in mpirun arguments that would be too overwhelming to end-users.

alsrgv:
I think it’d be great to have Spectrum MPI support. @nvcastet are you up for adding support to horovodrun and horovod.spark.run (for completeness)?

nvcastet:
@alsrgv @denfromufa Thank you for bring that up. I will talk with my team about it to see what we can do.

tgaddair:
Fixed by #1441 for Spectrum MPI.  Gloo is also now supported by #1181.

shailensobhee:
What about Intel MPI?

Output of `horovodrun`:

```
[...]
Extension tensorflow was built with MPI.
Open MPI/Spectrum MPI/MPICH not found in output of mpirun --version.
[...]
```

```
mpirun -V
[...]

PR #1441, opened 2019-10-11
Add Spectrum MPI support to horovodrun

nvcastet:
* Also, add '--mpi-args' flag to pass extra arguments to MPI.
In the case of LSF clusters, the slot numbers in the '-H' arguments can be
overwritten by the LSF MPI config. So to run on 2 boxes with 6 GPUs each, you
need to use: horovodrun --start-timeout 180 -np 12 --mpi-args="--map-by ppr:6:node" -H host1:6,host2:6 python mnist.py

### DIFF
--- a/horovod/run/mpi_run.py
+++ b/horovod/run/mpi_run.py
@@ -18,7 +18,15 @@ import six
 import traceback
 import sys
 import os
+from enum import Enum
 from horovod.run.common.util import env as env_util, safe_shell_exec, secret, codec
+
+# Supported MPI implementations
+_MPI_IMPLS = Enum('MPI_IMPLS', 'OMPI SMPI')
+# Open MPI Flags
+_OMPI_FLAGS = ['-mca pml ob1', '-mca btl ^openib']
+# Spectrum MPI Flags
+_SMPI_FLAGS = ['-gpu', '-disable_gdr']
 
 try:
     from shlex import quote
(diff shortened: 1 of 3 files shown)


nvcastet:
Implement #1164. Tested on Summit.
@alsrgv

horovod/run/run.py
```diff
@@ -389,6 +389,11 @@ def parse_args():
     parser.add_argument('-p', '--ssh-port', action='store', dest='ssh_port',
                         type=int, help='SSH port on all the hosts.')
 
+    parser.add_argument('--mpi-args', action='store', dest='mpi_args',
```
tgaddair: Let's put this in the 'library arguments' group (see `--mpi-threads-disable`).

horovod/run/mpi_run.py
```diff
@@ -76,26 +86,39 @@ def mpi_run(settings, common_intfs, env):
...
         common_intfs=','.join(common_intfs)) if common_intfs else ''
 
+    # Pick MPI implementation-specific flags
+    if mpi_impl == _MPI_IMPLS.SMPI:
+        mpi_args = _SMPI_FLAGS
+    else:
+        mpi_args = _OMPI_FLAGS
+
+    # On large cluster runs (e.g. Summit), we need extra settings to work around OpenMPI issues
+    if settings.num_hosts >= 64:
```
tgaddair: This feels like it should be a configurable arg to `horovodrun`.  Maybe `--large-cluster-hosts-threshold` with a default of 64?

nvcastet: I am hopping some of it get resolved in coming OpenMPI versions. We have seen those issues reported on 64 nodes or larger runs but never under that (related to the total number of daemons MPI needs to start on the boxes). So, it is a bootstrapping issue of MPI.
@tgaddair What is the advantage to have an argument for the user to pass it in? I am afraid if it is not useful, it might just be confusing.

horovod/run/mpi_run.py
```diff
@@ -35,29 +43,31 @@ def _is_open_mpi_installed():
...
     else:
         print("Was not able to run %s:\n%s" % (command, output_msg),
               file=sys.stderr)
-        return False
+        return False, None
 
 
 def mpi_run(settings, common_intfs, env):
-    if not _is_open_mpi_installed():
+    mpi_installed, mpi_impl = _is_open_mpi_installed()
```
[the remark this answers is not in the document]
nvcastet: Good points!

horovod/run/mpi_run.py
nvcastet: To be clear, those flags only impact the bootstrapping of the MPI processes but there is no impact on the processes once they run.

tgaddair: I see, thanks for clarifying.  My concern was that 64 seemed somewhat empirical and possibly implementation specific, and I wouldn't want a user to run into unexpected behavior when they scale up to 64 nodes.  But if it's a known issue with MPI that specifically affects jobs with greater than 64 nodes, then I think it's fine to hard-code it.

horovod/run/mpi_run.py
```diff
@@ -18,15 +18,21 @@
 import traceback
 import sys
 import os
+from enum import Enum
```
tgaddair: Looks like unit tests are failing due to this import (requires adding `enum34` to `setup.py`), but since we don't seem to be using this anymore, should be okay to just remove it.

PR #1443, opened 2019-10-11
Fix shell syntax.

rb-determined-ai:
Without this change, users who have `/bin/sh` as their default shell will see the following error:

`sh: 1: Syntax error: Bad fd number`


rb-determined-ai:
@tgaddair Should be good!

tgaddair:
Looks good, thanks again!

Issue #1123, opened 2019-06-05
TF 1.14 crashes with hwloc


byronyi:
**Environment:**
1. Framework: TensorFlow
2. Framework version: 1.14.0-rc0
3. Horovod version: 0.16.2
4. MPI version: OpenMPI 2.0.2
5. CUDA version: 10.0
6. NCCL version: 2.4.2
7. Python version: 3.5.3
8. OS and version: Debian 9.8

**Bug report:**
Please describe erroneous behavior you're observing and steps to reproduce it.

Use the in-stock [examples/tensorflow_mnist_estimator.py](https://github.com/horovod/horovod/blob/master/examples/tensorflow_mnist_estimator.py) and run with the following command:

```
mpirun --allow-run-as-root -n 1 --npernode 1 -bind-to none -hostfile /opt/tiger/hostfile --mca orte_tmpdir_base /opt/tiger/openmpi_tmp -x NCCL_DEBUG=INFO -x HOROVOD_WITH_TENSORFLOW -x JAVA_HOME -x PYPI -x HOROVOD_GPU_ALLREDUCE -x HTTP_PROXY -x LD_LIBRARY_PATH -x HTTPS_PROXY -x TENSORPACK -x PATH -x TENSORPACK_DATASET -x DATA --mca btl '^openib,smcuda' --mca btl_tcp_if_exclude docker0,lo -x NCCL_IB_DISABLE=1 -x NCCL_SOCKET_IFNAME=eth0 gdb --args python tensorflow_mnist_estimator.py
```

And it triggers segfault with the following backtrace:

```
#0  0x00007fff2dd36224 in hwloc_bitmap_and ()
   from /usr/local/lib/python3.5/dist-packages/tensorflow/python/../libtensorflow_framework.so.1
#1  0x00007fffd1019099 in opal_hwloc_base_filter_cpus ()
   from /usr/lib/x86_64-linux-gnu/libopen-pal.so.20
#2  0x00007fffcc05190e in ?? ()
   from /usr/lib/x86_64-linux-gnu/openmpi/lib/openmpi/mca_ess_pmi.so
#3  0x00007fffd545caa5 in orte_init ()
   from /usr/lib/x86_64-linux-gnu/libopen-rte.so.20
#4  0x00007fffd5957c4b in ompi_mpi_init ()
   from /usr/lib/x86_64-linux-gnu/libmpi.so.20
#5  0x00007fffd597a179 in PMPI_Init_thread ()
[...]

PR #1450, opened 2019-10-16
add tf1.4 hwloc issue to troubleshooting.rst

gibiansky:
I recently got bitten by [this issue](https://github.com/horovod/horovod/issues/1123) and it would be great if it were mentioned somewhere, since it seems likely to cause issues for many people using TF 1.14 and 1.15 from PyPi.

This PR adds a section to troubleshooting.rst that mentions this issue. I would have saved a bunch of time on this had this been there. 

If you prefer to write this somewhere else or in a different manner, feel free to close the PR and do it however is best. I just figured that since I was going to complain about this being "undocumented" on the issue, the least I can do is also make a PR :)

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -360,6 +360,27 @@ After ``hwloc`` is purged, `re-install Open MPI <https://www.open-mpi.org/faq/?c
 
 See `this issue <https://github.com/open-mpi/ompi/issues/4437>`__ for more details.
 
+segmentation fault with tensorflow 1.14 or higher mentioning `hwloc`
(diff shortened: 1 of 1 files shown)


gibiansky:
Signed the DCO and rebased on master.

tgaddair:
Hey @gibiansky, the build failed due to our Linter:

```
Warning, treated as error:
--
  | /workdir/docs/troubleshooting.rst:364:Title underline too short.
  |  
  | segmentation fault with tensorflow 1.14 or higher mentioning `hwloc`
  | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

Can you pad the underline to fit the length of the header?

PR #1463, opened 2019-10-22
Mention HOROVOD_MIXED_INSTALL for testing.

kit1980:


docs/contributors.rst
```diff
@@ -71,6 +71,8 @@ Horovod has unit tests for all frameworks you can run from the tests directory:
 
 **Note:** You will need PySpark and Java to run the Spark tests.
 
+**Note:** For Horovod compiled without specifiyng `HOROVOD_GPU_ALLREDUCE`, you will need to set environment variable `HOROVOD_MIXED_INSTALL=1` to prevent some tests from failing.
```
tgaddair: Good point!  That one other thing I'd point out here is that this only applies if your training framework was built with CUDA support, but Horovod was not.

PR #1465, opened 2019-10-23
support fp16 with ibm ddl

agemagician:
Hello,

IBM DDL support half precision training. I have add two lines to make it work with Horovod.


tgaddair:
Thanks for the PR @agemagician, could you sign the DCO (see below)?

PR #1476, opened 2019-10-28
Fixed start time set in Timeline to ensure timestamps are given relative to initialization of the TimelineWriter

### LINKED ISSUE #1472: How to convert "ts" field in timeline trace file to standard datetime?
**Environment:**
1. Framework: TensorFlow
2. Framework version: 1.14
3. Horovod version: 0.18.2
4. MPI version:
5. CUDA version:
6. NCCL version:
7. Python version:
8. OS and version:
9. GCC version:

**Your question:**
Hi there,
I am checking the timeline-file, following is part of the log
```
{"ph": "B", "name": "NEGOTIATE_ALLREDUCE", "ts": 354516603107, "pid": 5},
{"ph": "X", "name": "0", "ts": 354516603109, "pid": 5, "dur": 0},
{"name": "process_name", "ph": "M", "pid": 6, "args": {"name": "training/Adadelta_Allreduce/HorovodAllreduce_Adadelta_1_gradients_conv2d_BiasAdd_grad_BiasAddGrad_0"}},
{"name": "process_sort_index", "ph": "M", "pid": 6, "args": {"sort_index": 6}},
{"ph": "B", "name": "NEGOTIATE_ALLREDUCE", "ts": 354516608183, "pid": 6},
```
I am confused about the "ts" field, I think this is the "time-started", but how to interpret the number?
How to convert it to regular datetime? 
I checked the [timeline.cc](https://github.com/horovod/horovod/blob/master/horovod/common/timeline.cc) file, which shows the "ts" get a microsecond number. But divide `354516608183` by `1000` and then convert it to datetime, seems strange. Here is the output
```
from datetime import datetime
    ...: timestamp = 354663785644/1000.0
    ...: dt_object = datetime.fromtimestamp(timestamp)
    ...: print("dt_object =", dt_object)
    ...: print("type(dt_object) =", type(dt_object))
    ...:
    ...:
dt_object = 1981-03-28 16:43:05.644000
type(dt_object) = <class 'datetime.datetime'>
```
I think it is due to the fact that logging time always minused with `start_time`, so is there easy way to get the start time? ([timeline-L169](https://github.com/horovod/horovod/blob/master/horovod/common/timeline.cc#L169))

The reason I want to get the absolute time is to fuse multiple timeline traces.

tgaddair:
Fixes #1472.

Current timeline events have cryptic timestamps, e.g.:

```
{"ph": "B", "name": "NEGOTIATE_ALLREDUCE", "ts": 678417958358, "pid": 1},
```

With this fix, events properly represent the number of microseconds elapsed since initialization:

```
{"ph": "B", "name": "NEGOTIATE_ALLREDUCE", "ts": 4203464, "pid": 1},
```

### DIFF
--- a/horovod/common/timeline.cc
+++ b/horovod/common/timeline.cc
@@ -29,7 +29,7 @@ void TimelineWriter::Initialize(std::string file_name) {
   file_.open(file_name, std::ios::out | std::ios::trunc);
   if (file_.good()) {
     // Initialize the timeline with '[' character.
-    file_ << "[\n";
+    file_ << "[" << std::endl;
     healthy_ = true;
 
     // Spawn writer thread.
@@ -150,6 +150,8 @@ void Timeline::Initialize(std::string file_name, unsigned int horovod_size) {
   if (initialized_) {
     return;
   }
+
+  start_time_ = std::chrono::steady_clock::now();
 
   // Start the writer.
   writer_.Initialize(std::move(file_name));


Himscipy:
@tgaddair  Is this fix available with horovod version v0.18.1, since in my timeline.json I still see. 
`"ph": "B", "name": "NEGOTIATE_ALLREDUCE", "ts": 84070453580, "pid": 234` 
Also, when I open the File in chrome trace. I can see the time in "ms" . 

I am trying to write a script to fetch some specific quantities of interest. I am interested in getting the times in "ms" as well, is there a conversion which chrome trace is able to do which I am missing. Any insight will be really helpful I have seen your comments on #1472 as well regarding the issue

PR #1479, opened 2019-10-28
(minor change) AllreduceOp: use output tensor pointer and size

jessebenson:
For clarity, use output tensor size (`e.output->size()`) when writing to output tensor pointer (`e.output->data()`).

Note: for AllreduceOp, input tensor and output tensor have same shape/size, so this doesn't actually change any functionality.


tgaddair:
Thanks for the PR @jessebenson, can you sign the DCO for these changes?

PR #1480, opened 2019-10-29
Add CUDA Allgather implementation to support tensor fusion

jessebenson:
Allgather with CUDA tensors does not always work with tensor fusion/CUDA.

1. Build Horovod without HOROVOD_GPU_*
```
import torch
import horovod.torch as hvd

hvd.init()

// The following allgather will succeed because it will run on the CPU (manually cudaMemcpy to CPU, run CPU allgather) in pytorch/tensorflow/etc.
tensor = torch.cuda.FloatTensor(*([4])).fill_(hvd.rank())
result = hvd.allgather(tensor)

// Tensor fusion works because the CUDA memory is manually copied to/from CPU
// and AllgatherOp uses std::memcpy for tensor fusion
tensor1 = torch.cuda.FloatTensor(*([4])).fill_(hvd.rank())
tensor2 = torch.cuda.FloatTensor(*([4])).fill_(hvd.rank())
handle1 = hvd.allgather_async(tensor1)
handle2 = hvd.allgather_async(tensor2)
result1 = hvd.synchronize(handle1)
result2 = hvd.synchronize(handle2)
```

2. Build OpenMPI with CUDA support.  Build Horovod with HOROVOD_GPU_ALLREDUCE=MPI and HOROVOD_GPU_ALLGATHER=MPI.
```
import torch
import horovod.torch as hvd

hvd.init()

// The following allgather will fail because CUDA is not initialized
tensor = torch.cuda.FloatTensor(*([4])).fill_(hvd.rank())
result = hvd.allgather(tensor)

// The following allgather will succeed because MPI_CUDAAllreduce will first initialize CUDA
tensor = torch.cuda.FloatTensor(*([4])).fill_(hvd.rank())
dummy = hvd.allreduce(tensor)
result = hvd.allgather(tensor)

// The following allgather using tensor fusion will fail
// This will invoke AllgatherOp, which unconditionally calls std::memcpy on the pointers, even if they are allocated on a CUDA device
tensor1 = torch.cuda.FloatTensor(*([4])).fill_(hvd.rank())
tensor2 = torch.cuda.FloatTensor(*([4])).fill_(hvd.rank())
handle1 = hvd.allgather_async(tensor1)
handle2 = hvd.allgather_async(tensor2)
result1 = hvd.synchronize(handle1)
result2 = hvd.synchronize(handle2)
```

### DIFF
--- a/horovod/common/ops/cuda_operations.cc
+++ b/horovod/common/ops/cuda_operations.cc
@@ -92,9 +92,83 @@ void CUDAContext::WaitForEvents(std::queue<std::pair<std::string, cudaEvent_t>>&
   }
 }
 
+CUDAOpContext::CUDAOpContext(struct CUDAContext* context,
+                             HorovodGlobalState* global_state)
+    : cuda_context_(context), global_state_(global_state) {}
+
+void CUDAOpContext::InitCUDA(const std::vector<TensorTableEntry>& entries) {
+  auto& first_entry = entries[0];
+  cuda_context_->ErrorCheck("cudaSetDevice", cudaSetDevice(first_entry.device));
+
+  // Ensure stream is in the map before executing reduction.
+  cudaStream_t& stream = cuda_context_->streams[global_state_->current_nccl_stream][first_entry.device];
+  if (stream == nullptr) {
+    int greatest_priority;
+    cuda_context_->ErrorCheck("cudaDeviceGetStreamPriorityRange",
+                              cudaDeviceGetStreamPriorityRange(NULL, &greatest_priority));
+    cuda_context_->ErrorCheck("cudaStreamCreateWithPriority",
+                              cudaStreamCreateWithPriority(&stream, cudaStreamNonBlocking, greatest_priority));
+  }
+}
+
+void CUDAOpContext::InitCUDAQueue(const std::vector<TensorTableEntry>& entries, const Response& response) {
+  event_queue = std::queue<std::pair<std::string, cudaEvent_t>>();
+  stream = &cuda_context_->streams[global_state_->current_nccl_stream][entries[0].device];
+  host_buffer = nullptr;
+
+  if (global_state_->timeline.Initialized()) {
+    cuda_context_->RecordEvent(event_queue, QUEUE, *stream);
+  }
+}
+
+Status CUDAOpContext::FinalizeCUDAQueue(const std::vector<TensorTableEntry>& entries) {
+  // Use completion marker via event because it's faster than
+  // blocking cudaStreamSynchronize() in this thread.
+  cuda_context_->RecordEvent(event_queue, "", *stream);
+
+  auto& first_entry = entries[0];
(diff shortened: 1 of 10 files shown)


tgaddair:
Very nice!  Adding @abditag2 who worked on the original Allgather tensor fusion implementation.

horovod/common/ops/cuda_operations.cc
```diff
@@ -187,5 +187,101 @@ Status CUDAAllreduce::FinalizeCUDAQueue(const std::vector<TensorTableEntry>& ent
...
+                                               const void* buffer_data_at_offset, TensorTableEntry& e,
+                                               int64_t entry_offset, size_t entry_size) {
+  auto& first_entry = entries[0];
+  auto cuda_result = cudaMemcpyAsync((int8_t*)e.output->data() + entry_offset, buffer_data_at_offset,
+                                     entry_size, cudaMemcpyDeviceToDevice,
+                                     cuda_context_->streams[global_state_->current_nccl_stream][first_entry.device]);
+  cuda_context_->ErrorCheck("cudaMemcpyAsync", cuda_result);
+}
+
+void CUDAAllgather::InitCUDA(const std::vector<TensorTableEntry>& entries) {
```
jessebenson: These three methods are copied directly from CUDAAllreduce.  I want to de-dupe but I'm looking for guidance on code style:

1.  Composition - shared class similar to CUDAContext that encapsulates functionality as member variable.
1.  Multiple inheritance - class like `CUDAOp` and then we have `class CUDAAllreduce : public AllreduceOp, public CUDAOp`
1.  Templated inheritance - class like `template <typename Base> class CUDAOp : public Base` and then we have `class CUDAAllreduce : public CUDAOp<AllreduceOp>`
1.  Leave the code duplicated?

jessebenson:
I am also preparing a pull request to add "Reducescatter" as a first-class operation type.  There's a paper about a memory optimization for distributed training.  Instead of doing 'allreduce -> optimizer' they do 'reducescatter -> optimizer -> allgather'.  But to do that, I need reducescatter in Horovod :)

https://arxiv.org/abs/1910.02054

horovod/common/ops/cuda_operations.cc
tgaddair: Good question!  Looking through the code and how it's used, my suggestion would be to go with composition.  We could roll the three methods into a new class called `CUDAOpContext`.  Then, instead of subclasses like NCCLAllreduce and DDLAllreduce calling `InitCUDAQueue`, we have them call those methods on a protected member `cuda_op_context_`.  What do you think?

jessebenson: Yes, I like this approach.  I'll add that and update the pull request!

jessebenson:
I rebased on the latest master commit, and added the CUDAOpContext class for `CUDAAllreduce` and `CUDAAllgather` to share.

PR #1484, opened 2019-10-29
Adasum algorithm for allreduce

Tixxx:
### What is Adasum 

Scaling DNN training to many GPUs always comes at a convergence degradation. This is because with larger batch sizes, gradients are averaged and the learning rate per example is smaller. To address this, learning rate is usually scaled up but this can lead to divergence of model parameters. Adasum addresses these two issues without introducing any hyper-parameter. 

Suppose there are two almost-parallel gradients from two different GPUs, g1 and g2, and they need to be reduced as shown in the figure below. The two common practices for reductions are g1+g2, the gray vector or (g1+g2)/2, the green vector. g1+g2 may cause divergence of the model since it is effectively moving in the direction of g1 or g2 by two times the magnitude of g1 or g2. Therefore, generally (g1+g2)/2 is safer and more desired. 
pic1
 

Now consider the two orthogonal gradients g1 and g2 in the figure below. Since g1 and g2 are in two different dimensions and independent of each other, g1+g2 may not cause divergence. 
pic2
 

 

 

Finally, consider the third scenario where g1 and g2 are neither parallel nor orthogonal as shown in the figure below. In such a case, Adasum projects g2 on the orthogonal space of g1 (the pink vector) and adds that with g1 to produce the reduced vector. In this case, the final vector moves in each dimension only as much as each of g1 or g2 and therefore causes no divergence. 
pic3
 
This idea extends to many gradients as well. Suppose there are 2^n gradients coming from 2^n different GPUs. Adasum inductively takes pairs of gradients and reduces them using the method above until all of them are reduced into one gradient. 

### Highlights of code changes  

  

We provide an algorithmic interface which does not have any dependency on a particular communication library for extensibility. An MPI version of implementation of Adasum has been provided to support new operations we have added to Horovod. Here is the list of changes that we propose:  

  

- Adasum class in `horovod/common/ops/adasum/adasum.h`: Algorithmic interface of Adasum which contains the main logic.  

  

- AdasumMPI class in `horovod/common/ops/adasum/adasum_mpi.h` and `adasum_mpi.cc`: An MPI implementation of Adasum algorithm.  

  

- AdasumMPIAllreduceOp class in `horovod/common/ops/adasum_mpi_operations.h` and `adasum_mpi_operations.cc`: A new operation class that inherits from AdasumMPI and Horovod's AllreduceOp. This utilizes the fusion buffer to perform efficient Adasum reductions on CPU when HOROVOD_GPU_ALLREDUCE is set to None.  

  

- AdasumCudaAllreduceOp class in `horovod/common/ops/adasum_cuda_operations.h` and `adasum_cuda_operations.cc`: A new operation class that inherits from AdasumMPI and Horovod's NCCLAllreduce. This is a hierarchical operation that uses NCCL to perform intra-node sum-averaging and Adasum algorithm for inter-node reductions. This op requires Horovod to be compiled with HOROVOD_GPU_ALLREDUCE=NCCL  

  

- A new response and request type has been introduced in addition to existing ones:  

  

`enum ResponseType { ALLREDUCE = 0, ALLGATHER = 1, BROADCAST = 2, ADASUM = 3, ERROR = 4};`  

  

- A new environment variable `HOROVOD_ADASUM_MPI_CHUNK_SIZE` has been introduced to improve MPI communication efficiency for some platform configurations(i.e. Azure NC series machines + IntelMPI).  

  

   

  

In addition to the above changes in Horovod's common library, we also added a list of changes to framework layer for both Tensorflow and Pytorch to enable easy use of Adasum:  

  

- An enum that contains a list of allreduce operations has been introduced for users to select among Average, Sum or Adasum. This improves extensibility to add more ops in the future and backward compatibility.  

  

- An optional parameter `op` has been added to DistributedOptimizer and allreduce API for users to specify which operation to perform.  

  

- A new distributed optimizer has been added to both frameworks to support Adasum algorithm. Since the nature of Adasum requires it to operate on the full magnitude of the gradient, the newly added distributed optimizer uses the difference in magnitude of weights between before and after the optimizer performs a step to deliver a more accurate estimation. When `op=hvd.Adasum` is specified, the new optimizer will be used.  

  

   

  

  DistributedOptimizer example for Tensorflow:  

  

   

  

  ` opt = tf.train.AdamOptimizer(0.001)`  

  

  ` opt = hvd.DistributedOptimizer(opt, backward_passes_per_step=5, op=hvd.Adasum)`  

  

   

  

  Allreduce example for Tensorflow:  

  

   

  

  ` hvd.allreduce(tensor, op=hvd.Adasum)`  

  

   

  

  DistributedOptimizer example for Pytorch:  

 

  ` optimizer = optim.SGD(model.parameters(), lr=args.lr, momentum=args.momentum)`  

  

  `optimizer = hvd.DistributedOptimizer(optimizer, named_parameters=model.named_parameters(),  compression=compression, backward_passes_per_step = 5, op=hvd.Adasum)`  

  

 
  

  Allreduce example for Pytorch:  

  

   

  

  ` hvd.allreduce(tensor, op=hvd.Adasum)`  

  

  

  

### Additional notes  
- Adasum ensures correct convergence behavior even with large effective batch sizes. 
  

- As the number of ranks scales up, the learning rate does not need to be scaled if using CPU to do Adasum reduction. If HOROVOD_GPU_ALLREDUCE=NCCL flag is used to compile Horovod, Adasum needs the learning rate to be scaled by the number of GPUs **locally** on a node.  

  

- Pytorch training in fp16 format is not yet supported by this pull request. We are in the process of integrating Apex into the new optimizer to enabled full mixed precision training with Adasum in Pytorch. 

 

- When `HOROVOD_GPU_ALLREDUCE=NCCL` flag is used to compile Horovod and training is run on a single node, only averaging through NCCL library is used to perform reductions and no Adasum algorithm will take place in this configuration.


Tixxx:
closing this one and reopening another one with updated history

PR #1486, opened 2019-10-30
support user specify the NIC

ConeyLiu:
I have run horovod.spark in a three nodes cluster. However, it got the following warning:
```
WARNING: Open MPI failed to TCP connect to a peer MPI process.  This
should not happen.

Your Open MPI job may now hang or fail.

  Local host: sr505
  PID:        86258
  Message:    connect() to 10.1.2.104:1053 failed
  Error:      Operation now in progress (115)
```

And the job got to hang. This may be related to [link](https://github.com/open-mpi/ompi/issues/4963).

And also, different NIC can provide different performance. So, we could specify the NIC for using in this patch.

### DIFF
--- a/horovod/run/common/util/network.py
+++ b/horovod/run/common/util/network.py
@@ -85,13 +85,15 @@ class Wire(object):
 
 
 class BasicService(object):
-    def __init__(self, service_name, key):
+    def __init__(self, service_name, key, nic):
         self._service_name = service_name
         self._wire = Wire(key)
+        self._nic = nic
         self._server, _ = find_port(
             lambda addr: socketserver.ThreadingTCPServer(
                 addr, self._make_handler()))
         self._port = self._server.socket.getsockname()[1]
+        self._addresses = self._get_local_addresses()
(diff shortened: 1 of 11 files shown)


horovod/run/common/util/network.py
```diff
@@ -119,16 +121,24 @@ def _handle(self, req, client_address):
...
+            if self._nic and intf != self._nic:
+                continue
             for addr in intf_addresses:
                 if addr.family == socket.AF_INET:
                     if intf not in result:
                         result[intf] = []
                     result[intf].append((addr.address, self._port))
+        if not result:
+            raise NoValidAddressesFound(
+                'Can not find a available NIC matches the given one: {}'.format(self._nic))
```
tgaddair: Let's change this to:

```
if not result and self._nic:
        raise NoValidAddressesFound(`No available network interface found matching user provided interface: {}`.format(self._nic))
```

ConeyLiu:
thanks @tgaddair for the reviewing. The comments have addressed.

ConeyLiu:
thanks

PR #1494, opened 2019-11-04
Refactor horovod.spark.run()

EnricoMi:
- makes horovod.spark.run() use horovod.run.mpi_run() and exposes extra_mpi_args
- adds lots of tests

This PR is easiest to review commit by commit.


EnricoMi:
Thanks for the quick support!

PR #1496, opened 2019-11-04
Add Reducescatter operator

jessebenson:
Add **Reducescatter** operator.  From Nvidia documentation:
image

1. Implement Reducescatter operator on CPU and GPU (with MPI+CUDA)
1. Support tensor fusion with Reducescatter
1. Expose Reducescatter in Python through pytorch, tensorflow, keras, mxnet
1. Add unit tests for pytorch/tensorflow (similar set as covered by allreduce)
1. Updated `concepts.rst` documentation to describe operator.


jessebenson:
Many (but not all) of the "Run PyTests" runs are timing out when hitting the Reducescatter unit tests.  I am trying to understand what's causing this.  Any tips are appreciated!

kit1980:
The two problematic Join tests are the ones that test "not implemented" failures for allgather and broadcast. I think it's OK to disable them for this PR, I'll work on a proper fix separately.

jessebenson:
Looks like there may be a breaking change in tfhead.  The build image steps in unit tests are failing now:

```
ModuleNotFoundError: No module named 'tensorflow_core.keras'
```

tgaddair:
@jessebenson just triggered a rebuild.  TensorFlow's last nightly had a bug in it they've since rolled-back, so should be working now.

tgaddair:
Hey @jessebenson, there was another breaking change by TensorFlow that required a fix in #1515.  Can you rebase again?

jessebenson:
@tgaddair - will do.

I don't currently have ReduceScatter for MLSL, NCCL, or GLOO.  Those will take a bit longer.  MPI and GLOO ReduceScatter allow different receive counts per rank, while MLSL and NCCL require all ranks have same receive count.  I was planning to do a partial Reduce to last rank (if tensor doesn't evenly divide) to solve this, similar in principle to how Horovod currently does hierarchical APIs.

GLOO has ReduceScatter, but they don't expose an API to call - compare:
https://github.com/facebookincubator/gloo/blob/master/gloo/reduce_scatter.h
https://github.com/facebookincubator/gloo/blob/master/gloo/allgather.h#L71

jessebenson:
I think this pull request is in a good state to review.

The future work would be Reducescatter implementation for NCCL, GLOO, and the new Intel CCL.  MLSL is being removed.  However, GLOO doesn't have a proper public API for Reducescatter, CCL doesn't have Reducescatter at all - so that leaves NCCL.

horovod/tensorflow/mpi_ops.cc
```diff
@@ -480,5 +480,66 @@ Output
...
+    .Output("output: T")
+    .SetShapeFn([](shape_inference::InferenceContext* c) {
+      shape_inference::ShapeHandle output;
+      TF_RETURN_IF_ERROR(
+          c->ReplaceDim(c->input(0), 0, c->UnknownDim(), &output));
+      c->set_output(0, output);
+      return Status::OK();
+    })
+    .Doc(R"doc(
+Perform an MPI Reducescatter on a tensor. All other processes that do a
```
tgaddair: I know this is consistent with the other docstrings here, but we should remove the reference to MPI at some point.

horovod/tensorflow/__init__.py
```diff
@@ -118,6 +118,37 @@ def allreduce(tensor, average=None, device_dense='', device_sparse='',
         return new_tensor
 
 
+def reducescatter(tensor, average=True, device_dense='', compression=Compression.none):
```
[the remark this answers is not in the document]
jessebenson: Yes this seems perfectly reasonable.

horovod/tensorflow/mpi_ops.cc
jessebenson: Happy to update.

tgaddair:
Hey @jessebenson, any updates on the API consolidation?

jessebenson:
The `Allreduce` sort of supports both `average` and `op` now, and has some legacy logic to decide between them.  I'm wondering - should I support just `op` parameter?

tgaddair:
I would just support the new `op` param since this is new functionality.  Thanks for checking!

jessebenson:
It became a bit more involved to update all of mxnet/keras/tensorflow/pytorch.  Hopefully I was thorough and careful enough.  I also added a unit test to `test_torch.py` to verify passing `op=hvd.Adasum` will give an error.

horovod/tensorflow/mpi_ops.py
```diff
@@ -201,7 +201,7 @@ def _broadcast_grad(op, grad):
     return grad_reduced
 
 
-def _reducescatter(tensor, name=None):
+def _reducescatter(tensor, name=None, op=Sum):
```
tgaddair: Looks like everywhere else the default is `Average` but here it's `Sum`.  Is there a reason, or should we make it `Average` here as well?

jessebenson: So, `Average` needs to be handled in the higher-level `reducescatter()` in `__init__.py` (which is where the division happens).  I was trying to match the behavior of Allreduce, where `allreduce()` handles `Average` and passes `Sum` to `_allreduce()`.

`_allreduce()` and `_reducescatter()` defaults are `Sum`
`allreduce()` and `reducescatter()` defaults are `Average`

jessebenson: Let me make a small tweak to `reducescatter()` though.  Since the underlying `_reducescatter()` only supports `Sum` at the moment, I never actually pass the reduce op through to it.  However, I should pass something to it since any errors will get caught in `EnqueueTensorReduceScatter` anyway so it's "safe" to pass the op.  That way future people aren't confused why they pass `Max` and it doesn't show up in the C++ API.

jessebenson:
Some tests are failing now - likely a bad interaction with my changes and a recent change from master.  I'll have to investigate.

kit1980:
> Some tests are failing now - likely a bad interaction with my changes and a recent change from master.  I'll have to investigate.

After https://github.com/horovod/horovod/pull/1594 you need to set tensor sizes and data type for requests, like here: https://github.com/horovod/horovod/blob/master/horovod/common/controller.cc#L581

Also there is ongoing https://github.com/horovod/horovod/pull/1604

jessebenson:
@kit1980 - thanks, that is useful to know.  What are these tensor sizes used for?  Should it correspond to the fusion buffer size, or response size?

For Allreduce/Adasum, the input/output tensors are the same size.
For AllGather, the input (per rank) is size ~T and output is ~T*N.
For ReduceScatter, the input (per rank) is size T and output is ~T/N.
For Broadcast, it doesn't seem to set the tensor sizes (is it not needed?)

kit1980:
@jessebenson, currently the sizes are used in this way only for AllReduce and AdaSum.

The tensor sizes are used in https://github.com/horovod/horovod/blob/master/horovod/common/controller.cc#L642 and https://github.com/horovod/horovod/blob/master/horovod/common/controller.cc#L651

The sizes set in response in this case mean number of elements in the tensor. Not sure how this will work when input/output sizes are different.

Basically, the recent change was that `FuseResponses` used to call `tensor_queue_.GetTensorSizeAndType(response.tensor_names()[0],                                        tensor_size, dtype)`, and now `tensor_size` and `dtype` are from response directly (with a caveat that `tensor_sizes` in response are in the number of elements, so need to multiply on element size to get bytes).

legatoo:
Any updates of this PR?

maxhgerlach:
@legatoo, @ducviet00, and anybody else who's interested: I've revived this PR in #3299 and any feedback would be appreciated.

PR #1525, opened 2019-11-20
Consider CONTAINER_ID in host_hash

EnricoMi:
YARN containers do not share memory, so each YARN container must be considered its own 'host'.

Adds lots of documentation around how Horovod orchestrates Spark and MPI.


EnricoMi:
@tgaddair: done, all tests pass. Thanks!

PR #1532, opened 2019-11-22
Make Spark's Hadoop token file available to Python method

EnricoMi:
Python code run via horovod.spark.run() needs the
Hadoop token file to access secure HDFS. This is
made available to the Spark executor via environment
variable HADOOP_TOKEN_FILE_LOCATION. SparkTaskService
needs to pass this to orted, which runs the Python code.

### DIFF
--- a/test/test_spark.py
+++ b/test/test_spark.py
@@ -22,7 +22,6 @@ import os
 import pytest
 import re
 import subprocess
-import sys
 import time
 import torch
 import unittest
@@ -36,10 +35,7 @@ import horovod.torch as hvd
 
 from mock import MagicMock
 
-if sys.version_info[0] < 3:
-    from backports import tempfile
-else:
-    import tempfile
+from common import tempdir
 
 
 @contextlib.contextmanager
@@ -230,7 +226,7 @@ class SparkTests(unittest.TestCase):
             service = SparkTaskService(1, key, None)
             client = SparkTaskClient(1, service.addresses(), key, 3)
 
-            with tempfile.TemporaryDirectory() as d:
+            with tempdir() as d:
                 file = '{}/env'.format(d)
                 command = "env | grep -v '^PWD='> {}".format(file)
                 command_env = {"test": "value"}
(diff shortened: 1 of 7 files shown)


EnricoMi:
Build failed on unrelated issues, might be transient. Please retry the build checks.

test/test_spark.py
```diff
@@ -22,17 +22,25 @@
...
 
+from horovod.run.common.util import secret
 from horovod.run.mpi_run import _get_mpi_implementation_flags
 import horovod.spark
+from horovod.spark.task.task_service import SparkTaskService, SparkTaskClient
 import horovod.torch as hvd
 
 from mock import MagicMock
 
+if sys.version_info[0] < 3:
```
tgaddair: Rather than add a dependency on `backports.tempfile`, I think it would be easier to just write our own temp directory function.  We already did this in an upcoming PR (https://github.com/horovod/horovod/blob/spark_estimator/test/common.py#L63).  Maybe you can cherry pick that context manager and use it here?

EnricoMi: I know, this import and `backports.tempfile` dependency is ugly, but I'd prefer that over duplicating code, even if it is small.

A related question that may help me assess the situation: how long do you plan to support Python2.7? It will die in a month. Is it worth adding this dependency / your own implementation when this is only needed to support Python2.7, which will go away short term anyway?

tgaddair: Most likely we will drop Python 2.7 support at the end of the year, but that decision needs to be brought before the steering committee, in case anyone is stuck with Python 2.7 for whatever reason.

If you're uncomfortable adding that extra function, one option would be to simply skip this test for Python 2 by adding the following decorator:

```
@pytest.mark.skipif(sys.version_info < (3, 0), reason="tempfile.TemporaryDirectory only available in Python 3")
```

EnricoMi: Interesting alternative.

Since you are introducing the `TemporaryDirectory` anyway, I will use it. But it should get removed once Python2.7 support is dropped.

alasdairm-gr: @tgaddair is this okay to merge now?  It would be really useful to have in the next Horovod release

tgaddair: Sorry for the delay!  LGTM!

PR #1546, opened 2019-11-27
[ROCm] initial ROCm support

jeffdaily:
This adds appropriate hooks to setup.py and swaps CUDA APIs for HIP APIs via the preprocessor at build time.  Currently only supports TensorFlow build target.  A successful ROCm build uses the following environment variables
```
export HOROVOD_WITHOUT_MXNET=1
export HOROVOD_WITHOUT_PYTORCH=1
export HOROVOD_GPU_ROCM=1
export HOROVOD_WITH_TENSORFLOW=1
export HOROVOD_ROCM_PATH=/opt/rocm
export HOROVOD_GPU_ALLREDUCE=NCCL
```

### DIFF
--- a/horovod/common/ops/cuda_operations.h
+++ b/horovod/common/ops/cuda_operations.h
@@ -21,7 +21,34 @@
 #include <unordered_map>
 #include <vector>
 
+#if HAVE_CUDA
 #include <cuda_runtime.h>
+#elif HAVE_ROCM
+#include <hip/hip_runtime_api.h>
+// local hipify
+#define cudaDeviceGetStreamPriorityRange hipDeviceGetStreamPriorityRange
+#define cudaError_t hipError_t
+#define cudaEventBlockingSync hipEventBlockingSync
+#define cudaEventCreateWithFlags hipEventCreateWithFlags
+#define cudaEventDisableTiming hipEventDisableTiming
+#define cudaEventRecord hipEventRecord
+#define cudaEventSynchronize hipEventSynchronize
+#define cudaEvent_t hipEvent_t
+#define cudaGetDevice hipGetDevice
+#define cudaGetErrorString hipGetErrorString
+#define cudaMemcpyAsync hipMemcpyAsync
+#define cudaMemcpyDeviceToDevice hipMemcpyDeviceToDevice
+#define cudaMemcpyDeviceToHost hipMemcpyDeviceToHost
+#define cudaMemcpyHostToDevice hipMemcpyHostToDevice
+#define cudaSetDevice hipSetDevice
+#define cudaStreamCreateWithPriority hipStreamCreateWithPriority
+#define cudaStreamNonBlocking hipStreamNonBlocking
+#define cudaStreamSynchronize hipStreamSynchronize
+#define cudaStream_t hipStream_t
+#define cudaSuccess hipSuccess
+#else
+#error Must define either HAVE_CUDA or HAVE_ROCM
+#endif
 
 #include "collective_operations.h"
 #include "../thread_pool.h"
--- a/setup.py
+++ b/setup.py
@@ -431,7 +431,7 @@ def get_rocm_dirs(build_ext, cpp_flags):
     rocm_libs = ['hip_hcc']
     rocm_macros = [('__HIP_PLATFORM_HCC__',1)]
 
-    rocm_path = os.environ.get('HOROVOD_ROCM_PATH', '/opt/rocm')
+    rocm_path = os.environ.get('HOROVOD_ROCM_HOME', '/opt/rocm')
     rocm_include_dirs += [
             '%s/include' % rocm_path,
             '%s/hcc/include' % rocm_path,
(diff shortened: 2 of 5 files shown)


jeffdaily:
Looks like the failed tests were not related to changes made by this PR.  Can we retest?

tgaddair:
Hey @jeffdaily, thanks for the PR.  You may need to rebase following https://github.com/horovod/horovod/commit/c8c53a9c901e980e3f0a10634322580743bd17b8.  Then tests should hopefully succeed.

setup.py
```diff
@@ -440,27 +474,33 @@ def get_nccl_vals(build_ext, cuda_include_dirs, cuda_lib_dirs, cpp_flags):
     if nccl_lib_dir:
         nccl_lib_dirs += [nccl_lib_dir]
 
-    nccl_link_mode = os.environ.get('HOROVOD_NCCL_LINK', 'STATIC')
+    nccl_link_mode = os.environ.get('HOROVOD_NCCL_LINK', have_rocm and 'SHARED' or 'STATIC')
```
tgaddair: Can we change this to `'SHARED' if have_rocm else 'STATIC'`?

setup.py
```diff
@@ -440,27 +474,33 @@ def get_nccl_vals(build_ext, cuda_include_dirs, cuda_lib_dirs, cpp_flags):
...
+            #include <%s>
             #if NCCL_MAJOR < 2
             #error Horovod requires NCCL 2.0 or later version, please upgrade.
             #endif
             void test() {
                 ncclUniqueId nccl_id;
                 ncclGetUniqueId(&nccl_id);
             }
-            '''))
+            '''%(have_rocm and 'rccl.h' or 'nccl.h')))
```
tgaddair: Same here, let's prefer using Python's `if-else` construction.

setup.py
```diff
@@ -422,7 +422,41 @@ def get_cuda_dirs(build_ext, cpp_flags):
...
 
 
-def get_nccl_vals(build_ext, cuda_include_dirs, cuda_lib_dirs, cpp_flags):
+def get_rocm_dirs(build_ext, cpp_flags):
+    rocm_include_dirs = []
+    rocm_lib_dirs = []
+    rocm_libs = ['hip_hcc']
+    rocm_macros = [('__HIP_PLATFORM_HCC__',1)]
+
+    rocm_path = os.environ.get('HOROVOD_ROCM_PATH', '/opt/rocm')
```
tgaddair: Let's align with `HOROVOD_CUDA_HOME` and call this `HOROVOD_ROCM_HOME`.

setup.py
```diff
@@ -593,17 +633,21 @@ def get_common_options(build_ext):
...
+    have_cuda = False
+    have_rocm = False
+    cuda_include_dirs = cuda_lib_dirs = cuda_macros = []
     if gpu_allreduce or gpu_allgather or gpu_broadcast:
-        have_cuda = True
-        cuda_include_dirs, cuda_lib_dirs = get_cuda_dirs(build_ext, cpp_flags)
-    else:
-        have_cuda = False
-        cuda_include_dirs = cuda_lib_dirs = []
+        if os.environ.get('HOROVOD_GPU_ROCM'):
```
tgaddair: Instead of having the user specify `HOROVOD_GPU_ROCM`, how about we use the existing `HOROVOD_GPU_ALLREDUCE` env variable and have them set `HOROVOD_GPU_ALLREDUCE=RCCL`?

horovod/common/ops/cuda_operations.h
```diff
@@ -21,7 +21,34 @@
 #include <unordered_map>
 #include <vector>
 
+#if HAVE_CUDA
 #include <cuda_runtime.h>
+#elif HAVE_ROCM
+#include <hip/hip_runtime_api.h>
+// local hipify
+#define cudaDeviceGetStreamPriorityRange hipDeviceGetStreamPriorityRange
```
tgaddair: Seems we'll need to update this every time we use a new CUDA symbol somewhere.  Can you add tests to our CI to ensure we don't break something by mistake?

setup.py
jeffdaily: I'm typically in favor of simplifying, but in this case, shouldn't we also support building with ROCm support but without RCCL, i.e. ROCm with RDMA but without RCCL?

horovod/common/ops/cuda_operations.h
jeffdaily: This raises the more general question of how to support ROCm on existing CUDA code.  Without imposing that horovod change all CUDA APIs to HIP APIs, we can either "hipify" during the build via the preprocessor or hipify prior to the build via an additional script and pre-build step.

Since the number of actual CUDA APIs and types used currently are few, I opted for the preprocessor option.  We don't have a public CI supporting horovod at this time that would catch build errors for the ROCm path.  What kind of CI test do you propose would catch newly-added CUDA symbols?

jeffdaily:
> In addition to the stuff mentioned below, can you comment a bit about plans for fleshing out this feature? Will the hipify declarations be replaced with a less fragile solution long term? Do you have plans to support PyTorch and MXNet as well?

I commented inline regarding the hipify declarations.  Perhaps fragile in a different way, we could try and factor out the CUDA calls into a generic "GPU" interface that could be implemented by either CUDA or HIP APIs.  This is what we currently do in upstream TensorFlow, as much as we can.  If horovod developers determine a need for another CUDA API, it would need to be added to the declared GPU interface at which time it is a clear indicator that a HIP version is also needed.  This currently seems like a lot of work given that only a few CUDA symbols are currently used.

Unfortunately, I cannot comment on anything regarding business planning.

jeffdaily:
Replaced with #1632 .

PR #1554, opened 2019-12-02
Initial commit of Horovod Spark Estimator framework

tgaddair:
Horovod Spark Estimators provide a high level API for distributed training a Keras or PyTorch model directly on a Spark DataFrame.

Contributors:

- Travis Addair (@tgaddair)
- Fardin Abdi (@abditag2)


tgaddair:
@EnricoMi

examples/keras_spark_mnist.py
```diff
@@ -0,0 +1,124 @@
...
+if args.master:
+    conf.setMaster(args.master)
+elif args.num_proc:
+    conf.setMaster('local[{}]'.format(args.num_proc))
+spark = SparkSession.builder.config(conf=conf).getOrCreate()
+
+# Setup our store for intermediate data
+work_dir = args.work_dir
+hdfs_prefix = 'hdfs://'
+if work_dir.startswith(hdfs_prefix):
```
EnricoMi: This code could go into a class method of, lets say `Store.create(path)` that given a path returns an `HDFSStore` for a path starting with `hdfs://` and an `LocalStore` for a path starting with `file://`. This way, no-one needs to duplicate this 5 lines of code.

horovod/spark/common/backend.py
```diff
@@ -0,0 +1,59 @@
...
+        full_env = self._env or os.environ.copy()
+        if env:
+            full_env.update(env)
+
+        if 'CUDA_VISIBLE_DEVICES' in full_env:
+            # In TensorFlow 2.0, we set this to prevent memory leaks from the client.
+            # See https://github.com/tensorflow/tensorflow/issues/33168
+            del full_env['CUDA_VISIBLE_DEVICES']
+
+        return horovod.spark.run(fn, args=args, kwargs=kwargs,
```
EnricoMi: Method `horovod.spark.run` has a lot more useful arguments that might be required by some suers / in some environments, e.g. `extra_mpi_args`, `nic`, `verbose`.

horovod/spark/common/cache.py
```diff
@@ -0,0 +1,80 @@
...
+            content.validation_col == validation_col and \
+            content.store.exists(train_data_path) and \
+            (validation_split == 0.0 or content.store.exists(val_data_path))
+
+    def clear(self):
+        self._reset()
+
+    def _reset(self):
+        with self._lock:
+            self._entry_locks = dict()
```
EnricoMi: race condition: if while one thread calls into `acquire_lock_for_key` another thread calls `clear`, the `self._lock` is only held over lines 30-32, setting `self._entry_locks = dict` can happen right before `self._entry_locks[key]`, which will then throw an exception. That exception might not be very descriptive (acquiring the lock failed, but the exceptions says something about a missing key in a dictionary), if desired at all.

horovod/spark/common/constants.py
```diff
@@ -0,0 +1,30 @@
...
+
+ARRAY = 'array'
+CUSTOM_SPARSE = 'custom_sparse_format'
+NOCHANGE = 'nochange'
+
+MIXED_SPARSE_DENSE_VECTOR = 'mixed_sparse_dense_vector'
+SPARSE_VECTOR = 'sparse_vector'
+DENSE_VECTOR = 'dense_vector'
+
+TOTAL_BUFFER_MEMORY_CAP = 4
```
EnricoMi: should mention that this is in GiB

horovod/spark/common/constants.py
```diff
@@ -0,0 +1,30 @@
...
+ONE_GB = 1073741824
```
EnricoMi: should read `ONE_GIBIBYTE` or `ONE_GIB`

docs/spark.rst
```diff
@@ -3,13 +3,128 @@
...
+
+Installation
+~~~~~~~~~~~~
+
+When installing Horovod for usage with Spark, set environment variable ``HOROVOD_WITH_SPARK=1`` to install all
+Spark dependencies as well:
+
+.. code-block:: bash
+
+    $ HOROVOD_WITH_SPARK=1 ... pip install horovod
```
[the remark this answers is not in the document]
tgaddair: Great suggestion!  I admit I did not know about this feature.

examples/keras_spark_mnist.py
tgaddair: I like it!  Thanks for the suggestion.

EnricoMi:
I admit it is orthogonal to this PR. Can be done for other optional dependencies like pytorch and keras as well.


-------- Ursprüngliche Nachricht --------
Von: Travis Addair <notifications@github.com> 
Datum: 02.12.2019  22:16  (GMT+01:00) 
An: horovod/horovod <horovod@noreply.github.com> 
Cc: Enrico Minack <github@enrico.minack.dev>, Mention <mention@noreply.github.com> 
Betreff: Re: [horovod/horovod] Initial commit of Horovod Spark Estimator
  framework (#1554) 

@tgaddair commented on this pull request. In docs/spark.rst: > + +* Are using Keras (``tf.keras`` or ``keras``) or PyTorch for training. +* Want to train directly on a Spark DataFrame from ``pyspark``. +* Are using a standard gradient descent optimization process as your training loop. + +If for whatever reason the Estimator API does not meet your needs, the Run API offers more find-grained control. + +Installation +~~~~~~~~~~~~ + +When installing Horovod for usage with Spark, set environment variable ``HOROVOD_WITH_SPARK=1`` to install all +Spark dependencies as well: + +.. code-block:: bash + + $ HOROVOD_WITH_SPARK=1 ... pip install horovod Great suggestion! I admit I did not know about this feature. —You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

horovod/spark/common/backend.py
tgaddair: Good point.  I added a catch-call `**kwargs` argument to the `SparkBackend` that should allow users to pass in any keyword arguments used by `horovod.spark.run`.

horovod/spark/common/constants.py
tgaddair: Changed to `TOTAL_BUFFER_MEMORY_CAP_GIB`.

horovod/spark/common/constants.py
tgaddair: Changed to `BYTES_PER_GIB`.

horovod/spark/common/cache.py
tgaddair: Good catch!  I think having a per-key lock in this instance is overkill.  Removed and replaced with a single global lock.

horovod/spark/common/constants.py
EnricoMi: yes, bytes per GiB, even "more" correct

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+    for col_info in raw_col_info_list:
+        col_name = col_info[0]
+
+        all_col_types[col_name] = col_info[1][0]
+        col_shapes[col_name] = col_info[1][1]
+        col_max_sizes[col_name] = col_info[1][2]
+
+    # all the rows of each columns must have the same shape
+    for col in df.schema.names:
+        shape_set = col_shapes[col]
```
EnricoMi: if there is a column that has only `None`s, then `col_shapes` might not contain it

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+    }
+    """
+    all_col_types, col_shapes, col_max_sizes = _get_col_info(df)
+
+    metadata = dict()
+    for field in df.schema.fields:
+        col = field.name
+        col_types = all_col_types[col].copy()
+
+        if DenseVector in col_types and SparseVector in col_types:
```
EnricoMi: you can drop the first `if`clause, it is included in the second, because the bodies are identical:
`if dense and sparse or dense` is the same as `if dense`:
```
if DenseVector in col_types:
elif SparseVector in col_types: # and not dense
else: # neither dense nor sparse
```

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+        metadata[col] = {'spark_data_type': spark_data_type,
+                         'is_sparse_vector_only': is_sparse_vector_only,
+                         'shape': col_shapes[col],
+                         'intermediate_format': convert_to_target,
+                         'max_size': col_max_sizes[col]}
+
+    return metadata
+
+
+def to_petastorm_fn(schema_cols, metadata):
```
EnricoMi: this must be so expensive, how big are your datasets?

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+    ARRAY = constants.ARRAY
+    CUSTOM_SPARSE = constants.CUSTOM_SPARSE
+
+    # Convert Spark Vectors into arrays so Petastorm can read them
+    def to_petastorm(row):
+        import numpy as np
+        from pyspark import Row
+
+        fields = row.asDict().copy()
```
EnricoMi: When all columns are `NOCHANGE`, this is really hurting. Should be improved as the next step.

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+        metadata[col] = col_info
+
+    avg_row_size = total_byte_size / train_rows
+    return train_rows, val_rows, metadata, avg_row_size
+
+
+def prepare_data(num_processes, store, df, label_columns, feature_columns,
+                 validation_col=None, validation_split=0.0, sample_weight_col=None,
+                 partitions_per_process=10, verbose=False):
+    if validation_split and validation_col:
```
EnricoMi: Then you could have a single argument `validation` that can be either a String (interpreted as a validation column) or a float (interpreted as a validation split).

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+        raise ValueError("can only specify one of validation_col and validation_split")
+
+    num_partitions = num_processes * partitions_per_process
```
EnricoMi: `num_processes` and `partitions_per_process` must be larger than `0`

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+    if verbose:
+        print('num_partitions={}'.format(num_partitions))
+
+    train_data_path = store.get_train_data_path()
+    val_data_path = store.get_val_data_path()
+    if verbose:
+        print('train_data_path={}'.format(train_data_path))
+        print('val_data_path={}'.format(val_data_path))
+
+    for col in label_columns:
```
EnricoMi: can `label_columns` be empty or `None`?

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+        if col not in df.columns:
+            raise ValueError('Label column {} does not exist in this DataFrame'.format(col))
+
+    if feature_columns is None:
+        feature_columns = [col for col in df.columns if col not in set(label_columns)]
+
+    dataframe_hash = df.__hash__()
+
+    global _training_cache
```
EnricoMi: cache should be injected, not globally accessed

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+    key = (dataframe_hash, validation_split, validation_col, train_data_path, val_data_path)
+    with _training_cache.lock:
```
EnricoMi: This looks like a very long lock cycle when cache is cold, this will synchronize multiple threads. Can you please explain why your need to lock the cache anyway? What are the users of the cache / this method?

EnricoMi: I guess what you are trying to do is to avoid the same traning df is being written to parquet when it is accessed again while it is being created. Hence the per-key lock. This looks like a lazy creation of the parquet files.
Can you initially enumerate all training dataframes that will be used exactly once multi threaded, then you do not need any lock at all.

horovod/spark/common/util.py
EnricoMi: `validation_split` must be `[0.0, 1)`, is this checked where user code calls into horovod (if this is not that place)?

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+
+                # Approximate ratio of validation data to training data for proportionate scale
+                # of partitions
+                timeout_ms = 1000
+                confidence = 0.90
+                train_rows = train_df.rdd.countApprox(timeout=timeout_ms, confidence=confidence)
+                val_rows = val_df.rdd.countApprox(timeout=timeout_ms, confidence=confidence)
+                validation_ratio = val_rows / (val_rows + train_rows)
+
+            train_partitions = max(int(num_partitions * (1.0 - validation_ratio)),
```
EnricoMi: I think you should have at least `num_processes` partitions, not `partitions_per_process`, otherwise you are not utilizing all your executors / processes. Same for `val_partitions`.

setup.py
```diff
@@ -1430,6 +1431,7 @@ def build_extensions(self):
 
 
 require_list = ['cloudpickle', 'psutil', 'pyyaml', 'six']
+test_require_list = ['mock', 'pytest-forked']
```
EnricoMi: `pytest`?

horovod/spark/common/cache.py
```diff
@@ -0,0 +1,64 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from __future__ import absolute_import
+
+import threading
+
+
+class TrainingDataCache(object):
```
EnricoMi: should provide a method `get_key(df, validation, store)` that returns `df.__hash__(), validation, store.get_train_data_path(), store.get_val_data_path()`.

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+        if len(sizes) > 1 and not (SparseVector in all_col_types[col]):
+            raise ValueError(
+                "rows of column {col} have varying sizes. This is only allowed if datatype is "
+                "SparseVector or a mix of Sparse and DenseVector.".format(col=col))
+        col_max_sizes[col] = max(sizes)
+
+    return all_col_types, col_shapes, col_max_sizes
+
+
+def _get_metadata(df):
```
EnricoMi: should be tested in unit tests

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+        for idx, col, output_shape in zip(range(label_count), label_columns, output_shapes):
+            col_size = metadata[col]['shape']
+            output_size = abs(np.prod(output_shape))
+            if col_size != output_size:
+                raise ValueError('Label column \'{col}\' with size {label} must equal that of the '
+                                 'model output at index {idx} with size {output}'
+                                 .format(col=col, label=col_size, idx=idx, output=output_size))
+
+
+def _get_col_info(df):
```
EnricoMi: should be tested in unit tests

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+        return np.float64
+    elif dtype == LongType:
+        return np.int64
+    elif dtype == BooleanType:
+        return np.bool
+    else:
+        raise ValueError('Unrecognized data type: {}'.format(dtype))
+
+
+def check_shape_compatibility(metadata, feature_columns, label_columns,
```
EnricoMi: all errors raised here should be tested on unit tests

examples/keras_spark_mnist.py
tgaddair: I think this solution will work well for most open source users, who just want Store to mirror the behavior of PySpark when reading and writing files.  So I think it's good to simplify things behind this API.

examples/keras_spark_mnist.py
```diff
@@ -0,0 +1,124 @@
+from __future__ import print_function
```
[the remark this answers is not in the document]
tgaddair: My worry with doing that is that users will associate it with TensorFlow estimators or similar, or think of this as a "specialized" API instead of the default (for most users).

horovod/spark/common/store.py
```diff
@@ -0,0 +1,283 @@
...
+        return self._val_path
+
+    def get_test_data_path(self):
+        return self._test_path
+
+    def saving_runs(self):
+        return self._save_runs
+
+    def get_runs_path(self):
+        return self._get_path('runs')
```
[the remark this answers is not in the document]
tgaddair: Yes, good point, let's do that.

horovod/spark/common/store.py
```diff
@@ -24,6 +24,9 @@
 import pyarrow as pa
 
 
+HDFS_PREFIX = 'hdfs://'
```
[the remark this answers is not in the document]
tgaddair: Good point, I refactored it a bit more to better encapsulate this.

setup.py
```diff
@@ -1465,7 +1465,7 @@ def build_extensions(self):
               'numpy',
               'petastorm',
```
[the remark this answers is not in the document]
tgaddair: Good point, fixed.

horovod/spark/common/backend.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def run(self, fn, args=(), kwargs={}, env=None):
+        full_env = self._env or os.environ.copy()
+        if env:
+            full_env.update(env)
+
+        if 'CUDA_VISIBLE_DEVICES' in full_env:
+            # In TensorFlow 2.0, we set this to prevent memory leaks from the client.
+            # See https://github.com/tensorflow/tensorflow/issues/33168
+            del full_env['CUDA_VISIBLE_DEVICES']
```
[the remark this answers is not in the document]
tgaddair: Added more details to the comment.  Basically, we want TensorFlow to only allocate memory on the GPU in the training process.

horovod/spark/common/params.py
```diff
@@ -0,0 +1,300 @@
...
+    label_cols = Param(Params._dummy(), "label_cols", "label column names",
+                       typeConverter=TypeConverters.toListString)
+    validation_col = Param(Params._dummy(),
+                           "validation_col",
+                           "name of a column with 0 or 1 flag to specify using sample for validation",
+                           typeConverter=TypeConverters.toString)
+    callbacks = Param(Params._dummy(), 'callbacks', 'callbacks')
+    batch_size = Param(Params._dummy(), 'batch_size', 'batch size',
+                       typeConverter=TypeConverters.toInt)
+    epochs = Param(Params._dummy(), 'epochs', 'epochs')
```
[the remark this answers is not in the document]
tgaddair: Fixed, changed to `int`.

horovod/spark/common/params.py
```diff
@@ -0,0 +1,300 @@
...
+    train_steps_per_epoch = Param(Params._dummy(), 'train_steps_per_epoch',
```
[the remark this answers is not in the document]
tgaddair: Fixed.

horovod/spark/common/params.py
```diff
@@ -0,0 +1,300 @@
...
+                                  'number of training (batches) steps per epoch')
+    validation_steps_per_epoch = Param(Params._dummy(), 'validation_steps_per_epoch',
```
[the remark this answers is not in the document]
tgaddair: Fixed.

horovod/spark/common/serialization.py
```diff
@@ -0,0 +1,80 @@
...
+
+        basic_metadata = {"class": cls, "timestamp": int(round(time.time() * 1000)),
+                          "sparkVersion": sc.version, "uid": uid, "paramMap": json_params,
+                          "defaultParamMap": json_default_params}
+        if extra_metadata is not None:
+            basic_metadata.update(extra_metadata)
+        return json.dumps(basic_metadata, separators=[',', ':'])
+
+
+class HorovodParamsReader(DefaultParamsReader):
```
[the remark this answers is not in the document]
tgaddair: It comes from `BaseReadWrite`: https://github.com/apache/spark/blob/master/python/pyspark/ml/util.py#L98

horovod/spark/common/serialization.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def load(self, path):
+        metadata = DefaultParamsReader.loadMetadata(path, self.sc)
+        metadata['paramMap'] = self._deserialize_dict(metadata['paramMap'])
+        metadata['defaultParamMap'] = self._deserialize_dict(metadata['defaultParamMap'])
+        py_type = DefaultParamsReader._DefaultParamsReader__get_class(metadata['class'])
+        # Instantiate KerasModel class
```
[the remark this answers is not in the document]
tgaddair: Yes, you're right.  Fixed.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,293 @@
...
+class LocalStore(PrefixStore):
+    def __init__(self, prefix_path, *args, **kwargs):
+        super(LocalStore, self).__init__(prefix_path, *args, **kwargs)
+        self._fs = pa.LocalFileSystem()
+
+    def get_filesystem(self):
+        return self._fs
+
+    def exists(self, path):
+        return os.path.exists(path)
```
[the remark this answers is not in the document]
tgaddair: Good point, fixed.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,293 @@
...
+    def exists(self, path):
+        return self._hdfs.exists(path)
+
+    def read(self, path):
+        with self._hdfs.open(path, 'rb') as f:
+            return f.read()
+
+    def get_petastorm_path_fn(self):
+        def get_path(path):
+            return 'hdfs://' + path
```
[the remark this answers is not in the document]
tgaddair: Fixed.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,293 @@
...
+        self._fs = pa.LocalFileSystem()
+
+    def get_filesystem(self):
+        return self._fs
+
+    def exists(self, path):
+        return os.path.exists(path)
+
+    def read(self, path):
+        with open(path, 'rb') as f:
```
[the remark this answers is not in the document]
tgaddair: Fixed.

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+    # Check for model and input type incompatibility. Columns must have the same size
+    # (total number of elements) of the corresponding inputs.
+    feature_count = len(feature_columns)
+    if input_shapes is not None:
+        if feature_count != len(input_shapes):
+            raise ValueError('Feature column count {features} must equal '
+                             'model inputs count {inputs}'
+                             .format(features=feature_count, inputs=len(input_shapes)))
+
+    if all(metadata[col]['shape'] for col in feature_columns):
```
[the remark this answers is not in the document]
tgaddair: @abditag2 can you clarify the purpose of this line.  Can it be removed?

horovod/spark/common/util.py
tgaddair: Added check.

horovod/spark/common/util.py
tgaddair: Should be safe to remove the global keyword here.

setup.py
tgaddair: Added.

horovod/spark/common/util.py
tgaddair: Yes, good catch.

horovod/spark/common/util.py
tgaddair: No, added check.

horovod/spark/common/util.py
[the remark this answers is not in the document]
EnricoMi: I am not thinking about removing it, otherwise checks below will throw exceptions. I am more wondering if there should be a check that all `feature_columns` and `label_columns` are in `metadata`. Only then it can be removed (here and further down).

horovod/spark/common/store.py
```diff
@@ -0,0 +1,293 @@
...
+    def get_logs_path(self, run_id):
+        raise NotImplementedError()
+
+    def get_checkpoint_filename(self):
+        raise NotImplementedError()
+
+    def get_logs_subdir(self):
+        raise NotImplementedError()
+
+    def exists(self, path):
```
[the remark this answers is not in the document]
tgaddair: In its current form you're definitely right, but I want to keep Store generic enough that it could potentially support storage APIs other than filesystems (like some kind of blob storage DB).  Does that seem reasonable to you?

horovod/spark/common/store.py
```diff
@@ -0,0 +1,293 @@
...
+    def read(self, path):
+        with self._hdfs.open(path, 'rb') as f:
+            return f.read()
+
+    def get_petastorm_path_fn(self):
+        def get_path(path):
+            return 'hdfs://' + path
+        return get_path
+
+    def get_local_output_dir_fn(self, run_id):
```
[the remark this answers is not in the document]
tgaddair: Good point, added optional `temp_dir` arg to allow for customizing the temporary directory.

horovod/spark/common/util.py
tgaddair: I like the idea of merging `validation_split` and `validation_col`.  @abditag2 what do you think?

I'll also add the check on the validation split.  Thanks for the suggestion.

horovod/spark/common/util.py
```diff
@@ -0,0 +1,530 @@
...
+    if not store.exists(train_data_path):
+        raise ValueError("{} path does not exist on the store".format(train_data_path))
+
+    train_data = pq.ParquetDataset(train_data_path, filesystem=store.get_filesystem())
+    schema = train_data.schema.to_arrow_schema()
+
+    train_rows = 0
+    total_byte_size = 0
+    for piece in train_data.pieces:
+        metadata = piece.get_metadata()
```
ConeyLiu: what's the difference between `metadata` and `file_metadata`? It seems equally:
```python3
def get_metadata(self):
        """
        Returns the file's metadata

        Returns
        -------
        metadata : FileMetaData
        """
        f = self.open()
        return f.metadata

    def open(self):
        """
        Returns instance of ParquetFile
        """
        reader = self.open_file_func(self.path)
        if not isinstance(reader, ParquetFile):
            reader = ParquetFile(reader, **self.file_options)
        return reader

```

horovod/spark/common/util.py
```diff
@@ -0,0 +1,530 @@
...
+    if train_rows > total_byte_size:
+        raise ValueError('Found {} bytes in {} rows.  Training dataset may be corrupted.'
+                         .format(total_byte_size, train_rows))
+
+    val_rows = 0
+    if store.exists(validation_data_path):
+        val_data = pq.ParquetDataset(validation_data_path, filesystem=store.get_filesystem())
+        for piece in val_data.pieces:
+            metadata = piece.get_metadata()
+            val_rows += metadata.num_rows
```
ConeyLiu: Do we need to check the validation data as the train data?

horovod/spark/common/util.py
tgaddair: Done.

horovod/spark/common/util.py
```diff
@@ -0,0 +1,539 @@
...
+    elif dtype == LongType:
+        return np.int64
+    elif dtype == BooleanType:
+        return np.bool
+    else:
+        raise ValueError('Unrecognized data type: {}'.format(dtype))
+
+
+def check_shape_compatibility(metadata, feature_columns, label_columns,
+                              input_shapes=None, output_shapes=None):
```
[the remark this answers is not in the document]
tgaddair: Yes, we skip checking `output_shape` for PyTorch, for example.  Fixed the issue of not checking for `None`.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,321 @@
...
+            'checkpoint_filename': self.get_checkpoint_filename(),
+            'logs_subdir': self.get_logs_subdir(),
+            'get_petastorm_path': self.get_petastorm_path_fn(),
+            'get_local_output_dir': self.get_local_output_dir_fn(run_id),
+            'sync': self.sync_fn(run_id)
+        }
+
+    @staticmethod
+    def create(work_dir):
+        if HDFSStore.matches(work_dir):
```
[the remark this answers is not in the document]
tgaddair: Agreed!  Good suggestion.

horovod/spark/common/util.py
tgaddair: Good catch!  Fixed.

horovod/spark/common/util.py
tgaddair: Good point.  Added TODO for followup PR to remove this when schema is all scalar.

horovod/spark/common/util.py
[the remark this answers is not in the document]
tgaddair: Ah, good point.  Added an additional check called `Params._check_params` to assert this condition.

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+            input_size = abs(np.prod(input_shape))
+            if col_size != input_size:
+                raise ValueError(
+                    'Feature column \'{col}\' with size {feature} must equal that of the '
+                    'model input at index {idx} with size {input}'
+                    .format(col=col, feature=col_size, idx=idx, input=input_size))
+
+    label_count = len(label_columns)
+    if output_shapes is not None and all(metadata[col]['shape'] for col in label_columns):
+        if label_count != len(output_shapes):
```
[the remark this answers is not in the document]
tgaddair: Refactored this function to remove these checks and moved the metadata check to `Params._check_params`.

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+            output_size = abs(np.prod(output_shape))
+            if col_size != output_size:
+                raise ValueError('Label column \'{col}\' with size {label} must equal that of the '
+                                 'model output at index {idx} with size {output}'
+                                 .format(col=col, label=col_size, idx=idx, output=output_size))
+
+
+def _get_col_info(df):
+    """
+    Infer the type and shape of all the columns.
```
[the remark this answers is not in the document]
tgaddair: Added note.

horovod/spark/common/util.py
tgaddair: Yes, this can get pretty expensive.  This is also the approached recommended by Databricks for their HorovodRunner.  The good news is Databricks and the Petastorm folks are working on a solution, so should get better in the future.

horovod/spark/common/util.py
tgaddair: This lock is in place in the event that the user wants to train multiple estimators in parallel without corrupting the cache.  Currently, this would be difficult to do, but we've been working on a system that allows you to do this with a thin API.

horovod/spark/common/util.py
```diff
@@ -0,0 +1,518 @@
...
+    """
+
+    def get_meta(row):
+        row_dict = row.asDict()
+        row_schema = []
+        for col_name, data_col in row_dict.items():
+            dtype = type(data_col)
```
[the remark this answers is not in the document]
tgaddair: `data_col` can be `NullType` in PySpark, which is okay.  It cannot, as far as I can tell, be `None`.  Added unit tests to test `NullType`.

horovod/spark/common/util.py
tgaddair: If the column is all `None`, then it will still have size and shape of 1.  Added tests to verify this.

horovod/spark/common/util.py
tgaddair: Added.

horovod/spark/common/util.py
tgaddair: Fixed.

horovod/spark/common/cache.py
tgaddair: Done.

horovod/spark/common/util.py
tgaddair: Done.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,321 @@
...
+        return os.path.join(self.get_run_path(run_id), self.get_logs_subdir()) \
+            if self._save_runs else None
+
+    def get_checkpoint_filename(self):
+        return 'checkpoint.h5'
+
+    def get_logs_subdir(self):
+        return 'logs'
+
+    def normalize_path(self, path):
```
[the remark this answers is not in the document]
tgaddair: Good point.  Renamed to `get_full_path_fn`.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,321 @@
...
+        return self._hdfs
+
+    def exists(self, path):
+        return self._hdfs.exists(path)
+
+    def read(self, path):
+        with self._hdfs.open(path, 'rb') as f:
+            return f.read()
+
+    def get_petastorm_path_fn(self):
```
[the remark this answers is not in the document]
tgaddair: Done.

horovod/spark/common/util.py
tgaddair: Good idea.  Added additional checks for the validation data.

horovod/spark/common/util.py
tgaddair: Hmm, after testing seems there's no difference.  Possible there was a difference in older versions of pyarrow, as the API changes a lot, but since we're pinned to newer versions due to Petastorm, I removed the redundant file_metadata.

horovod/spark/common/util.py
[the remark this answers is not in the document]
tgaddair: Apparently there's a bug in Python 2.7 (not 3) related to pickling NoneType, so added a workaround.

EnricoMi:
Great progress, sorry for keeping you so busy.

I like the general architecture of this PR, how `Estimator` and its implementations and `RemoteTrainer` work together. There were a few assumptions that I made when going through the code, that made it very hard to understand the concepts because the assumptions turned out to be wrong:

* It took me some time to find out that `Store` stores only a single `fit` dataframe. With `fit` dataframe I mean the training and optional validation dataframe, as in `df` given to `fit(df)`.
* Further, the concept that `TrainingDataCache` handles the tuple (dataframe, store) is important to understand the code. The same dataframe but a different store would be created a second time and not be considered "cached".
* An estimator is bound to a store, and since a `Store` stores only a single `fit` dataframe, that store can only be reused by other estimators if they are `fit`ted on the very same dataframe. Or put otherwise, an estimator cannot be used with multiple `fit` dataframes, though `estimator.fit(dataframe)` looks like you could just do that.
* The `Store` seems to store model specific information like checkpoints. So I am not sure if other `Estimator`s should re-use a store (though there is the concept of run-id, which I also did not fully understand the impact and semantics of).

**Maybe some more information should be documented at respected classes.**

However, the current setup of `store`, `TrainingDataCache` and `Estimator` allows to do the following things:

```
keras_estimator1 = hvd.KerasEstimator(store=store)
keras_estimator2 = hvd.KerasEstimator(store=store)

keras_estimator1.fit(train_df2)
keras_estimator2.fit(train_df2)
```

Or use the same estimator for multiple dataframes:
```
keras_estimator.fit(train_df)  // prepares train_df and writes to training path in store, then fits
keras_estimator.fit(train_df2) // will overwrite training path in store, then fits
keras_estimator.fit(train_df)  // cache says train_df is still cached, so prepared train_df2 will be used to fit
```

Maybe the Estimator should not be bound to a single store if `fit` has a side affect on the `Store`. Maybe `fit` should better be `def fit(self, df, store)`.

I think the `fit` dataset and the `Store` form an entity, they should be represented by a class, e.g. `FitDataset`, which is now `Store` plus a single dataframe: `FitDataset(prefix_path, df)`. Then there could be `LocalFitDataset` and `HDFSFitDataset`.

Unfortunately, the store also stores model-specific information, e.g. the checkpoints, which somehow belong to the `fit` dataset as well. So the `Store` is also a `ModelStore`, which stores model specific information. You could argue that the `FitDataset` belongs to the model, but then you could not reuse the prepared `FitDataset` across alternative models.

So I think `Store` should be split up into the model-specific bit and the `fit` dataset bit. Then `Estimator(model_store, fit_dataset).fit()` allows:
```
fit_dataset = FitDataset(path, df)

keras_estimator1 = hvd.KerasEstimator(model_store=model_store1, fit_dataset=fit_dataset)
keras_estimator2 = hvd.KerasEstimator(model_store=model_store2, fit_dataset=fit_dataset)

keras_estimator1.fit()
keras_estimator2.fit()
```

The line `fit_dataset = FitDataset(path, df)` could prepare the dataset when it does not exist, so no cache is required anymore and the `fit` methods can be called concurrently.

**So the `Estimator`, `Store` and `fit` dataframe all form a difficult relationship which I think needs some refactoring or much better documentation and guarding so they are not accidentally misused.**

horovod/spark/common/store.py
```diff
@@ -0,0 +1,314 @@
...
+            'logs_path': self.get_logs_path(run_id),
+            'checkpoint_filename': self.get_checkpoint_filename(),
+            'logs_subdir': self.get_logs_subdir(),
+            'get_full_path': self.get_full_path_fn(),
+            'get_local_output_dir': self.get_local_output_dir_fn(run_id),
+            'sync': self.sync_fn(run_id)
+        }
+
+    @staticmethod
+    def create(work_dir):
```
EnricoMi: `work_dir` should read `prefix_path`, as `work_dir` is a concept of the application, e.g. `keras_spark_rossmann.py`.

tgaddair: Good catch.  Fixed.

horovod/spark/common/util.py
```diff
@@ -0,0 +1,538 @@
...
+                    raise ValueError(
+                        'Validation DataFrame does not any samples with validation param {}'
+                        .format(validation))
+                if verbose:
+                    print('val_rows={}'.format(val_rows))
+
+            _training_cache.put(key, (train_rows, val_rows, metadata, avg_row_size, dataset_idx))
+
+    try:
+        yield train_rows, val_rows, metadata, avg_row_size, dataset_idx
```
EnricoMi: The `key` could map to `dataset_idx` and `dataset_idx` to `train_rows, val_rows, metadata`, then `yield dataset_idx` is sufficient here and calling code can get this through extra method `get_dataset_properties(dataset_idx)`

The `key` is deterministic given `df` and `store`. But this determinism is currently not needed, as it maps to a pseudo-random `dataset_idx`. So that `dataset_idx` feels like the actual key that the cache uses. Might all simplify if you consider cross-application `caching` feature.

horovod/spark/common/cache.py
```diff
@@ -0,0 +1,81 @@
...
+    def create_key(self, df, store, validation):
+        return df.__hash__(), store.get_train_data_path(), store.get_val_data_path(), validation
+
+    def set_in_use(self, key, in_use):
+        if in_use:
+            self._keys_in_use[key] += 1
+        else:
+            self._keys_in_use[key] -= 1
+
+    def create_data_paths(self, key, store):
```
EnricoMi: this method is not thread safe, should be marked as such

horovod/spark/common/cache.py
```diff
@@ -0,0 +1,81 @@
...
+        _, _, _, validation = key
+        idx = 0
+        while True:
+            train_data = store.get_train_data_path(idx)
+            val_data = store.get_val_data_path(idx)
+
+            last_key = self._data_to_key.get((train_data, val_data))
+            if self._keys_in_use[last_key] > 0 and \
```
EnricoMi: When the same dataframe is used first with `validation = False` and later with `validation = True`, then the `idx` of the first one will be reused though the key is still in use. Training data will be overwritten while they are used.

Is there an expected situation where `last_key` is in use and `idx` should be used? Shouldn't always be `idx += 1` when `last_key` is in use? Given there is a `@context_manager` around `prepare_data`.

horovod/spark/common/cache.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def set_in_use(self, key, in_use):
+        if in_use:
+            self._keys_in_use[key] += 1
+        else:
+            self._keys_in_use[key] -= 1
+
+    def create_data_paths(self, key, store):
+        _, _, _, validation = key
+        idx = 0
```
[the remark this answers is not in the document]
tgaddair: We do have support for this using the `fit_on_parquet` method of the estimator.  It does require that the user setup their store to explicitly reference the correct saved DataFrame path.

Databricks is working on an API that will simplify the DataFrame -> Petastorm Parquet conversion so that it should be easier to perform in two steps.  Though as you mentioned, users can choose to use `prepare_data` for now.

horovod/spark/common/cache.py
tgaddair: Good point.  I was originally thinking that we could use an earlier index if the data was deleted by some external system, but that adds additional complexity without any apparent benefit.  Removed.

horovod/spark/common/cache.py
tgaddair: Added comment on usage.

horovod/spark/common/util.py
tgaddair: Yes, I like that design better.  I've refactored the cache to have a cleaner separation between keys and dataset indices.

tgaddair:
@EnricoMi, thank you for the excellent feedback. Please take another look when you get a chance.

Regarding your earlier point about the coupling of DataFrames and stores, I agree that the behavior you pointed out was unintuitive and not by design. In the new design, we create separate datasets to ensure that such coupling does not occur (stores are treated as simply an API to access paths in a data storage layer).

Regarding the API changes you suggested, one thing we are forced into by basing this on the Estimator API is that `fit` can only take a DataFrame and optional parameter map as input.  This is to ensure that these Estimators can be used as Pipelines where we do not have control over the calls to fit directly.

Hopefully the new design should result in more expected behavior, let me know what you think.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,334 @@
...
+    def create(prefix_path):
+        if HDFSStore.matches(prefix_path):
+            return HDFSStore(prefix_path)
+        else:
+            return LocalStore(prefix_path)
+
+
+class PrefixStore(Store):
+    def __init__(self, prefix_path, train_path=None, val_path=None, test_path=None, runs_path=None, save_runs=True):
+        self.prefix_path = self.get_normalized_path(prefix_path)
```
ConeyLiu: why should we nomalize the `prefix_path`? Given `prefix_path` as 'hdfs://horovod:8888/path', after normalized the `prefix_path` should be `horovod:8888/path`. Then `train_path` will be 'horovod:8888/path/train_data'. Should this as expect? I guess this path should not be work for spark.write.

horovod/spark/common/util.py
```diff
@@ -0,0 +1,546 @@
...
+            if verbose:
+                print('using cached dataframes for key: {}'.format(key))
+                print('train_data_path={}'.format(train_data_path))
+                print('train_rows={}'.format(train_rows))
+                print('val_data_path={}'.format(val_data_path))
+                print('val_rows={}'.format(val_rows))
+        else:
+            dataset_idx = _training_cache.next_dataset_index(key)
+            train_data_path = store.get_train_data_path(dataset_idx)
+            val_data_path = store.get_val_data_path(dataset_idx)
```
ConeyLiu: The `train_data_path ` and `val_data_path ` should be appended by `filesystem_prefix`, by calling `store.get_full_path` ?

horovod/spark/common/store.py
ConeyLiu: I see, should wrap with `get_full_path `.

test/test_spark_torch.py
```diff
@@ -0,0 +1,356 @@
...
+            backend = CallbackBackend()
+            with local_store() as store:
+                store.get_train_data_path = lambda v=None: store._train_path
+                store.get_val_data_path = lambda v=None: store._val_path
+
+                with util.prepare_data(backend.num_processes(),
+                                       store,
+                                       df,
+                                       feature_columns=['features'],
+                                       label_columns=['y']):
```
ConeyLiu: should we respect the verbose settings?

horovod/spark/common/util.py
```diff
@@ -0,0 +1,546 @@
...
+    elif validation:
+        raise ValueError('Unrecognized validation type: {}'.format(type(validation)))
+
+    return train_df, val_df, validation_ratio
+
+
+@contextlib.contextmanager
+def prepare_data(num_processes, store, df, label_columns, feature_columns,
+                 validation=None, sample_weight_col=None,
+                 partitions_per_process=10, verbose=False):
```
ConeyLiu: maybe here `verbose` use `int` for consistent with user settings.

horovod/spark/keras/estimator.py
```diff
@@ -0,0 +1,520 @@
...
+        partitions_per_process = self.getPartitionsPerProcess()
+
+        with util.prepare_data(backend.num_processes(),
+                               store,
+                               df,
+                               label_columns=label_columns,
+                               feature_columns=feature_columns,
+                               validation=validation,
+                               sample_weight_col=sample_weight_col,
+                               partitions_per_process=partitions_per_process) as dataset_idx:
```
ConeyLiu: same as here

horovod/spark/common/cache.py
```diff
@@ -0,0 +1,91 @@
...
+import threading
+
+
+class TrainingDataCache(object):
+    def __init__(self):
+        self.lock = threading.Lock()
+        self._reset()
+
+    def create_key(self, df, store, validation):
+        return df.__hash__(), store.get_train_data_path(), store.get_val_data_path(), validation
```
ConeyLiu: We should use the full path here. Because `hdfs://path` and `file://path` can get the same key

horovod/spark/common/util.py
```diff
@@ -0,0 +1,546 @@
...
+            train_partitions = max(int(num_partitions * (1.0 - validation_ratio)),
+                                   num_processes)
+            if verbose:
+                print('train_partitions={}'.format(train_partitions))
+
+            train_df \
+                .coalesce(train_partitions) \
+                .write \
+                .mode('overwrite') \
+                .parquet(train_data_path)
```
ConeyLiu: what about saving failed?  The path should be created if failed. Should we wrap this with `try catch` and delete the failed path.

horovod/spark/common/util.py
```diff
@@ -0,0 +1,546 @@
...
+                val_partitions = max(int(num_partitions * validation_ratio),
+                                     num_processes)
+                if verbose:
+                    print('val_partitions={}'.format(val_partitions))
+
+                val_df \
+                    .coalesce(val_partitions) \
+                    .write \
+                    .mode('overwrite') \
+                    .parquet(val_data_path)
```
ConeyLiu: same as here

horovod/spark/common/store.py
```diff
@@ -0,0 +1,334 @@
...
+        self._train_path = train_path or self._get_path('train_data')
+        self._val_path = val_path or self._get_path('val_data')
+        self._test_path = test_path or self._get_path('test_path')
+        self._runs_path = runs_path or self._get_path('runs')
+        self._save_runs = save_runs
+        super(PrefixStore, self).__init__()
+
+    def get_train_data_path(self, idx=None):
+        return '{}.{}'.format(self._train_path, idx) if idx is not None else self._train_path
```
ConeyLiu: what about a more readable name? such as `{}_horovod_spark_estimator_intermediate_file_{}`

horovod/spark/common/util.py
```diff
@@ -0,0 +1,546 @@
...
+    if total_rows > total_byte_size:
+        raise ValueError('Found {} bytes in {} rows; {} dataset may be corrupted.'
+                         .format(total_byte_size, total_rows, dataset_id))
+
+    return total_rows, total_byte_size
+
+
+def get_simple_meta_from_parquet(store, label_columns, feature_columns, sample_weight_col, dataset_idx=None):
+    train_data_path = store.get_train_data_path(dataset_idx)
+    validation_data_path = store.get_val_data_path(dataset_idx)
```
ConeyLiu: here should also prepend with schema

horovod/spark/common/store.py
tgaddair: You're correct, this should be using the full path.  I've refactored Store to work with full paths by default, and only use the shortened localized paths in specific situations where it's needed (like interacting with a Pyarrow Filesystem object).

horovod/spark/common/util.py
tgaddair: Yes, good point.  Fixed.  See above for details.

test/test_spark_torch.py
tgaddair: You mean from pytest?  I am not sure of an easy way to do that, but happy to do so if you know how!

horovod/spark/common/cache.py
tgaddair: Good point.  Fixed.

horovod/spark/common/util.py
tgaddair: Fixed.

horovod/spark/common/util.py
tgaddair: Done.

horovod/spark/keras/estimator.py
tgaddair: Fixed.

test/test_spark_torch.py
tgaddair: Ah, I think you mean in the TorchEstimator.  Fixed.

horovod/spark/common/store.py
tgaddair: Renamed to `intermediate_train_data` (etc.).  In most cases, the user should specify a path that is unique to their Horovod training (for output data like checkpoints), so I don't think we need to be too verbose.

horovod/spark/common/util.py
tgaddair: I think it would be difficult to cleanup this path, because during writing Spark will write to a unique temporary file (if I remember correctly), then move it when writing completes.  I am not sure how to get the name of that temp file.

In most cases, it shouldn't be an issue because we will reuse the same paths for successive attempts.

horovod/spark/common/util.py
tgaddair: See above.

horovod/spark/common/util.py
ConeyLiu: OK, that makes sense.

horovod/spark/common/store.py
ConeyLiu: reasonable

horovod/spark/common/util.py
```diff
@@ -0,0 +1,545 @@
...
+
+    def merge(x, y):
+        dtypes, shapes, sizes = x
+        new_dtypes, new_shapes, new_sizes = y
+        dtypes.update(new_dtypes)
+        shapes.update(new_shapes)
+        sizes.update(new_sizes)
+        return dtypes, shapes, sizes
+
+    raw_col_info_list = df.rdd.flatMap(get_meta).reduceByKey(merge).collect()
```
ConeyLiu: I think we should not collect this data directly. It should be a big data if the data format is `SparseVector`. The following check and merge process should be done in spark. We just need to collect the data `pair (shape, max_size)`.

tgaddair: Good point. I changed the `merge` function to only track the min and max sizes and shapes, which will give us all the same information (max size, are the shapes and sizes all uniform), but requires only tracking at most two elements instead of unbounded.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,338 @@
...
+            'checkpoint_path': self.get_checkpoint_path(run_id),
+            'logs_path': self.get_logs_path(run_id),
+            'checkpoint_filename': self.get_checkpoint_filename(),
+            'logs_subdir': self.get_logs_subdir(),
+            'get_local_output_dir': self.get_local_output_dir_fn(run_id),
+            'sync': self.sync_fn(run_id)
+        }
+
+    @staticmethod
+    def create(prefix_path):
```
ConeyLiu: we should match this arguments list as HDFSStore.__ini__

horovod/spark/common/store.py
```diff
@@ -0,0 +1,338 @@
...
+                    with open(local_path, 'rb') as f:
+                        hdfs.upload(hdfs_path, f)
+                    uploaded[local_path] = modified_ts
+
+        return fn
+
+    def _get_filesystem_fn(self):
+        host = self._host
+        port = self._port
+        user = self._user
```
ConeyLiu: If `user` is empty, we should use the current os user name.

horovod/spark/common/store.py
tgaddair: Added `*args` and `**kwargs` to match all arguments.

horovod/spark/common/store.py
tgaddair: This will be handled by Pyarrow.  In order to prevent users misusing the user variable in some way, I changed this to a `self._hdfs_kwargs` variable to encapsulate all these args.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,338 @@
...
+        self.prefix_path = self.get_full_path(prefix_path)
+        self._train_path = train_path or self._get_path('intermediate_train_data')
+        self._val_path = val_path or self._get_path('intermediate_val_data')
+        self._test_path = test_path or self._get_path('intermediate_test_data')
+        self._runs_path = runs_path or self._get_path('runs')
+        self._save_runs = save_runs
+        super(PrefixStore, self).__init__()
+
+    def exists(self, path):
+        return self.get_filesystem().exists(self.get_localized_path(path))
```
ConeyLiu: I have test this locally, it reported errors. So here and following should be full path?

tgaddair: Hey @ConeyLiu, what was the error, and what was the path that was passed in?  Using full paths do not work with the PyArrow `HadoopFilesystem` class.

tgaddair: I think I see where your error is coming from.  I'm assuming you're using full URLs like `hdfs://namenode01:8020/user/test/horovod`.  I updated the `HDFSStore` to work with these strings for the `prefix_path`, let me know if it solves your issue.

EnricoMi:
> @EnricoMi, thank you for the excellent feedback. Please take another look when you get a chance.
I will go over it once more Monday morning. Good work!

horovod/spark/common/store.py
ConeyLiu: sorry for the late reply. It works great.

horovod/spark/torch/estimator.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+class TorchEstimatorParamsReadable(MLReadable):
+    @classmethod
+    def read(cls):
+        """Returns a DefaultParamsReader instance for this class."""
+        return TorchEstimatorParamsReader(cls)
+
+
+class TorchEstimator(Estimator, EstimatorParams, TorchEstimatorParamsWritable,
```
EnricoMi: has 4 methods in common with `KerasEstimator`: `_get_or_create_backend`, `fit_on_parquet`, `_fit_on_parquet`, `_fit`, `_has_checkpoint`. Maybe worth moving them into common base class.

setup.py
```diff
@@ -1444,6 +1445,7 @@ def build_extensions(self):
 
 
 require_list = ['cloudpickle', 'psutil', 'pyyaml', 'six']
+test_require_list = ['mock', 'pytest', 'pytest-forked']
```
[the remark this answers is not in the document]
tgaddair: Good catch, fixed.  `pytest-forked` is necessary if the user wants to run `test_spark_keras.py` and `test_spark_torch.py` at once in the same process.  This is due to conflicts between importing TensorFlow, PyTorch, and Pyarrow together in the same env.

docs/spark.rst
```diff
@@ -3,13 +3,130 @@
...
+processing (from Spark DataFrames to deep learning datasets), model training loop, model checkpointing, metrics
+collection, and distributed training.
+
+We recommend using Horovod Spark Estimators if you:
+
+* Are using Keras (``tf.keras`` or ``keras``) or PyTorch for training.
+* Want to train directly on a Spark DataFrame from ``pyspark``.
+* Are using a standard gradient descent optimization process as your training loop.
+
+If for whatever reason the Estimator API does not meet your needs, the Run API offers more find-grained control.
```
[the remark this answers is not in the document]
tgaddair: Fixed.

horovod/spark/common/cache.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def set_in_use(self, key, in_use):
+        if in_use:
+            self._keys_in_use[key] += 1
+        else:
+            self._keys_in_use[key] -= 1
+
+    def next_dataset_index(self, key):
+        """Finds the next available `dataset_idx` given the (key, store) pair.
```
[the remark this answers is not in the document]
tgaddair: You are correct, removed.

horovod/spark/common/cache.py
```diff
@@ -0,0 +1,91 @@
...
+
+class TrainingDataCache(object):
+    def __init__(self):
+        self.lock = threading.Lock()
+        self._reset()
+
+    def create_key(self, df, store, validation):
+        return df.__hash__(), store.get_train_data_path(), store.get_val_data_path(), validation
+
+    def set_in_use(self, key, in_use):
```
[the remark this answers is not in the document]
tgaddair: I like it!  Done.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,385 @@
...
+    def get_local_output_dir_fn(self, run_id):
+        run_path = self.get_localized_path(self.get_run_path(run_id))
+
+        @contextlib.contextmanager
+        def local_run_path():
+            if not os.path.exists(run_path):
+                try:
+                    os.makedirs(run_path)
+                except OSError:
+                    # Race condition from workers on the same host: ignore
```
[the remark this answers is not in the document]
tgaddair: Added check if directory exists.  Permissions errors will now get bubbled up correctly.  I like the idea of making it only user writable, so set to `0x755`.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,385 @@
...
+                    pass
+            yield run_path
+
+        return local_run_path
+
+    def sync_fn(self, run_id):
```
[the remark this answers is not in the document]
tgaddair: Added comment explaining that this function is a no-op because `local_run_path` is the same as `root_path`.  Added assertion to enforce this.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,385 @@
...
+    def exists(self, path):
+        raise NotImplementedError()
+
+    def read(self, path):
+        raise NotImplementedError()
+
+    def get_local_output_dir_fn(self, run_id):
+        raise NotImplementedError()
+
+    def sync_fn(self, run_id):
```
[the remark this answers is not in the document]
tgaddair: Added.

horovod/spark/common/util.py
```diff
@@ -0,0 +1,564 @@
...
+        val_rows = val_df.rdd.countApprox(timeout=timeout_ms, confidence=confidence)
+        validation_ratio = val_rows / (val_rows + train_rows)
+    elif validation:
+        raise ValueError('Unrecognized validation type: {}'.format(type(validation)))
+
+    return train_df, val_df, validation_ratio
+
+
+@contextlib.contextmanager
+def prepare_data(num_processes, store, df, label_columns, feature_columns,
```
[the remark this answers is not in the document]
tgaddair: Done.

horovod/spark/common/cache.py
```diff
@@ -0,0 +1,91 @@
...
+    def get_dataset(self, key):
+        return self._key_to_dataset[key]
+
+    def get_dataset_properties(self, dataset_idx):
+        return self._dataset_properties[dataset_idx]
+
+    def set_dataset_properties(self, dataset_idx, props):
+        self._dataset_properties[dataset_idx] = props
+
+    def is_cached(self, key, store):
```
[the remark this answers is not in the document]
tgaddair: Clarified with comment, let me know if this makes sense:

```Returns true if the key is in the cache and its paths exist in the store already.```

File: horovod/spark/torch/estimator.py (hunk 66 above)
Comment (tgaddair) [reply horovod/spark/torch/estimator.py]: Good point.  Abstracted into `HorovodEstimator` in `horovod.spark.common.estimator`.

File: docs/spark.rst [hunk 76]
```diff
@@ -3,13 +3,130 @@
...
+format interpretable by the training framework, and distributing the training using Horovod.  The user only needs to
+provide a Keras or PyTorch model, and the Estimator will do the work of fitting it to the DataFrame.
+
+After training, the Estimator returns a Transformer representation of the trained model.  The model transformer can
+be used like any Spark ML transformer to make predictions on an input DataFrame, writing them as new columns in the
+output DataFrame.
+
+Estimators can be used to track experiment history through model checkpointing, hot-start retraining, and metric
+logging (for Tensorboard) using the Estimator ``Store`` abstraction.  Stores are used for persisting all training
+artifacts including intermediate representations of the training data.  Horovod comes included with stores for HDFS
```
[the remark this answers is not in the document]
tgaddair: I agree, changed to `Horovod natively supports stores for HDFS and local filesystems`.

horovod/spark/common/store.py
```diff
@@ -0,0 +1,385 @@
...
+        self.prefix_path = self.get_full_path(prefix_path)
+        self._train_path = train_path or self._get_path('intermediate_train_data')
+        self._val_path = val_path or self._get_path('intermediate_val_data')
+        self._test_path = test_path or self._get_path('intermediate_test_data')
+        self._runs_path = runs_path or self._get_path('runs')
+        self._save_runs = save_runs
+        super(PrefixStore, self).__init__()
+
+    def exists(self, path):
+        return self.get_filesystem().exists(self.get_localized_path(path))
```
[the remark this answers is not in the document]
tgaddair: Changed usage in the `TrainingCache` to check new store method `is_parquet_dataset`, which will try to create a PyArrow `ParquetDataset` object.  If the parquet files are malformed then an exception will be raised, which will be caught to return False.

Other usages of `exists` can be for files other than Parquet datasets, or where specific exceptions should be raised.

docs/spark.rst
```diff
@@ -46,34 +163,15 @@ A toy example of running a Horovod job in Spark is provided below:
...
-To run the example, please install the following dependencies:
-
-*  ``pyspark``
-*  ``petastorm >= 0.7.0``
-*  ``h5py >= 2.9.0``
-*  ``tensorflow-gpu >= 1.12.0`` (or ``tensorflow >= 1.12.0``)
-*  ``horovod >= 0.15.3``
-
-Run the example:
+A more complete example can be found in `keras_spark_rossmann_advanced.py <../examples/keras_spark_rossmann_advanced.py>`__, which
```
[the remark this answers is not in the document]
tgaddair: Renamed to `keras_spark_rossmann_estimator.py` and `keras_spark_rossmann_run.py`.

tgaddair:
Thanks for the feedback @EnricoMi!  Addressed your comments and added docs explaining how to use `fit_on_parquet`.

horovod/spark/common/store.py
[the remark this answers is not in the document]
EnricoMi: I thought `local_run_path` contains the run id and is unique per estimator run, whereas `root_path` is the root of the `PrefixStore`? Please correct me if I am wrong.

horovod/spark/common/store.py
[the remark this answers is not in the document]
EnricoMi: Can this be made configurable? Or is the permissions of the store's root set by the user sufficient to restrict any read and write access as desired? It probably is.

horovod/spark/common/store.py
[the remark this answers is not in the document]
EnricoMi: Due to the atomic move, there will be no malformed parquet files if writing fails. For `ParquetDataset`, the existing path will be empty (it should ignore the `_temporary` subdirectory), which should translate into an empty parquet file. This is all `is_parquet_dataset` should / will see in that situation.

horovod/spark/common/store.py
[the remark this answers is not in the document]
tgaddair: Yes, if the user does not expose read/execute permissions on parent directories, then they will not be able to read these files either.  I think that's sufficient.

horovod/spark/common/store.py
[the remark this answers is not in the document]
tgaddair: The name was being overloaded.  I renamed it to `local_run_path` in `sync_fn`.  Hopefully that clarifies things.

horovod/spark/common/store.py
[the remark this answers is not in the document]
tgaddair: I've tested this with `ParquetDataset`.  If there is a `_temporary` directory underneath the root of the path, then creating the `ParquetDataset` will fail and `is_parquet_dataset` will return `False`.  That's because a subdirectory is only considered to be valid if it defines a partition of the form `/some/path/dataset/split=key`.  Because `_temporary` does not have this form, it is not considered to be a valid Parquet dataset.

tgaddair:
Hey @EnricoMi, thanks again for the quick review.  Addressed comments, let me know what you think.

horovod/spark/common/store.py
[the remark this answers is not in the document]
EnricoMi: sounds like pyarrow deviates here from how scala `spark.read.parquet` works, but sounds like it is well tested on your end

tgaddair:
@EnricoMi thanks so much for the detailed review.  I agree, the code is in a much better place now, and for a feature this big we definitely want to make sure we get the APIs right.

jhagege:
@tgaddair Hi, looks great.
I noticed you are using the pytest-forked to run tests in a Deep Learning environment.
I am looking for a pattern to get rid of the forked plugin, but can't do so so far since running without it leaves some "state" between each run of a sequence of tests, therefore making tests not reliable to run serially.
The reason I am not a big fan of forked plugin is that it makes it uneasy to troubleshoot when there are failing tests due to the very scarce logs that get outputted (most of them get captured by pytest-forked).
I'm wondering if you found any workaround to any of the two issues. 
Happy to exchange ideas on the topic, thanks much for any help.

PR #1571, opened 2019-12-06
Fix Join op bugs with cache and fusion.

kit1980:
After testing Join Op (original PR https://github.com/horovod/horovod/pull/1058) more thoroughly, I found several issues related to fusion and caching. This PR adds fixes the issues and adds more extensive tests.


kit1980:
@romerojosh, could you take a look at this? @tgaddair suggested to ask you for review, because the PR is cache-related.

kit1980:
There is some strange interaction between this and `test_delta_optimizer` test. If I remove that test, everything passes, otherwise `test_horovod_join_allreduce` results in "Segmentation fault Address not mapped". Investigating.

kit1980:
> There is some strange interaction between this and test_delta_optimizer test. If I remove that test, everything passes, otherwise test_horovod_join_allreduce results in "Segmentation fault Address not mapped". Investigating.

This is fixed.

kit1980:
Not sure why the CI is failling on this, every test seems to pass, but in the end`Error: The command exited with status 123`
Daily build is also failing since yesterday, but in a different way.

tgaddair:
@kit1980, looks like builds failures are coming from a segfault [here](https://buildkite.com/horovod/horovod/builds/1527#e0c9e6b6-4561-4a0b-9d0f-7566bdb3b0b7/209-1827) in `test_horovod_allreduce_error`.  You can see from this successful test [here](https://buildkite.com/horovod/horovod/builds/1516#0015b888-4c89-4467-8cbf-159f863c1706/207-2150) that it should be raising a precondition error [here](https://github.com/horovod/horovod/blob/master/horovod/common/controller.cc#L422), but instead it is going forward with the allreduce, but failing.

horovod/common/controller.cc
```diff
@@ -324,7 +333,7 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down,
       if ((response.response_type() == Response::ResponseType::ALLREDUCE ||
            response.response_type() == Response::ResponseType::ADASUM) &&
           (int)response.devices().size() == size_) {
-        response_cache_.put(response, tensor_queue_);
+        response_cache_.put(response, tensor_queue_, state.joined);
```
romerojosh: I think we need to be careful here in the cache handling. I see that if you have this `state.joined` flag set, the cache doesn't save any of the parameters from the tensor table entry that are not normally stored within the responses (`device`, `dtype`, `shape`), since you know that the tensors won't be in the table. This is tricky because it now puts the cache in an inconsistent state across workers (for a particular response, some workers will have the additional tensor information, while others will not). We need to make sure that when the join operation is concluded, the cache is in a globally consistent state across all workers. 

For example, will this work in the following scenarios? (assume there are two ranks, either joining or submitting tensor A, each tuple is one set of operations submitted):
1. (join, A), (join, join) (A, A)
2. (A, A), (join, A), (join, join), (A, A)

The difference between the two cases is that in the first case, worker 0 does not have a preexisting cache entry for A, while in case 2, worker 0 does have a prexisting cached entry. 

On a related note, I can't seem to find where `state.joined` is reset back to false?

horovod/common/controller.cc
```diff
@@ -156,7 +164,8 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down,
 
   bool need_communication = true;
   if (response_cache_.capacity() > 0 &&
-      !cache_coordinator.uncached_in_queue()) {
+      !cache_coordinator.uncached_in_queue() &&
+      !cache_coordinator.just_joined()) {
```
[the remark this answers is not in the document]
kit1980: Join is different from other ops because it's done rank-by-rank, at different times.
All ranks need to communicate if any other rank just sent Join request, because this may be the last joined rank and Join response will be sent by the communicator.
These other ranks will not have the Join request in their own queue, so the cache is used to let them know someone else just joined.

horovod/common/controller.cc
kit1980: `state.joined` reset back to false on execute:
https://github.com/horovod/horovod/blob/master/horovod/common/ops/collective_operations.cc#L222

horovod/common/controller.cc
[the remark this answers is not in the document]
romerojosh: Thanks for the explanation! I still don't think the additional status is necessary though. If a worker has a Join request in their queue, it will set the `uncached_in_queue()` status flag since the Join requests are not cached. This will trigger the communication path with the coordinator when any worker has a Join request in their queue. If it is the last joined rank, the `uncached_in_queue()` flag will be set, the coordinator communication will trigger, and the Join response should be created/sent to all workers. This seems to result in the same behavior as the `just_joined()` flag.

horovod/common/controller.cc
kit1980: > (join, A), (join, join) (A, A)
> (A, A), (join, A), (join, join), (A, A)

I think I need to explain the Join op more.

After a rank sends Join request, it does not send any horovod requests (can still communicate with the coordinator, but with empty request list) until it gets Join response (which is sent by the coordinator after all ranks sent Join request). After getting Join response, every rank reset their Join state and continue as normal.

So "(join, A), (join, join) (A, A)" is not possible, but "(join, A), (-, join), (A, A)" is possible, there will be communication of every of these 3 steps.

"(A, A), (join, A), (join, join), (A, A)" will be "(A, A), (join, A), (-, join), (A, A)". Second step will use cache, and on the 0-th rank the cache info for the tensor A will be updated: `params.joined = true` and params.shape will be empty https://github.com/horovod/horovod/pull/1571/files#diff-9b02a25ea514cc98deff8de867ffd54dR177. So on the forth step the shape of the A tensor in the 0-th rank cache will be different from the shape in the request, so cache invalidation will happen.

horovod/common/controller.cc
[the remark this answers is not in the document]
kit1980: I think you're correct. But I need to verify/implement "set the uncached_in_queue() status flag" part - I don't think it works for Join requests now. This should be easy and doesn't require additional flags.

horovod/common/controller.cc
romerojosh: Your notation is much clearer. I was just trying to indicate that a rank was still in a join state, not that it'd repeatedly send join requests.

So to focus on the `(A, A), (join, A), (-, join), (A, A)` case:
- "Second step will use cache, and on the 0-th rank the cache info for the tensor A will be updated: `params.joined = true` and params.shape will be empty" <-- Does this operate correctly given this block of code? https://github.com/horovod/horovod/blob/9723db4302d8692a0c2631a0c4c83d4be70e002b/horovod/common/response_cache.cc#L102
In the `put_` method, the existing cache entry is just moved if the cache state is HIT. Since A already exists in the cache, I don't think the params will be updated. 
- I don't see how the `params.joined = true` state is reset. On the cycle that rank 1 submits the join, at the time the responses are recached here:
https://github.com/horovod/horovod/blob/9723db4302d8692a0c2631a0c4c83d4be70e002b/horovod/common/controller.cc#L336
I think both workers will have `state.joined = true` still right, since the state isn't reset until the after the operation is executed?

horovod/common/controller.cc
```diff
@@ -98,6 +100,12 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down,
     }
   }
 
+  if (state.joined && response_cache_.capacity() > 0) {
+    for (uint32_t bit : response_cache_.list_all_bits()) {
+      cache_coordinator.record_hit(bit);
```
romerojosh: Noticed a subtle issue with this logic here to set all cache bits if `state.joined = true`. On the cycle where the last worker submits a join, all workers will signal that all responses in the cache are hit, and will proceed to retrieve process every response in the cache. The extra status flag may be useful for this case, except you'd want to catch if all workers are in a join state, and if so, erase all cache hits.

kit1980: > Noticed a subtle issue with this logic here to set all cache bits if state.joined = true. On the cycle where the last worker submits a join, all workers will signal that all responses in the cache are hit, and will proceed to retrieve process every response in the cache. The extra status flag may be useful for this case, except you'd want to catch if all workers are in a join state, and if so, erase all cache hits.

On the cycle where the last worker submits a join, `need_communication` is true, so responses from hits are generated on the coordinator, which knows the `joined_size` and can just not generate these responses.

kit1980:
Hi @romerojosh,
Thank you for looking into details of the PR.
I've just pushed an update based on your comments:
- don't use any extra status flags, instead just use `uncached_in_queue`
- don't use `params.joined` for cache, instead create zero-tensors and use their info
- prevent generating responses for all tensors in cache on the cycle when all ranks join

Also fixed several fusion-related issues.
Please take another look.

kit1980:
All CI tests are passing now.

horovod/common/controller.cc
```diff
@@ -121,6 +128,9 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down,
     // a shutdown. This function removes any invalid cache entries, if they
     // exist.
     CoordinateCacheAndState(cache_coordinator);
+    if (state.joined && !cache_coordinator.invalid_bits().empty()) {
+      cache_coordinator.set_uncached_in_queue(true);
+    }
```
romerojosh: This code here is setting the uncached_in_queue flag *after* the cache_coordinator has been synced in `CoordinateCacheAndState`, which is not good to do. However, I don't think this code is necessary. In the case where there is an invalid bit set, `uncached_in_queue` is already set during the processing of local requests in this else case here: https://github.com/horovod/horovod/blob/0d34859957b79ab07fe39bb4f755cefa7d84c986/horovod/common/controller.cc#L89
If there are any entries to be invalidated, this `uncached_in_queue` should already be true.

horovod/common/response_cache.cc
```diff
@@ -55,8 +55,8 @@ ResponseCache::CacheState ResponseCache::cached(const Request& message) const {
     uint32_t cache_bit = it->second;
     auto& cache_params = std::get<1>(*cache_iters_[cache_bit]);
     return (cache_params.device == message.device() &&
-            cache_params.dtype == message.tensor_type() &&
-            cache_params.shape == message.tensor_shape())
+                cache_params.dtype == message.tensor_type() &&
```
romerojosh: Nit: Remove extra whitespace.

kit1980: done

horovod/common/controller.cc
kit1980: Thanks. Definitely not needed, forgot to delete after some experiments. Removed.

kit1980:
@romerojosh , I printed out cache parameters for the tensor for the 2 cases.
1. (join, A), (-, join) (A, A) - empty cache on both ranks for all steps

2. (A, A), (join, A), (-, join), (A, A)
After first step and for the rest of the steps:
device 0 dtype 7 shape 5 5
device 1 dtype 7 shape 5 5

horovod/common/controller.cc
```diff
@@ -632,30 +646,36 @@ ResponseList Controller::FuseResponses(std::deque<Response>& responses) {
     responses.pop_front();
     int64_t tensor_size = 0;
     DataType dtype;
-    if (response.response_type() == Response::ResponseType::ALLREDUCE || 
+    if (response.response_type() == Response::ResponseType::ALLREDUCE ||
         response.response_type() == Response::ResponseType::ADASUM) {
       // Attempt to add more responses to this fused response.
 
-      // found_tensor can be false for ranks that did Join.
-      bool found_tensor = tensor_queue_.GetTensorSizeAndType(response.tensor_names()[0], tensor_size, dtype);
```
romerojosh: @kit1980 In some recent testing, I noticed that tensor fusion is currently not functional (i.e. no tensors are being fused). It appears the issue is that the `dtype` variable above is no longer being set to anything. Was the line calling `GetTensorSizeAndType` removed by mistake?

kit1980: Yes, my mistake. The result `found_tensor` in no longer needed because of the `joined` parameter, but the call is needed. 

I thought there is a test for fusion? I'll try to fix this for both join and non-join ASAP.

kit1980: There is a test indeed, `test_horovod_allreduce_async_fused` for PyTorch. But looks like it doesn't actually test the tensors are fused (and it probably can't via Python API?)

tgaddair: Yeah, unfortunately that test assumes fusion buffering is occurring, and then tests to ensure that it produces correct results.  If fusion buffering is not working, then it will do nothing.  Once we move our build system to CMake, I want to introduce some C++-specific unit tests, of which this would be one of them, but for now it is difficult to test automatically.

kit1980: I've sent a PR fixing this: https://github.com/horovod/horovod/pull/1594

Issue #1521, opened 2019-11-19
NCCL broadcast for the Horovod broadcast operation


nvcastet:
Is there a way to use NCCL Broadcast `ncclBroadcast` for the Horovod broadcast? If not, is there a reason?
I may be wrong but from the code, it seems only Gloo, MLSL, or MPI can be picked.

tgaddair:
Hey @nvcastet, that's a good question.  I'm not aware of any technical reason it hasn't been implemented.  Because most broadcast calls tend not to fall on the performance-critical path (i.e., training loop), it hasn't been a high priority in the past.  That said, more users are exploring broadcasting model state once per epoch or more frequently, in which case that might start to change.  Are you interested in contributing this as a PR?

PR #1579, opened 2019-12-10
Add NCCL Broadcast operation

nvcastet:
Horovod needs to be compiled with HOROVOD_GPU_BROADCAST=NCCL.
Implements #1521.


jessebenson:
Good to see more NCCL ops!

tgaddair:
Hey @nvcastet, looks like some unit tests are failing:

```
ERROR:tensorflow:2 root error(s) found.
(0) Unknown: Type uint8 is not supported in NCCL mode.
[[node HorovodBroadcast_Cast_0 (defined at /usr/local/lib/python3.6/dist-packages/tensorflow_core/python/framework/ops.py:1748) ]]
[[All/_7]]
(1) Unknown: Type uint8 is not supported in NCCL mode.
[[node HorovodBroadcast_Cast_0 (defined at /usr/local/lib/python3.6/dist-packages/tensorflow_core/python/framework/ops.py:1748) ]]
0 successful operations.
0 derived errors ignored.
```

nvcastet:
@tgaddair It is finally ready for review :)

horovod/common/ops/nccl_operations.h
```diff
@@ -38,15 +38,12 @@ struct NCCLContext {
...
+class NCCLOp {
 public:
-  NCCLAllreduce(NCCLContext* nccl_context, CUDAContext* cuda_context,
-                HorovodGlobalState* global_state)
-      : CUDAAllreduce(cuda_context, global_state),
-        nccl_context_(nccl_context){};
-
-  Status Execute(std::vector<TensorTableEntry>& entries,
-                 const Response& response) override;
+  NCCLOp(NCCLContext* nccl_context, HorovodGlobalState* global_state)
```
tgaddair: In a previous PR (https://github.com/horovod/horovod/pull/1480/files#diff-5798de9e6fdc83e7f7bf0fa8344a56c0R65) we opted to use delegation to solve this problem by introducing `CUDAOpContext`.  It looks like we could do something similar here, which would make things in our codebase more consistent.  What do you think?

horovod/common/wire/message.fbs
```diff
@@ -28,7 +28,8 @@ enum DataType:byte {
     HOROVOD_FLOAT16 = 6,
     HOROVOD_FLOAT32 = 7,
     HOROVOD_FLOAT64 = 8,
-    HOROVOD_BOOL = 9
+    HOROVOD_BOOL = 9,
+    HOROVOD_BYTE = 10
```
tgaddair: I'm a little confused by `HOROVOD_BYTE`.  In what cases would it be used in place of `HOROVOD_UINT8`?

horovod/common/ops/nccl_operations.h
nvcastet: I agree to try to keep things consistent through the code base. Let me give it a try.

horovod/common/wire/message.fbs
nvcastet: I was under the impression that we needed to keep the enums consistent between `enum DataType` in [message.h](https://github.com/horovod/horovod/blob/master/horovod/common/message.h) and `enum DataType` in [message.fbs](https://github.com/horovod/horovod/blob/master/horovod/common/wire/message.fbs)/[message_generated.h](https://github.com/horovod/horovod/blob/master/horovod/common/wire/message_generated.h). Am i incorrect? `HOROVOD_BYTE` was already in [message.h](https://github.com/horovod/horovod/blob/master/horovod/common/message.h).

tgaddair: Ah, I see what happened.  The `HOROVOD_BYTE` dtype was added in https://github.com/horovod/horovod/commit/aa605d6de1e07d1184e8793b6d4f65cee9d4efb3 to abstract an API call to `MPI_Bcast`, but was subsequently simplified in https://github.com/horovod/horovod/commit/669eb9e0dde25f7475e6a31630ae4b9a704ae106.

The result is that `HOROVOD_BYTE` is no longer used for anything, so I think we can safely remove it.

PR #1584, opened 2019-12-13
Added support for Spark 3 TaskContext to fetch GPU resources per task

tgaddair:
@tgravescs

### DIFF
--- a/examples/keras_spark3_rossmann.py
+++ b/examples/keras_spark3_rossmann.py
@@ -30,6 +30,7 @@ DATA_LOCATION = 'file://' + os.getcwd()
 # Location of outputs on local filesystem (without file:// prefix).
 LOCAL_SUBMISSION_CSV = 'submission.csv'
 LOCAL_CHECKPOINT_FILE = 'checkpoint.h5'
+DISCOVERY_SCRIPT = 'get_gpu_resources.sh'
 
 # Spark clusters to use for training. If set to None, uses current default cluster.
 #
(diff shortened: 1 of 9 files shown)


examples/keras_spark3_rossmann.py
```diff
@@ -0,0 +1,567 @@
+# Copyright 2017 onwards, fast.ai, Inc.
```
EnricoMi: This 567 lines script deviates from `examples/keras_spark_rosspann.py` in only 11 rows. Differences should only be spark3 specific, but I suspect both files will diverge quite quickly on other changes.

I suggest to put some github actions in place that compare both scripts and flag up unexpected deviations, as put in place for `README.rst` and `docs/summary.rst`. I am happy to create a PR for this once this is merged into master.

abditag2:
After @EnricoMi 's comments, LGTM.

tgravescs:
the changes look good for local-cluster and standalone deployment.

If you want to support yarn or k8s the configs for the gpu scheduling will be slightly different.

Note that your default - local-cluster[2,1,1024] is using 2 workers and thus is relying on you also having 2 GPUs on the host.  If you try to run on a host without at least 2 GPUs, it will fail with a error saying not enough GPU addresses available.

examples/keras_spark3_rossmann.py
tgaddair: Good idea.  At some point we may want to further consolidate things so that we don't have so many nearly identical scripts.  Once Spark 3 is out, we may just replace the old examples or add special handling for older versions of Spark.

PR #1592, opened 2019-12-17
Add oneCCL to cpu test suite

shirosankaku:
Signed-off-by: Yana Shchyokotova <yana.shchyokotova@intel.com>

We noticed that MLSL was excluded from cpu test suit some time ago. Since oneCCL changes was based on current Horovod master branch state, this change remained unnoticed. This PR is intended to fix this.

### DIFF
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -22,6 +22,8 @@ tests=( \
        test-cpu-openmpi-py3_6-tf2_0_0-keras2_3_1-torch1_3_0-mxnet1_5_0-pyspark2_4_0 \
        test-cpu-openmpi-py3_6-tfhead-kerashead-torchhead-mxnethead-pyspark2_4_0 \
        test-cpu-mpich-py3_6-tf1_14_0-keras2_3_1-torch1_3_0-mxnet1_5_0-pyspark2_4_0 \
+       test-cpu-oneccl-py3_6-tf1_14_0-keras2_3_1-torch1_3_0-mxnet1_5_0-pyspark2_4_0 \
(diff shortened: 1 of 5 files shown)


tgaddair:
Hey @shirosankaku, looks like tests are now passing.  Is this ready to go once TRACE messages are removed and other tests are restored?

shirosankaku:
Hello, @tgaddair . Actually, this test passing doesn't mean that the issue has gone away. The problem is floating. And we need to investigate it more

shirosankaku:
Hello, @tgaddair !
Could you, please, review our last changes for test suite?

shirosankaku:
@tgaddair ?

shirosankaku:
Hello, @tgaddair !
Thank you for reviewing this PR!
As far as I understood, you are asking about oneccl and oneccl-ofi test suites difference?
The difference is in environment. The regular oneccl uses default environment via MPI, the oneccl-ofi uses OFI transport. No MPI involved.

EnricoMi:
Is there a reason why

    oneccl_env=$(echo ${oneccl_env//:/ })

is not simply

    oneccl_env=${oneccl_env//:/ }

Is the `echo` about trimming white spaces?

This `oneccl_env` really needs some refactoring, it looks it is about injecting environment variables where it actually injects bash code. Very confusing.

shirosankaku:
Hello, @EnricoMi !
You are right,
`oneccl_env=${oneccl_env//:/ }`
should be enough... Actually, this will remove extra space in other test suit command lines in case when oneccl_env is empty (in all the other cases except for oneccl). This will look cleaner. 
Should I create one more PR with this fix?

As for the bash code injection, oneccl_env contains `source` command and file content change. That's all. File mpirun_command sets env variables and does echo, env, which, ldd commands to see the actual environment setting the tests are being launched into.

PR #1594, opened 2019-12-18
Restore fusion functionality broken by #1571

kit1980:
I verified that fusion is happening via logs for both Join and normal case.

### DIFF
--- a/horovod/common/controller.cc
+++ b/horovod/common/controller.cc
@@ -507,12 +507,7 @@ Response Controller::ConstructResponse(std::string& name, int joined_size) {
     }
   }
 
-  // If there is at least one rank that requested Join, communicate tensor sizes
-  // in the response, because joined ranks don't have this info.
-  // If caching is enabled, the sizes info needs to be communicated even if
-  // there are no currently joined ranks, for possible future use.
-  if ((joined_size > 0 || response_cache_.capacity() > 0) &&
-      (message_type == Request::ALLREDUCE || message_type == Request::ADASUM)) {
+  if (message_type == Request::ALLREDUCE || message_type == Request::ADASUM) {
     TensorShape tensor_shape;
     for (auto dim : requests[0].tensor_shape()) {
       tensor_shape.AddDim(dim);
@@ -582,22 +577,18 @@ Response Controller::ConstructResponse(std::string& name, int joined_size) {
     }
   } else if (message_type == Request::ALLREDUCE) {
     response.set_response_type(Response::ALLREDUCE);
-    if (joined_size > 0 || response_cache_.capacity() > 0) {
-      for (auto dim : tensor_sizes) {
-        response.add_tensor_size(dim);
-      }
-      response.set_tensor_type(data_type);
+    for (auto dim : tensor_sizes) {
+      response.add_tensor_size(dim);
     }
+    response.set_tensor_type(data_type);
   } else if (message_type == Request::BROADCAST) {
     response.set_response_type(Response::BROADCAST);
   } else if (message_type == Request::ADASUM) {
     response.set_response_type(Response::ADASUM);
-    if (joined_size > 0 || response_cache_.capacity() > 0) {
-      for (auto dim : tensor_sizes) {
-        response.add_tensor_size(dim);
-      }
-      response.set_tensor_type(data_type);
+    for (auto dim : tensor_sizes) {
+      response.add_tensor_size(dim);
     }
+    response.set_tensor_type(data_type);
   }
   response.set_devices(devices);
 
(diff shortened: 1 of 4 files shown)


horovod/common/controller.cc
```diff
@@ -658,8 +668,8 @@ ResponseList Controller::FuseResponses(std::deque<Response>& responses,
 
         std::vector<TensorTableEntry> entries_for_join;
         if (joined) {
-          tensor_queue_.GetTensorEntriesFromResponse(response, entries_for_join,
-                                                     joined);
+          tensor_queue_.GetTensorEntriesFromResponse(new_response,
+                                                     entries_for_join, joined);
```
romerojosh: Thanks for the quick PR to fix fusion!

@tgaddair and I were discussing this and we think the fusion code here can be simplified a bit. Would it be possible to get both the `dtype` and `new_tensor_size` directly from the `new_response`, instead of generating `new_entry`? This would remove special treatment for join in `FuseResponses` and the code below to determine if fusion is possible would look something like:
```
        int64_t new_tensor_size = new_response.tensor_sizes()[0];
        if (response.response_type() == new_response.response_type() &&
            response.devices() == new_response.devices() &&
            response.tensor_type() == new_response.tensor_type() &&
            tensor_size + new_tensor_size <= TensorFusionThresholdBytes())
```
This would require making sure the type and tensor size fields are set in the response are set unconditionally here (and also for ADASUM): https://github.com/horovod/horovod/blob/master/horovod/common/controller.cc#L585
but we think that is alright to do.

With this, you can also remove the method `tensor_queue_.GetTensorSizeAndType`, since I don't think it is used anywhere else.

kit1980: Sent an update with these changes.

kit1980: I've realized that there are couple of other things I can remove, and `FuseResponses()` doesn't even need `joined` parameter. Doing this now...

kit1980: Updated.

horovod/common/controller.cc
```diff
@@ -628,53 +619,43 @@ void Controller::CoordinateCacheAndState(CacheCoordinator& cache_coordinator) {
...
     for (auto bit : cache_coordinator.cache_hits()) {
       auto& response = response_cache_.peek_response(bit);
       timeline_.NegotiateEnd(response.tensor_names()[0]);
     }
   }
 }
 
-ResponseList Controller::FuseResponses(std::deque<Response>& responses,
-                                       bool joined) {
+ResponseList Controller::FuseResponses(std::deque<Response>& responses) {
```
tgaddair: This needs to be updated in the header as well.

kit1980: Forgot to add to the commit, fixed.

horovod/common/controller.cc
```diff
@@ -628,53 +619,45 @@ void Controller::CoordinateCacheAndState(CacheCoordinator& cache_coordinator) {
...
-                                                     joined);
-        }
-
-        const auto& new_entry =
-            joined
-                ? entries_for_join[0]
-                : tensor_queue_.GetTensorEntry(new_response.tensor_names()[0]);
-        int64_t new_tensor_size = new_entry.tensor->size();
-
+        int64_t new_tensor_size = new_response.tensor_sizes().empty()
```
romerojosh: Nice catch @kit1980! If the tests pass with this change, I suggest modifying this so that rather than creating and setting the `new_tensor_size` variable out here with this check on `new_response.tensor_sizes().empty()`, just use `new_response.tensor_sizes()[0]` directly in place of `new_tensor_size` in code below. The first check on `response.response_type() == new_response.response_type()` in the if statement should guard against attempting to access the size for non allreduce messages.

kit1980: Hi @romerojosh. The tests passed with the current code in this PR.
I tried your suggestion, but then the tests fail after error cases, like duplicate name tests, (but not every time...)

romerojosh: Yes, you are right. The segfault in the updated version occurs in the else case where `new_tensor_size` is being used (which becomes problematic if replaced with `new_response.tensor_sizes()[0]` directly):
https://github.com/horovod/horovod/blob/f042aeca6b9e9a273b43fd15fd22e7dac970827e/horovod/common/controller.cc#L672

The existing solution is fine with me, but @tgaddair, maybe you can also chime in on this?

tgaddair: I think this approach is fine as well.

horovod/common/controller.cc
```diff
@@ -628,53 +619,45 @@ void Controller::CoordinateCacheAndState(CacheCoordinator& cache_coordinator) {
...
+                                      : new_response.tensor_sizes()[0];
         if (response.response_type() == new_response.response_type() &&
             response.devices() == new_response.devices() &&
-            dtype == new_entry.tensor->dtype() &&
+            response.tensor_type() == new_response.tensor_type() &&
             tensor_size + new_tensor_size <= TensorFusionThresholdBytes()) {
           // These tensors will fuse together well.
           tensor_size += new_tensor_size;
           response.add_tensor_name(new_response.tensor_names()[0]);
+          LOG(DEBUG) << "Fused tensor " << new_response.tensor_names()[0]
```
tgaddair: I think we should either remove this or add it for the Allgather case as well.  I would lean towards removing, personally.

kit1980: Removed.

PR #1604, opened 2019-12-20
Correction to tensor size used for fusion in #1594

romerojosh:
The updated code in #1594 that was merged introduced another error in fusion. In particular, based on my suggestion to simplify, it was modified to use `response.tensor_sizes()` to compute sizes for fusion instead of generating new tensor entries. However, entries in `response.tensor_sizes()` are in elements, not bytes, causing the fusion logic for allreduce to be incorrect (may create fused responses larger than allocated buffer).  

~~I have reverted to the original version of #1594 to deal with this.~~
Multiplication of `response.tensor_sizes()` by `GetTypeSize()` corrects this.

### DIFF
--- a/horovod/common/controller.cc
+++ b/horovod/common/controller.cc
@@ -639,7 +639,7 @@ ResponseList Controller::FuseResponses(std::deque<Response>& responses) {
         response.response_type() == Response::ResponseType::ADASUM) {
       // Attempt to add more responses to this fused response.
 
-      tensor_size = response.tensor_sizes()[0];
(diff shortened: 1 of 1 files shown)


romerojosh:
@tgaddair @kit1980 Can you take a look? I still feel like there should be a cleaner option to handle this, instead of allocating new temporary entries just to get the size of the tensor for the join case here. 
An alternative option to rolling back to this implementation could be to introduce a size of operation for Horovod data types and using it to correct the current broken implementation. Not sure if it would be good to bake in our own assumptions on type sizes though, rather than depend on the framework.

PR #1606, opened 2019-12-22
Fixed pytorch model serialization bug

abditag2:


tgaddair:
How is this different from #1605?

Issue #1428, opened 2019-10-04
horovodrun with --hostfile option is not working


apeforest:
**Environment:**
1. Framework: MXNet
2. Framework version: 1.5.1
3. Horovod version: 0.18.2
4. MPI version: 3.1.4
5. CUDA version: 10.0
6. NCCL version: 2.4.6
7. Python version: 3.7.3
8. OS and version: Ubuntu 16.04
9. GCC version: 5.4

**Checklist:**
1. Did you search issues to find if somebody asked this question before? Yes
2. If your question is about hang, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/running.rst)?
3. If your question is about docker, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/docker.rst)?
4. Did you check if you question is answered in the [troubleshooting guide](https://github.com/horovod/horovod/blob/master/docs/troubleshooting.rst)?

**Bug report:**
horovodrun with --hostfile is not working if there are multiple hosts in the format:
172.31.1.0 slots=8
172.31.2.0 slots=8

The culprit is this PR: https://github.com/horovod/horovod/pull/1181

alsrgv:
@apeforest, I think it may be hard to support all possible options in the hostfile, what are the most common that we should target?

apeforest:
Fixed by https://github.com/horovod/horovod/pull/1607

PR #1607, opened 2019-12-23
Fixed hostfile bug during horovodrun command

karan6181:
Fixed the bug [#1428](https://github.com/horovod/horovod/issues/1428)

cc @apeforest 
Signed-off-by: Karan Jariwala <karankjariwala@gmail.com>


karan6181:
@apeforest Added a test for the same. 
cc @tgaddair

PR #1609, opened 2019-12-24
Fixed issue with broadcasting optimizer state for params that don't require grads, added back utility function to broadcast any object

### LINKED ISSUE #1608: RuntimeError: For integral input tensors, argument alpha must not be a floating point number
**Environment:**
1. Framework: (TensorFlow, Keras, PyTorch, MXNet): Pytorch
2. Framework version: 1.3.0
3. Horovod version: 0.16.1
4. MPI version: 3.1.2 (?)
5. CUDA version: 10.2
6. NCCL version: ? (probably not important)
7. Python version: 2.7
8. OS and version: Ubuntu 14.04.4 (probably not important)
9. GCC version: 4.8.4 (probably not important)

**Bug report:**
Horovod does not like it when my module has integer parameters, even if they do not require a gradient. It says

```python
  File "/.../mwe.py", line 30, in <module>
    hvd.broadcast_optimizer_state(optimizer, root_rank=0)
  File "/.../python2.7/dist-packages/horovod-0.16.1-py2.7-linux-x86_64.egg/horovod/torch/__init__.py", line 261, in broadcast_optimizer_state
    super(optimizer.__class__, optimizer).step()
  File "/.../torch/optim/sgd.py", line 106, in step
    p.data.add_(-group['lr'], d_p)
RuntimeError: For integral input tensors, argument alpha must not be a floating point number.

```

MWE
```python
import torch
import torch.nn as nn
import torch.optim as optim

import horovod.torch as hvd


class A(nn.Module):

    def __init__(self, a, b):
        super(A, self).__init__()
        self.a = nn.Parameter(a.int(), requires_grad=False)   # change to a.float() and hvd is happy
        self.b = nn.Parameter(b)

    def forward(self, x):
        return torch.index_select(self.b, 0, self.a.long()) * x


hvd.init()

a = torch.Tensor([1, 3])
b = torch.rand(4)

model = A(a, b).cuda()

optimizer = optim.SGD(model.parameters(), lr=0.01)
optimizer = hvd.DistributedOptimizer(optimizer, named_parameters=model.named_parameters())

hvd.broadcast_parameters(model.state_dict(), root_rank=0)
hvd.broadcast_optimizer_state(optimizer, root_rank=0)  # <-- hvd is sad :(

model.train()

for i in range(1000):

    x = torch.Tensor(torch.rand(2)).cuda()

    optimizer.zero_grad()

    loss = torch.mean((model(x) - 0) ** 2)

    loss.backward()
    optimizer.step()

    if hvd.rank() == 0:
        print(i, loss.detach())

```

tgaddair:
Fixes #1608.

### DIFF
--- a/horovod/torch/__init__.py
+++ b/horovod/torch/__init__.py
@@ -497,7 +497,7 @@ def broadcast_optimizer_state(optimizer, root_rank):
     if len(state_dict['state']) == 0:
         for group in optimizer.param_groups:
             for p in group['params']:
-                if p.requires_grad and id(p) not in optimizer.state_dict()['state']:
+                if p.requires_grad and id(p) not in state_dict['state']:
                     p.grad = p.data.new(p.size()).zero_()
         # This function accepts a torch.optim.Optimizer or a DistributedOptimizer
         # wrapped around a torch optimizer. Calling step() with a DistributedOptimizer
--- a/horovod/torch/__init__.py
+++ b/horovod/torch/__init__.py
@@ -19,7 +19,11 @@ from __future__ import division
 from __future__ import print_function
 
 from contextlib import contextmanager
+
+import io
 import warnings
+
+import cloudpickle
 
 from horovod.common.util import check_extension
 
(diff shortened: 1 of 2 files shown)


una-dinosauria:
LGTM. Thanks for the quick fix!

PR #1613, opened 2019-12-25
Correct keras_spark_rossmann_estimator

ConeyLiu:
Expose backend kw args in order to pass extra mpi args to backend. And two small fixes.

### DIFF
--- a/examples/keras_spark_rossmann_estimator.py
+++ b/examples/keras_spark_rossmann_estimator.py
@@ -284,7 +284,7 @@ test_min_date = test_df.agg(F.min(test_df.Date)).collect()[0][0]
 test_max_date = test_df.agg(F.max(test_df.Date)).collect()[0][0]
 one_year = datetime.timedelta(365)
 train_df = train_df.withColumn('Validation',
-                               (train_df.Date < test_min_date - one_year) | (train_df.Date >= test_max_date - one_year))
+                               (train_df.Date > test_min_date - one_year) & (train_df.Date <= test_max_date - one_year))
 
 # Determine max Sales number.
 max_sales = train_df.agg(F.max(train_df.Sales)).collect()[0][0]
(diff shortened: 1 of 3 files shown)


ConeyLiu:
It seems there are other errors in `keras_spark_rossmann_estimator`. I will check it again.

ConeyLiu:
It's ready to review.

ConeyLiu:
When I run the example `keras_spark_rossmann_estimator`. I found there are some warning:
```
WARNING:tensorflow:Entity <function make_petastorm_dataset.<locals>.<lambda> at 0x7f807c0a9950> could not be transformed and will be executed as-is. Please report this to the AutoGraph team. When filing the bug, set the verbosity to 10 (on Linux, `export AUTOGRAPH_VERBOSITY=10`) and attach the full output. Cause: expected exactly one node node, found []
```

And 
```
[1,3]<stderr>:Error in sys.excepthook:
[1,3]<stderr>:
[1,3]<stderr>:Original exception was:
```

Horovod is the master branch and tensorflow is 2.0 version.

tgaddair:
Hey @ConeyLiu, thanks for the PR.

Instead of adding this parameter to the estimator, I'm wondering if it would work for you to use the existing `backend` parameter like this:

```
backend = SparkBackend(num_proc, ...)
estimator = KerasEstimator(backend=backend, ...)
```

I apologize that it isn't documented yet, but the intent is that `num_proc` is just a simplified way of specifying `backend`.

I'll definitely sync with the Petastorm team to figure out what's going on with that warning you mentioned.

ConeyLiu:
very thanks @tgaddair, that works for me. Sorry for not noticing this usage. I have modified the code to just correct the example. Thanks again.

ConeyLiu:
thanks a lot.

PR #1632, opened 2020-01-03
Initial support for ROCm: rename CUDA to GPU et al, pimpl for GPU details

jeffdaily:
- add ROCm and ROCm TensorFlow support to setup.py
- Renamed symbols and files
  - CUDA -> GPU
  - Cuda -> Gpu
  - cuda -> gpu
  - HAVE_CUDA -> HAVE_GPU --- iff generically referring to any GPU device
- CUDAContext becomes GPUContext and uses pimpl pattern to decouple
  CUDA and HIP implementations of GPU interface.
- New GPUContext methods
  - StreamCreate
  - StreamSynchronize
  - GetDevice, SetDevice
  - MemcpyAsyncD2D, MemcpyAsyncH2D, MemcpyAsyncD2H

Signed-off-by: Jeff Daily <jeff.daily@amd.com>

### DIFF
--- a/horovod/common/ops/adasum_gpu_operations.cc
+++ b/horovod/common/ops/adasum_gpu_operations.cc
@@ -13,30 +13,30 @@
 // limitations under the License.
 // =============================================================================
 
-#include "adasum_cuda_operations.h"
+#include "adasum_gpu_operations.h"
 
 namespace horovod {
 namespace common {
 
-AdasumCudaAllreduceOp::AdasumCudaAllreduceOp(MPIContext* mpi_context,
-                                             NCCLContext* nccl_context,
-                                             CUDAContext* cuda_context,
-                                             HorovodGlobalState* global_state)
+AdasumGpuAllreduceOp::AdasumGpuAllreduceOp(MPIContext* mpi_context,
+                                           NCCLContext* nccl_context,
+                                           GPUContext* gpu_context,
+                                           HorovodGlobalState* global_state)
     : AdasumMPI(mpi_context, global_state),
-      NCCLAllreduce(nccl_context, cuda_context, global_state, Communicator::LOCAL) {
+      NCCLAllreduce(nccl_context, gpu_context, global_state, Communicator::LOCAL) {
   // Pre-allocate host buffer size equal to the fusion buffer length
   current_host_buffer_length =
       global_state->parameter_manager.TensorFusionThresholdBytes();
-  cuda_op_context_.host_buffer = (uint8_t*)malloc(current_host_buffer_length);
+  gpu_op_context_.host_buffer = (uint8_t*)malloc(current_host_buffer_length);
 }
 
-AdasumCudaAllreduceOp::~AdasumCudaAllreduceOp() {
-  if (cuda_op_context_.host_buffer != nullptr) {
-    free(cuda_op_context_.host_buffer);
+AdasumGpuAllreduceOp::~AdasumGpuAllreduceOp() {
+  if (gpu_op_context_.host_buffer != nullptr) {
+    free(gpu_op_context_.host_buffer);
   }
 }
-Status AdasumCudaAllreduceOp::Execute(std::vector<TensorTableEntry>& entries,
-                                      const Response& response) {
+Status AdasumGpuAllreduceOp::Execute(std::vector<TensorTableEntry>& entries,
+                                     const Response& response) {
   if (entries.empty()) {
     return Status::OK();
   }
@@ -48,14 +48,14 @@ Status AdasumCudaAllreduceOp::Execute(std::vector<TensorTableEntry>& entries,
   return NcclHierarchical(entries, response);
 }
 
-uint8_t* AdasumCudaAllreduceOp::GetHostBuffer(uint64_t buffer_length) {
-  return CheckBufferAndReallocate((uint8_t**)&cuda_op_context_.host_buffer,
+uint8_t* AdasumGpuAllreduceOp::GetHostBuffer(uint64_t buffer_length) {
+  return CheckBufferAndReallocate((uint8_t**)&gpu_op_context_.host_buffer,
                                   buffer_length, current_host_buffer_length);
 }
 
 Status
-AdasumCudaAllreduceOp::NcclHierarchical(std::vector<TensorTableEntry>& entries,
-                                        const Response& response) {
+AdasumGpuAllreduceOp::NcclHierarchical(std::vector<TensorTableEntry>& entries,
+                                       const Response& response) {
   auto& first_entry = entries[0];
 
   // Determine GPU IDs of the devices participating in this communicator.
@@ -65,9 +65,9 @@ AdasumCudaAllreduceOp::NcclHierarchical(std::vector<TensorTableEntry>& entries,
   for (size_t rank : global_state_->controller->GetLocalCommRanks()) {
     nccl_device_map.push_back(response.devices()[rank]);
   }
-  cuda_op_context_.InitCUDA(entries);
+  gpu_op_context_.InitGPU(entries);
   nccl_op_context_.InitNCCLComm(entries, nccl_device_map);
-  cuda_op_context_.InitCUDAQueue(entries, response);
+  gpu_op_context_.InitGPUQueue(entries, response);
   const void* fused_input_data;
   void* buffer_data;
   size_t buffer_len;
@@ -76,9 +76,9 @@ AdasumCudaAllreduceOp::NcclHierarchical(std::vector<TensorTableEntry>& entries,
   if (entries.size() > 1) {
     MemcpyInFusionBuffer(entries, fused_input_data, buffer_data, buffer_len);
     if (global_state_->timeline.Initialized()) {
-      cuda_context_->RecordEvent(cuda_op_context_.event_queue,
+      gpu_context_->RecordEvent(gpu_op_context_.event_queue,
                                  MEMCPY_IN_FUSION_BUFFER,
-                                 *cuda_op_context_.stream);
+                                 *gpu_op_context_.stream);
     }
   } else {
     fused_input_data = first_entry.tensor->data();
@@ -157,12 +157,12 @@ AdasumCudaAllreduceOp::NcclHierarchical(std::vector<TensorTableEntry>& entries,
     auto nccl_result = ncclReduceScatter(
         fused_input_data, buffer_data_at_rank_offset,
         (size_t)num_elements_per_rank, GetNCCLDataType(first_entry.tensor),
-        ncclSum, *nccl_op_context_.nccl_comm_, *cuda_op_context_.stream);
+        ncclSum, *nccl_op_context_.nccl_comm_, *gpu_op_context_.stream);
 
     nccl_context_->ErrorCheck("ncclReduceScatter", nccl_result);
     if (global_state_->timeline.Initialized()) {
-      cuda_context_->RecordEvent(cuda_op_context_.event_queue,
-                                 NCCL_REDUCESCATTER, *cuda_op_context_.stream);
+      gpu_context_->RecordEvent(gpu_op_context_.event_queue,
+                                 NCCL_REDUCESCATTER, *gpu_op_context_.stream);
     }
   }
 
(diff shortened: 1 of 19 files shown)


jeffdaily:
It's not clear to me whether the current test failures are related to the changes in this PR.

tgaddair:
Hey @jeffdaily, thanks for the new PR.  Overall, I like the approach to abstracting GPU operations in this way.

One thing I noticed is it looks like the DDL codepaths haven't been updated to use the new `gpu_operations.h`.  Those aren't currently under unit test, but the changes should be fairly straightforward.

There was also a PR landed yesterday that added NCCL broadcast support which will require a rebase.

For the unit tests, it looks like at least one test was failing because the Horovod MXNet plugin failed to compile, so we may need to dig into why that is.  I'm rerunning some of the other tests to see if it was a transient error.

I'll try and put together inline feedback as well.  I think it would be good to have @nvcastet and @romerojosh take a look at this PR if they have the bandwidth.

jeffdaily:
Rebased to pick up NCCL broadcast change.  Also updated the DDL sources, though I did not test them.

horovod/common/ops/gpu_operations.h
```diff
@@ -14,70 +14,91 @@
...
 namespace common {
 
-struct CUDAContext {
-  cudaError_t GetCudaEvent(cudaEvent_t* event);
-
-  cudaError_t ReleaseCudaEvent(cudaEvent_t event);
+class GPUContext {
+public:
+  GPUContext();
+  ~GPUContext();
```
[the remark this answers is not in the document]
jeffdaily: I ran into an issue desribed here https://herbsutter.com/gotw/_100/ .  Specifically: 

> You still need to write the visible class’ destructor yourself and define it out of line in the implementation file, even if normally it’s the same as what the compiler would generate. This is because although both unique_ptr and shared_ptr can be instantiated with an incomplete type, unique_ptr’s destructor requires a complete type in order to invoke delete (unlike shared_ptr which captures more information when it’s constructed). By writing it yourself in the implementation file, you force it to be defined in a place where impl is already defined, and this successfully prevents the compiler from trying to automatically generate the destructor on demand in the caller’s code where impl is not defined.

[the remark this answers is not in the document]
nvcastet: Thanks @jeffdaily for the explanation and the link.

jeffdaily:
@tgaddair are the last two test failures transient or do they indicate a real issue?  Also, this PR currently does not integrate any kind of CI build for ROCm yet.  What is your acceptance criteria?

tgaddair:
@jeffdaily one of the errors appears to be real:
https://buildkite.com/horovod/horovod/builds/1844#c24a6237-3f92-4cc5-9310-bcf714008444

The MXNet test for mixed install is consistently failing because the Horovod MXNet plugin did not compile successfully.

Once we figure that out, I think we should be safe to land.

jeffdaily:
@tgaddair  I tried building the docker container for the mixed install so I could see how horovod is failing to build the mxnet extension, and I'm seeing the following error:

```
  CMake Error: The source directory "/tmp/pip-req-build-7pquo6gf/third_party/gloo" does not exist.
  Specify --help for usage, or press the help button on the CMake GUI.
  INFO: Unable to build MXNet plugin, will skip it.
...
  error: None of TensorFlow, PyTorch, or MXNet plugins were built. See errors above.
  Building wheel for horovod (setup.py): finished with status 'error'
  ERROR: Failed building wheel for horovod
```

jeffdaily:
@tgaddair disregard previous comment concerning the dockerfile build.  I forgot to fetch submodules prior to attempting the dockerfile build.

I was able to reproduce the mxnet build failure.  Fix will be ready shortly.

jeffdaily:
@tgaddair fixed the MXNet test for mixed install.  Other errors have popped up now that were passing in the previous attempt.

tgaddair:
Hey @jeffdaily, looks like test failures may be transient.  Rerunning them now.

horovod/common/ops/adasum_gpu_operations.cc
```diff
@@ -172,33 +172,30 @@ AdasumCudaAllreduceOp::NcclHierarchical(std::vector<TensorTableEntry>& entries,
...
-      cuda_context_->RecordEvent(cuda_op_context_.event_queue, NCCL_REDUCE,
-                                 *cuda_op_context_.stream);
+      gpu_context_->RecordEvent(gpu_op_context_.event_queue, NCCL_REDUCE,
+                                 *gpu_op_context_.stream);
     }
   }
 
   if (global_state_->controller->IsHomogeneous() || is_root_rank) {
-    // cudaHostAlloc is significantly slower than malloc.  Pre-allocating
+    // gpuHostAlloc is significantly slower than malloc.  Pre-allocating
```
tgaddair: For consistency with other comments and NVIDIA documentation link below, this should be left as `cudaHostAlloc`.

horovod/common/ops/nccl_operations.cc
```diff
@@ -279,34 +279,32 @@ NCCLHierarchicalAllreduce::Execute(std::vector<TensorTableEntry>& entries,
...
     nccl_context_->ErrorCheck("ncclReduce", nccl_result);
     if (global_state_->timeline.Initialized()) {
-      cuda_context_->RecordEvent(cuda_op_context_.event_queue, NCCL_REDUCE, *cuda_op_context_.stream);
+      gpu_context_->RecordEvent(gpu_op_context_.event_queue, NCCL_REDUCE, *gpu_op_context_.stream);
     }
   }
 
   if (global_state_->controller->IsHomogeneous() || is_root_rank) {
-    // cudaHostAlloc is significantly slower than malloc.  Pre-allocating
+    // gpuHostAlloc is significantly slower than malloc.  Pre-allocating
```
tgaddair: Same above: `gpuHostAlloc` -> `cudaHostAlloc`.

tgaddair:
Hey @jeffdaily, took a closer look over the PR and discussed it with some other Horovod contributors.  Overall, I think the abstraction layer is good.  

The one thing I think we should still address is the possibility for the ROCm/RCCL and CUDA/NCCL APIs to diverge in the future.  With the pimpl pattern, we're locked in to the API exposed by the GPUContext.  In this design how would a contributor add functionality specific to one of these frameworks but not the other?

horovod/common/ops/adasum_gpu_operations.cc
jeffdaily: Done.

horovod/common/ops/nccl_operations.cc
jeffdaily: Done.

jeffdaily:
I am rather at a loss to suggest how to address a future ROCm/RCCL and CUDA/NCCL API divergence.  With respect to HIP (ROCm) versus CUDA APIs, parity is intended by design.  I am not privy to the CUDA development roadmap, so I can't predict new or deprecated APIs.  Are there any plans for Horovod to use less common CUDA APIs?

Concerning RCCL versus NCCL, again parity is by design in order to make it easier to support AMD hardware through the same interface.

I think if there ever exists some CUDA or NCCL feature that isn't reflected by HIP or RCCL, we have the option of separating implementation details within the same source file using `#if HAVE_ROCM`, or a source file that is compiled for CUDA/NCCL only.  It would need to be documented as Horovod feature X is not supported for ROCm builds.

GPUDirect, or ROCmRDMA, is very well supported with AMD hardware.

jeffdaily:
Rebased again to resolve conflicts from https://github.com/horovod/horovod/commit/d4d54bc7e772c1511a2634d6c0d7c6fc2456599f

setup.py
```diff
@@ -608,17 +649,21 @@ def get_common_options(build_ext):
...
+    have_cuda = False
+    have_rocm = False
+    gpu_include_dirs = gpu_lib_dirs = gpu_macros = []
     if gpu_allreduce or gpu_allgather or gpu_broadcast:
-        have_cuda = True
-        cuda_include_dirs, cuda_lib_dirs = get_cuda_dirs(build_ext, cpp_flags)
-    else:
-        have_cuda = False
-        cuda_include_dirs = cuda_lib_dirs = []
+        if os.environ.get('HOROVOD_GPU_ROCM'):
```
tgaddair: Can we get rid of this environment variable and ask the user to set `HOROVOD_GPU_ALLREDUCE=RCCL` instead?

tgaddair:
@jeffdaily I'm fine punting on the API divergence thing for now.  I think the right longterm solution is to create separate NCCL vs RCCL derived classes with separate CUDA and ROCM implementations of the `GPUContext`.  But I think perhaps we can cross that bridge when we come to it, or doing it in a followup PR.

nvcastet:
@jeffdaily @tgaddair Using derived classes would also cause issues because even if we use the base type through the code, we will still need to construct the object with the derived type and to avoid compilation errors, we would need have to have a bunch of "ifdef" which is not really clean code in modern C++. Were you thinking of another way to do that?
OOP polymorphism seems to work better when everything can get built and also in this case, both APIs will not be used in the same build.
The C++11 version of the pimpl pattern seems to be a good fit.
With any interface/encapsulation design, we have will have the potentiality of divergence in implementations. If one implementation does not support a feature, we can always document it or throw a not_implemented exception printing a clear error message.
Is anyone thinking of a better pattern? @jeffdaily is pimpl the main design pattern AMD has been using in open source projects where a single GPU API existed? I would assume so for C++ code.

setup.py
jeffdaily: I think at best we could use `HOROVOD_GPU_ALLREDUCE=RCCL` to automatically set a missing `HOROVOD_GPU_ROCM`, but shouldn't we still support builds where RCCL is not used but we still build horovod for ROCm and assume a GPUDirect-capable MPI install?  Is this still a use case for CUDA?

nvcastet: Correct, now we can use cuda-aware MPI without NCCL if we say (HOROVOD_GPU_ALLREDUCE=MPI or HOROVOD_GPU_BROADCAST=MPI).

jeffdaily:
@nvcastet unfortunately, there is not one main design pattern for supporting AMD hardware in open source projects.

As one might expect, long-established open source projects tend to resist adding HIP APIs, and certainly not fully converting code bases to use HIP alone even though HIP can compile to CUDA.  That said, I've been on the teams to support TensorFlow, PyTorch, and now Horovod -- and each project maintainer has dictated a different approach.

TensorFlow already had a device abstraction layer to differentiate between CPU and GPU devices.  The GPU was itself further abstracted as a stream executor.  That was all done prior to our involvement, but it made the most sense to integrate our ROCm stack as an implementation of the stream executor abstract interface.  Even so, there are plenty of cases where APIs or device features are not the same.  For example, CUDNN has either different behavior or a different API interface compared to our MIOpen, and so such code must be protected with `#if TENSORFLOW_USE_ROCM`.  In summary, TensorFlow uses both this preprocessor macro as well as separately compiled sources even though attempts were made to use abstract interfaces for the GPU device and libraries (BLAS, CUDNN, etc).

PyTorch has mandated that the code base is "hipified" during the build, so a"hipify" python script was developed and maintained as part of the PyTorch code base.  There are very few pure-HIP source files; the rest of the PyTorch source looks like any other CUDA-based project.  But, macros can still be used to project CUDA-only code and features.  We have further been mandated to use this hipify script for building all ROCm PyTorch extensions.  Those changes are prepared in a separate branch that will become a new PR if this current PR is accepted.

setup.py
tgaddair: That's a good point.  How about we change it to `HOROVOD_GPU=ROCM`.  That way we leave open the possibility of other APIs as well, and potentially to auto-infer if not set.

tgaddair:
@nvcastet Currently we use preprocessor macros to determine which set of ops to use [here](https://github.com/horovod/horovod/blob/master/horovod/common/operations.cc#L142).  Otherwise, the code is isolated into its own classes / files by framework so it's easier for developers to reason about the code paths.

I think that if the APIs ever diverged, we would necessarily need to either do something similar, or mix the preprocessor macros into the `nccl_operations.cc` file, the latter of which I think is less ideal from a maintainability standpoint.

So long as the APIs are identical, I agree the pimpl pattern is cleaner.  And since there isn't any indication that they will diverge at the moment, I also agree it's the better solution for now.

setup.py
jeffdaily: I'm assuming you would like the current behavior to remain the same, i.e., assume that CUDA is selected.  So something like the following?

```python
    have_cuda = False
    have_rocm = False
    gpu_include_dirs = gpu_lib_dirs = gpu_macros = []
    if gpu_allreduce or gpu_allgather or gpu_broadcast:
        gpu_type = os.environ.get('HOROVOD_GPU', 'CUDA')
        if gpu_type == 'CUDA':
            have_cuda = True
            gpu_include_dirs, gpu_lib_dirs = get_cuda_dirs(build_ext, cpp_flags)
        elif gpu_type == 'ROCM':
            have_rocm = True
            gpu_include_dirs, gpu_lib_dirs, gpu_macros = get_rocm_dirs(build_ext, cpp_flags)
        else:
            raise DistutilsError("Unknown HOROVOD_GPU type '%s'" % gpu_type)
```

tgaddair: Yes, that's exactly what I had in mind.  Does that sound good to you?

jeffdaily: Done.

PR #1636, opened 2020-01-06
Bring back mxnet nightly for testing

eric-haibin-lin:


tgaddair:
@eric-haibin-lin, please try rebasing off of master.  There is an issue with nightly torch that was causing unit test failures.

PR #1649, opened 2020-01-09
Deprecate DDL backend

nvcastet:
Transitioning to NCCL backend for the GPU collective communication library.


nvcastet:
@tgaddair

PR #1665, opened 2020-01-14
Bump version to 0.19.0

tgaddair:


tgaddair:
> Do we need update any changelog when bumping versions?

We will, yes.  There will be a page at https://github.com/horovod/horovod/releases/tag/v0.19.0 that has the release notes.

PR #1670, opened 2020-01-15
Fix HOROVOD_GPU_BROADCAST setup check.

kit1980:


nvcastet:
Thanks for catching that @kit1980!

PR #1671, opened 2020-01-15
Enable GPU inference for e2e example keras_spark3_rossmann.py

chuanlihao:
Currently only CPU inference is available for end-to-end example keras_spark3_rossmann.py

This PR enables GPU inference for that example:
- There is a flag indicating whether to infer on GPU. It's disabled by default.
- GPU inference runs on a configurable Spark 3.0 cluster.
- This PR also demonstrates Spark 3.0 GPU scheduling.

Besides, there's another pending git commit to enable batch inference on GPU: https://github.com/chuanlihao/horovod/commit/7d5a5690f988a500fbd80dbdeb1609a7a8f90dc6

### DIFF
--- a/examples/keras_spark3_rossmann.py
+++ b/examples/keras_spark3_rossmann.py
@@ -494,33 +494,36 @@ def train_fn(model_bytes):
             return history.history, f.read()
 
 
+def set_gpu_conf(conf):
+    # This config will change depending on your cluster setup.
+    #
+    # 1. Standalone Cluster
+    # - Must configure spark.worker.* configs as below.
+    #
+    # 2. YARN
+    # - Requires YARN 3.1 or higher to support GPUs
+    # - Cluster should be configured to have isolation on so that
+    #   multiple executors don’t see the same GPU on the same host.
+    # - If you don’t have isolation then you would require a different discovery script
+    #   or other way to make sure that 2 executors don’t try to use same GPU.
+    #
+    # 3. Kubernetes
+    # - Requires GPU support and isolation.
+    # - Add conf.set(“spark.executor.resource.gpu.discoveryScript”, DISCOVERY_SCRIPT)
+    # - Add conf.set(“spark.executor.resource.gpu.vendor”, “nvidia.com”)
+    conf = conf.set("spark.test.home", os.environ.get('SPARK_HOME'))
+    conf = conf.set("spark.worker.resource.gpu.discoveryScript", DISCOVERY_SCRIPT)
+    conf = conf.set("spark.worker.resource.gpu.amount", 1)
+    conf = conf.set("spark.task.resource.gpu.amount", "1")
+    conf = conf.set("spark.executor.resource.gpu.amount", "1")
+    return conf
+
+
 # Create Spark session for training.
 conf = SparkConf().setAppName('training')
 if TRAINING_CLUSTER:
     conf.setMaster(TRAINING_CLUSTER)
-
-# This config will change depending on your cluster setup.
-#
-# 1. Standalone Cluster
-# - Must configure spark.worker.* configs as below.
-#
-# 2. YARN
-# - Requires YARN 3.1 or higher to support GPUs
-# - Cluster should be configured to have isolation on so that
-#   multiple executors don’t see the same GPU on the same host.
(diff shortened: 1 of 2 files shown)


examples/keras_spark3_rossmann.py
```diff
@@ -554,11 +587,18 @@ def fn(rows):
...
-        # Do not use GPUs for prediction, use single CPU core per task.
-        config = tf.ConfigProto(device_count={'GPU': 0})
-        config.inter_op_parallelism_threads = 1
-        config.intra_op_parallelism_threads = 1
-        K.set_session(tf.Session(config=config))
+        if GPU_INFERENCE_ENABLED:
+            from pyspark import TaskContext
+            config = tf.ConfigProto()
+            config.gpu_options.allow_growth = True
+            config.gpu_options.visible_device_list = TaskContext.get().resources()['gpu'].addresses[0]
```
anfeng: You should use: get_available_devices()[0]
See examples/keras_spark3_rossmann.py for a reference

examples/keras_spark3_rossmann.py
```diff
@@ -554,11 +587,18 @@ def fn(rows):
         import tensorflow as tf
         import tensorflow.keras.backend as K
 
-        # Do not use GPUs for prediction, use single CPU core per task.
-        config = tf.ConfigProto(device_count={'GPU': 0})
-        config.inter_op_parallelism_threads = 1
-        config.intra_op_parallelism_threads = 1
-        K.set_session(tf.Session(config=config))
+        if GPU_INFERENCE_ENABLED:
+            from pyspark import TaskContext
```
anfeng: from horovod.spark.task import get_available_devices

examples/keras_spark3_rossmann.py
```diff
@@ -543,8 +549,35 @@ def train_fn(model_bytes):
 conf = SparkConf().setAppName('prediction') \
     .setExecutorEnv('LD_LIBRARY_PATH', os.environ.get('LD_LIBRARY_PATH')) \
     .setExecutorEnv('PATH', os.environ.get('PATH'))
-if LIGHT_PROCESSING_CLUSTER:
-    conf.setMaster(LIGHT_PROCESSING_CLUSTER)
+
+if GPU_INFERENCE_ENABLED:
+    if GPU_INFERENCE_CLUSTER:
+        conf.setMaster(GPU_INFERENCE_CLUSTER)
+    # This config will change depending on your cluster setup.
```
tgaddair: Since this is copy-pasted from above, we could move it into a helper function: `set_gpu_conf(conf) -> conf`

examples/keras_spark3_rossmann.py
chuanlihao: Thanks for reviewing. But I don't think that will work.

For training, "horovod.spark.run(train_fn, args, ...)" is invoked, which will setup the whole environment (e.g., set GPU resources for each worker) and trigger distributed training. That's why you could call "horovod.spark.task.get_available_devices" in each worker.

For transform, it's just a pure Spark application. You cannot call "horovod.spark.task.get_available_devices", because Horovod initialization is not involved and "horovod.spark.task.task_info.set_resources" is not even invoked. Now that it's a pure Spark application, why not get GPU resources in Spark way by calling "TaskContext.get().resources()['gpu']"?

@tgaddair, please correct me if I were wrong, thanks in advance.

examples/keras_spark3_rossmann.py
chuanlihao: Please see above discussion, thanks.

examples/keras_spark3_rossmann.py
chuanlihao: Thanks for reviewing.

Updated the code. Please help review again, thanks.

examples/keras_spark3_rossmann.py
tgaddair: Yes, you're correct @chuanlihao.  For inference we are not running Horovod, just using a Spark UDF.  I would add a comment explaining this distinction, because it's not obvious.

chuanlihao: Thanks for your confirmation. Added a comment for this distinction in line 601, commit 4.

anfeng:
The revise PR looks good

PR #1675, opened 2020-01-16
Fixed issue with ThreadPool not joining on reinitialization

tgaddair:
Recreating the ThreadPool without joining was resulting in errors of the form:

```
Thu Jan 16 16:21:02 2020[0]<stderr>:terminate called without an active exception
Thu Jan 16 16:21:02 2020[0]<stderr>:Aborted
```

This PR allows `ThreadPool::create` to be called during the second init by joining and clearing threads on shutdown.

Also fixes an issue with mixed install (using CUDA without any GPU Horovod ops) where Horovod CUDA headers would be included and the `HAVE_CUDA` macro would be set, but implementation files would not be included, causing symbol lookup errors at runtime.

### DIFF
--- a/test/test_torch.py
+++ b/test/test_torch.py
@@ -76,6 +76,26 @@ class TorchTests(unittest.TestCase):
            types = [t for t in types if t in ccl_supported_types]
         return types
 
+    def test_horovod_reinit(self):
+        """Test that Horovod can init -> shutdown -> init successfully."""
+        mpi_rank, _ = mpi_env_rank_and_size()
+        gloo_rank = int(os.getenv('HOROVOD_RANK', -1))
+
+        is_mpi = gloo_rank == -1
+        if is_mpi:
+            # Only applies for Gloo
+            return
+
+        hvd.init()
+        rank, size = hvd.rank(), hvd.size()
+
+        hvd.shutdown()
+        hvd.init()
+        rank2, size2 = hvd.rank(), hvd.size()
+
+        assert rank == rank2
+        assert size == size2
+
     def test_horovod_rank(self):
         """Test that the rank returned by hvd.rank() is correct."""
         mpi_rank, _ = mpi_env_rank_and_size()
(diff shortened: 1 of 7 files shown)


tgaddair:
@EnricoMi any idea why the docs workflows are failing?  An issue on GitHub's end?

EnricoMi:
> @EnricoMi any idea why the docs workflows are failing? An issue on GitHub's end?
Yes, looks like checks didn't even start, just rerun the checks.

tgaddair:
> > @EnricoMi any idea why the docs workflows are failing? An issue on GitHub's end?
> 
> Yes, looks like checks didn't even start, just rerun the checks.

Thanks, seems to be working now (was down for about an hour, it seems).

test/test_torch.py
```diff
@@ -76,6 +76,26 @@ def filter_supported_types(self, types):
...
         return types
 
+    def test_horovod_reinit(self):
+        """Test that Horovod can init -> shutdown -> init successfully."""
+        mpi_rank, _ = mpi_env_rank_and_size()
+        gloo_rank = int(os.getenv('HOROVOD_RANK', -1))
+
+        is_mpi = gloo_rank == -1
+        if is_mpi:
+            # Only applies for Gloo
```
romerojosh: Good catch on this one and glad you added a test for reinitialization! Does this issue only impact gloo?

tgaddair: In the case of MPI, it seems that shutdown is permanent:

```
*** The MPI_Query_thread() function was called after MPI_FINALIZE was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort
```

The reinit support is relevant for Gloo because we need to do it for fault tolerance.

romerojosh: Yep, you are absolutely right. Shutdown invokes `MPI_Finalize` which is permanent. Looks good!

PR #1682, opened 2020-01-21
Fixes the pickle incompatibility between python 2 and 3

abditag2:


horovod/run/util/cache.py
```diff
@@ -75,11 +68,26 @@ def put(self, key, val):
...
+    def _dump(self, content):
+        with open(self._cache_file, 'wb') as cf:
+            cloudpickle.dump(content, cf, protocol=self._pickle_protocol)
+
+    def _load(self):
+        with open(self._cache_file, 'rb') as cf:
+            try:
+                content = cloudpickle.load(cf)
+            except Exception as e:
+                print(
```
tgaddair: Should we just delete the file automatically?

abditag2: good idea. Will do.

PR #1683, opened 2020-01-21
Convert single loss item to list in torch estimators

abditag2:


horovod/spark/torch/remote.py
```diff
@@ -36,9 +36,12 @@
 
 
 def RemoteTrainer(estimator, metadata, last_checkpoint_state, run_id, dataset_idx):
+    # Utility functions
+    make_list_if_not_list = make_list_if_not_list_fn()
```
tgaddair: Since we're using this function before entering the remote code, no need to make it a generator function.

horovod/spark/torch/remote.py
```diff
@@ -548,3 +550,15 @@ def calculate_loss(outputs, labels, loss_weights, loss_fns, sample_weights=None)
         return loss
 
     return calculate_loss
+
+
+def make_list_if_not_list_fn():
```
tgaddair: I would move this to `horovod/spark/common/util.py`.

test/test_spark_torch.py
```diff
@@ -354,3 +354,20 @@ def test_torch_direct_parquet_train(self):
                     transformer = est.fit_on_parquet()
                     predictions = transformer.transform(df)
                     assert predictions.count() == df.count()
+
+    def test_make_list_if_not_list(self):
```
tgaddair: If we move the function then this should be in `test_spark.py`.

horovod/spark/torch/remote.py
```diff
@@ -36,9 +36,12 @@
 
 
 def RemoteTrainer(estimator, metadata, last_checkpoint_state, run_id, dataset_idx):
+    # Utility functions
+    make_list_if_not_list = make_list_if_not_list_fn()
+
     # Estimator parameters
-    loss_fns_pre_train = estimator.getLoss()
-    loss_constructors = estimator.getLossConstructors()
+    loss_fns_pre_train = make_list_if_not_list(estimator.getLoss())
```
tgaddair: What if you have multiple label cols but only want to specify a single loss.  Do we need to ensure that the list length is equal to the number of label cols?

tgaddair: We have tests in `test_spark_torch.py` that do not provide loss as a list, but the tests are passing.  Based on this, I'm not sure how that's possible, do you?

abditag2: If we specify a single loss for a multi-output model, that loss will be used for all the outputs.

horovod/spark/common/params.py
```diff
@@ -156,6 +156,8 @@ def getStore(self):
         return self.getOrDefault(self.store)
 
     def setLoss(self, value):
+        if not isinstance(value, list):
```
tgaddair: I'm not a fan of this solution because it results in counter-intuitive behavior:

```
estimator.setLoss(loss)
estimator.getLoss() == loss  # False, but expected True
```

We should do the wrapping etc. only at train time.

horovod/spark/torch/remote.py
```diff
@@ -56,13 +56,12 @@ def RemoteTrainer(estimator, metadata, last_checkpoint_state, run_id, dataset_id
...
 
     # If loss weight is not provided, use equal loss for all the labels
-    label_columns = estimator.getLabelCols()
     loss_weights = estimator.getLossWeights()
     if not loss_weights:
         num_labels = len(label_columns)
         loss_weights = [float(1) / num_labels for _ in range(num_labels)]
 
-    # Utility functions
+    # More utility functions
```
tgaddair: Just `Utility functions`.

horovod/spark/common/params.py
abditag2: I see your point, will revert back.

horovod/spark/torch/remote.py
abditag2: fixed.

test/test_spark_torch.py
abditag2: moved.

horovod/spark/torch/remote.py
abditag2: moved.

horovod/spark/torch/remote.py
abditag2: In the init function of the TorchEstimator, we had a check if loss and loss_constructors were a list, and if not, we made them a list. This problem came up when we use est.setLoss().

horovod/spark/torch/remote.py
abditag2: NA anymore.

test/test_spark_torch.py
```diff
@@ -343,14 +343,16 @@ def test_torch_direct_parquet_train(self):
...
                         optimizer=optimizer,
-                        loss=loss,
                         input_shapes=[[2]],
                         feature_cols=['features'],
                         label_cols=['y'],
                         batch_size=1,
                         epochs=3,
                         verbose=2)
 
+                    # To make sure that setLoss works with non-list loss.
```
tgaddair: Looks like code paths are the same for constructor and this function now, right?  Was the issue before that users were calling `setLoss` which did not coerce the param into a list?

test/test_spark.py
```diff
@@ -772,3 +772,19 @@ def fn():
         with spark_session('test_get_available_devices', gpus=2):
             res = horovod.spark.run(fn, env={'PATH': os.environ.get('PATH')}, verbose=0)
             self.assertListEqual([(['0'], 0), (['1'], 1)], res)
+
+
+    def test_make_list_if_not_list(self):
```
tgaddair: `test_to_list`

test/test_spark_torch.py
abditag2: I believe so. Passing the loss directly during instantiation works fine as we had it in the tests.

tgaddair: I think we can move this back to the constructor, then.

test/test_spark_torch.py
```diff
@@ -343,14 +343,16 @@ def test_torch_direct_parquet_train(self):
...
+                    est.setLoss(loss)
+
```
tgaddair: Can you add tests for `_calculate_loss_fn` in `remote.py`?  I think this problem could have been caught before if we had testing for multi-label scenarios.

horovod/spark/torch/remote.py
```diff
@@ -56,8 +56,7 @@ def RemoteTrainer(estimator, metadata, last_checkpoint_state, run_id, dataset_id
     train_minibatch = train_minibatch_fn if train_minibatch_fn else _train_minibatch_fn()
 
     # If loss weight is not provided, use equal loss for all the labels
-    label_columns = estimator.getLabelCols()
-    loss_weights = estimator.getLossWeights()
+    loss_weights = to_list(estimator.getLossWeights())
```
tgaddair: I don't think this will work.  If you look in https://github.com/horovod/horovod/blob/master/horovod/spark/torch/remote.py#L536 you'll see that we expect the length of loss_weights to always match the length of the labels.  We should so something similar to what we do on line 62 here.

horovod/spark/torch/remote.py
```diff
@@ -37,8 +37,8 @@
 
 def RemoteTrainer(estimator, metadata, last_checkpoint_state, run_id, dataset_idx):
     # Estimator parameters
-    loss_fns_pre_train = estimator.getLoss()
-    loss_constructors = estimator.getLossConstructors()
+    loss_fns_pre_train = to_list(estimator.getLoss())
```
tgaddair: In https://github.com/horovod/horovod/blob/master/horovod/spark/torch/remote.py#L522 we pad the list of loss fns on every call to `calculate_loss`.  Why not do it here?  We could use `num_labels` to ensure the loss / loss_constructors / loss_weights are the right length.  We could even add it to `to_list`:

```
loss_fns_pre_train = to_list(estimator.getLoss(), length=num_labels)
```

horovod/spark/common/util.py
```diff
@@ -587,3 +587,23 @@ def get_dataset_properties(dataset_idx):
...
+
+
+def to_list(var, length):
+    if var is None:
+        return None
+
+    if not isinstance(var, list):
+        var = [var]
+
+    # If only one loss function or loss_constructor is passed by user, use
```
tgaddair: Since this is in `common`, we should not reference loss and loss_constructors.

horovod/spark/torch/remote.py
```diff
@@ -54,13 +52,21 @@ def RemoteTrainer(estimator, metadata, last_checkpoint_state, run_id, dataset_id
     user_verbose = estimator.getVerbose()
     train_minibatch_fn = estimator.getTrainMinibatchFn()
     train_minibatch = train_minibatch_fn if train_minibatch_fn else _train_minibatch_fn()
+    loss_fns_pre_train = to_list(estimator.getLoss(),
```
tgaddair: Let's pull `num_labels` out from line 63 and put it just above line 55 (`loss_fns_pre_train = ...`).  That way, we can just reference `num_labels` instead of `len(label_columns)`.

test/test_spark_torch.py
abditag2: added.

horovod/spark/torch/remote.py
abditag2: Moved it.

horovod/spark/common/util.py
abditag2: addressed

PR #1689, opened 2020-01-25
Prevent launching Horovod ops in MXNet DistributedOptimizer/DistributedTrainer if running on a single rank.

romerojosh:
The current MXNet `DistributedOptimizer`/`DistributedTrainer` implementations will unnecessarily launch Horovod ops in the case where a script is run on a single rank. This PR adds checks to prevent this from occurring.

### DIFF
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -51,6 +51,8 @@ class DistributedOptimizer(mx.optimizer.Optimizer):
         return self._optimizer.create_state_multi_precision(index, weight)
 
     def _do_allreduce(self, index, grad):
+        if size() == 1: return
+
         if isinstance(index, (tuple, list)):
             for i in range(len(index)):
                 allreduce_(grad[i], average=False,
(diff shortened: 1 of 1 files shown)


karan6181:
Thank You for your contributions! Is it possible to have a unittest for this?

romerojosh:
@karan6181 I’m not sure it is possible to add a unit test for this one. The issue is that in the Python API, there isn’t a way to tell if Horovod ops have been launched or not.

romerojosh:
@apeforest mind taking a look at this one?

@taddair do you think it would a good idea to add single GPU testing in order to make sure any code unique to `size == 1` (like what is introduced in this PR) is functional?

tgaddair:
> @tgaddair do you think it would a good idea to add single GPU testing in order to make sure any code unique to `size == 1` (like what is introduced in this PR) is functional?

I think that's a good idea.  Most of the unit tests are designed to work with 2 GPUs, so we would need to create a separate set of tests for single-process.  It may be sufficient to just run the MNIST examples without `mpirun` or `horovodrun` as a start.

Also, do you think it's necessary to use GPU, or can we get away with a CPU test (to reduce costs)?  I think the code paths should be nearly the same.

romerojosh:
I'd agree on both points. MNIST is probably all we need since I think the `size == 1` special cases only occur in the `DistributedOptimizer` wrappers, not the lower level ops. CPU testing only should also be adequate.

romerojosh:
@tgaddair Can we merge this one and add a separate PR to deal with adding single worker testing?

tgaddair:
@romerojosh Sounds good.  I'll go ahead and merge.

Issue #1679, opened 2020-01-21
DUPLICATE_NAME_ERROR when training a GAN system with mxnet


feevos:
Dear all, 

I am trying to train a "modified" GAN system, based on mxnet. It is basically a domain adaptation problem, where there are two networks, say a Generator G, and a discriminator T.  The whole training process is wrapped in a class, where it holds the trainers and the networks. This is a portion of the code where I define relative horovod+mxnet functions/trainers etc: 

```python
        # @@@@@@@@@@@@@@@@@@@ DEFINE THESE IN CONFIGURATION DICTIONARY @@@@@@@@@@@@@@@@@@@
        # @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@        
        self.netG =  self.config[C.C_MODEL_G]# Features NET
        self.netT =  self.config[C.C_MODEL_T] # Talahassee network 

        self.ctxG =  self.config[C.C_CTX_G]
        self.ctxT =  self.config[C.C_CTX_T]


        # Initialize networks -- takes care the case where there is no flname to load params.
        self._initialize_network(self.netG, self.config[C.C_INITIALIZER_G], self.ctxG, self.config[C.C_FLNM_G_LOAD_PARAMS])
        self._initialize_network(self.netT, self.config[C.C_INITIALIZER_T], self.ctxT, self.config[C.C_FLNM_T_LOAD_PARAMS])

        # This should take place ONLY if the parameters were randomly initialized 
        # @@@@@@@@@@@@ HVD BROADCAST ALL PARAMETERS TO ALL WORKERRS @@@@@@@@@@@@@@@
        if self.config[C.C_FLNM_G_LOAD_PARAMS] is None and self.netG.collect_params() is not None:
                hvd.broadcast_parameters(self.netG.collect_params(), root_rank=0)

        if self.config[C.C_FLNM_T_LOAD_PARAMS] is None and self.netT.collect_params() is not None:
                hvd.broadcast_parameters(self.netT.collect_params(), root_rank=0)
        # @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@




        # Initialize Trainer object - if you want to load previously saved states. 
        # TODO: Add functional definition 
[...]

PR #1696, opened 2020-01-29
Use parameter names when scheduling allreduces in DistributedTrainer. 

romerojosh:
#1186 introduced a sort of the parameter list to ensure the parameters were submitted to Horovod in a consistent order across workers when using `DistributedTrainer` so that requests to Horovod are given globally consistent names. This sort should not be applied in the case that the parameter list originates from a `ParameterDict` in MXNet as that data structure is already ordered (based on the graph construction). 

A particular reason this sort should only be applied when absolutely necessary is that the sort by name produces `priority` values based on a name-based ordering, which is unrelated to the underlying computation. Preserving the order of the parameter list from a `ParameterDict` results in assigned priorities consistent with those applied in the native MXNet `Trainer`. 

Update: Instead of conditionally applying the sort, new solution is to just use the parameter names when scheduling the allreduces so that the order of submission no longer matters.

### DIFF
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -102,11 +102,10 @@ class DistributedTrainer(mx.gluon.Trainer):
     def _allreduce_grads(self):
         if size() == 1: return
 
-        # sort needed for Python < 3.6 is not guaranteed
-        for i, param in enumerate(sorted(self._params, key=lambda p: p.name)):
+        for i, param in enumerate(self._params):
             if param.grad_req != 'null':
                 allreduce_(param.list_grad()[0], average=False,
-                           name=str(i), priority=-i)
+                           name=param.name, priority=-i)
 
 
 # Wrapper to inject Horovod broadcast after parameter initialization


horovod/mxnet/__init__.py
```diff
@@ -99,11 +99,16 @@ def __init__(self, params, optimizer, optimizer_params=None):
         # average in allreduce, has better performance. 
         self._scale /= size()
 
+        self._keep_param_order = isinstance(params, mx.gluon.parameter.ParameterDict)
```
eric-haibin-lin: Thanks for adding the improvement. For some rare case, users define the parameter in nondeterministic orders. That would lead to different order of insertion to ParameterDict. I'm not sure how to easily workaround that, unless all workers exchange the parameter order and name at initialization. What do you think?

romerojosh: Actually, I think there might be a better solution. Right now, the requests to Horovod are being named with `str(i)` which is what is causing the issue with the ordering of the loop. If we instead name the requests with the actual parameter names instead, the order they are submitted will not matter. In the case of inconsistently ordered parameters, the priorities will be set differently across workers, but that would also happen in the native `trainer.py` with kvstore. Does this alternate solution sound alright to you?

eric-haibin-lin: Oh you're right. I like this solution much better. 
BTW - do you find the priority helpful in scheduling horovod allreduce? Any perf benchmark u can share?

romerojosh: Ok great! I will update the PR.

I do not currently have any extensive perf benchmarks for this.

horovod/mxnet/__init__.py
```diff
@@ -102,11 +102,10 @@ def __init__(self, params, optimizer, optimizer_params=None):
     def _allreduce_grads(self):
         if size() == 1: return
 
-        # sort needed for Python < 3.6 is not guaranteed
-        for i, param in enumerate(sorted(self._params, key=lambda p: p.name)):
+        for i, param in enumerate(self._params):
             if param.grad_req != 'null':
                 allreduce_(param.list_grad()[0], average=False,
-                           name=str(i), priority=-i)
+                           name=param.name, priority=-i)
```
apeforest: Do we need a sanity check of no same name here? Otherwise, we may end up with two allreduce operation with the same name?

romerojosh: The backend will already generate an error for duplicate names if it detects them so we should not need an additional check here.

romerojosh:
This PR may also partially fix #1679. In that issue, the two `DistributedTrainer`s in the GAN are using duplicate names since the allreduce names are assigned as `str(i)` for both. By using the parameters names to schedule the allreduces, this _may_ be fixed, assuming that the parameters for each network in the GAN have unique names.

PR #1700, opened 2020-01-30
Pinned Pandas version to 0.24.2

abditag2:
Petastorm is not compatible with the newly released Pandas==1.0.0 and causes the tests to fail. We will remove this pinned version after petastorm is fixed.

### DIFF
--- a/Dockerfile.test.cpu
+++ b/Dockerfile.test.cpu
@@ -97,7 +97,7 @@ RUN pip install ${TENSORFLOW_PACKAGE}
 
 # Install Keras.
 # Pin scipy<1.4.0: https://github.com/scipy/scipy/issues/11237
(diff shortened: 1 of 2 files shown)


tgaddair:
@nvcastet looks like there was a transient failure in jenkins.  Do you know if there's an easy way to re-run (without creating a new diff)?

nvcastet:
@tgaddair I restarted the build. But you can go to the PR URL: https://powerci.osuosl.org/job/Horovod_PPC64LE_GPU_PIPELINE/view/change-requests/job/PR-1700/ and click `Build Now` on the left side. You need to login and be an admin to see the button appear.

tgaddair:
Ah, thanks!  I missed the fact that I wasn't logged in, but I see the button now.

PR #1702, opened 2020-01-31
Update tests to petastorm==0.8.2 to fix pandas==1.0.0 compatibility

abditag2:
After fix is confirmed and new petastorm version is released, we will update the petastorm version and merge.


tgaddair:
We'll probably need to wait on v0.8.2 for https://github.com/uber/petastorm/pull/481 to be merged in.

PR #1698, opened 2020-01-29
Fix tests_required in setup.py

EnricoMi:

PR #1705, opened 2020-01-31
Add dev to extras_require in setup.py

EnricoMi:
This simplifies initial setup of an development environment by invoking
```
pip install -e .[dev]
```
Rather than extracting the dependencies from documentation.

Note: this contains #1698, which is unrelated to this PR and should go into master first.

### DIFF
--- a/Dockerfile.test.cpu
+++ b/Dockerfile.test.cpu
@@ -6,6 +6,7 @@ FROM ubuntu:${UBUNTU_VERSION}
 ARG UBUNTU_VERSION=16.04
 ARG MPI_KIND=OpenMPI
 ARG PYTHON_VERSION=3.6
+# NOTE: keep versions in sync with setup.py extras_require{'dev'}:
 ARG TENSORFLOW_PACKAGE=tensorflow-cpu==1.15.0
 ARG KERAS_PACKAGE=keras==2.2.4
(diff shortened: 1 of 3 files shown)


DEKHTIARJonathan:
I don't think this PR shall be merged as-is.
Python packages provides: `extras_require` to plan for this.
A development environment shall be installed as followed: `pip install -e ".[dev]"`.

In this case, as Horovod contains also C files, they need to be rebuilt if modified.
Requirements files can be used to specify dependencies and read in `setup.py`.|

I would argue that using requirement files is probably a better practice than stating dependencies in `setup.py` as doing so tends to clutter the git history for this file.

EnricoMi:
@DEKHTIARJonathan I have updated the `setup.py` with a `dev` target to install recommended versions. This simplifies docs and codifies dev dependencies. What do you think?

DEKHTIARJonathan:
@tgaddair what do u think?
I like it like this IMHO it's more "clean" than our current states.

tgaddair (approved):
This is a good start.  Going forward, I think it would be good to find ways to add more deps here so we can ensure the local and test environments are as close as possible.  One way might be to add a `requirements_minimal.txt` file that has pinned versions of all dependencies (from `pip freeze`), and have `pip install .[dev]` and the the min CPU test use that file.  Something like that.

PR #1711, opened 2020-02-05
tensorboardX: catch the exception when tensorboardX isn't installed

pragupta:
Signed-off-by: Prachi Gupta <pragupta@us.ibm.com>

### DIFF
--- a/examples/pytorch_imagenet_resnet50.py
+++ b/examples/pytorch_imagenet_resnet50.py
@@ -11,6 +11,7 @@ import horovod.torch as hvd
 import os
 import math
 from tqdm import tqdm
+from distutils.version import LooseVersion
 
 # Training settings
 parser = argparse.ArgumentParser(description='PyTorch ImageNet Example',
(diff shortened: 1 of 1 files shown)


nvcastet:
@tgaddair It it to make `tensorboardX` optional for this example, if it is not installed.

examples/pytorch_imagenet_resnet50.py
```diff
@@ -85,7 +84,11 @@
 verbose = 1 if hvd.rank() == 0 else 0
 
 # Horovod: write TensorBoard logs on first worker.
-log_writer = tensorboardX.SummaryWriter(args.log_dir) if hvd.rank() == 0 else None
+try:
+    import tensorboardX
```
tgaddair: PyTorch 1.2 and above support Tensorboard natively, which we use in the [TorchEstimator](https://github.com/horovod/horovod/blob/master/horovod/spark/torch/remote.py#L25).  Maybe we can check that here as well?  Something like:

```
try:
    if LooseVersion(torch.__version__) >= LooseVersion('1.2.0'):
        from torch.utils.tensorboard import SummaryWriter
    else
        from tensorboardX import SummaryWriter
    log_writer = SummaryWriter(args.log_dir) if hvd.rank() == 0 else None
except ImportError as e:
    log_writer = None
```
```

nvcastet: Nice, i did not know about `torch.utils.tensorboard`. @praguta And `except ImportError:` instead of `except ImportError as e:`

PR #1718, opened 2020-02-09
mxnet: Remove unnecessary wait_to_read

eric-haibin-lin:
@apeforest @romerojosh I don't know why this `wait_to_read` was needed. Shall we remove it?

### DIFF
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -114,7 +114,6 @@ def _append_broadcast_init(param, root_rank):
     def wrapped_init_impl(self, *args, **kwargs):
         init_impl(*args, **kwargs)
         broadcast_(self.data(), root_rank=root_rank)
-        self.data().wait_to_read()
     return wrapped_init_impl
 
 
@@ -151,8 +150,3 @@ def broadcast_parameters(params, root_rank=0):
     # Run broadcasts.
     for i, tensor in enumerate(tensors):
         broadcast_(tensor, root_rank, str(i))
-
-    # Make sure tensors pushed to MXNet engine get processed such that all
-    # workers are synced before starting training.
-    for tensor in tensors:
-        tensor.wait_to_read()


romerojosh:
@eric-haibin-lin Can you rebase on the current master to fix tests?

apeforest:
Yes, I think we can remove it. Could you run the imagenet example once to make sure it doesn't break?

tgaddair:
Any update on this @eric-haibin-lin?  Would be great to get this into the next release.

horovod/mxnet/__init__.py
```diff
@@ -151,8 +150,3 @@ def broadcast_parameters(params, root_rank=0):
     # Run broadcasts.
     for i, tensor in enumerate(tensors):
         broadcast_(tensor, root_rank, str(i))
-
-    # Make sure tensors pushed to MXNet engine get processed such that all
-    # workers are synced before starting training.
-    for tensor in tensors:
-        tensor.wait_to_read()
```
apeforest: I remember there was a case if one worker received the tensor and started the forward-backward computation but other workers have not yet received the tensors. If this early worker finishes the batch well ahead of other workers there would be some problem of Horovod handling such non-synchronized situation. @tgaddair is this still a limitation of Horovod backend?

eric-haibin-lin: Does that happen when training with only 1 batch of data?? What error will be printed on such a case?

romerojosh: @apeforest I'm not sure I follow the desync situation you are describing. If a worker begins forward/backward earlier than all other workers, it will still submit allreduce operations to the MXNet engine, which will not complete until all other workers submit and launch the matching allreduce ops. This should keep them synchronized.

apeforest: @romerojosh You are right. I think this should be fine.

PR #1728, opened 2020-02-18
Use backend api instead of tensorflow api in keras callbacks.

bethune-bryant:
I ran into a type error with `hvd.callbacks.MetricAverageCallback()`. It turns out it was because `tf.Variable` did not respect the type specified with keras `set_floatx`. Using the backend API fixed the issue, and it's probably good to do that every where that's possible.

@nvcastet @bnemanich

### DIFF
--- a/horovod/_keras/callbacks.py
+++ b/horovod/_keras/callbacks.py
@@ -52,8 +52,8 @@ class MetricAverageCallbackImpl(object):
         self.device = device
 
     def _make_variable(self, metric, value):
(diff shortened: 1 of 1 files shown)


tgaddair:
Thanks for the PR @bethune-bryant!  Could you rebase on master?  It should fix issues with building the docs.

bethune-bryant:
@tgaddair Thanks, I rebased and it passed.

PR #1736, opened 2020-02-26
Ping mxnet nightly to the lastest path

apeforest:
MXNet is now publishing nightly to a webpage. Update tests to the correct path.

@karan6181 @eric-haibin-lin please review.

### DIFF
--- a/Dockerfile.test.cpu
+++ b/Dockerfile.test.cpu
@@ -114,7 +114,7 @@ RUN pip install "Pillow<7.0" --no-deps
 
 # Install MXNet.
 RUN if [[ ${MXNET_PACKAGE} == "mxnet-nightly" ]]; then \
-        pip install -pre mxnet -f https://dist.mxnet.io/python/all; \
+        pip install --pre mxnet -f https://dist.mxnet.io/python/all; \
     else \
         pip install ${MXNET_PACKAGE} ; \
     fi
--- a/Dockerfile.test.cpu
+++ b/Dockerfile.test.cpu
@@ -116,7 +116,7 @@ RUN pip install "Pillow<7.0" --no-deps
 RUN if [[ ${MXNET_PACKAGE} == "mxnet-nightly" ]]; then \
         pip install -pre mxnet -f https://dist.mxnet.io/python/all; \
     else \
-        pip install ${MXNET_PACKAGE} \
+        pip install ${MXNET_PACKAGE} ; \
     fi
 
 # Install Horovod.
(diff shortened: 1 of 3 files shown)


Dockerfile.test.cpu
```diff
@@ -113,7 +113,11 @@ RUN if [[ ${PYTORCH_PACKAGE} == "torch-nightly" ]]; then \
 RUN pip install "Pillow<7.0" --no-deps
 
 # Install MXNet.
-RUN pip install ${MXNET_PACKAGE}
+RUN if [[ ${MXNET_PACKAGE} == "mxnet-nightly" ]]; then \
+        pip install -pre mxnet -f https://dist.mxnet.io/python/all; \
+    else \
+        pip install ${MXNET_PACKAGE} \
```
tgaddair: I think you're missing a `;` here.

apeforest: Thanks for the good catch!

Dockerfile.test.cpu
```diff
@@ -113,7 +113,11 @@ RUN if [[ ${PYTORCH_PACKAGE} == "torch-nightly" ]]; then \
 RUN pip install "Pillow<7.0" --no-deps
 
 # Install MXNet.
-RUN pip install ${MXNET_PACKAGE}
+RUN if [[ ${MXNET_PACKAGE} == "mxnet-nightly" ]]; then \
+        pip install -pre mxnet -f https://dist.mxnet.io/python/all; \
```
tgaddair: On this line, it should be `--pre` instead of `-pre`.

Dockerfile.test.gpu
```diff
@@ -96,7 +96,11 @@ RUN if [[ ${PYTORCH_PACKAGE} == "torch-nightly" ]]; then \
 RUN pip install "Pillow<7.0" --no-deps
 
 # Install MXNet.
-RUN pip install ${MXNET_PACKAGE}
+RUN if [[ ${MXNET_PACKAGE} == "mxnet-nightly" ]]; then \
+        pip install -pre mxnet_cu100 -f https://dist.mxnet.io/python/all; \
```
tgaddair: Tests for the nightly builds are running on CUDA 10.1.  Should this be `mxnet_cu101`?

Dockerfile.test.cpu
apeforest: Fixed.

Dockerfile.test.gpu
apeforest: Updated to cu101

tgaddair:
Hey @apeforest, looks like MXNet nightly is causing tests to fail: https://buildkite.com/horovod/horovod/builds/2116

Can you investigate?

apeforest:
Looking into it

apeforest:
I don't find issue when I build horovod locally with mxnet nightly. One possible issue is that MXNet stopped supporting python 2.7 since the latest release. I will make changes to the build script.

apeforest:
@tgaddair Created another PR https://github.com/horovod/horovod/pull/1746 to skip mxnet nightly when Python version is 2.7

PR #1739, opened 2020-02-26
Fix bug in cache dump

nvcastet:


tgaddair:
Thanks for the fix @nvcastet!  @abditag2 can you review?

PR #1744, opened 2020-02-27
Add '--tcp' flag to horovodrun for tcp only communication

nvcastet:


tgaddair:
Hey @nvcastet, can you rebase off master?  There was a fix for MXNet just landed.

horovod/run/run.py
```diff
@@ -559,6 +559,8 @@ def parse_args():
                                        help='Extra MPI arguments to pass to mpirun. '
                                        'They need to be passed with the equal sign to avoid parsing issues. '
                                        'e.g. --mpi-args="--map-by ppr:6:node"')
+    group_library_options.add_argument('--tcp', action='store_true', dest='tcp_flag',
+                                       help='If this flag is set, TCP is only used for communication.')
```
bethune-bryant: It may be a knit-pick, but it might be clearer for the help text to read:
`'If this flag is set, only TCP is used for communication.'`

PR #1746, opened 2020-02-27
Do not run mxnet nightly with py2.7

apeforest:
MXNet nightly stopped supporting python 2.7. This PR will test horovod with mxnet 1.5 when Python version is 2.7


apeforest:
@tgaddair Please review. Thanks

PR #1750, opened 2020-02-28
support data transformation  before train and validation in torch estimators

abditag2:
Some traninig jobs require to apply custom transformations to each batch before passing it to train or validation step. For instance, for training resnet model, each batch is cropped and scaled. this diff enables such transformations: 

an example transformation is:
 
```
def _transform_row(row):
    transform = transforms.Compose([
        transforms.RandomResizedCrop(224),
        transforms.RandomHorizontalFlip(),
        transforms.ToTensor(),
        transforms.Normalize(mean=[0.485, 0.456, 0.406],
                             std=[0.229, 0.224, 0.225])
    ])
    codec = CompressedImageCodec('png')
    class_to_index = {val[0]: int(key) for key, val in index_to_class.items()}

    row['image'] = row['image'].apply(lambda row: transform(Image.fromarray(codec.decode("image", row))).flatten().tolist())
    row['noun_id'] = row['noun_id'].apply(lambda val: class_to_index[val])
    return row
```
with 
```
index_to_class = {"0": ["n01440764", "tench"], "1": ["n01443537", "goldfish"],
                  "2": ["n01484850", "great_white_shark"],
                  "3": ["n01491361", "tiger_shark"],
					...
                  "999": ["n15075141", "toilet_tissue"]}
```

### DIFF
--- a/horovod/spark/keras/estimator.py
+++ b/horovod/spark/keras/estimator.py
@@ -140,6 +140,9 @@ class KerasEstimator(HorovodEstimator, KerasEstimatorParamsReadable,
         train_steps_per_epoch: Number of steps to train each epoch. Useful for testing that model trains successfully.
                                Defaults to training the entire dataset each epoch.
         validation_steps_per_epoch: Number of validation steps to perform each epoch.
+        transformation_fn: Optional custom function to execute before passing the loaded data into the train and validation steps.
+                This is useful for tasks such as cropping and rescaling of image samples before train starts.
+
     """
 
     custom_objects = Param(Params._dummy(), 'custom_objects', 'custom objects')
@@ -169,7 +172,8 @@ class KerasEstimator(HorovodEstimator, KerasEstimatorParamsReadable,
                  partitions_per_process=None,
                  run_id=None,
                  train_steps_per_epoch=None,
-                 validation_steps_per_epoch=None):
+                 validation_steps_per_epoch=None,
+                 transformation_fn=None):
 
         super(KerasEstimator, self).__init__()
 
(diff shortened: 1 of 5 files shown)


horovod/spark/keras/estimator.py
```diff
@@ -140,6 +140,9 @@ class KerasEstimator(HorovodEstimator, KerasEstimatorParamsReadable,
         train_steps_per_epoch: Number of steps to train each epoch. Useful for testing that model trains successfully.
                                Defaults to training the entire dataset each epoch.
         validation_steps_per_epoch: Number of validation steps to perform each epoch.
+        transformation_fn: Optional custom function to execute before passing the loaded data into the train and validation steps.
```
tgaddair: Here we should explain how this works a little more.  A reference to the Petastorm documentation would be useful.  Something like: `Optional function that takes a row as its parameter and returns a modified row that is then fed into the train or validation step.  This transformation is applied after batching.  See Petastorm [TransformSpec](link) for more details`.

horovod/spark/torch/estimator.py
```diff
@@ -119,6 +119,8 @@ class TorchEstimator(HorovodEstimator, TorchEstimatorParamsWritable,
         train_steps_per_epoch: Number of steps to train each epoch. Useful for testing that model trains successfully.
                                Defaults to training the entire dataset each epoch.
         validation_steps_per_epoch: Number of validation steps to perform each epoch.
+        transformation_fn: Optional custom function to execute before passing the loaded data into the train and validation steps.
```
tgaddair: Same as above.

PR #1751, opened 2020-02-28
Add extra build flag 'HOROVOD_BUILD_ARCH_FLAGS'

nvcastet:
Add env variable to specify architecture-specific compiler flags.
For example, to be compatible with the older CPU architecture 'Sandy Bridge' and tune for 'Broadwell':
HOROVOD_BUILD_ARCH_FLAGS="-march=sandybridge -mtune=broadwell" ... pip install horovod

Signed-off-by: Nicolas V Castet <nvcastet@us.ibm.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -118,17 +118,17 @@ def get_supported_instruction_set_flags(flags_to_check):
 def get_cpp_flags(build_ext):
     last_err = None
     default_flags = ['-std=c++11', '-fPIC', '-O2', '-Wall', '-fassociative-math', '-ffast-math', '-ftree-vectorize', '-funsafe-math-optimizations']
-    avx_env_vars = os.environ.get('HOROVOD_BUILD_ARCH_FLAGS')
-    avx_fma_flags = get_supported_instruction_set_flags(['-mf16c', '-mavx', '-mfma']) if avx_env_vars is None else avx_env_vars.split()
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -118,7 +118,8 @@ def get_supported_instruction_set_flags(flags_to_check):
 def get_cpp_flags(build_ext):
     last_err = None
     default_flags = ['-std=c++11', '-fPIC', '-O2', '-Wall', '-fassociative-math', '-ffast-math', '-ftree-vectorize', '-funsafe-math-optimizations']
-    avx_fma_flags = get_supported_instruction_set_flags(['-mf16c', '-mavx', '-mfma'])
+    avx_env_vars = os.environ.get('HOROVOD_BUILD_ARCH_FLAGS')
+    avx_fma_flags = get_supported_instruction_set_flags(['-mf16c', '-mavx', '-mfma']) if avx_env_vars is None else avx_env_vars.split()
```
romerojosh: Looks good to me but I wonder if we should rename these variables to something more generic, like `avx_env_vars` -> `build_arch_flags_env` and `avx_fma_flags` -> `build_arch_flags`?

tgaddair: +1, looks great, I think making the variable names consistent with the environment variable is a good idea.

nvcastet: good catch

PR #1753, opened 2020-03-01
Import ABC from collections.abc instead of collections for Python 3.9 compatibility

### LINKED ISSUE #1752: Importing ABC directly from collections was deprecated and will be removed in Python 3.10. Use collections.abc instead.
**Bug report:**

```
test/test_torch.py
1114:                if not isinstance(p, collections.Iterable):
horovod/torch/__init__.py
526:        if isinstance(x, collections.Iterable):
```

tirkarthi:
Fixes #1752


tirkarthi:
Thanks @tgaddair for the review and merge.

PR #1763, opened 2020-03-03
[ROCm] add support for ROCm PyTorch extensions

jeffdaily:


tgaddair:
Hey @jeffdaily, can you rebase off master to fix Buildkite tests?

setup.py
```diff
@@ -1182,6 +1182,35 @@ def is_torch_cuda_v2(build_ext, include_dirs, extra_compile_args):
...
+        pattern = re.compile(r'-D(\w+)=?(\w+)?')
+        return [pattern.match(flag).groups() for flag in COMMON_HIPCC_FLAGS if pattern.match(flag)]
+    except:
+        return []
+
+
+def is_torch_rocm_v2(build_ext, include_dirs, extra_compile_args):
+    try:
+        from torch.utils.cpp_extension import include_paths
+        rocm_macros = get_torch_rocm_macros()
```
tgaddair: If this return `[]` due to an exception, does this effectively just compile CUDA and return True (if CUDA is available)?

jeffdaily: The `return []` is for the exception case in `get_torch_rocm_macros()`, yes.  The exception could occur if `COMMON_HIPCC_FLAGS` is not present in the installed torch module, indicating the torch version is too old and does not have our commits that added HIP extension support.  If the exception does not occur, i.e. `COMMON_HIPCC_FLAGS` is present, then the `test_compile` in `is_torch_rocm_v2` should fail for CUDA and CPU builds.

tgaddair: Thanks, Jeff.  To clarify: you're saying that if the exception occurs (old version of PyTorch), then this will succeed if CUDA is available, which in turn will set the `HAVE_ROCM` macro, correct?  If that's true, that would not seem correct.  It seems that the desired behavior would be: if `get_torch_rocm_macros()` raises an exception, then skip the build for `is_torch_rocm_v2()` and return False.  Am I missing something?

jeffdaily: Sorry for the delay in responding.  Considering `get_torch_rocm_macros()`, the exception should occur if and only if the pytorch version is old.  `get_torch_rocm_macros()` is used in two places. 

First, it is used in `is_torch_rocm_v2()` so that the compilation test can use the correct macros for ROCm.  That means `is_torch_rocm_v2()` can fail if either pytorch is old (such that `COMMON_HIPCC_FLAGS` is not present), or pytorch is new enough but was not compiled for ROCm (i.e., THH.h header not present), or some other unknown error has occurred that I can't predict.

`get_torch_rocm_macros()` is used again in the body of `build_torch_extension_v2`.  But this occurs only in the branch where we are compiling horovod with `HOROVOD_GPU=ROCM` and `is_torch_rocm_v2()` are both `True`.

Yes, `get_torch_rocm_macros()` can still succeed if compiling for CUDA when pytorch is sufficiently new.  `get_torch_rocm_macros()` represents a necessary but not sufficient condition for indicating a ROCm build, but otherwise will not affect CUDA.  If compiling for CUDA, the compiler check in `is_torch_rocm_v2()` will definitely fail, regardless of `get_torch_rocm_macros()` succeeding, and set `have_rocm` to `False`.

I hope that better answers your question. Thanks.

tgaddair: Thanks for the explanation @jeffdaily.

PR #1764, opened 2020-03-04
fixed reshaping bug in TorchModel

abditag2:


tgaddair:
Can you rebase off master to fix Buildkite tests?

PR #1765, opened 2020-03-04
Update MXNet nightly cpu build

apeforest:
MXNet nightly use mkldnn CPU build by default


tgaddair:
Hey @apeforest, can you rebase off master to fix Buildkite tests?

PR #1790, opened 2020-03-12
Fixes bug with sample weight in torch

abditag2:


tgaddair:
Thoughts on unit test failure?

PR #1803, opened 2020-03-19
Make host assignment prefer large hosts

EnricoMi:
The `get_host_assignments` method says:

> This function will try to allocate as many as possible processes on the same host to leverage local network.

When hosts are heterogeneous in size then jobs should be assigned to large nodes first. For equi-sized hosts, assignment should continue to be done in the order of given hosts.


horovod/run/common/util/hosts.py
```diff
@@ -87,7 +87,8 @@ def get_host_assignments(hosts, min_np, max_np=None):
     cross_sizes = collections.defaultdict(int)
 
     # allocate processes into slots
-    for host_idx, host_info in enumerate(hosts):
+    sorted_hosts = sorted(hosts, key=lambda host: host.slots, reverse=True)
```
tgaddair: In the elastic case we need to be careful about preserving the order that we assign ranks after their processes have been started.  This is because we need to ensure that the oldest worker is always assigned rank 0 (for the purpose of sending the most up-to-date state to the new workers).

So one way we might be able to achieve this would be to track the timestamp that the host was first assigned, and then use that as the first sorting dimension.  So all hosts that were assigned together initially will then be ordered as a subgroup by number of slots.

What do you think?

EnricoMi: You are right, for the elastic assignment, this code will have to change completely anyway.

Would it suffice to pick the oldest host as the first and sort all other as above? I guess we are over-optimizing the heterogeneous-host-sizes-scenario here.

So `HostInfo` should get that timestamp field (e.g. named `discovered`) and then we sort on `(host.discovered, host.slots)`.

Should that assignment after some hosts dropped and new hosts were discovered minimise the number of hosts that change their rank (given size did not change)? E.g. on host dropped and another one was discoverd as a replacement. Then this new host has the highest timestamp but should it replace the dropped host in the host assignment rank order? In that case the assignment needs the previous `SlotInfo`s.

EnricoMi:
A requirement is to keep the order of hosts as they become available, so this is not a working improvement.

PR #1805, opened 2020-03-19
Add LSF and jsrun support to horovodrun

nvcastet:
Example to run on a LSF cluster (e.g. Summit):
horovodrun python train.py

Also, perform cpu/mem process binding to get the best performance.

Contributors:
@bethune-bryant
@nvcastet

Signed-off-by: Nicolas V Castet <nvcastet@us.ibm.com>

### DIFF
--- a/horovod/run/run.py
+++ b/horovod/run/run.py
@@ -22,8 +22,6 @@ import sys
 import re
 import textwrap
 
-from socket import AF_INET
-from psutil import net_if_addrs
 try:
     from shlex import quote
 except ImportError:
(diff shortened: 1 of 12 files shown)


horovod/run/run.py
```diff
@@ -779,47 +793,49 @@ def _run(args):
         if settings.verbose >= 2:
             print('SSH was successful into all the remote hosts.')
 
-    if len(remote_host_names) > 0:
-        if settings.verbose >= 2:
-            print('Testing interfaces on all the hosts.')
+    common_intfs = None
+    # Skipping interface discovery for LSF cluster as it slows down considerably the job start
+    if not lsf.LSFUtils.using_lsf():
```
tgaddair: This block is getting pretty heavily nested.  Maybe we can pull this out into a utility function `_get_common_interfaces`, where the first couple lines are a precondition check:

```
if lsf.LSFUtils.using_lsf():
    return None
```

What do you think?

nvcastet: Yes I agree, code will be cleaner. I will do that.

PR #1808, opened 2020-03-20
Added support for inclusive NIC names

### LINKED ISSUE #1724: Horovod.run.run on multiple hosts kills jupyter notebook session
**Environment:**
1. Framework: (**TensorFlow**, Keras, PyTorch, MXNet)
2. Framework version: 2.0
3. Horovod version: 0.19.0
4. MPI version: 
```
mpirun --version
mpirun (Open MPI) 4.0.2
```
5. CUDA version: 10.2
6. NCCL version: 2.5.6
7. Python version: 2.7
8. OS and version: Centos 7,7
9. GCC version:
```
$ gcc --version
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39)
```
**Checklist:**
1. Did you search issues to find if somebody asked this question before? No
2. If your question is about hang, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/running.rst)? Don't think it's about hang
3. If your question is about docker, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/docker.rst)? Not about docker
4. Did you check if you question is answered in the [troubleshooting guide](https://github.com/horovod/horovod/blob/master/docs/troubleshooting.rst)? Couldn't find it

**Bug report:**
Please describe errorneous behavior you're observing and steps to reproduce it.

I have horovod.run.run working on 1 server with four GPUs finally after the issues I had in #1691. However, now when I try to run horovod.run.run on multiple servers with multiple GPUs, my Jupyter notebook runs for a few seconds and then crashes. The error message is:
```
The kernel appears to have died. It will restart automatically.
```
Here's my first cell:
```
import horovod
import horovod.run
import tensorflow as tf

def test_main():
    # Horovod: initialize Horovod.
    hvd.init()
    # Horovod: pin GPU to be used to process local rank (one GPU per process)
    gpus = tf.config.experimental.list_physical_devices('GPU')
    for gpu in gpus:
        tf.config.experimental.set_memory_growth(gpu, True)
    if gpus:
        tf.config.experimental.set_visible_devices(gpus[hvd.local_rank()], 'GPU')
    print(f"Hello world from {hvd.local_rank()}")
```
Then I make my horovod.run.run call in my second cell:
```
horovod.run.run(test_main, np=2, hosts="host1:4, host2:4", verbose=True, disable_cache=True)
```
I do get a little output before it crashes:

> Filtering local host names.
> Remote host found: host1
> Checking ssh on all remote hosts.
> SSH was successful into all the remote hosts.
> Testing interfaces on all the hosts.
> Launched horovodrun server.
> Attempted to launch horovod task servers.
> Waiting for the hosts to acknowledge.
> Notified all the hosts that the registration is complete.
> Waiting for hosts to perform host-to-host interface checking.

After this it immediately crashes. I figured there might be a brief wait, but there is not. Looking at the code, the crash appears to be occurring in [run.py)[https://github.com/horovod/horovod/blob/master/horovod/run/run.py#L252].

I can run this script with just mpi and it works on host1:4 and host2:4 just fine, so I don't think it's that. If y'all have any ideas, let me know. 

I'd like to scale horovod.run.run up to a _few_ more than just these two nodes if possible. :)

abditag2:
Fixes #1724.


horovod/run/run.py
```diff
@@ -609,6 +612,12 @@ def parse_args():
                                   help='Run Horovod using the MPI controller. This will '
                                        'be the default if Horovod was built with MPI support.')
 
+    parser.add_argument('--btl_tcp_if_include', action='store', dest='btl_tcp_if_include',
```
tgaddair: There is already a parameter `--network-interface` that does something very similar.  So I would suggest reusing it.  However, there are two things we'd need to change to make it work the way this PR does:

1. `--network-interface` should accept a comma-separate list of NICs
2. Existing behavior still checks that `--network-interface` is common among all hosts, so we should skip this step (as you do here).

abditag2: Thanks for pointing it out.

horovod/run/run.py
```diff
@@ -427,8 +430,10 @@ def parse_args():
                              'HOROVOD_START_TIMEOUT can also be used to '
                              'specify the initialization timeout.')
 
-    parser.add_argument('--network-interface', action='store', dest='nic',
-                        help='Specify the network interface used for communication.')
+    parser.add_argument('--network-interfaces', action='store', dest='nics',
```
tgaddair: Let's avoid changing the name of this parameter.  Some users are already using it, and we don't want to suddenly break them.  I think it's okay if it's called "nic" and it accepts a comma-separated list.

tgaddair:
@abditag2 please rebase off master, as there was a recent change to the `run.py` file.

horovod/run/run.py
abditag2: changed.

horovod/run/run.py
```diff
@@ -254,8 +254,10 @@ def parse_args():
                              'HOROVOD_START_TIMEOUT can also be used to '
                              'specify the initialization timeout.')
 
-    parser.add_argument('--network-interface', action='store', dest='nic',
-                        help='Specify the network interface used for communication.')
+    parser.add_argument('--network-interface', action='store', dest='nics',
+                        help='Network interfaces that can be used for communication separated by '
+                             'comma. If not specified, Horovod will find the common NICs among all '
+                             'the workers and use it; example, --nics eth0,eth1.')
```
tgaddair: Example should read: --network-interface "eth0,eth1"

PR #1818, opened 2020-03-24
Check jsrun is available before launching via it.

EnricoMi:


horovod/run/run.py
```diff
@@ -667,6 +667,9 @@ def _launch_job(args, remote_host_names, settings, common_intfs, command):
         if not mpi_built(verbose=(settings.verbose >= 2)):
             raise ValueError('MPI support has not been built.  If this is not expected, ensure MPI is installed '
                              'and reinstall Horovod with HOROVOD_WITH_MPI=1 to debug the build error.')
+        if not lsf.LSFUtils.using_lsf() or not is_jsrun_installed():
```
nvcastet: In this code path, the user picks explicitly `jsrun` with the `jsrun` flag. We check inside `js_run(...)` if both `MPI` and `jsrun` are installed and print the appropriate error message. It follows the same pattern as `mpi_run(...)` where we do the pre-condition checks for executables at the top of the `run` function. Does that make sense?

The code path at line 676 is when `horovodrun` picks automatically the launcher for the user, so we need an `if statement` to pick `jsrun` over `mpirun`.

EnricoMi: I see, the `_launch_job` method checks `is_built` semantics, kind of is-meant-to-be-used and the actual run method checks availability in terms of installation. I am happy to leave it as it is then.

When looking into `_launch_job`, it is strange that the else branch checks for `lsf.LSFUtils.using_lsf()` but the `elif args.use_jsrun:` branch and `js_run` do not. What is the exact meaning of `using_lsf()` and why is it not check in all js paths?

nvcastet: This is a good question. LSF is a popular workload management platform/job scheduler in HPC clusters. Some LSF installs like in the Summit supercomputer also have a tool called `jsrun` that launches the processes more efficiently but still uses MPI.

So you can have LSF environments without jsrun (in this case mpirun is used) but you cannot have an environment having `jsrun` that is not LSF.

We could technically add the check for LSF in the `use_jsrun`/`js_run` path.

EnricoMi: Then lets only check for `lsf.LSFUtils.using_lsf()` in `_launch_job`, this aligns best with `use_gloo` and `use_mpi` and the `else:` branches in that method, the actual `*_run` methods then check for installation of used commands. I have adjusted this PR accordingly.

horovod/run/run.py
```diff
@@ -676,6 +676,9 @@ def _launch_job(args, remote_host_names, settings, nics, command):
         if not mpi_built(verbose=(settings.verbose >= 2)):
             raise ValueError('MPI support has not been built.  If this is not expected, ensure MPI is installed '
                              'and reinstall Horovod with HOROVOD_WITH_MPI=1 to debug the build error.')
+        if not lsf.LSFUtils.using_lsf():
+            raise ValueError('MPI support with jsrun is not available.  Please, make sure you are running '
```
nvcastet: Thanks!
I would say:
`Horovod did not detect a LSF job. 'jsrun' launcher can only be used in that environment. Please, pick a different launcher for other environments.`

EnricoMi: That sounds better! Changed it to:
`Horovod did not detect an LSF job.  The jsrun launcher can only be used in that environment. Please, pick a different launcher for other environments.`

PR #1820, opened 2020-03-25
Remove method injection from testing.

EnricoMi:
Replaces method injection with better `mock.patch` approach in unit testing.

Removes `run_func` from `mpi_run` and `js_run`.
Removes `exec_command` from `gloo_run`.
Removes `execute` from `_get_mpi_implementation`.

Conflicts with #1807, should be merged later.


EnricoMi:
Sorry for the bad code design that I have introduced. This fixes it.

PR #1830, opened 2020-03-27
Add Join op (with support for AllReduce and TensorFlow )

Richie-yan:
This PR is based on the problem #832 and implemented hvd.join() on the Tensorflow level.
The author @kit1980 has implemented join logic on the horovod framework and support for Pytorch. So this PR is based on the Join function provided by author kit1980 , adding join op support for Tensorflow


kit1980:
Hi,
May I ask why do you keep closing/opening new PRs with the same content?

Richie-yan:
Sorry, I have troubled you due to improper operation. The content is somewhat changed

Richie-yan:
> Hi,
> May I ask why do you keep closing/opening new PRs with the same content?

Hi,This is the first time I have no experience with PR. Can you help me see why CI reports an error? I don't see how the error is related to the code I mentioned? Or that the code I mentioned should not cause such an error

horovod/tensorflow/mpi_ops.cc
```diff
@@ -248,11 +278,87 @@ TFOpContext::AllocateOutput(common::TensorShape shape,
   return ConvertStatus(status);
 }
 
+int GetDeviceID(OpKernelContext* context);
+
+template <class T>
+void Allocate(int64_t num_elements, int device_, ::tensorflow::Tensor& zero_tensor) {
```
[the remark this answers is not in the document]
Richie-yan: I started to use zeros-like, but I got an error when compiling horovod. The zeros-like op of tensorflow should not be open to the public. Do you have any better suggestions? Or is there something wrong with my compilation?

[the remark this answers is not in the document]
tgaddair: I would suggest using [OpKernelContext::allocate_temp](https://github.com/tensorflow/tensorflow/blob/bd0d204700df1a1a245b0593a11efd8ede139311/tensorflow/core/framework/op_kernel.h#L254) instead.  We already use `allocate_persistent` in some places, so the code should be similar.  The only extra step is to make sure all values are properly set to 0.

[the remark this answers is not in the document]
Richie-yan: Ok, I will change to OpKernelContext :: allocate_persistent based on your suggestion.
Is my current method okay for setting all values to 0?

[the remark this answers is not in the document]
tgaddair: I would look at how TensorFlow does it internally and mirror their approach.  For example, here's how they assign values to zero in the ZeroLike op: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/util/tensor_ops_util.h#L45

@romerojosh, do you have any suggestions on how best to approach this?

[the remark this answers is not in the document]
romerojosh: To initialize to zero, you can use `memset` for CPU buffers and `cudaMemset` for the GPU buffers directly.

iamyangchen:
I'm wondering if there's any update on this? I'm encountering the same issue and this PR will greatly help.

Richie-yan:
Hi，iamyangchen
I ’ve been busy recently, ignoring pr, I will modify the code as soon as possible, and then review, and strive to update in May

Richie-yan:
@tgaddair 
I modified the code and hope to review

horovod/tensorflow/mpi_ops.py
```diff
@@ -199,3 +199,7 @@ def _broadcast_grad(op, grad):
     if rank() != root_rank:
         return grad_reduced * 0
     return grad_reduced
+
+#add join for tf
```
romerojosh, line 203: Nit: Please remove this comment.

horovod/tensorflow/mpi_ops.cc
```diff
@@ -480,5 +544,41 @@ Output
                `tensor` on root rank.
 )doc");
 
+//add join op for  tf
```
romerojosh, line 547: Nit: Please remove this comment as well.

horovod/tensorflow/mpi_ops.cc
```diff
@@ -248,11 +276,47 @@ TFOpContext::AllocateOutput(common::TensorShape shape,
...
+  auto hvd_context = std::make_shared<TFOpContext>(context_);
+  if (device_ != CPU_DEVICE_ID){
+    tf_attribute.set_on_host(false);
+  } else{
+    tf_attribute.set_on_host(true);
+  }
+
+  Status status = context_->allocate_persistent(tf_data_type, ::tensorflow::TensorShape({num_elements}), zero_tensor.get(), &unused, tf_attribute);
+   
+  if (device_ != CPU_DEVICE_ID){
```
romerojosh, line 298: Please add space after if conditions and opening bracket in this file like this: `if (device_ != CPU_DEVICE_ID) {`

Same applies to else. Should be `} else {`.

horovod/tensorflow/mpi_ops.cc
```diff
@@ -248,11 +276,47 @@ TFOpContext::AllocateOutput(common::TensorShape shape,
...
+#if HAVE_GPU
+    cudaMemset((void*)zero_tensor->AccessTensor(hvd_context->GetKernelContext())->tensor_data().data(), 0,
```
romerojosh, line 300: Actually, just found something we should fix here. This should be a `cudaMemsetAsync` and pass in the stream from `device_context->stream()` or else the `cudaMemset` will launched in the default CUDA stream and cause undesired device-wide synchronization behavior. I think using this particular stream should ensure the memset occurs after the TF tensor is allocated. The call needs to be something like:
```
auto device_context = context_->op_device_context();
auto stream = (device_context != nullptr) ? AsGpuStreamValue(device_context->stream()) : 0;
cudaMemsetAsync((void*)zero_tensor->AccessTensor(hvd_context->GetKernelContext())->tensor_data().data(), 0, 
                zero_tensor->AccessTensor(hvd_context->GetKernelContext())->tensor_data().size(), stream);
```
@tgaddair can you also comment on this? I am not as well versed in this part of the code, but this seems to be the behavior we want. I'm also not quite clear on when `context_->op_device_context()` could be `nullptr` if running on GPU.

tgaddair: Nice catch!  I don't think it should ever be `nullptr`, as we get the device from the [GetDeviceID](https://github.com/horovod/horovod/blob/master/horovod/tensorflow/mpi_ops.cc#L264), which always returns a valid GPU ID or the CPU ID (-1).  If we're on CPU, then we don't hit this code path, so we shouldn't need to check for that.

Maybe we can raise an error if the `device_context` is not valid?

romerojosh: We could raise an error, but we also have a similar check for `nullptr` here in the existing code and would need to do the same thing: https://github.com/horovod/horovod/pull/1830/files#diff-ab99f007f05dd3de4771697e8192483aR315

Since `device_context` shouldn't ever be `nullptr` in any user case running on GPU (I assume this would only show up during code development), I don't think it needs a runtime error.

Richie-yan: Okay, do I have to be consistent with similar checks in existing code? Check whether device_context is nullptr

tgaddair: Good question.  We should check for nullptr to be consistent.

Richie-yan: @romerojosh 
The AsGpuStreamValue function was introduced in tensorflow version 1.12.1: https://github.com/tensorflow/tensorflow/blob/v1.12.1/tensorflow/stream_executor/gpu/gpu_stream.h
so for < tensorflow version 1.12.1, it seems that is AsCUDAStreamValue function:
https://github.com/tensorflow/tensorflow/blob/v1.12.0/tensorflow/stream_executor/cuda/cuda_stream.h
What are your suggestions for this situation?

tgaddair: Good catch.  Let's do it the same way we do it for PyTorch.  We can define a macro called `TENSORFLOW_VERSION` which we use like [this](https://github.com/horovod/horovod/blob/master/horovod/torch/ready_event.cc#L17).  This macro can be defined in setup.py like we do [here](https://github.com/horovod/horovod/blob/master/setup.py#L1272).  There's a function called [parse_version](https://github.com/horovod/horovod/blob/master/setup.py#L1007) you can use to generate the numerical representation of the version string.

Richie-yan: Ok, the corresponding result is: for different tensorflow versions, call their corresponding functions

tgaddair: Yes, and we can do this conditional logic at compile time via macros.

Richie-yan:
Hi, @tgaddair 
The environment reports an error: horovod / tensorflow / mpi_ops.cc: 31: 10: fatal error: tensorflow / stream_executor / gpu / gpu_stream.h: No such file or directory
I am a bit confused why this file cannot be found?

Richie-yan:
@romerojosh 
Can you help me see this problem? It seems that tensorflow does not expose the tensorflow / stream_executor / gpu / gpu_stream.h interface

romerojosh:
@Richie-yan, yes, it looks like the `gpu_stream.h` header is not included in the TF installation. What we can do, though this is a bit of a hack, is just forward declare the function that we need at the top of the file instead of trying to include the missing header like this:
```
// Forward declaration of AsGpuStreamValue
#if TENSORFLOW_VERSION >= 1012001000
namespace stream_executor {
namespace gpu {
cudaStream_t AsGpuStreamValue(Stream* stream);
} // namespace stream_executor
} // namespace gpu
#else
namespace stream_executor {
namespace cuda {
cudaStream_t AsCUDAStreamValue(Stream* stream);
} // namespace stream_executor
} // namespace cuda
#endif
```
@tgaddair, can you think of a better option? I'm not exactly sure why this header isn't added to the installation.

Richie-yan:
Hi, @romerojosh 
Which tf version corresponds to the tfhead in the test-gpu-openmpi-py3_6-tfhead-kerashead-torchhead-mxnethead-XXX test environment?
 In this test environment, an error is reported：
tensorflow.python.framework.errors_impl.NotFoundError: /usr/local/lib/python3.6/dist-packages/horovod/tensorflow/mpi_lib.cpython-36m-x86_64-linux-gnu.so: undefined symbol: _ZN15stream_executor4cuda17AsCUDAStreamValueEPNS_6StreamE
My local test is tf1.10 version, which can be compiled and tested

romerojosh:
tfhead is a nightly version of TensorFlow, which should have the newer `AsGPUStreamValue` available, not `AsCUDAStreamValue`, but the error is showing that for the build using tfhead, the old function prototype is being declared and subsequently not found at runtime. 
I think the issue is in the version checking code here, where the version is conditionally set to `1004000000`:
https://github.com/horovod/horovod/blob/1cbcea5a344947cc366133a604a4464e08420850/setup.py#L80-L85
It should work if you remove this conditional setting of `version`.

Richie-yan:
Hi, @romerojosh 
Please review my current pr when you are convenient

romerojosh:
LGTM! @tgaddair, go ahead and merge if you approve of the current solution forward declaring the required TF helper functions.

iamyangchen:
@Richie-yan Thank you for the change!

I'm wondering if you could provide a minimal example to show how this would work? I'm assuming this should have the same usage as `horovod.torch.join` which says "All ranks that did not call join() continue to process allreduce operations. This function blocks Python thread until all ranks join."

I have tried something like this
```
...
        try:
            while not self.coord.should_stop():
                self.session.run(...)
        except tf.errors.OutOfRangeError:
            hvd.join()
            self.coord.request_stop()
```
and it seems `hvd.join()` returns immediately and my script would crash. Feel like I must have missed something.

Richie-yan:
Hi, @iamyangchen 
You can refer to the following usage methods:
```
#First create a join_hook
class HorovodJoinHook(session_run_hook.SessionRunHook):
     def begin(self):
         self._join_op = hvd.join()
     def end(self, session):
         session.run(self._join_op)

def join_hook():
     return HorovodJoinHook()

#Then pass the hook to the estimator interface:
hvd_join_hook = join_hook()
mnist_classifier.train(
         input_fn=train_input_fn,
         steps=20000 // hvd.size(),
         hooks=[logging_hook, bcast_hook, hvd_join_hook])
```

kit1980:
@Richie-yan I don't understand your example.

It looks like every worker does the same number of steps (`steps=20000 // hvd.size()`), why do you even need `join` in this case?

Richie-yan:
@kit1980 
The example provided above focuses on how to use the join function. In tensorflow's estimator interface, although the step of each rank is set to be the same, if there is currently a data imbalance situation, when one of the ranks does not have data At the time, the training will be launched, at this time the join hook will be called
In other words, in two cases, the estimator interface will exit, the data has been consumed or the specified step has been reached

iamyangchen:
@Richie-yan Thanks for the suggestion. I tried the session hook approach, and it worked. 

However, I think it doesn't work with horovod's autotune option. If I launch the job with `horovodrun --autotune ...`, then the job would fail with this error message:

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  _Map_base::at
*** Process received signal ***
Signal: Aborted (6)
Signal code:  (-6)
[ 0] [1,3]<stderr>:/lib/x86_64-linux-gnu/libc.so.6(+0x3ef20)[0x7fe835e1df20]
[ 1] [1,3]<stderr>:/lib/x86_64-linux-gnu/libc.so.6(gsignal+0xc7)[0x7fe835e1de97]
[ 2] [1,3]<stderr>:/lib/x86_64-linux-gnu/libc.so.6(abort+0x141)[0x7fe835e1f801]
[ 3] [1,3]<stderr>:/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x8c957)[0x7fe834ce6957]
[ 4] [1,3]<stderr>:/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x92ab6)[0x7fe834cecab6]
[ 5] [1,3]<stderr>:/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x92af1)[0x7fe834cecaf1]
[ 6] [1,3]<stderr>:/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x92d24)[0x7fe834cecd24]
[ 7] [1,3]<stderr>:/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x8e82f)[0x7fe834ce882f]
[ 8] [1,3]<stderr>:/usr/local/lib/python2.7/dist-packages/horovod/tensorflow/mpi_lib.so(_ZN7horovod6common11TensorQueue25GetTensorDataForAutotunerERKNS0_12ResponseListERSt6vectorISsSaISsEE+0x26c)[0x7fe74689504c]
[ 9] [1,3]<stderr>:/usr/local/lib/python2.7/dist-packages/horovod/tensorflow/mpi_lib.so(+0x60990)[0x7fe746873990]
[10] [1,3]<stderr>:/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0xbd66f)[0x7fe834d1766f]
[11] /lib/x86_64-linux-gnu/libpthread.so.0(+0x76db)[0x7fe835bc76db]
[12] [1,3]<stderr>:/lib/x86_64-linux-gnu/libc.so.6(clone+0x3f)[0x7fe835f0088f]
*** End of error message ***
```

EvenOldridge:
Is this functionality available / compatible with keras?

PR #1839, opened 2020-04-01
Make train fn see Spark worker's environment

EnricoMi:
MPI runs the training function via `orted` in the directory where `mpirun` is executed on the driver. Each Spark worker has a unique working directory. That working directory contains important files provided by Spark to the worker and python code.

This refactors how `SparkTaskService` passes environment variables to `BasicTaskService`. Used to be environment variable names where `BasicTaskService` picks the value from the worker's `os.environ`, where `SparkTaskService` now provides a dictionary with the override values. This dict is initialized with `os.environ` so `SparkTaskService` uses the entire worker's env. This env is overwriten by the command request's env. `HorovodRunTaskService` uses request's env only.

This further provides `HOROVOD_SPARK_WORK_DIR` with the working directory of Spark workers to `mpirun_exec_fn` which then changes to that directory. `HOROVOD_SPARK_PYTHONPATH` can be used to add some paths to worker's `PYTHONPATH`. Providing a `PYTHONPATH` via command request env would overwrite the entire `PYTHONPATH` on the worker side.

`_HOROVOD_SECRET_KEY` is not sent through `mpirun -x` anymore because it is redundant. The `SparkTaskService` knows this key already and provides it as an environment var to the command it executes (`orted`).

This change might break peoples **Spark** training scripts when they use python code from the current working directory where they get started. If this path is available on the workers then changing the working directory on the worker will make that code inaccessible. Users should provide that path to the trainings script through `HOROVOD_SPARK_PYTHONPATH`. This will then be added to worker's `PYTHONPATH` and `sys.path`.

When workers cannot access the directory from where the training script gets exected, the driver can now submit python files to the workers via `spark-submit --py-files` or `session.sparkContext.addPyFile()`.

In contrast to classic MPI, Spark over MPI does not want to push driver's environment variables to the workers as their environment may differ. Spark workers provide a fully working environment already. Thus giving no env to `horovod.spark.mpi_run` does not fall back to `os.environ` as `horovod.run.mpi_run` does.

### DIFF
--- a/test/test_spark.py
+++ b/test/test_spark.py
@@ -43,7 +43,7 @@ import horovod.spark
 import horovod.torch as hvd
 
 from horovod.common.util import gloo_built, mpi_built
-from horovod.run.common.util import secret
+from horovod.run.common.util import codec, secret
 from horovod.run.mpi_run import is_open_mpi
 from horovod.spark.common import constants, util
 from horovod.spark.common.store import HDFSStore
@@ -265,7 +265,7 @@ class SparkTests(unittest.TestCase):
         def mpi_impl_flags(tcp):
             return ["--mock-mpi-impl-flags"], ["--mock-mpi-binding-args"]
 
-        def gloo_exec_command_fn(driver_addresses, settings, env):
+        def gloo_exec_command_fn(driver_addresses, key, settings, env):
             def _exec_command(command, alloc_info, event):
                 return 1, 1.0
             return _exec_command
@@ -1092,6 +1092,7 @@ class SparkTests(unittest.TestCase):
                         'HADOOP_TOKEN_FILE_LOCATION=path',
                         'HOROVOD_SPARK_WORK_DIR={cwd}'.format(cwd=os.getcwd()),
                         'PYTHONPATH=pypath',
+                        '{}={}'.format(secret.HOROVOD_SECRET_KEY, codec.dumps_base64(key)),
                         'other=values',
                         'test=value'
                     ]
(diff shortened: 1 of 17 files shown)


test/test_spark.py
```diff
@@ -68,12 +70,24 @@ def __init__(self, *args, **kwargs):
...
         if int(os.getenv('OMPI_COMM_WORLD_RANK', 0)) != 0 or int(os.getenv('HOROVOD_RANK', 0)) != 0:
-            # Running in MPI or Gloo with rank > 0, ignore.
-            # Purposefully skip these silently
-            return
+            self.skipTest("Not testing for rank > 0, tests start their own MPI / Gloo setup")
 
-        super(SparkTests, self).run(result)
+        # ... and remove all HOROVOD_* environment variables.
+        # Pretend horovodrun is not there.
+        env = os.environ.copy()
```
tgaddair, line 83: We should probably just change `test_spark.py` to run without mpirun/horovodrun, similar to what we do for `test_spark_torch.py`, etc.

EnricoMi: Good idea, thanks, have done this: https://github.com/horovod/horovod/pull/1839/commits/d956d623d14c10e3e0f3f0d9ac0091df393f9ea7

I have also added a test that compares `.buildkite/gen_pipeline.sh` with an expected yaml file, so we can track changes in the generated pipeline in git and PRs: https://github.com/horovod/horovod/pull/1839/commits/fccc745b442a0f93c1ecbed30b8c53b5d42a3f04

PR #1845, opened 2020-04-04
fix memory leak

gaurav1086:


horovod/common/ops/ccl_operations.cc
```diff
@@ -187,6 +187,8 @@ Status CCLAllgather::Execute(std::vector<TensorTableEntry>& entries, const Respo
   timeline.ActivityStartAll(entries, ALLOCATE_OUTPUT);
   Status status = AllocateOutput(entries, response, entry_component_sizes, recvcounts);
   if (!status.ok()) {
+    delete[] recvcounts;
+    delete[] displcmnts;
     return status;
```
EnricoMi, line 192: shouldn't these also be freed?

```
delete[] entry_component_sizes;
delete[] entry_component_offsets;
```

EnricoMi:
@tgaddair @nvcastet these issues should be checked by some automatic memory leak detection tool, at some point even as a guthub check. Which tools do prefer?

Alternatively use smart pointers. Is than an option?

horovod/common/ops/ccl_operations.cc
gaurav1086: @EnricoMi yes.

nvcastet:
@tgaddair @EnricoMi I think because we can compile with c++11, I agree that we should push for smart pointers to own the memory allocations, it is also what is recommended by the C++ standard committee folks.
I can work on a PR to go through the code to clean that up, if nobody has starting that yet.

nvcastet:
`Valgrind` is usually the tool for memory leaks with C++/C.

gaurav1086:
> @tgaddair @nvcastet these issues should be checked by some automatic memory leak detection tool, at some point even as a guthub check. Which tools do prefer?
> 
> Alternatively use smart pointers. Is than an option?

I tried changing them to smart pointers, but it entailed changes in a lot of function (signatures), hence I chose to discontinue.

EnricoMi:
> I can work on a PR to go through the code base to clean that up, if nobody has starting that yet.

Happy with that. I'll merge this into master in the meantime.

PR #1846, opened 2020-04-04
Add [tensorflow-cpu] extra

mengxr:
to simplify CPU-only build:

requirements.txt:

~~~
tensorflow-cpu==2.1.*
horovod[tensorflow-cpu]==0.19.*
~~~

tensorflow-cpu package is around 100MB while tensorflow or tensorflow-gpu packages are around 400MB

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1572,6 +1572,7 @@ test_require_list = ['mock', 'pytest', 'pytest-forked']
 
 # framework dependencies
 tensorflow_require_list = ['tensorflow']
+tensorflow_cpu_require_list = ['tensorflow-cpu']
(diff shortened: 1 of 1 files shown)


EnricoMi:
@tgaddair should we remove `[tensorflow]` alltogether and use `[tensorflow-cpu]` and `[tensorflow-gpu]` only? Should we replace `tensorflow` in `all-frameworks` with `tensorflow-cpu`?

tgaddair:
@EnricoMi going forward TensorFlow is going to stick with having a single `tensorflow` package for both CPU and GPU support.  So I think it makes sense to keep all three, but use `tensorflow` for `all-frameworks`.

PR #1849, opened 2020-04-07
Initial bare-metal implementation of elastic mode for fault tolerance and auto-scaling

### LINKED ISSUE #1907: Elastic mode to enable fault tolerance and auto-scaling
[RFC](https://docs.google.com/document/d/15ZoHA5AeSI_boeyIBapg9WPXKrYXMRvPytPzQWTCTn4/edit?usp=sharing)

This feature covers the bare-metal implementation of a fault tolerant and auto-scaling capability / API for Horovod built on top of the Gloo controller.

tgaddair:
Resolves #1907. 

[RFC](https://docs.google.com/document/d/15ZoHA5AeSI_boeyIBapg9WPXKrYXMRvPytPzQWTCTn4/edit?usp=sharing).


horovod/run/elastic/registration.py
```diff
@@ -0,0 +1,138 @@
...
+    def _record_state(self, host, slot, state):
+        if self._driver.finished():
+            logging.info('driver finished, ignoring registration: {}[{}] = {}'.format(host, slot, state))
+            return self._rendezvous_id
+
+        key = (host, slot)
+        with self._lock:
+            if key in self._states:
+                logging.info('key exists, reset barrier: {}[{}] = {}'.format(host, slot, state))
+                self._barrier.reset()
```
EnricoMi, line 79: Strange situation, why is this not an exception and why is barrier reset sufficient?

horovod/run/elastic/worker.py
```diff
@@ -0,0 +1,116 @@
...
+             nic=None, hostname=None, local_rank=None):
+        with self._lock:
+            if self._service:
+                return
+
+            rendezvous_addr = rendezvous_addr or os.environ.get(HOROVOD_GLOO_RENDEZVOUS_ADDR)
+            if not rendezvous_addr:
+                return
+
+            rendezvous_port = rendezvous_port if rendezvous_port is not None else \
```
EnricoMi, line 59: `rendezvous_port or int(os.environ.get(HOROVOD_GLOO_RENDEZVOUS_PORT))`

horovod/run/elastic/worker.py
```diff
@@ -0,0 +1,116 @@
...
+                int(os.environ.get(HOROVOD_GLOO_RENDEZVOUS_PORT))
+            nic = nic or os.environ.get(HOROVOD_GLOO_IFACE)
+            hostname = hostname or os.environ.get(HOROVOD_HOSTNAME)
+            local_rank = local_rank if local_rank is not None else \
```
EnricoMi, line 63: `local_rank or int(os.environ.get(HOROVOD_LOCAL_RANK))`

horovod/run/common/util/safe_shell_exec.py
```diff
@@ -84,7 +84,7 @@ def forward_stream(src_fd, dst_stream, prefix, index):
                     line_buffer = ''
 
 
-def execute(command, env=None, stdout=None, stderr=None, index=None, event=None):
+def execute(command, env=None, stdout=None, stderr=None, index=None, events=None):
```
EnricoMi, line 87: can we rename

    def set_sigterm_received(signum, frame):

to

    def set_event_on_signal(signum, frame):

in this file as we did in `horovod/run/gloo_run.py`?

horovod/common/exceptions.py
```diff
@@ -0,0 +1,25 @@
...
+# ==============================================================================
+
+from __future__ import absolute_import
+
+
+class HorovodInternalError(RuntimeError):
+    pass
+
+
+class HostsUpdatedException(RuntimeError):
```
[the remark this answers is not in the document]
tgaddair: Done.

docs/elastic.rst
```diff
@@ -0,0 +1,144 @@
...
+If a worker fails during training, the other workers will catch a ``HorovodInternalError`` that is raised when
+attempting to perform the next collective operation.  This will trigger a rollback to the beginning of the
+synchronized function.
+
+Rollback process is as follows:
+
+1. Catch exception within the hvd.elastic.run decorator.
+2. Restore last committed state (to clear any half-updated parameters).
+3. Reinitialize Horovod context performing a new round of rendezvous.
+4. Synchronize state among the workers by broadcasting from the new worker-0.
```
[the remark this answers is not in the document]
tgaddair: One of the assumptions we make is that the driver may exist on different hardware from the workers.  As such, it could have a relatively slow NIC, no GPU, etc.  So we want to avoid sending data to the driver if possible.

docs/elastic.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Rollback process is similar to worker failure, but the last committed state will not be restored (as there is no
+partial state that could have been lost).  However, broadcast is still necessary for new workers to obtain the most
+updated state.
+
+During rendezvous, older workers will take priority in being assigned worker-0 status to ensure that the state that
+is broadcast is up to date.
+
+
+Running with horovodrun
```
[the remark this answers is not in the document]
tgaddair: Added examples for TensorFlow v1 / v2 and PyTorch.

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,233 @@
...
+        self._wait_hosts_cond.acquire()
+        try:
+            while not self._has_available_slots(self._discovered_hosts.count_available_slots(), min_np, max_np):
+                self._wait_hosts_cond.wait(tmout.remaining())
+                tmout.check_time_out_for('minimum number of hosts to become available')
+        finally:
+            self._wait_hosts_cond.release()
+
+    def _has_available_slots(self, slots, min_np, max_np):
+        return slots >= min_np and (max_np is None or slots <= max_np)
```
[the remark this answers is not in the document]
tgaddair: Good point.  Not sure why I coded it up that way.  Fixed!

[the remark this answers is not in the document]
tgaddair: Ah I see, this was for a test.  I'll make a fix to the test to better encapsulate this logic.

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,233 @@
...
+        timestamp = _epoch_time_s()
+        for (host, slot), client in self._worker_clients.items():
+            try:
+                client.notify_hosts_updated(timestamp)
+            except:
+                if self._verbose >= 2:
+                    print('WARNING: failed to notify {}[{}] of host updates'
+                          .format(host, slot))
+
+    def _update_assigned_hosts(self):
```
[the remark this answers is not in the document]
tgaddair: Combined and added comments.

horovod/run/task_fn.py
```diff
@@ -31,7 +31,7 @@ def _task_fn(index, driver_addresses, settings):
         task.wait_for_initial_registration(settings.timeout)
         # Tasks ping each other in a circular fashion to determine interfaces
         # reachable within the cluster.
-        next_task_index = (index + 1) % settings.num_hosts
+        next_task_index = (index + 1) % num_hosts
```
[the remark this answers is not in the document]
tgaddair: Good catch!

test/test_elastic_driver.py
```diff
@@ -0,0 +1,360 @@
...
+        super(HostDiscoverySequence, self).__init__()
+
+    def find_available_hosts_and_slots(self):
+        try:
+            return self.iterator.next()
+        except:
+            return self.end
+
+
+class ElasticDriverTests(unittest.TestCase):
```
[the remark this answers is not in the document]
tgaddair: Self-consistent within a file is fine, I think.  `self.assert*` provides several nice convenience functions, and is generally considered the correct solution.  I'll change this file to use `self.assert*`.

horovod/run/runner.py
```diff
@@ -632,32 +650,74 @@ def _run(args):
...
+                                                run_func_mode=args.run_func is not None,
+                                                nics=args.nics)
+
+    if not gloo_built(verbose=(settings.verbose >= 2)):
+        raise ValueError('Gloo support is required to use elastic traiing, but has not been built.  Ensure CMake is '
+                         'installed and reinstall Horovod with HOROVOD_WITH_GLOO=1 to debug the build error.')
+
+    env = os.environ.copy()
+    config_parser.set_env_from_args(env, args)
+    gloo_run_elastic(settings, env, args.command)
```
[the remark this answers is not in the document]
tgaddair: Yes, the `run` method will need special consideration.  I agree it's better to address it in a separate PR (same for Spark).  I'll create an issue to track.

horovod/common/exceptions.py
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from __future__ import absolute_import
+
+
+class HorovodInternalError(RuntimeError):
```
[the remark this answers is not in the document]
tgaddair: Added comment.

horovod/run/elastic/worker.py
tgaddair: `local_rank == 0` is valid, but unfortunately `or` will still treat it as `False`.

horovod/run/common/util/safe_shell_exec.py
tgaddair: Done.

docs/elastic.rst
```diff
@@ -0,0 +1,299 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Elastic training jobs are started using the ``horovodrun`` command line tool. The major difference when launching
+elastic jobs is that hosts are not specified explicitly, but instead **discovered** at runtime.  The most general way
+to allow Horovod to discover available hosts is to provide a ``--host-discovery-script`` when launching the job:
+
+.. code-block:: bash
+
+    $ horovodrun -np 8 --host-disocvery-script discover_hosts.sh python train.py
```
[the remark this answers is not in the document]
tgaddair: Can you clarify in what situations this would be useful?

horovod/run/elastic/worker.py
EnricoMi: right, falsy values are so tricky in Python

horovod/run/common/util/safe_shell_exec.py
```diff
@@ -103,11 +103,11 @@ def execute(command, env=None, stdout=None, stderr=None, index=None, event=None)
 
         sigterm_received = threading.Event()
```
EnricoMi, line 104: this also needs renaming to `signal_received`

.buildkite/gen-pipeline.sh
```diff
@@ -94,23 +92,28 @@ run_mpi_pytest() {
...
-    exclude_keras_if_needed="| sed 's/test_keras.py//g' | sed 's/test_tensorflow_keras.py//g'"
+    exclude_keras="| sed 's/test_keras.py//g' | sed 's/test_tensorflow_keras.py//g'"
   else
-    exclude_keras_if_needed="| sed 's/[a-z_]*tensorflow2[a-z_.]*//g'"
+    exclude_keras="| sed 's/[a-z_]*tensorflow2[a-z_.]*//g'"
   fi
 
-  local exclude_interactiverun="| sed 's/test_interactiverun.py//g' | sed 's/test_spark_keras.py//g' | sed 's/test_spark_torch.py//g'"
+  local exclude_elastic=""
+  if [[ ${test} == *"py2_"* ]]; then
```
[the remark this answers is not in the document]
tgaddair: Yes, the `Barrier` feature used in the driver does not exist in Python 2.  So rather than finding a less elegant solution, I thought it would be better to just drop Python 2 support :).

.buildkite/gen-pipeline.sh
```diff
@@ -143,7 +146,7 @@ run_mpi_integration() {
   fi
 
   run_test "${test}" "${queue}" \
-    ":python: Test PyTorch MNIST (${test})" \
+    ":fire: Test PyTorch MNIST (${test})" \
```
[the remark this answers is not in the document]
tgaddair: Just makes the icon look (kinda) like the PyTorch logo.  Since we have a lot of tests in Buildkite, it's useful to distinguish them visually.

docs/elastic.rst
```diff
@@ -0,0 +1,299 @@
...
+                    state.commit()
+            state.batch = 0
+
+    def on_state_reset():
+        optimizer.lr.assign(lr * hvd.size())
+
+    state = hvd.elastic.TensorFlowKerasState(model, optimizer, batch=0, epoch=0)
+    state.register_reset_callbacks([on_state_reset])
+    train(state)
+
```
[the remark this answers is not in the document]
tgaddair: Added.

docs/elastic.rst
```diff
@@ -0,0 +1,299 @@
...
+5. Resume training by executing the underlying training function.
+
+During rendezvous, older workers will take priority in being assigned worker-0 status to ensure that the state that
+is broadcast is up to date.
+
+
+Elastic TensorFlow
+~~~~~~~~~~~~~~~~~~
+
+TensorFlow v1 Example:
```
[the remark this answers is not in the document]
tgaddair: Good idea!  There is a way to do it, though it only shows up in the read-the-docs, not in GitHub.  See: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/code_blocks.html#highlight-lines

Added.

examples/elastic/tensorflow_keras_mnist_elastic.py
```diff
@@ -0,0 +1,87 @@
...
+def on_state_reset():
+    tf.keras.backend.set_value(model.optimizer.lr, lr * hvd.size())
+
+
+state = hvd.elastic.KerasState(model, batch=100, epoch=0)
+state.register_reset_callbacks([on_state_reset])
+
+callbacks = [
+    # Horovod: elastic training callbacks to update and commit state.
+    hvd.elastic.CommitStateCallback(state),
```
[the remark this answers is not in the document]
tgaddair: I like the idea of making it explicit and trying not hide too much from the user, but I'm definitely open to the idea of adding more convenience wrappers.  I think this is something we'll need to brainstorm a bit along with the idea of wrapping more logic into the `DistributedOptimizer`.

docs/elastic.rst
```diff
@@ -0,0 +1,299 @@
...
+   through these callbacks.
+
+   Callbacks are called after Horovod has reinitialized, but before state is synchronized across the workers.
+
+The reset process following a ``HorovodInternalError`` (failure) or ``HostsUpdatedInterrupt`` (add/remove request) is as follows:
+
+1. Catch exception within the ``hvd.elastic.run`` decorator.
+2. Restore last committed state if ``HorovodInternalError`` was raised.
+3. Reinitialize Horovod context performing a new round of rendezvous.
+4. Synchronize state among the workers by broadcasting from the new worker-0.
```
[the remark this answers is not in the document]
tgaddair: Very good point.  It's not currently coded that way, but it should be.  I'll make an update.

test/test_elastic_driver.py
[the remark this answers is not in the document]
tgaddair: Actually, looking at the file, I think this is an example where plain assert is cleaner, so I'm going to stick with it here for now (though open to refactor in a separate PR).

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def _has_available_slots(self, slots, min_np, max_np):
+        return slots >= min_np and (max_np is None or slots <= max_np)
+
+    def _activate_hosts(self, min_np):
+        logging.info('wait for available hosts: {}'.format(min_np))
+        self.wait_for_available_hosts(min_np)
+        new_assigned_hosts = self._update_assigned_hosts()
+        self._worker_registry.reset(self.world_size())
+        for host in new_assigned_hosts:
```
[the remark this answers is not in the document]
tgaddair: Yes, this was an oversight.  I changed the code to properly launch processes for all slots that were not previously assigned.  Let me know if that looks good.

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,233 @@
...
+            logging.info('start worker processes: {}'.format(host))
+            self._start_worker_processes(host)
```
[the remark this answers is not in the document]
tgaddair: Fixed.

horovod/run/elastic/worker.py
tgaddair: Same as local_rank.

horovod/run/elastic/discovery.py
```diff
@@ -0,0 +1,139 @@
...
+    def __init__(self, discovery_script, slots):
+        self._discovery_script = discovery_script
+        self._default_slots = slots
+        super(HostDiscoveryScript, self).__init__()
+
+    def find_available_hosts_and_slots(self):
+        stdout = six.StringIO()
+        exit_code = safe_shell_exec.execute(self._discovery_script, stdout=stdout)
+        if exit_code != 0:
+            raise RuntimeError('Failed to execute discovery script: {}. Exit code: {}'
```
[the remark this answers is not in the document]
tgaddair: Updated docs.  Also changed behavior to raise an error only on the first call (due to misconfiguration).  Otherwise to retry.

horovod/run/elastic/registration.py
tgaddair: Good question. I've added comments to elaborate on this code.  Basically, the idea here is that after a worker records READY, it is still possible that it can fail somehow before the rendezvous barrier completes.  In that case, we want it to change its vote from READY to FAILURE.

It's not sufficient to just have the failing thread update the state to FAILURE and then return, it also needs to wait for the barrier to complete so it can decide if it needs to record its results.  This necessitates that we reset the barrier, as otherwise we will have double counted the failing worker in the barrier (once for its READY thread, and once for the FAILURE thread).

Hope that makes sense, let me know if it still needs clarification (and please take a look at updated comments).

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def has_rank_assignment(self, host, slot):
+        if self._discovered_hosts.is_blacklisted(host):
+            return False
+        return host in self._host_assignments and len(self._host_assignments[host]) > slot
+
+    def get_available_hosts(self):
+        return self._discovered_hosts.get_available_hosts()
+
+    def wait_for_available_hosts(self, min_np, max_np=None):
```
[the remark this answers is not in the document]
tgaddair: This is a good point.  I think the best way to handle this situation (for now) is to simply fail the job if none of the hosts from the previous host assignments is found in the current host assignments.  In the future, we can come up with a more robust solution to this problem.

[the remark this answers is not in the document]
tgaddair: Updated.

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,233 @@
...
+        for host in new_assigned_hosts:
+            logging.info('start worker processes: {}'.format(host))
+            self._start_worker_processes(host)
+
+    def _discover_hosts(self):
+        while not self._shutdown.is_set():
+            self._wait_hosts_cond.acquire()
+            try:
+                if self._discovered_hosts.update_available_hosts():
+                    self._notify_workers_host_changes()
```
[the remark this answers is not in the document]
tgaddair: Good point, I've added a precondition check to `_notify_workers_host_changes()` to early return if (1) no assigned hosts have been removed and (2) there are no added hosts that can be assigned (`max_np` slots already assigned or no available hosts were added).

We do already notify workers if an assigned host is removed from the available hosts, at which point it can be gracefully removed without needing to fail on allreduce.  The worker will kill itself after rendezvous when it received -1 for its rank.

You brought up a good optimization that can be made in a follow-up PR.  Currently, if a host is removed then added back before a reset event, then the reset will still occur (workers will raise an exception).  But ideally, the driver should send a notification to cancel the reset in such a case.

horovod/common/gloo/gloo_context.cc
```diff
@@ -140,6 +154,55 @@ void GlooContext::Initialize(const std::string& gloo_iface) {
     LOG(DEBUG) << "no rendezvous server provided, assuming single process execution";
   }
 
+  bool elastic = GetBoolEnvOrDefault(HOROVOD_ELASTIC, false);
```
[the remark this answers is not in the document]
tgaddair: True, but for non-elastic we still want to initialize via env vars, as the code paths for doing it dynamically are particular to the ElasticDriver.  We could refactor that, but it would be a big change, better left to an optional follow-up PR.

horovod/common/elastic.py
```diff
@@ -0,0 +1,161 @@
...
+        updated = False
+        while not self._host_messages.empty():
+            timestamp = self._host_messages.get()
+            if timestamp > self._last_updated_timestamp:
+                self._last_updated_timestamp = timestamp
+                updated = True
+
+        # In order to ensure all workers raise the exception at the same time, we need to sync
+        # the updated state across all the workers.
+        updated = self._sync_host_updates(updated)
```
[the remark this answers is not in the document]
tgaddair: Good point.  Further optimization is to avoid notifying any ranks except 0 in the first place, which I have also changed.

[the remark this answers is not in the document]
tgaddair: And yes, it is possible (prior to my change) that other workers could receive notifications before worker 0.  The barrier is used to ensure they all raise the exception at the same time. 
 Now that worker 0 is the only one to receive notifications, this logic is simplified somewhat.

docs/elastic.rst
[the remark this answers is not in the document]
tgaddair: Thinking about this further, we may want to make this optimization in a follow-up PR, as it's currently difficult to determine whether a worker was removed or not (from the individual worker perspective).  Initially I considered just comparing previous world size to current world size, but that would not be correct if workers were both added and removed between rendezvous.

horovod/run/elastic/driver.py
```diff
@@ -174,14 +172,14 @@ def _notify_workers_host_changes(self):
...
         # We need to ensure this list preserves relative order to ensure the oldest hosts are assigned lower ranks.
-        new_assigned_hosts = []
         self._assigned_hosts = self._discovered_hosts.filter_available_hosts(self._assigned_hosts)
         current_hosts = set(self._assigned_hosts)
         for host in self.get_available_hosts():
             if host not in current_hosts:
-                new_assigned_hosts.append(host)
                 self._assigned_hosts.append(host)
 
         # Adjust the host assignments to account for added / removed hosts
```
EnricoMi, line 209: When old hosts get additional slots then some hosts later in `self._assigned_hosts` will drop out of `self._host_assignments`, but their processes will continue running. I think I saw some code that handles this situation where processes do not get a rank though they are running. So they are turned idle I guess.

EnricoMi: I think `self._assigned_hosts` should be renamed as the naming is too similar to `self._host_assignment`. I think `self._available_hosts` describes it best (as they are populated via `self._assigned_hosts = self._discovered_hosts.filter_available_hosts(self._assigned_hosts)` plus `self.get_available_hosts()`).

horovod/run/elastic/driver.py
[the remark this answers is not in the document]
EnricoMi: Please create an issue for that optimization, or we will forget about it. You could label all those issues with a `elastic` label.

horovod/run/elastic/driver.py
EnricoMi: You mentioned that code in one of your comments:

> We do already notify workers if an assigned host is removed from the available hosts, at which point it can be gracefully removed without needing to fail on allreduce. The worker will kill itself after rendezvous when it received -1 for its rank.

So this is handled.

horovod/run/elastic/registration.py
EnricoMi: So the reset is only for the single host? It resets the count of that host only and not the entire barrier?

horovod/run/elastic/driver.py
```diff
@@ -178,6 +178,11 @@ def _notify_workers_host_changes(self):
 
         timestamp = _epoch_time_s()
         for (host, slot), client in self._worker_clients.items():
```
EnricoMi, line 180: can the elastic driver keep a reference to the `rank = 0` worker client to avoid this iteration?

horovod/run/elastic/driver.py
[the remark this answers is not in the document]
EnricoMi: I saw the fix, looks good now!

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,272 @@
...
+                logging.warning(str(e))
+            finally:
+                self._wait_hosts_cond.release()
+            first_update = False
+            self._shutdown.wait(DISCOVER_HOSTS_FREQUENCY_SECS)
+
+    def _notify_workers_host_changes(self):
+        current_hosts = self._host_assignments.keys()
+        available_hosts = self.get_available_hosts()
+        if not current_hosts - available_hosts and not self._can_assign_hosts(available_hosts - current_hosts):
```
EnricoMi, line 175: This does not cater for the situation where a `current_host` looses slots. But workers will experience exceptions on slot failure anyway.

Reassignment of hosts ans slots is initiated by workers experiencing `HostsUpdatedInterrupt` or an `HorovodInternalError` right? So there is no code here that changes `host_assignment` before workers know it will change. Otherwise we would need to sync this condition with other code in the driver (and add a comment to link those two places).

horovod/run/elastic/driver.py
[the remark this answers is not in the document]
tgaddair: Here we go: https://github.com/horovod/horovod/issues/1872

horovod/run/runner.py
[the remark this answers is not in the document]
tgaddair: https://github.com/horovod/horovod/issues/1873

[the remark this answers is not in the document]
tgaddair: https://github.com/horovod/horovod/issues/1874

horovod/common/gloo/gloo_context.cc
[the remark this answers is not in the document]
tgaddair: https://github.com/horovod/horovod/issues/1875

horovod/run/elastic/registration.py
tgaddair: It does reset the entire barrier, but because of the logic that handles the `BrokenBarrierException`, all the other ranks will re-enter the barrier.  Only the duplicated rank will raise an exception because its state has been overridden.

docs/elastic.rst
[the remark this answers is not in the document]
tgaddair: https://github.com/horovod/horovod/issues/1876

horovod/run/elastic/driver.py
tgaddair: Renamed to `ordered_available_hosts` for clarity.  Let me know if that makes sense.

horovod/run/elastic/driver.py
tgaddair: Good point.  I think the best way to handle this is to recompute the host assignments and do a 1:1 comparison between current assignments and proposed next assignments.  If they're not different, early return.  I've made this change.

horovod/run/elastic/driver.py
tgaddair: Fixed.  Added rank_assignments to index directly into worker clients.  Let me know what you think.

horovod/run/elastic/registration.py
EnricoMi: I still don't fully understand this situation. Do other reviewers understand it with the given comments only @abditag2 @nvcastet @romerojosh ?

horovod/run/elastic/driver.py
EnricoMi: very good name!

horovod/run/elastic/driver.py
EnricoMi: This feels safer.

horovod/run/elastic/driver.py
EnricoMi: Nice fix! Maybe `slot_info` should be renamed to `coordinator_slot_info`.

horovod/run/runner.py
```diff
@@ -631,32 +651,74 @@ def _run(args):
...
+                                                ssh_port=args.ssh_port,
+                                                extra_mpi_args=args.mpi_args,
+                                                key=secret.make_secret_key(),
+                                                timeout=tmout,
+                                                output_filename=args.output_filename,
+                                                run_func_mode=args.run_func is not None,
+                                                nics=args.nics)
+
+    if not gloo_built(verbose=(settings.verbose >= 2)):
+        raise ValueError('Gloo support is required to use elastic training, but has not been built.  Ensure CMake is '
```
sboshin, line 714: Will this solution ever work for MPI or is this going to only focus on gloo

mshiryaev:
@tgaddair - can this functionality work with CCL backend? If no then what are requirements for backend to enable it?

horovod/run/runner.py
EnricoMi: Elastic Horovod is Gloo only for now. If we are ever making it work with MPI, it will be a separate PR which will change this line.

nvcastet: MPI will not work in the sense you cannot reinitialize the world communicator without killing the processes. And by MPI standard, if one MPI process dies, the entire job dies.

horovod/run/elastic/registration.py
nvcastet: Yes there seems to be a short time window where a worker could fail waiting at the barrier, then it would update its state a 2nd time with failure and we don't want 2 threads coming from same worker at the barrier.
@tgaddair Do we have a unit test case for this scenario?

horovod/common/ops/nccl_operations.cc
```diff
@@ -41,8 +41,9 @@ ncclDataType_t GetNCCLDataType(const std::shared_ptr<Tensor> tensor) {
   }
 }
 
-void NCCLContext::ErrorCheck(std::string op_name, ncclResult_t nccl_result) {
+void NCCLContext::ErrorCheck(std::string op_name, ncclResult_t nccl_result, ncclComm_t& nccl_comm) {
   if (nccl_result != ncclSuccess) {
+    ncclCommAbort(nccl_comm);
```
nvcastet, line 46: As a comment, because NCCL ops are async, the error may not necessary be for this specific op but for a previous one: `this function may also return error codes from previous, asynchronous launches.`

Also, at the places we are waiting in our codebase for NCCL ops to finish, we probably need to call `ncclCommGetAsyncError()` to check for specific NCCL async errors related to network communications, etc... See https://docs.nvidia.com/deeplearning/sdk/nccl-developer-guide/docs/usage/communicators.html#asynchronous-errors-and-error-handling for code snippet.

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def _start_worker_process(self, slot_info):
+        create_worker_fn = self._create_worker_fn
+        shutdown_event = self._shutdown
+        host_event = self._discovered_hosts.get_host_event(slot_info.hostname)
+
+        def run_worker():
+            res = create_worker_fn(slot_info, [shutdown_event, host_event])
+            exit_code, timestamp = res
+            self._handle_worker_exit(slot_info, exit_code, timestamp)
```
EnricoMi, line 259: Image the following situation:
- We have a host with two slots and start the workers.
- Then the first slot becomes unavailable and the first worker gets pre-empted.
- Later a second slot becomes available again. In host assignment, the second local rank will be added to pending slots so that the second local rank will be started again.
- Then we have two worker running with the same slot_info. Both will register their exit codes with the second local rank and none will register a successful exit code for the first local rank.

horovod/run/runner.py
```diff
@@ -631,32 +651,74 @@ def _run(args):
...
         finally:
             run_func_server.shutdown_server()
     else:
         command = args.command
-        _launch_job(args, remote_host_names, settings, nics, command)
+        _launch_job(args, settings, nics, command)
         return None
 
 
+def _run_elastic(args):
```
EnricoMi, line 678: In elastic, 'args.np` must be `> 1` when `args.max_np > args.np` and `args.nics is None`. Otherwise, the task-to-task address registration will be performed on a single host and `next_task_index = (index + 1) % num_hosts` makes the task connect to itself. This is fine in static mode because there will be no task-to-task communication, but in elastic mode with `args.max_np > args.np` there will be other tasks comming up later and then the detected common interfaces might be wrong then.

A simple exception when this `args` situation occurs should help to avoid this situation. Starting from a single worker with `args.max_np > args.np` is not very realistic anyway.

horovod/run/elastic/worker.py
```diff
@@ -0,0 +1,116 @@
...
+HOROVOD_LOCAL_RANK = 'HOROVOD_LOCAL_RANK'
+
+
+class HostsUpdatedRequest(object):
+    """Notifies worker that the set of available hosts/slots has changed."""
+    def __init__(self, timestamp):
+        self.timestamp = timestamp
+
+
+class HostsUpdatedResponse(object):
```
EnricoMi, line 39: empty response, why not using `network.AckResponse`?

horovod/run/elastic/worker.py
EnricoMi: `rendezvous_port` can be `None` here when this env var is not set (which should be set by the launching code), but exception will be something like `could not connect to http://127.0.0.1:None/worker_addresses/hostname:local_rank`, so a meaningful exception that this env var is missing would improve the situation

EnricoMi:
I need some clarification on the main loops here:

On the driver side, `launch_gloo_elastic` sets up the `ElasticDriver`:
- it calls `driver.start` which starts the first `num_proc` workers
- it then calls `driver.get_results()` which waits on all `worker_thread.join()`

Those worker threads become the main loop on the driver side:
- before a worker thread terminates, it calls `ElasticDriver._handle_worker_exit` -> `WorkerStateRegistry._record_state` which waits on the `_barrier` so that all worker threads come to wait on the barrier.
- the last thread to enter the barrier will call `WorkerStateRegistry._on_workers_recorded` which stops the driver or continues the elastic loop:
  - it calls into `driver._resume`
  - this starts new worker threads
- then all worker threads continue to terminate, none adds its result to `driver._results`, they terminate and `launch_gloo_elastic` joins those terminated threads
- since there are new worker threads, the `launch_gloo_elastic` is still waiting for those and the elastic loop is spinning

On the worker side, the training function is decorated with `hvd.elastic.run`, which decorates it with `horovod.common.elastic.run_fn`:
- this main loop only terminates on `HostsUpdatedInterrupt`, it keeps spinning on `HorovodInternalError`
- I presume down-scaling through `HorovodInternalError` is handled by `HorovodBasics.shutdown` and `HorovodBasics.init()` reshaping the cluster and continuing the work with the remaining hosts
- up-scaling to new hosts makes all workers terminate and spin the elastic loop as describe above

A few questions:

1) Is it really necessary to restart all workers when we add a few new workers? Can't this be handled the way `HorovodInternalError` is handled? The driver needs to launch the new workers only and existing workers integrate them into the horovod cluster.

2) On up-scaling, `ElasticDriver._activate_hosts` -> `_start_worker_processes(pending_slots)` starts the new workers, but which path code 'restarts' the remaining workers? They won't be in `pending_slots`. All workers need to be started here I think.

horovod/run/runner.py
tgaddair: Yes, there are no plans at this time to explore MPI due to the reasons @nvcastet mentioned.  Elastic mode was actually the primary motivation for adding Gloo as an alternative to MPI :)

tgaddair:
> @tgaddair - can this functionality work with CCL backend? If no then what are requirements for backend to enable it?

From my understanding of CCL, it should be possible.  The requirements for working with elastic mode are:

1. No dependency on MPI (in other words, work with Gloo controller).
2. Raises an exception on failure (can recover from failure).
3. Communicators can be recreated dynamically at runtime (as worker membership changes).

horovod/run/elastic/registration.py
tgaddair: Good question, yes, we have a test for this in `test_elastic_driver.py::ElasticDriverTests::test_rank_and_size_with_host_failure`.  I'll modify the comments with @nvcastet's phrasing, as I think that's more clear.

horovod/run/elastic/driver.py
tgaddair: Renamed.

horovod/run/elastic/driver.py
tgaddair: Yes, I'll add a comment that this should be enforced (for future modifiers).

horovod/run/gloo_run.py
```diff
@@ -234,79 +184,122 @@ def _exec_command(command, alloc_info, event):
...
+
+    exec_command = _exec_command_fn(settings)
+    event = register_shutdown_event()
+    run_command = get_run_command(command, server_ip, nics, global_rendezv_port, elastic=True)
+    create_worker = _create_elastic_worker_fn(exec_command, run_command, env, event)
+
+    driver.start(settings.num_proc, create_worker)
+    res = driver.get_results()
+    driver.stop()
+
```
EnricoMi, line 298: I think we need to shut down the rendezvous server here:

        rendezvous.stop_server()

horovod/common/elastic.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+class ObjectState(State):
+    """State for simple Python objects.
+
+    Every object is specified as a keyword argument, and will be assigned as an attribute.
+
+    Args:
+        bcast_object: Horovod broadcast object function used to sync state dictionary.
+        kwargs: Properties to sync, will be exposed as attributes of the object.
```
EnricoMi, line 119: argument `get_rank` is not documented

horovod/common/elastic.py
```diff
@@ -0,0 +1,169 @@
...
+
+def run_fn(func, reset):
+    @functools.wraps(func)
+    def wrapper(state, *args, **kwargs):
+        notification_manager.init()
+        notification_manager.register_listener(state)
+
+        try:
+            reset_required = False
+            while True:
```
EnricoMi, line 154: `reset()` and `state.on_reset()` can be moved to the end of the loop, which won't be executed when `return func(...)` returns. No need for `reset_required` then.

tgaddair:
> 1. What adds new workers to the Horovod cluster? Only `ElasticDriver.resume` -> `_activate_hosts` -> `_start_worker_processes(pending_slots)` starts the new workers (`pending_slots`) when hosts/slots become available. But `driver.resume` is only called when all workers terminate (`WorkerStateRegistry._on_workers_recorded), which requires all workers to leave the main loop.

`WorkerStateRegistry._on_workers_recorded` triggers when all workers are either terminated OR register as `READY`.  Workers in READY state are still running (`worker_thread` has not joined), but are in the `hvd.init()` awaiting new rank information.

> 2. If only a single worker throws an exception it might not be detected because driver-side threads are joined one by one.

The threads are joined one by one when we collect results at the very end (`get_results()`).  Before that, a single worker can throw an exception resulting in a failure, and record itself as `FAILURE`.  When that happens, the other workers will fail on the next `allreduce` call with a `HorovodInternalError`, which will cause them to reset and record themselves in the `READY` state.  Only if all workers record `FAILURE` or at least one worker records `SUCCESS` do we proceed to join on all the worker threads and collect results (joining ensures that we do not return the results before all threads have written their exit codes).

> 3. If all workers would leave their main loop, which path code 'restarts' the remaining workers? They won't be in `pending_slots`. All workers need to be started here I think.

Not sure I fully understand the question, but if all workers leave the main loop (i.e., exit), then that signals that training has finished, so no new workers will be started.  In the current implementation, we do not restart workers.  If a process on a host fails, we blacklist that host (in the future, we will add a cooldown for restarting workers).

horovod/run/elastic/discovery.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def count_blacklisted_slots(self):
+        return sum([self.get_slots(host) for host, meta in self._host_state.items() if meta.is_blacklisted()])
+
+    def get_host_event(self, host):
+        return self._host_state[host].get_event()
+
+
+class HostDiscovery(object):
+    def find_available_hosts_and_slots(self):
```
EnricoMi, line 98: Isn't `hosts = set(slots.keys())` always the case? would it be sufficient to return slots only then?

horovod/run/elastic/discovery.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def blacklist(self, host):
+        if not self._host_state[host].is_blacklisted():
+            logging.warning('blacklist failing host: {}'.format(host))
+        self._host_state[host].blacklist()
+
+    def is_blacklisted(self, host):
+        return self._host_state[host].is_blacklisted()
+
+    def count_blacklisted_slots(self):
```
EnricoMi, line 90: I found that this method fails when hosts get removed from the `_available_slots` (because the discovery script does not list the host anymore). The host is still in `_host_state` but `get_slots(host)` fails with a `KeyError`.

Should this method call `get_slots(host)` only `if host in self._available_slots`?
Or should `get_slots(host)` return `0 if host not in self._available_slots`?

Would that change pose a problem for `WorkerStateRegistry._on_workers_recorded` where it says `# If there are no active hosts that aren't blacklisted, treat this as job failure` (which is btw. very difficult to understand)?

horovod/run/elastic/registration.py
```diff
@@ -0,0 +1,149 @@
...
+            logging.error('failure count == {} -> stop running'.format(self._size))
+            self._driver.stop()
+            return
+
+        # Check for failures, and add them to the blacklisted hosts list
+        failures = self.get(FAILURE)
+        for host, slot in failures:
+            self._discovered_hosts.blacklist(host)
+
+        # If there are no active hosts that aren't blacklisted, treat this as job failure
```
EnricoMi, line 138: this comment is really hard to understand, it is double-negated, please rephrase

tgaddair: Reworded.

horovod/run/elastic/discovery.py
tgaddair: Seems to be caused by a race condition.  In `count_blacklisted_slots`, we iterate over `self._available_hosts`, which should be 1:1 with `self._available_slots`, but it is possible that the discovery thread modifies this structure during the iteration.  

I propose two changes:

1. Use a default value if missing for `get_slots` of `0`.
2. Guard access to `available_hosts` and `available_slots` with a mutex.

I'll make this change.

tgaddair: Okay, this turned into a larger refactor to ensure that the host update thread doesn't conflict with any operations on the set of known hosts.  This was accomplished by moving all operations to a separate class that is atomically updated.

horovod/run/elastic/discovery.py
tgaddair: Yes, this was a holdover from an earlier design.  Refactored and removed the `hosts` variables.

horovod/common/elastic.py
tgaddair: Much cleaner!

horovod/run/gloo_run.py
tgaddair: Good catch!

horovod/run/elastic/worker.py
tgaddair: Good point.  Holdover from an earlier implementation.

horovod/run/runner.py
tgaddair: I added some additional logic to `gloo_run.py` and `wait_for_available_hosts()` to address this.

horovod/run/elastic/driver.py
tgaddair: In the current implementation, this won't be a problem because of how aggressive we are at blacklisting hosts for any failure.  In the future, once we flesh out the blacklist logic, I agree, this is something we need to be mindful of.  I will add a task to track the improved blacklisting.

tgaddair: #1926

tgaddair:
@EnricoMi please take another look when you get a chance.

horovod/common/ops/nccl_operations.cc
tgaddair: I'm putting together a follow-up PR to address the async error handling, in order to avoid making this PR any more complex.  Let me know if that sounds reasonable, and if the current PR looks good as a first step.

docs/elastic.rst
```diff
@@ -0,0 +1,358 @@
...
```
zw0610, line 84: Is it possible that the `new worker-0` is the newly added worker whose weights are not synchronized with what has be obtained from the previous training?

docs/elastic.rst
```diff
@@ -0,0 +1,358 @@
...
+Running with horovodrun
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Elastic training jobs are started using the ``horovodrun`` command line tool. The major difference when launching
+elastic jobs is that hosts are not specified explicitly, but instead **discovered** at runtime.  The most general way
+to allow Horovod to discover available hosts is to provide a ``--host-discovery-script`` when launching the job:
+
+.. code-block:: bash
+
+    $ horovodrun -np 8 --host-discovery-script discover_hosts.sh python train.py
```
zw0610, line 298: Will this `discover_hosts.sh` script be reused after `HorovodInternalError` or `HostsUpdatedInterrupt` is raised?

tgaddair: Hey @zw0610, yes, the `discover_hosts.sh` is used by the driver, while `HorovodInternalError` and `HostsUpdatedInterrupt` are raised on the workers.  On the driver side, it will continue calling `discover_hosts.sh` forever to update the set of known hosts.  On the worker side, if one of those exceptions is raised, a reset event will happen, where the results of the `discover_hosts.sh` script will be used to add or remove workers from the group.

zw0610: I see. Thanks a lot :)

horovod/run/elastic/discovery.py
EnricoMi: please document the contract: now it is a dict(hostname -> number of slots)

horovod/run/elastic/discovery.py
```diff
@@ -0,0 +1,166 @@
...
+        self._hosts_state[host].blacklist()
+
+    def is_blacklisted(self, host):
+        return self._hosts_state[host].is_blacklisted()
+
+    def get_host_event(self, host):
+        return self._hosts_state[host].get_event()
+
+    @staticmethod
+    def order_available_hosts(available_hosts, prev_host_assignment_order):
```
EnricoMi, line 117: perfect method for a unit test

horovod/run/elastic/discovery.py
```diff
@@ -0,0 +1,166 @@
...
+        # Use the host_assignment_order as it does not contain blacklisted hosts
+        return sum([self.get_slots(host) for host in self._host_assignment_order])
+
+    def update(self, hosts_state):
+        self._host_assignment_order = [host for host in self._host_assignment_order
+                                       if not hosts_state[host].is_blacklisted()]
+        return self
+
+
+class HostManager(object):
```
EnricoMi, line 82: with a mocked discovery, the whole class should be nicely unit-testable (changing set of discovered hosts, changing order of hosts, blacklisting hosts)

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,283 @@
...
+        while not self._shutdown.is_set():
+            self._wait_hosts_cond.acquire()
+            try:
+                if self._host_manager.update_available_hosts():
+                    self._notify_workers_host_changes(self._host_manager.current_hosts)
+                    self._wait_hosts_cond.notify_all()
+            except RuntimeError as e:
+                if first_update:
+                    # Misconfiguration, fail the job immediately
+                    raise
```
EnricoMi, line 166: this exception is raised in a thread, this won't affect the driver, no shutdown I think (a unit test `test_shutdown_on_initial_discovery_failure` in `test_elastic_driver` would be nice)

test/test_elastic_driver.py
```diff
@@ -0,0 +1,351 @@
...
+        for rank, (slot_info, updated_slot_info) in rank_results.items():
+            assert updated_slot_info.size == 4, rank
+            assert updated_slot_info.rank == slot_info.rank, rank
+            assert updated_slot_info.local_size == slot_info.local_size, rank
+            assert updated_slot_info.local_rank == slot_info.local_rank, rank
+            assert updated_slot_info.cross_size == 2, rank
+            assert updated_slot_info.cross_rank == slot_info.cross_rank, rank
+
+    @mock.patch('horovod.run.elastic.driver.DISCOVER_HOSTS_FREQUENCY_SECS', 0.01)
+    def test_wait_for_available_hosts(self):
```
EnricoMi, line 175: another test with `min_hosts=2` and `min_np=2` with the same slots sequence as this one to show that discovery waits for the second host even though the first has enough slots

test/test_elastic_driver.py
```diff
@@ -0,0 +1,351 @@
...
+        """Tests that driver blocks until the min number of slots are available."""
+        slots = [{'host-1': 4},
+                 {'host-1': 4, 'host-2': 8},
+                 {'host-1': 4, 'host-2': 8, 'host-3': 4}]
+        discovery = HostDiscoverySequence(slots)
+
+        driver = ElasticDriver(mock.Mock(), discovery, min_np=2, max_np=12)
+        driver.wait_for_available_hosts(min_np=10)
+        assert driver._host_manager.current_hosts.count_available_slots() >= 10
```
EnricoMi, line 184: Can we assert here that `wait_for_available_hosts` returned on the second slots sequence, not the third one? I guess this is hard because discovery polls the sequences in a separate thread at high frequency. Even the returned `current_hosts` might come from the third slots sequence if thread is too quick.

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,283 @@
...
+                self._wait_hosts_cond.release()
+            first_update = False
+            self._shutdown.wait(DISCOVER_HOSTS_FREQUENCY_SECS)
+
+    def _notify_workers_host_changes(self, current_hosts):
+        # Assignments are required to be stable via contract
+        next_host_assignments, _ = self._get_host_assignments(current_hosts)
+        if next_host_assignments == self._host_assignments:
+            # Skip notifying workers when host changes would not result in changes of host assignments
+            logging.debug('no host changes, skipping notifications')
```
EnricoMi, line 179: `no host assignment changes, skipping ...`

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,283 @@
...
+        if not coordinator_client:
+            logging.debug('no coordinator client, skipping notifications')
+            return
+
+        timestamp = _epoch_time_s()
+        try:
+            coordinator_client.notify_hosts_updated(timestamp)
+        except:
+            if self._verbose >= 2:
+                logging.exception('failed to notify {}[{}] of host updates'
```
EnricoMi, line 198: the exception message might be useful here, given we bothered setting `verbose=2`

horovod/run/gloo_run.py
```diff
@@ -234,79 +184,132 @@ def _exec_command(command, alloc_info, event):
...
+            raise RuntimeError('Horovod detected that one or more processes exited with non-zero '
                                'status, thus causing the job to be terminated. The first process '
                                'to do so was:\nProcess name: {name}\nExit code: {code}\n'
                                .format(name=name, code=exit_code))
 
 
-def gloo_run(settings, remote_host_names, nics, env, server_ip, command):
+def _get_min_start_hosts(settings):
+    # This function exists for the purpose of mocking in tests
+    return 2 if settings.max_np > settings.num_proc and not settings.nics else 1
```
EnricoMi, line 268: `settings.max_np` can be unbound (see https://github.com/horovod/horovod/pull/1849/files#diff-0bc78cae4a6c24827df277600992eedeR358-R360) and `None > int` throws `TypeError: '>' not supported between instances of 'NoneType' and 'int'`.

Should be `if is_auto_scale and not settings.nics` with `is_auto_scale = settings.min_np < settings.num_proc or settings.max_np is None or settings.max_np > settings.num_proc` (assuming `settings.min_np` cannot be `None` here).

horovod/run/gloo_run.py
```diff
@@ -234,79 +184,132 @@ def _exec_command(command, alloc_info, event):
...
+
+
+def gloo_run_elastic(settings, env, command):
+    # Make the output directory if it does not exist
+    if settings.output_filename:
+        _mkdir_p(settings.output_filename)
+
+    rendezvous = RendezvousServer(settings.verbose)
+    driver = ElasticDriver(rendezvous, settings.discovery,
+                           settings.min_np, settings.max_np,
```
EnricoMi, line 286: There is no way to set `start_timeout` here, defaults to 10 minutes, `settings` should have a way to wire this number through to this place. The existing `settings.timeout` is a `timeout.Timeout` instance.

EnricoMi: `ElasticDriver` could reuse `settings.timeout` as this measures time since the start of the horovod cluster.

horovod/common/elastic.py
```diff
@@ -0,0 +1,166 @@
...
+        last_updated_timestamp = prev_timestamp = self._last_updated_timestamp
+        while not self._host_messages.empty():
+            timestamp = self._host_messages.get()
+            if timestamp > last_updated_timestamp:
+                last_updated_timestamp = timestamp
+
+        # In order to ensure all workers raise the exception at the same time, we need to sync
+        # the updated state across all the workers.
+        # TODO(travis): this should be a max allreduce to account for changes in rank 0
+        prev_timestamp, self._last_updated_timestamp = self._bcast_object((prev_timestamp, last_updated_timestamp))
```
EnricoMi, line 91: I am getting a `RuntimeError` in `_bcast_object` with torch when a node in the cluster went down. This is not handled by the `@run` main loop as it only handles `HorovodInternalError` and `HostsUpdatedInterrupt`:
```
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/data/enrico/Work/git/horovod-GR-2/horovod/spark/task/gloo_exec_fn.py", line 30, in <module>
    main(codec.loads_base64(sys.argv[1]), codec.loads_base64(sys.argv[2]))
  File "/data/enrico/Work/git/horovod-GR-2/horovod/spark/task/gloo_exec_fn.py", line 23, in main
    task_exec(driver_addresses, settings, 'HOROVOD_RANK', 'HOROVOD_LOCAL_RANK')
  File "/data/enrico/Work/git/horovod-GR-2/horovod/spark/task/__init__.py", line 61, in task_exec
    result = fn(*args, **kwargs)
  File "/data/enrico/Work/git/horovod-GR-2/test/test_spark_torch.py", line 569, in fn
    return train(state, dir)
  File "/data/enrico/Work/git/horovod-GR-2/horovod/common/elastic.py", line 159, in wrapper
    return func(state, *args, **kwargs)
  File "/data/enrico/Work/git/horovod-GR-2/test/test_spark_torch.py", line 538, in train
    state.commit()
  File "/data/enrico/Work/git/horovod-GR-2/horovod/common/elastic.py", line 75, in commit
    self.check_host_updates()
  File "/data/enrico/Work/git/horovod-GR-2/horovod/common/elastic.py", line 93, in check_host_updates
    prev_timestamp, self._last_updated_timestamp = self._bcast_object((prev_timestamp, last_updated_timestamp))
  File "/data/enrico/Work/git/horovod-GR-2/horovod/torch/functions.py", line 219, in broadcast_object
    broadcast_(sz, root_rank, name + '.sz')
  File "/data/enrico/Work/git/horovod-GR-2/horovod/torch/mpi_ops.py", line 456, in broadcast_
    handle = broadcast_async_(tensor, root_rank, name)
  File "/data/enrico/Work/git/horovod-GR-2/horovod/torch/mpi_ops.py", line 434, in broadcast_async_
    return _broadcast_async(tensor, tensor, root_rank, name)
  File "/data/enrico/Work/git/horovod-GR-2/horovod/torch/mpi_ops.py", line 344, in _broadcast_async
    tensor, output, root_rank, name.encode() if name is not None else _NULL)
RuntimeError: Horovod has been shut down. This was caused by an exception on one of the ranks or an attempt to allreduce, allgather or broadcast a tensor after one of the ranks finished execution. If the shutdown was caused by an exception, you should see the exception in the log before the first shutdown message.
```

I am happy with getting this fixed in a separate PR after merging this beast into master.

horovod/run/elastic/registration.py
```diff
@@ -0,0 +1,151 @@
...
+    def record_ready(self, host, slot):
+        return self._record_state(host, slot, READY)
+
+    def record_success(self, host, slot):
+        return self._record_state(host, slot, SUCCESS)
+
+    def record_failure(self, host, slot):
+        return self._record_state(host, slot, FAILURE)
+
+    def _record_state(self, host, slot, state):
```
EnricoMi, line 73: I am running into the following situation:

- one worker fails
- the other worker resets
- registration registered all worker states so barrier fires (`_activate_hosts()`)
- then Spark kills the remaining worker because there were too many failures
- horovod has to timeout and does not shutdown immediately

```
[18:34:57        I registration.py:91 _record_state()] record state: special-bed85116180244240386e924d50a5a6e[0] = READY
[18:34:57        I rsh.py:70 rsh()] rsh exit code 1 for host special-f7aaa54d65dccc3b31a5088196cb3347 slot 0
[18:34:57        I registration.py:91 _record_state()] record state: special-f7aaa54d65dccc3b31a5088196cb3347[0] = FAILURE
[18:34:57        I registration.py:124 _on_workers_recorded()] all 2 workers recorded
[18:34:57        W discovery.py:107 blacklist()] blacklist failing host: special-f7aaa54d65dccc3b31a5088196cb3347
[18:34:57        I driver.py:154 _activate_hosts()] wait for available hosts: 2
20/05/05 18:35:01 WARN PythonRunner: Incomplete task 1.1 in stage 0 (TID 1) interrupted: Attempting to kill Python Worker
[18:35:01        I rsh.py:70 rsh()] rsh exit code None for host special-bed85116180244240386e924d50a5a6e slot 0
[18:35:01        I registration.py:89 _record_state()] key exists, reset barrier: special-bed85116180244240386e924d50a5a6e[0] = FAILURE
20/05/05 18:35:01 WARN TaskSetManager: Lost task 1.0 in stage 0.0 (TID 1, localhost, executor driver): TaskKilled (Stage cancelled)
```

Would you recommend to set the Spark failure limit to ∞ and rely on Horovods elastic failure count limit (is there one, I remember it being mentioned somewhere)? Or is this a situation the barrier reset should address?

I am happy to see this being addressed in a separate PR.

docs/elastic.rst
```diff
@@ -0,0 +1,358 @@
...
+
+   Because the sync function uses collective ops, and upon worker add the active workers will not reset from before this
+   function, *no Horovod collective ops (broadcast, allreduce, allgather, etc.) can be called before this function*.
+
+2. Place all variables that need to be kept in sync between worker replicas (model parameters, optimizer state, epoch and batch numbers, etc.) into a ``hvd.elastic.State`` object.
+
+   Standard state implementations are provided for TensorFlow, Keras, and PyTorch.  However, it may be necessary in some cases to override
+   the base ``hvd.elastic.State`` object to handle broadcasting custom types.
+
+3. Periodically call ``state.commit()`` to backup a copy of your state in memory.
```
EnricoMi, line 51: Is it a requirement to run a distributed operation like `hvd.allgather` just before a commit, since commit is not failing when nodes dropped out of the cluster? Otherwise workers will successfully continue to perform `state.commit` after some nodes failed, and new nodes will start from states later than where the workers terminated, but state will not contain global (`allgather`) knowledge.

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,283 @@
...
+                current_hosts = self._host_manager.current_hosts
+                if current_hosts.count_available_slots() >= min_np and len(current_hosts.available_hosts) >= min_hosts:
+                    return current_hosts
+                self._wait_hosts_cond.wait(tmout.remaining())
+                tmout.check_time_out_for('minimum number of hosts to become available')
+        finally:
+            self._wait_hosts_cond.release()
+
+    def _activate_hosts(self, min_np):
+        logging.info('wait for available hosts: {}'.format(min_np))
```
EnricoMi, line 149: Actually, we are not waiting for `min_np` hosts here but slots. There is a place now where we wait for `min_hosts`. This naming / wording here can be misleading, maybe `workers` is better than `hosts` here as we use that term in other places already

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,283 @@
...
+                # Transient error, retry until timeout
+                logging.warning(str(e))
+            finally:
+                self._wait_hosts_cond.release()
+            first_update = False
+            self._shutdown.wait(DISCOVER_HOSTS_FREQUENCY_SECS)
+
+    def _notify_workers_host_changes(self, current_hosts):
+        # Assignments are required to be stable via contract
+        next_host_assignments, _ = self._get_host_assignments(current_hosts)
```
EnricoMi, line 176: `_get_host_assignments` throws an `ValueError` when `current_hosts` is insufficient for a `min_np` assignment, which kills the discovery thread and might shutdown the whole app. An insufficient assignment is allowed from this method's point of view sometimes. The exception should be caught, ignored and make the method return. Unit test would be nice.

docs/elastic.rst
EnricoMi: I found that state.commit calls into

        self.save()
        self.check_host_updates()

where the latter performs a broadcast operation. Does that one fail if any node other then `rank == 0` dropped out of the cluster? Does this synchronize all workers?

Still, node `rank == 0` may have committed some batches that have not been done on other workers, and on reset those batches will not be re-done on other workers, they will be skipped. Is that a general problem or is this training-script specific and therefore needs to be considered on a per-script basis?

test/integration/data/elastic_torch_main.py
```diff
@@ -0,0 +1,136 @@
...
+
+exit_schedule = json.loads(args.exit_schedule) if args.exit_schedule else {}
+
+
+def check_exit(epoch, batch):
+    key = str((epoch, batch))
+    if key in exit_schedule:
+        ranks_to_exit = exit_schedule[key]
+        if start_rank in ranks_to_exit:
+            raise RuntimeError('check_rank and exit epoch={} batch={} start_rank={} rank={}'
```
EnricoMi, line 72: This only tests an exception in the training loop but not a lost worker node. Both cases should be tested, a new `--exit-mode=[exception|kill]` could be added.

Changing this to `psutil.Process(os.getpid()).kill()` makes the elastic horovod job fail and not recover. The other workers terminate on a `gloo::IoException`:
```
Mon May 11 11:44:15 2020[0]<stderr>:Killed
Mon May 11 11:44:15 2020[3]<stderr>:terminate called after throwing an instance of 'gloo::IoException'
Mon May 11 11:44:15 2020[3]<stderr>:  what():  [horovod/third_party/gloo/gloo/transport/tcp/pair.cc:563] Read error [127.0.0.1]:16802: Connection reset by peer
```

Calling `sys.exit(1)` does not make the horovod job fail because some graceful shutdown is still happening.

test/integration/data/elastic_torch_main.py
```diff
@@ -0,0 +1,136 @@
...
+@hvd.elastic.run
+def train(state):
+    state.rendezvous += 1
+    while state.epoch < args.epochs:
+        print('epoch {} batch {}'.format(state.epoch, state.batch))
+
+        while state.batch < args.batches_per_epoch:
+            check_exit(state.epoch, state.batch)
+
+            optimizer.zero_grad()
```
EnricoMi, line 99: should `optimizer` and `model` be accessed through `state`?

test/integration/data/elastic_torch_main.py
```diff
@@ -0,0 +1,136 @@
...
+                    time.sleep(0.1)
+
+        state.epoch += 1
+        state.batch = 0
+        state.commits += 1
+        state.commit()
+
+
+def on_state_reset():
+    for param_group in optimizer.param_groups:
```
EnricoMi, line 130: here as well: `state.optimizer`?

test/integration/data/elastic_torch_main.py
tgaddair: Good catch.  The problem here was that we were not properly handling failures in the control plane.  Should be fixed now.

test/integration/data/elastic_torch_main.py
tgaddair: It doesn't have to be, since we don't re-assign the variable in `state`.

test/integration/data/elastic_torch_main.py
tgaddair: Same as above.  In this case we would need to reorder some things, since we create state after this callback function.

horovod/run/gloo_run.py
EnricoMi: @tgaddair I have addressed this with commit d842e2a1a29200c372f67a263156d2599226eb78, please review

EnricoMi:
Looks like commit 5f204eab3c05839e3e6cc742c7e09c35199b9989 broke tensorflow.

horovod/run/elastic/discovery.py
tgaddair: Good idea.  Added tests.

horovod/run/elastic/driver.py
tgaddair: Fixed and added test.

test/test_elastic_driver.py
tgaddair: Usually it returns on the third sequence just because the threads progress at different rates.  We could test this behavior by removing the third sequence, but the behavior is the same.

horovod/run/elastic/driver.py
tgaddair: `logging.exception` will also print out the exception that was caught along with whatever else you provide.

horovod/run/gloo_run.py
tgaddair: That's fair.  I've also changed behavior to raise an error if the user attempts to run in fault tolerance mode with only one host.

horovod/run/gloo_run.py
tgaddair: Looks good, thanks!

horovod/common/elastic.py
tgaddair: Nice catch, added some additional checks to raise `HorovodInternalError` when we call into the Horovod lib, which should take care of this.

horovod/run/elastic/registration.py
tgaddair: For now, I think setting an unbounded failure limit is the right approach.  Added https://github.com/horovod/horovod/issues/1947 to track the reset limit in Horovod.

docs/elastic.rst
tgaddair: If a failure occurs during broadcast / commit, then a `HorovodInternalError` should be raised, which will result in a reset event (this can happen if one of the other workers fail, as you suggested).  

Because Horovod synchronizes on the allreduce, we can feel confident that if we call commit at the end of the batch, we should not need to worry about workers being out of sync.  So it should not be possible for some workers to get ahead of others.  If a worker fails in between allreduce and commit, then we should still be okay, because the gradient updates of the failing worker should still be applied to the other non-failing workers.

horovod/run/elastic/driver.py
tgaddair: Changed the name to `_activate_workers` and `wait_for_available_slots`.

horovod/run/elastic/driver.py
tgaddair: Very good catch, this should have been caught by existing unit tests, but there was a bug with the way we were generating the sequence of hosts.  This should be fixed now, so we now have proper coverage of this case.

tgaddair: In this case, we do still need to notify workers if the previous assignments were valid, so we should interpret it as an empty set of assignments.

tgaddair: Fixed.

horovod/run/elastic/driver.py
```diff
@@ -0,0 +1,299 @@
...
+
+    def get_coordinator_info(self):
+        return self._rank_assignments.get(0)
+
+    def has_rank_assignment(self, host, slot):
+        if self._host_manager.is_blacklisted(host):
+            return False
+        return host in self._host_assignments and len(self._host_assignments[host]) > slot
+
+    @property
```
EnricoMi, line 132: Why a `@property`, why not `world_size`, `coordinator_info`, `results` and `finished` as well? In any case, should not return the mutable dict `self._host_assignments` but a copy of it, deep copy not necessary here

PR #1861, opened 2020-04-10
rsh returns exit code for horovod.spark.gloo_run

EnricoMi:
This completes the new Spark over Gloo implementation by making `exec_command` in `horovod.spark.gloo_run` wait for the command executed via `rsh` and return its exit code. Further it reacts on the given `event` and aborts the command in that case.

For this, the `BasicTaskClient` can send `WaitForCommandExitCodeRequest` which blocks until command terminates on the service and returns its exit code. Turns the polling loop that sends a `CommandTerminatedRequest` request every second into an event-base notification by pushing the loop into the service. In static Horovod over Spark these requests are sent over localhost, but elastic Horovod on Spark will use this to wait for training functions' termination, which would send a request per second to each task service over the network.

`BasicService` and `BasicClient` support concurrent requests. `horovod.spark.runner._task_fn` requires `service.shutdown` to wait for all running requests to finish. Every Spark worker runs `_task_fn`, which runs the task service. It waits on command completion and then shuts down the service again. In Spark over Gloo and Elastic Horovod over Spark (which uses gloo) we will launch those commands via `rsh`, which also waits for task completion to retrieve the command's exit code. If `service.shutdown` would not wait for those requests to finish, the driver would not retrieve exit codes and dies. This is only supported by setting  `BasicService._server._block_on_close = True`. This wait has no timeout so all requests are expected to terminate logically. There are unit tests that tests for this via `BasicTaskService` and `BasicTaskClient` /  `BasicService` and `BasicClient`.

**Note**: this includes https://github.com/horovod/horovod/pull/1860, https://github.com/horovod/horovod/pull/1869 and https://github.com/horovod/horovod/pull/1870 so should be reviewed / merged after that PR.

### DIFF
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -279,28 +279,32 @@ run_spark_integration() {
   # Horovod Spark Estimator tests
   if [[ ${test} != *"tf1_1_0"* && ${test} != *"tf1_6_0"* && ${test} != *"torch0_"* && ${test} != *"mpich"* && ${test} != *"oneccl"* ]]; then
     if [[ ${test} != *"tf2"* && ${test} != *"tfhead"* ]]; then
-      run_test "${test}" "${queue}" \
-        ":spark: Spark Keras Rossmann Run (${test})" \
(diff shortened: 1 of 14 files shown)


EnricoMi:
tests succeeded earlier (https://buildkite.com/horovod/horovod/builds/2596) but failed due to timeout after merging master

PR #1868, opened 2020-04-13
Added metadata cache when fit_on_parquet

abditag2:


horovod/spark/common/store.py
```diff
@@ -184,6 +184,12 @@ def get_val_data_path(self, idx=None):
     def get_test_data_path(self, idx=None):
         return '{}.{}'.format(self._test_path, idx) if idx is not None else self._test_path
 
+    def get_data_metadata_path(self, path):
+        localized_path = self.get_localized_path(path)
```
yariv, line 188: os.path.join could help here

PR #1869, opened 2020-04-14
Move thread patterns into in_thread and on_event

EnricoMi:
- Create a (daemon) thread with a single line: in_thread
- Create a thread that executes a function on an event: on_event
- Create a method that delays a parallel function call (for testing): delay
- Clean-up wait-on-event thread in safe_shell_exec

### DIFF
--- a/horovod/run/common/util/safe_shell_exec.py
+++ b/horovod/run/common/util/safe_shell_exec.py
@@ -145,7 +145,7 @@ def execute(command, env=None, stdout=None, stderr=None, index=None, event=None)
     stop = None
     if event is not None:
         stop = threading.Event()
-        on_event(event, os.kill, (middleman_pid, signal.SIGTERM), stop=stop)
+        on_event(event, os.kill, (middleman_pid, signal.SIGTERM), stop=stop, silent=True)
 
     try:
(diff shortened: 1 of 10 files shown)


horovod/run/common/util/safe_shell_exec.py
```diff
@@ -142,26 +135,17 @@ def kill_executor_children_if_sigterm_received():
...
     # TODO: Currently this requires explicitly declaration of the event and signal handler to set
     #  the event (gloo_run.py:_launch_jobs()). Need to figure out a generalized way to hide this behind
     #  interfaces.
+    stop = None
     if event is not None:
-        bg_thread = threading.Thread(target=kill_middleman_if_master_thread_terminate)
-        bg_thread.daemon = True
-        bg_thread.start()
+        stop = threading.Event()
+        on_event(event, os.kill, (middleman_pid, signal.SIGTERM), stop=stop)
```
tgaddair, line 148: Should this set `silent=True` to be consistent with previous `kill_middleman_if_master_thread_terminate()` behavior?

EnricoMi: You are totally right, this one place was the reason I added it, but refactoring around must have dropped that argument here. Well spotted.

PR #1879, opened 2020-04-15
changed DataLoader to BatchDataLoader 

abditag2:
Has to be merged after this PR is merged and released in Petastorm: https://github.com/uber/petastorm/pull/540

### DIFF
--- a/Dockerfile.test.gpu
+++ b/Dockerfile.test.gpu
@@ -69,9 +69,6 @@ RUN if [[ ${MPI_KIND} != "None" ]]; then \
         pip install mpi4py; \
     fi
 
-# Install PetaStorm
-RUN pip install petastorm
-
 ### END OF CACHE ###
 COPY . /horovod
 
--- a/Dockerfile.test.gpu
+++ b/Dockerfile.test.gpu
@@ -69,6 +69,9 @@ RUN if [[ ${MPI_KIND} != "None" ]]; then \
         pip install mpi4py; \
     fi
 
+# Install PetaStorm
+RUN pip install petastorm
+
 ### END OF CACHE ###
 COPY . /horovod
 
--- a/horovod/spark/torch/remote.py
+++ b/horovod/spark/torch/remote.py
@@ -253,6 +253,8 @@ def RemoteTrainer(estimator, metadata, last_checkpoint_state, run_id, dataset_id
                             for col in label_columns]
 
                         sample_weights = row.get(sample_weight_col, None)
+                        if sample_weights is not None:
+                            sample_weights = sample_weights.float()
                         if cuda_available:
                             inputs = [input.cuda() for input in inputs]
                             labels = [label.cuda() for label in labels]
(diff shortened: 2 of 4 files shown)


tgaddair:
@abditag2 Petastorm v0.9.0 has been released.  Is this ready to go?

Dockerfile.test.gpu
```diff
@@ -69,6 +69,9 @@ RUN if [[ ${MPI_KIND} != "None" ]]; then \
         pip install mpi4py; \
     fi
 
+# Install PetaStorm
+RUN pip install petastorm
```
tgaddair, line 73: This is not necessary, we already install petastorm.

setup.py
```diff
@@ -1579,7 +1579,7 @@ def build_extensions(self):
 keras_require_list = ['keras>=2.0.8,!=2.0.9,!=2.1.0,!=2.1.1']
 pytorch_require_list = ['torch']
 mxnet_require_list = ['mxnet>=1.4.1']
-spark_require_list = ['h5py>=2.9', 'numpy', 'petastorm==0.8.2', 'pyarrow>=0.15.0', 'pyspark>=2.3.2']  # Petastorm 0.7.7 is not compatible with pyarrow<0.15.0
+spark_require_list = ['h5py>=2.9', 'numpy', 'petastorm==0.9.0', 'pyarrow>=0.15.0', 'pyspark>=2.3.2']  # Petastorm 0.7.7 is not compatible with pyarrow<0.15.0
```
tgaddair, line 1582: There's no need to pin to a specific version anymore (that was only necessary when there was a `pyarrow` incompatibility.  You can just change this to: `petastorm>=0.9.0`.

Also, we can remove this comment about Petastorm 0.7.7

horovod/spark/torch/remote.py
```diff
@@ -237,9 +237,9 @@ def save_checkpoint():
                                     **reader_factory_kwargs) \
                     if should_validate else empty_batch_reader() as val_reader:
 
-                    train_loader = DataLoader(train_reader,
-                                              batch_size=batch_size,
-                                              shuffling_queue_capacity=shuffle_buffer_size)
+                    train_loader = BatchedDataLoader(train_reader,
```
tgaddair, line 240: Is BatchedDataLoader strictly better than DataLoader in all cases?  Are there scenarios where DataLoader will perform better than BatchedDataLoader?

fps7806: From my experiments it is strictly better, the only reason we still have DataLoader is that certain data types (string/decimals) are not supported by BatchedDataLoader.

setup.py
abditag2: removed the comment and petastorm version.

setup.py
```diff
@@ -1579,7 +1579,7 @@ def build_extensions(self):
 keras_require_list = ['keras>=2.0.8,!=2.0.9,!=2.1.0,!=2.1.1']
 pytorch_require_list = ['torch']
 mxnet_require_list = ['mxnet>=1.4.1']
-spark_require_list = ['h5py>=2.9', 'numpy', 'petastorm==0.8.2', 'pyarrow>=0.15.0', 'pyspark>=2.3.2']  # Petastorm 0.7.7 is not compatible with pyarrow<0.15.0
+spark_require_list = ['h5py>=2.9', 'numpy', 'petastorm', 'pyarrow>=0.15.0', 'pyspark>=2.3.2']
```
tgaddair, line 1582: We need to specify `petastorm>=0.9.0` (lower bound).  Otherwise, the user could have an older version already installed, pip won't update it, and then `BatchedDataLoader` will fail to import.

abditag2: fair point. Added.

PR #1894, opened 2020-04-22
Use parameter names when scheduling broadcasts in MXNet broadcast_parameters.

romerojosh:
Fixes a bug introduced by #1718. In that commit, some syncs were removed in `broadcast_parameters`. This led to an issue where back to back calls to `broadcast_parameters` would possibly crash due to duplicate name usage, as the broadcast calls within that function use `str(i)` for naming, where `i` is just an index in the list of tensors. 
These back to back calls commonly arise when running symbolic models in MXNet, for example in the MXNet RN50 example script:
```
    # Horovod: fetch and broadcast parameters
    (arg_params, aux_params) = mod.get_params()
    if arg_params is not None:
        hvd.broadcast_parameters(arg_params, root_rank=0)
    if aux_params is not None:
        hvd.broadcast_parameters(aux_params, root_rank=0)
```

To fix this, we can replace the `str(i)` naming for the broadcasts with the actual parameter names. 

On a similar note, I also added parameter name usage to the broadcasts in `wrapped_init_impl` since I also think this is required. Unless there is a guaranteed ordering of deferred initializations, letting Horovod generate default names, which are based on an atomically increasing index, for these broadcasts is unsafe.

### DIFF
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -151,4 +151,4 @@ def broadcast_parameters(params, root_rank=0):
 
     # Run broadcasts.
     for tensor, name in zip(tensors, names):
-        broadcast_(tensor, root_rank, name=name)
+        broadcast_(tensor, root_rank, name=str(name))
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -113,7 +113,7 @@ def _append_broadcast_init(param, root_rank):
     init_impl = getattr(param, '_init_impl')
     def wrapped_init_impl(self, *args, **kwargs):
         init_impl(*args, **kwargs)
-        broadcast_(self.data(), root_rank=root_rank)
+        broadcast_(self.data(), root_rank=root_rank, name=self.name)
     return wrapped_init_impl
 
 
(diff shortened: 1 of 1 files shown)


romerojosh:
@eric-haibin-lin, please also take a look at this.

romerojosh:
Added a change to pass `test_horovod_broadcast_grad`, but it leads me to a question: For `broadcast_parameters`, do we consider having `params` be a dictionary with non-string keys a valid use of this function? The test uses integer keys, but that doesn't seem to be a typical use case (if we assume `params` should have string names as keys and parameters as values, which is consistent with the output from `Module.get_params()` or `Block.collect_params()`). The function arguments are not very well documented outside of that typical use scenario.

horovod/mxnet/__init__.py
```diff
@@ -148,5 +150,5 @@ def broadcast_parameters(params, root_rank=0):
         raise ValueError('invalid params of type: %s' % type(params))
 
     # Run broadcasts.
-    for i, tensor in enumerate(tensors):
-        broadcast_(tensor, root_rank, str(i))
+    for tensor, name in zip(tensors, names):
+        broadcast_(tensor, root_rank, name=str(name))
```
apeforest, line 154: Is name here always a string?

romerojosh: This is related to my comment above, where `test_horovod_broadcast_grad` was using integer keys in the passed in params dictionary, so this conversion to string was necessary to pass that test. However, I think passing a dictionary with non-string keys to this method should actually be considered user error because the params dictionaries from `Module.get_params()` or `Block.collect_params()` have string keys right?

PR #1899, opened 2020-04-23
Remove torchvision from pytorch extra

### LINKED ISSUE #1898: Hard Torchvision Dependency
Hi, 
we're trying to build a docker image for lightning, which also installs horovod as an extra dependency.

We noted that in https://github.com/horovod/horovod/blob/master/setup.py#L1578 you have a hard torch vision dependency.

We don't want a hard dependency on torch vision, since guys doing NLP or other stuff don't need it. 

So I just wanted to ask, whether you need a hard dependency on it and why?

Best,
Justus

tgaddair:
Fixes #1898.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1575,7 +1575,7 @@ tensorflow_require_list = ['tensorflow']
 tensorflow_cpu_require_list = ['tensorflow-cpu']
 tensorflow_gpu_require_list = ['tensorflow-gpu']
 keras_require_list = ['keras>=2.0.8,!=2.0.9,!=2.1.0,!=2.1.1']
-pytorch_require_list = ['torch','torchvision']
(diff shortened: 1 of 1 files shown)


EnricoMi:
Should we add `torchvision` somewhere in the documentation, somewhere those examples are mentioned? It is mentioned in the contributors guide but people looking for examples might not look there.

tgaddair:
> Should we add `torchvision` somewhere in the documentation, somewhere those examples are mentioned? It is mentioned in the contributors guide but people looking for examples might not look there.

Good point.  I think we could add a README / requirements.txt file for the examples, so users know what dependencies they need to run them successfully.  We could also use this as an opportunity to provide better instructions for running ResNet50 training, etc.

See: #1901.

PR #1639, opened 2020-01-07
Fix a crash bug in MXNet due to early free of tensor object

apeforest:
This PR includes the following changes:
1) Avoid early destruction of NDArray invoked by Python garbage collection by making a copy using shared_ptr
2) Simplify code by removing unnecessary template class
3) Code refactoring
4) Fixed a bug in ResizeND mentioned in https://github.com/horovod/horovod/issues/1409

This PR fixes https://github.com/horovod/horovod/issues/1533

@eric-haibin-lin @karan6181 @yuxihu Please help review!

PR #1908, opened 2020-04-25
mxnet: add fp16 compression

vycezhong:
since horovod for mxnet supports allreduce for temporary tensors https://github.com/horovod/horovod/pull/1639 , a simple wrapper like this is sufficient.  

```python
            if param.grad_req != 'null':
                compressed, ctx = self._compression.compress(param.list_grad()[0])
                allreduce_(compressed, average=False,
                           name=param.name, priority=-i)
                param._grad[0] = self._compression.decompress(compressed, ctx)
```

### DIFF
--- a/horovod/mxnet/compression.py
+++ b/horovod/mxnet/compression.py
@@ -66,7 +66,7 @@ class FP16Compressor(Compressor):
 
 
 class Compression(object):
-    """Optional gradient compression algorithm used during push_pull."""
+    """Optional gradient compression algorithm used during allreduce."""
 
     """Do not compress the gradients. This is the default."""
     none = NoneCompressor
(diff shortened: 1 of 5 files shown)


tgaddair:
This is awesome!  Thanks for the PR @vycezhong!  @apeforest @romerojosh @yuxihu can one of you review?

horovod/mxnet/compression.py
```diff
@@ -0,0 +1,75 @@
...
+        """Upcasts the tensor to the initialization dtype."""
+        tensor_decompressed = tensor
+        dtype = ctx
+        if 'float' in str(dtype):
+            tensor_decompressed = tensor.astype(dtype, copy=False)
+        return tensor_decompressed
+
+
+class Compression(object):
+    """Optional gradient compression algorithm used during push_pull."""
```
romerojosh, line 69: Change `push_pull` to `allreduce` here to be consistent with Horovod terminology.

apeforest:
> Implementation looks good to me, just left a minor comment. Should this feature also be implemented for `DistributedOptimizer`? @apeforest, any thoughts on this?

I guess for now DistributedTraining is fine. Gluon API is going to be the default training API in MXNet going forward.

Adnios:
How about this pr? It seems that Horovod has not yet added compression for mxnet

PR #1915, opened 2020-04-27
Modified safe_shell_exec to use multiprocess spawning instead of forking to prevent deadlocks

tgaddair:

### DIFF
--- a/horovod/run/common/util/safe_shell_exec.py
+++ b/horovod/run/common/util/safe_shell_exec.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 
+import multiprocessing
 import os
 import psutil
 import re
@@ -24,6 +25,7 @@ import time
 
 from horovod.run.util.threads import in_thread, on_event
 
+_PY3 = sys.version_info[0] == 3
 GRACEFUL_TERMINATION_TIME_S = 5
 
 
@@ -42,24 +44,27 @@ def terminate_executor_shell_and_children(pid):
             pass
 
     # Wait for graceful termination.
-    time.sleep(GRACEFUL_TERMINATION_TIME_S)
+    gone, alive = psutil.wait_procs(p.children(), timeout=GRACEFUL_TERMINATION_TIME_S)
 
     # Send STOP to executor shell to stop progress.
     p.send_signal(signal.SIGSTOP)
 
     # Kill children recursively.
-    for child in p.children(recursive=True):
+    for child in alive:
         try:
             child.kill()
         except psutil.NoSuchProcess:
             pass
 
     # Kill shell itself.
-    p.kill()
+    p.terminate()
+    try:
+        p.wait(timeout=GRACEFUL_TERMINATION_TIME_S)
+    except psutil.TimeoutExpired:
+        p.kill()
 
 
-def forward_stream(src_fd, dst_stream, prefix, index):
-
+def forward_stream(src_stream, dst_stream, prefix, index):
     def prepend_context(line, rank, prefix):
         localtime = time.asctime(time.localtime(time.time()))
         return '{time}[{rank}]<{prefix}>:{line}'.format(
@@ -69,75 +74,98 @@ def forward_stream(src_fd, dst_stream, prefix, index):
             line=line
         )
 
-    with os.fdopen(src_fd, 'r') as src:
-        line_buffer = ''
-        while True:
-            text = os.read(src.fileno(), 1000)
-            if not isinstance(text, str):
-                text = text.decode('utf-8')
-            if not text:
-                break
+    line_buffer = ''
+    while True:
+        text = os.read(src_stream.fileno(), 1000)
+        if not isinstance(text, str):
+            text = text.decode('utf-8')
+        if not text or len(text) == 0:
+            break
 
-            for line in re.split('([\r\n])', text):
-                line_buffer += line
-                if line == '\r' or line == '\n':
-                    if index is not None:
-                        line_buffer = prepend_context(line_buffer, index, prefix)
+        for line in re.split('([\r\n])', text):
+            line_buffer += line
+            if line == '\r' or line == '\n':
+                if index is not None:
+                    line_buffer = prepend_context(line_buffer, index, prefix)
 
-                    dst_stream.write(line_buffer)
-                    dst_stream.flush()
-                    line_buffer = ''
+                dst_stream.write(line_buffer)
+                dst_stream.flush()
+                line_buffer = ''
 
-        # flush the line buffer if it is not empty
-        if len(line_buffer):
-            if index is not None:
-                line_buffer = prepend_context(line_buffer, index, prefix)
-            dst_stream.write(line_buffer)
-            dst_stream.flush()
+    # flush the line buffer if it is not empty
+    if len(line_buffer):
+        if index is not None:
+            line_buffer = prepend_context(line_buffer, index, prefix)
+        dst_stream.write(line_buffer)
+        dst_stream.flush()
+
+    src_stream.close()
 
 
-def execute(command, env=None, stdout=None, stderr=None, index=None, event=None):
-    # Make a pipe for the subprocess stdout/stderr.
-    (stdout_r, stdout_w) = os.pipe()
-    (stderr_r, stderr_w) = os.pipe()
-
-    # Make a pipe for notifying the child that parent has died.
-    (r, w) = os.pipe()
-
-    middleman_pid = os.fork()
-    if middleman_pid == 0:
-        # Close unused file descriptors to enforce PIPE behavior.
-        os.close(w)
-        os.setsid()
-
-        executor_shell = subprocess.Popen(command, shell=True, env=env,
-                                          stdout=stdout_w, stderr=stderr_w)
-
-        sigterm_received = threading.Event()
-
-        def set_sigterm_received(signum, frame):
-            sigterm_received.set()
-
-        signal.signal(signal.SIGINT, set_sigterm_received)
-        signal.signal(signal.SIGTERM, set_sigterm_received)
-
-        def kill_executor_children_if_parent_dies():
-            # This read blocks until the pipe is closed on the other side
-            # due to the process termination.
-            os.read(r, 1)
-            terminate_executor_shell_and_children(executor_shell.pid)
-
-        in_thread(kill_executor_children_if_parent_dies)
-
-        on_event(sigterm_received, terminate_executor_shell_and_children, args=(executor_shell.pid,))
-
-        exit_code = executor_shell.wait()
-        os._exit(exit_code)
+def _exec_middleman(command, env, exit_event, stdout, stderr, rw):
+    stdout_r, stdout_w = stdout
+    stderr_r, stderr_w = stderr
+    r, w = rw
 
     # Close unused file descriptors to enforce PIPE behavior.
-    os.close(r)
-    os.close(stdout_w)
-    os.close(stderr_w)
+    stdout_r.close()
+    stderr_r.close()
+    w.close()
+    os.setsid()
+
+    executor_shell = subprocess.Popen(command, shell=True, env=env,
+                                      stdout=stdout_w, stderr=stderr_w)
+
+    def kill_executor_children_on_event():
+        exit_event.wait()
+        terminate_executor_shell_and_children(executor_shell.pid)
+
+    in_thread(kill_executor_children_on_event)
+
+    def kill_executor_children_if_parent_dies():
+        # This read blocks until the pipe is closed on the other side
+        # due to parent process termination (for any reason, including -9).
+        os.read(r.fileno(), 1)
+        terminate_executor_shell_and_children(executor_shell.pid)
+
+    in_thread(kill_executor_children_if_parent_dies)
+
+    exit_code = executor_shell.wait()
+    if exit_code < 0:
+        # See: https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html
+        exit_code = 128 + abs(exit_code)
+
+    sys.exit(exit_code)
+
+
+def execute(command, env=None, stdout=None, stderr=None, index=None, events=None):
+    ctx = multiprocessing.get_context('spawn') if _PY3 else multiprocessing
+
+    # When this event is set, signals to the middleman that it should terminate its children
+    # and exit.
+    exit_event = ctx.Event()
+
+    # Make a pipe for the subprocess stdout/stderr.
+    (stdout_r, stdout_w) = ctx.Pipe()
+    (stderr_r, stderr_w) = ctx.Pipe()
+
+    # This Pipe is how we ensure that the executed process is properly terminated (not orphaned) if
+    # the parent process is hard killed (-9). If the parent (this process) is killed for any reason,
+    # this Pipe will be closed, which can be detected by the middleman. When the middleman sees the
+    # closed Pipe, it will issue a SIGTERM to the subprocess executing the command. The assumption
+    # here is that users will be inclined to hard kill this process, not the middleman.
+    (r, w) = ctx.Pipe()
+
+    middleman = ctx.Process(target=_exec_middleman, args=(command, env, exit_event,
+                                                          (stdout_r, stdout_w),
+                                                          (stderr_r, stderr_w),
+                                                          (r, w)))
+    middleman.start()
+
+    # Close unused file descriptors to enforce PIPE behavior.
+    r.close()
+    stdout_w.close()
+    stderr_w.close()
 
     # Redirect command stdout & stderr to provided streams or sys.stdout/sys.stderr.
     # This is useful for Jupyter Notebook that uses custom sys.stdout/sys.stderr or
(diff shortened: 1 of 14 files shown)


horovod/run/common/util/safe_shell_exec.py
```diff
@@ -42,24 +44,27 @@ def terminate_executor_shell_and_children(pid):
             pass
 
     # Wait for graceful termination.
-    time.sleep(GRACEFUL_TERMINATION_TIME_S)
+    gone, alive = psutil.wait_procs(p.children(), timeout=GRACEFUL_TERMINATION_TIME_S)
 
     # Send STOP to executor shell to stop progress.
     p.send_signal(signal.SIGSTOP)
 
     # Kill children recursively.
```
EnricoMi, line 52: comment still says `recursively`, can / should we still kill those alive children recursively?

horovod/run/common/util/safe_shell_exec.py
```diff
@@ -69,75 +74,104 @@ def prepend_context(line, rank, prefix):
...
-            if not isinstance(text, str):
-                text = text.decode('utf-8')
-            if not text:
-                break
+    line_buffer = ''
+    while True:
+        text = os.read(src_stream.fileno(), 1000)
+        if not isinstance(text, str):
+            text = text.decode('utf-8')
+        if not text or len(text) == 0:
```
EnricoMi, line 82: `not text` tests for `None` or empty, right? `len(text) == 0` tests for empty again, maybe better to be explicit what `not text` is expected to test

horovod/run/common/util/safe_shell_exec.py
```diff
@@ -69,75 +74,104 @@ def prepend_context(line, rank, prefix):
...
+                                      stdout=stdout_w, stderr=stderr_w)
 
-        executor_shell = subprocess.Popen(command, shell=True, env=env,
-                                          stdout=stdout_w, stderr=stderr_w)
+    def kill_executor_children_on_event():
+        exit_event.wait()
+        terminate_executor_shell_and_children(executor_shell.pid)
 
-        sigterm_received = threading.Event()
+    in_thread(kill_executor_children_on_event)
```
EnricoMi, line 123: can be replaced with `on_event(exit_event, terminate_executor_shell_and_children)`

horovod/run/common/util/safe_shell_exec.py
```diff
@@ -69,75 +74,104 @@ def prepend_context(line, rank, prefix):
...
+def _create_event(ctx):
+    # We need to expose this method for internal testing purposes, so we can mock it out to avoid
+    # leaking semaphores.
+    return ctx.Event()
+
+
+def execute(command, env=None, stdout=None, stderr=None, index=None, events=None):
+    ctx = multiprocessing.get_context('spawn') if _PY3 else multiprocessing
+
+    # When this event is set, signals to the middleman that it should terminate its children
```
EnricoMi, line 150: should read `signal`, maybe remove newline even if line exceeds line limit a tiny bit

horovod/run/common/util/safe_shell_exec.py
```diff
@@ -150,22 +184,22 @@ def kill_executor_children_if_parent_dies():
...
 
-    # TODO: Currently this requires explicitly declaration of the event and signal handler to set
+    # TODO: Currently this requires explicitly declaration of the events and signal handler to set
     #  the event (gloo_run.py:_launch_jobs()). Need to figure out a generalized way to hide this behind
     #  interfaces.
-    stop = None
-    if event is not None:
-        stop = threading.Event()
-        on_event(event, os.kill, (middleman_pid, signal.SIGTERM), stop=stop, silent=True)
+    stop = threading.Event()
```
EnricoMi, line 190: since `stop` is always set to an `Event` (was `None` before), the `if` clause in the `finally` below can be removed and `stop.set` can always be done. If there are no events, that is redundant but simplifies the code.

horovod/run/common/util/safe_shell_exec.py
```diff
@@ -150,22 +184,22 @@ def kill_executor_children_if_parent_dies():
...
+    events = events or []
+    for event in events:
+        on_event(event, exit_event.set, stop=stop, silent=True)
 
     try:
-        res, status = os.waitpid(middleman_pid, 0)
+        middleman.join()
     except:
         # interrupted, send middleman TERM signal which will terminate children
```
EnricoMi, line 198: what kind of interrupts can this be? SIGINT?

horovod/run/common/util/safe_shell_exec.py
```diff
@@ -69,75 +74,104 @@ def prepend_context(line, rank, prefix):
...
+    # leaking semaphores.
+    return ctx.Event()
+
+
+def execute(command, env=None, stdout=None, stderr=None, index=None, events=None):
+    ctx = multiprocessing.get_context('spawn') if _PY3 else multiprocessing
+
+    # When this event is set, signals to the middleman that it should terminate its children
+    # and exit.
+    exit_event = _create_event(ctx)
```
EnricoMi, line 152: we can't use `threading.Event` here because `middleman` is a separate thread (IPC), right?

horovod/run/common/util/safe_shell_exec.py
```diff
@@ -69,75 +74,104 @@ def prepend_context(line, rank, prefix):
...
 
-        # flush the line buffer if it is not empty
-        if len(line_buffer):
-            if index is not None:
-                line_buffer = prepend_context(line_buffer, index, prefix)
-            dst_stream.write(line_buffer)
-            dst_stream.flush()
+    # flush the line buffer if it is not empty
+    if len(line_buffer):
+        if index is not None:
```
EnricoMi, line 97: these four lines in this `if` body are identical to the for loop, can them be moved into a `def`?

test/data/run_safe_shell_exec.py
```diff
@@ -0,0 +1,18 @@
...
+
+class FakeEvent(object):
+    def wait(self):
+        time.sleep(999)
+
+
+if __name__ == '__main__':
+    cmd = ' '.join([sys.executable] + sys.argv[1:])
+
+    # Mock out the event to avoid leaking semaphores
```
EnricoMi, line 15: why would this leak semaphores, why would this be problematic here, and are tests still testing properly with this mock?

test/test_run.py
```diff
@@ -346,17 +350,49 @@ def test_safe_shell_exec_interrupts_on_event(self):
...
         start = time.time()
-        self._do_test_safe_shell_exec('sleep {}'.format(sleep), 143, '', 'Terminated\n', interrupt)
+        self._do_test_safe_shell_exec('sleep {}'.format(sleep), 143, '', '', interrupt)
         duration = time.time() - start
 
         self.assertGreaterEqual(duration, 1.0)
         self.assertLess(duration, 2.0 + safe_shell_exec.GRACEFUL_TERMINATION_TIME_S, 'sleep should not finish')
         self.assertGreater(sleep, 2.0 + safe_shell_exec.GRACEFUL_TERMINATION_TIME_S, 'sleep should allow for GRACEFUL_TERMINATION_TIME_S')
 
+    def test_safe_shell_exec_interrupts_if_parent_shutdown(self):
```
EnricoMi, line 360: `if` -> `on`

test/test_run.py
```diff
@@ -346,17 +350,49 @@ def test_safe_shell_exec_interrupts_on_event(self):
...
+        child_script = os.path.join(os.path.dirname(__file__), 'data/sleep.py')
+
+        with temppath() as logfile:
+            # It's important that this executes in an entirely different interpreter with as little shared
+            # state as possible, to avoid issues with the semaphore tracker.
+            cmd = ' '.join([sys.executable, parent_script, child_script, str(sleep), logfile])
+            p = subprocess.Popen(cmd, shell=True)
+
+            # Wait until the child script has written its PID to the logfile
+            wait(lambda: os.path.exists(logfile), timeout=5)
```
EnricoMi, line 372: the file might appear but when reading is still empty, to avoid this race condition (flaky test), `sleep.py` should write into `logfile.tmp`, close it and rename it to `logfile`

test/test_run.py
```diff
@@ -346,17 +350,49 @@ def test_safe_shell_exec_interrupts_on_event(self):
...
+
+            parent = psutil.Process(p.pid)
+            with open(logfile, 'r') as f:
+                child = psutil.Process(int(f.read()))
+
+            assert parent.is_running()
```
EnricoMi, line 378: `test_run.py` uses `self.assert*` notation, so here `self.assertTrue(parent.is_running())`

horovod/run/common/util/safe_shell_exec.py
tgaddair: Good point.  Added additional logic to get children of the alive processes.

horovod/run/common/util/safe_shell_exec.py
tgaddair: Good point.  Changed to check for None before the decode, and remove the length check here.

horovod/run/common/util/safe_shell_exec.py
tgaddair: Yes, SIGINT would be the most common.

horovod/run/common/util/safe_shell_exec.py
tgaddair: We need to use a `multiprocessing.Event` because the middleman is a separate process.  The multiprocessing event is identical to the threading event, except that works over separate processes instead of threads.  However, kill -9 on a subprocess will do really bad things with global state when using an Event like this, which is why we stub it out into a mockable function.

test/data/run_safe_shell_exec.py
tgaddair: I'll add this as a comment:

Python runs a global process called Semaphore Tracker that is used across processes spawned with the multiprocessing API.  If a process spawned by multiprocessing is hard killed while holding semaphores, then they will not be properly cleaned up and will be effectively leaked.  Note that this is only a problem at test time, as in practice the process that is hard-killed will not be the child of yet another process.

test/test_run.py
tgaddair: Good point!  Fixed.

test/data/run_safe_shell_exec.py
EnricoMi: In the unit test you run the child process `run_safe_shell_exec.py` via `subprocess.Popen`. Will the Semaphore Tracker still recognize the semaphore, created in that process for the child-child process? Only `run_safe_shell_exec.py` will use the `multiprocessing` API.

horovod/run/common/util/safe_shell_exec.py
```diff
@@ -42,24 +44,32 @@ def terminate_executor_shell_and_children(pid):
...
             pass
 
     # Kill shell itself.
-    p.kill()
-
+    p.terminate()
+    try:
+        p.wait(timeout=GRACEFUL_TERMINATION_TIME_S)
+    except psutil.TimeoutExpired:
+        p.kill()
```
EnricoMi, line 69: this `p.kill()` might raise an `psutil.NoSuchProcess` as well

horovod/run/common/util/safe_shell_exec.py
```diff
@@ -69,75 +79,103 @@ def prepend_context(line, rank, prefix):
...
+        os.read(r.fileno(), 1)
+        terminate_executor_shell_and_children(executor_shell.pid)
 
-        executor_shell = subprocess.Popen(command, shell=True, env=env,
-                                          stdout=stdout_w, stderr=stderr_w)
+    in_thread(kill_executor_children_if_parent_dies)
 
-        sigterm_received = threading.Event()
+    exit_code = executor_shell.wait()
+    print('exit_code: {}'.format(exit_code))
```
EnricoMi, line 138: is this `print` meant to stay?

horovod/run/common/util/safe_shell_exec.py
```diff
@@ -42,24 +44,32 @@ def terminate_executor_shell_and_children(pid):
             pass
 
     # Wait for graceful termination.
-    time.sleep(GRACEFUL_TERMINATION_TIME_S)
+    gone, alive = psutil.wait_procs(p.children(), timeout=GRACEFUL_TERMINATION_TIME_S)
 
     # Send STOP to executor shell to stop progress.
     p.send_signal(signal.SIGSTOP)
```
EnricoMi, line 50: Why `SIGSTOP`? stops a process that can be continued later. Is this about 'freezing' it so it does not spawn new children?

test/test_run.py
```diff
@@ -346,17 +350,50 @@ def test_safe_shell_exec_interrupts_on_event(self):
...
+            parent = psutil.Process(p.pid)
+            with open(logfile, 'r') as f:
+                child_pid = int(f.read())
+            child = psutil.Process(child_pid)
+
+            self.assertTrue(parent.is_running())
+            self.assertTrue(child.is_running())
+
+            # Hard kill the parent process
+            parent.kill()
```
EnricoMi, line 383: The `parent` is actually not `run_safe_shell_exec.py`, it is an intermediate shell. Killing the shell might make `run_safe_shell_exec.py` a zombie process, `_exec_middleman` will not see the `PIPE` breaking.

test/test_run.py
```diff
@@ -346,17 +350,50 @@ def test_safe_shell_exec_interrupts_on_event(self):
...
+            parent.wait(timeout=safe_shell_exec.GRACEFUL_TERMINATION_TIME_S)
+            p.wait()
+
+            # Child process will exit when pipe breaks
+            child.wait(timeout=safe_shell_exec.GRACEFUL_TERMINATION_TIME_S)
```
EnricoMi, line 388: in worst case `child` can take up to `2 * safe_shell_exec.GRACEFUL_TERMINATION_TIME_S`:

- `psutil.wait_procs(p.children(), timeout=GRACEFUL_TERMINATION_TIME_S)`
- `p.wait(timeout=GRACEFUL_TERMINATION_TIME_S)`

Plus give it an extra second so the test does not out-race the middleman.

EnricoMi:
Here is an ASCII graph of the `test_run.py`'s IPC:

```
test_run.py -subprocess.Popen()-> shell -> python (run_safe_shell_exec.py) -ctx.Process()-> _exec_middleman -subprocess.Popen()-> shell -> python (sleep.py) -> children
 |                                                                                           |
 +- KILL -----------------------> shell                                                      |
                                                                                            broken pipe
                                                                                             |
                                                                                             +- TERM ------------------------------------> python (sleep.py)
                                                                                             +- wait ------------------------------------> python (sleep.py)
                                                                                             +- STOP ---------------------------> shell
                                                                                             +- KILL ---------------------------------------------------------> children
                                                                                             +- KILL ------------------------------------> python (sleep.py)
                                                                                             +- TERM ---------------------------> shell
                                                                                             +- wait ---------------------------> shell
                                                                                             +- KILL ---------------------------> shell

```

As mentioned in https://github.com/horovod/horovod/pull/1915#discussion_r417187758, `test_run.py` should kill `python`, not the intermediate shell.

horovod/run/common/util/safe_shell_exec.py
tgaddair: I believe so.  @alsrgv may be able to comment further.

horovod/run/common/util/safe_shell_exec.py
tgaddair: Good catch!

horovod/run/common/util/safe_shell_exec.py
tgaddair: No.  Removed.

test/test_run.py
tgaddair: Very good catch.  Added an additional logfile for the parent.

tgaddair:
> Here is an ASCII graph of the `test_run.py`'s IPC:
> 
> ```
> test_run.py -subprocess.Popen()-> shell -> python (run_safe_shell_exec.py) -ctx.Process()-> _exec_middleman -subprocess.Popen()-> shell -> python (sleep.py) -> children
>  |                                                                                           |
>  +- KILL -----------------------> shell                                                      |
>                                                                                             broken pipe
>                                                                                              |
>                                                                                              +- TERM ------------------------------------> python (sleep.py)
>                                                                                              +- wait ------------------------------------> python (sleep.py)
>                                                                                              +- STOP ---------------------------> shell
>                                                                                              +- KILL ---------------------------------------------------------> children
>                                                                                              +- KILL ------------------------------------> python (sleep.py)
>                                                                                              +- TERM ---------------------------> shell
>                                                                                              +- wait ---------------------------> shell
>                                                                                              +- KILL ---------------------------> shell
> ```
> 
> As mentioned in [#1915 (comment)](https://github.com/horovod/horovod/pull/1915#discussion_r417187758), `test_run.py` should kill `python`, not the intermediate shell.

Nice!  What tool did you use to generate this?

EnricoMi:
> Nice! What tool did you use to generate this?

A text editor and browsing the code ;-)

horovod/run/common/util/safe_shell_exec.py
alsrgv: STOP essentially freezes the process, it's not supposed to be scheduled to be executed until it's resumed.

EnricoMi: and the purpose of doing so in this method is? to avoid having it spawn new children? please add a comment to the code with the reason to STOP here.

alsrgv: Yeah, that's exactly right.

tgaddair: Updated comment.

test/data/run_safe_shell_exec.py
tgaddair: No, subprocess.Popen will spawn its own semaphore tracker.

EnricoMi:
Did you test the `test_safe_shell_exec_interrupts_on_parent_shutdown` fails with the old implementation of `safe_shell_exec`?

tgaddair:
> Did you test the `test_safe_shell_exec_interrupts_on_parent_shutdown` actually fails with the old implementation of `safe_shell_exec`?

Good question.  Yes, without the middleman this test will timeout.

PR #1918, opened 2020-04-28
Fix Spark tests by ensuring that process 0 always reports the earliest timestamp

tgaddair:

### DIFF
--- a/test/test_spark.py
+++ b/test/test_spark.py
@@ -44,6 +44,7 @@ import horovod.torch as hvd
 
 from horovod.common.util import gloo_built, mpi_built
 from horovod.run.common.util import codec, secret
+from horovod.run.common.util.env import get_env_rank_and_size
 from horovod.run.mpi_run import is_open_mpi
 from horovod.spark.common import constants, util
 from horovod.spark.common.store import HDFSStore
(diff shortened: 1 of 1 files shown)


test/test_spark.py
```diff
@@ -411,7 +411,7 @@ def _do_test_spark_run_with_gloo(self, args=(), kwargs={}, num_proc=1, extra_mpi
         def fn():
             return 1
 
-        exec_command = mock.MagicMock(return_value=(1, 1.0))
+        exec_command = mock.MagicMock(side_effect=lambda: (1, os.environ.get('HOROVOD_RANK') + 1.0))
```
EnricoMi, line 414: How is the 'earliest' timestamp relevant?

EnricoMi: found it myself:

    for name, value in sorted(res.items(), key=lambda item: item[1][1]):

wasn't aware of the earliest, thought it was sorted by rank.

EnricoMi: I fixed that earlier in another place by changing the exception match (`Process name: [0-9]+`):

    def test_spark_run_with_non_zero_exit_with_gloo(self):
        expected = '^Gloo job detected that one or more processes exited with non-zero ' \
                   'status, thus causing the job to be terminated. The first process ' \
                   'to do so was:\nProcess name: [0-9]+\nExit code: 1$'
        self.do_test_spark_run_with_non_zero_exit(use_mpi=False, use_gloo=True,
                                                  expected=expected)

If you keep your approach, please change `Process name: [0-9]+` back to `Process name: 0` in `test_spark_run_with_non_zero_exit_with_gloo`.

tgaddair: Fixed.

PR #1923, opened 2020-04-29
Sync Batch Norm for PyTorch

### LINKED ISSUE #1384: Sync Batch Norm for Pytorch
There are sync batch norm implementations for PyTorch distributed and Apex but can't seem to find one for Horovod. Is there a technical reason why there isn't one?

alsrgv:
Implementation of https://pytorch.org/docs/stable/nn.html#syncbatchnorm using Horovod.  Current version uses optimized CUDA kernels written for PyTorch and have two different invocations because they changed.  We can consider making our own implementation if it turns out to be a hassle.

## Why Sync Batch Norm?
As evidenced by https://arxiv.org/abs/1804.07612, small batches are great for training neural networks.  However, https://arxiv.org/abs/1803.08494 noted that multi-GPU training with small-batch BatchNorm is detrimental to performance.

SyncBatchNorm improves training where each worker can only hold few examples (1..4) and total number of workers is not too high - at which point it starts to lose its regularization abilities.

Fixes #1384

### DIFF
--- a/horovod/torch/sync_batch_norm.py
+++ b/horovod/torch/sync_batch_norm.py
@@ -24,6 +24,11 @@ import torch.nn.functional as F
 from torch.nn.modules.batchnorm import _BatchNorm
 
 
+# Backward compat for old PyTorch
+if not hasattr(torch.jit, 'unused'):
+    torch.jit.unused = lambda x: x
+
+
 class SyncBatchNorm(_BatchNorm):
     """
     Applies synchronous version of N-dimensional BatchNorm.  In this version, normalization
--- a/horovod/torch/sync_batch_norm.py
+++ b/horovod/torch/sync_batch_norm.py
@@ -52,6 +52,19 @@ class SyncBatchNorm(_BatchNorm):
         if input.dim() < 2:
             raise ValueError('expected at least 2D input (got {}D input)'.format(input.dim()))
 
+    def _run_bn(self, input):
+        return F.batch_norm(
+            input, self.running_mean, self.running_var, self.weight, self.bias,
+            self.training or not self.track_running_stats, self.momentum, self.eps)
+
+    @torch.jit.unused
+    def _maybe_run_sync_bn(self, input):
+        if size() == 1:
+            return self._run_bn(input)
+        return _SyncBatchNorm.apply(
+            input, self.weight, self.bias, self.running_mean, self.running_var,
+            self.eps, self.momentum)
+
     def forward(self, input):
         # currently only GPU input is supported
         if not input.is_cuda:
@@ -62,14 +75,10 @@ class SyncBatchNorm(_BatchNorm):
         if self.training and self.track_running_stats:
             self.num_batches_tracked = self.num_batches_tracked + 1
 
-        if (not self.training and self.track_running_stats) or size() == 1:
-            return F.batch_norm(
-                input, self.running_mean, self.running_var, self.weight, self.bias,
-                self.training or not self.track_running_stats, self.momentum, self.eps)
+        if not self.training and self.track_running_stats:
+            return self._run_bn(input)
         else:
-            return _SyncBatchNorm.apply(
-                input, self.weight, self.bias, self.running_mean, self.running_var,
-                self.eps, self.momentum)
+            return self._maybe_run_sync_bn(input)
 
 
 class _SyncBatchNorm(Function):
--- a/horovod/torch/sync_batch_norm.py
+++ b/horovod/torch/sync_batch_norm.py
@@ -29,6 +29,9 @@ if not hasattr(torch.jit, 'unused'):
     torch.jit.unused = lambda x: x
 
 
+_SYNC_BN_V2 = LooseVersion(torch.__version__) >= LooseVersion('1.6.0')
+
+
 class SyncBatchNorm(_BatchNorm):
     """
     Applies synchronous version of N-dimensional BatchNorm.  In this version, normalization
@@ -49,6 +52,8 @@ class SyncBatchNorm(_BatchNorm):
             module tracks the running mean and variance, and when set to `False`,
             this module does not track such statistics and always uses batch
             statistics in both training and eval modes. Default: `True`
+    
+    NOTE: only GPU input tensors are supported in the training mode.
     """
     def __init__(self, num_features, eps=1e-5, momentum=0.1, affine=True, track_running_stats=True):
         super().__init__(num_features, eps, momentum, affine, track_running_stats)
@@ -71,7 +76,7 @@ class SyncBatchNorm(_BatchNorm):
             self.eps, self.momentum)
 
     def forward(self, input):
-        # currently only GPU input is supported
+        # currently only GPU input is supported by underlying kernel from PyTorch
         if not input.is_cuda:
             raise ValueError('SyncBatchNorm expected input tensor to be on GPU')
 
@@ -106,11 +111,11 @@ class _SyncBatchNorm(Function):
         mean_all = synchronize(mean_handle)
         invstd_all = synchronize(invstd_handle)
 
-        if LooseVersion(torch.__version__) < LooseVersion('1.6.0'):
+        if _SYNC_BN_V2:
+            counts_for_bngswc = count_all.view(-1).float().to(input.device)
+        else:
             # backwards compatibility
             counts_for_bngswc = count_all.view(-1).tolist()
-        else:
-            counts_for_bngswc = count_all.view(-1).float().to(input.device)
 
         # calculate global mean & invstd
         mean, invstd = torch.batch_norm_gather_stats_with_counts(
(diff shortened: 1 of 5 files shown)


horovod/torch/sync_batch_norm.py
```diff
@@ -0,0 +1,187 @@
...
+        count_handle = allgather_async(count.unsqueeze(0), name='sync_batch_norm.count')
+        mean_handle = allgather_async(mean.unsqueeze(0), name='sync_batch_norm.mean')
+        invstd_handle = allgather_async(invstd.unsqueeze(0), name='sync_batch_norm.invstd')
+
+        # wait on the async communication to finish
+        count_all = synchronize(count_handle)
+        mean_all = synchronize(mean_handle)
+        invstd_all = synchronize(invstd_handle)
+
+        if LooseVersion(torch.__version__) < LooseVersion('1.6.0'):
```
tgaddair, line 109: Loose version check are somewhat expensive.  Maybe move this to a constant?

horovod/torch/sync_batch_norm.py
```diff
@@ -0,0 +1,187 @@
...
+        self.save_for_backward(input, weight, mean, invstd, count_all)
+
+        # apply element-wise normalization
+        return torch.batch_norm_elemt(input, weight, bias, mean, invstd, eps)
+
+    @staticmethod
+    def backward(self, grad_output):
+        grad_output = grad_output.contiguous()
+        saved_input, weight, mean, invstd, count_all = self.saved_tensors
+        grad_input = grad_weight = grad_bias = None
```
tgaddair, line 136: Seems `grad_weight` and `grad_bias` are always set.

horovod/torch/sync_batch_norm.py
```diff
@@ -0,0 +1,187 @@
...
+
+        # calculate local stats as well as grad_weight / grad_bias
+        sum_dy, sum_dy_xmu, grad_weight, grad_bias = torch.batch_norm_backward_reduce(
+            grad_output,
+            saved_input,
+            mean,
+            invstd,
+            weight,
+            self.needs_input_grad[0],
```
tgaddair, line 145: Can we explode the `self.needs_input_grad` tuple to something like:

```
need_input_grad, need_weight_grad, need_bias_grad, _, _, _ = self.needs_input_grad
```

for clarity?

horovod/torch/sync_batch_norm.py
```diff
@@ -0,0 +1,187 @@
...
+        if self.needs_input_grad[0]:
+            # synchronizing stats used to calculate input gradient.
+            sum_dy_handle = allreduce_async(sum_dy, op=Sum, name='sync_batch_norm.sum_dy')
+            sum_dy_xmu_handle = allreduce_async(sum_dy_xmu, op=Sum, name='sync_batch_norm.sum_dy_xmu')
+
+            # wait on the async communication to finish
+            sum_dy = synchronize(sum_dy_handle)
+            sum_dy_xmu = synchronize(sum_dy_xmu_handle)
+
+            if LooseVersion(torch.__version__) < LooseVersion('1.6.0'):
```
tgaddair, line 159: Same as above, consider moving this to a constant.

horovod/torch/sync_batch_norm.py
```diff
@@ -0,0 +1,187 @@
...
+    @torch.jit.unused
+    def _maybe_run_sync_bn(self, input):
+        if size() == 1:
+            return self._run_bn(input)
+        return _SyncBatchNorm.apply(
+            input, self.weight, self.bias, self.running_mean, self.running_var,
+            self.eps, self.momentum)
+
+    def forward(self, input):
+        # currently only GPU input is supported
```
tgaddair, line 74: Can you add a TODO stating what would be needed to get it work on CPU?

horovod/torch/sync_batch_norm.py
```diff
@@ -0,0 +1,187 @@
...
+if not hasattr(torch.jit, 'unused'):
+    torch.jit.unused = lambda x: x
+
+
+class SyncBatchNorm(_BatchNorm):
+    """
+    Applies synchronous version of N-dimensional BatchNorm.  In this version, normalization
+    parameters are synchronized across workers during forward pass.  This is very useful in
+    situations where each GPU can fit a very small number of examples.
+
```
tgaddair, line 37: Can you include a comment in the docstring about requiring the tensor to be on GPU?

test/test_torch.py
```diff
@@ -1667,6 +1667,59 @@ def test_horovod_join_broadcast(self):
                 ret = hvd.join(hvd.local_rank())
             else:
                 ret = hvd.join()
+    
+    def test_horovod_sync_batch_norm(self):
```
tgaddair, line 1671: May also want to include a skip if Python version is < 3 (integration tests passing because we have no Python 2 GPU tests anymore).

horovod/torch/sync_batch_norm.py
alsrgv: This is constrained by PyTorch internal kernels that we use. They're only implemented on GPU.  If we want to make it work on CPU, we should consider rolling our own implementation w/o dependency on PyTorch internals.

horovod/torch/sync_batch_norm.py
alsrgv: Fixed

horovod/torch/sync_batch_norm.py
alsrgv: Fixed

horovod/torch/sync_batch_norm.py
alsrgv: Fixed

horovod/torch/sync_batch_norm.py
alsrgv: Fixed

horovod/torch/sync_batch_norm.py
alsrgv: Done

test/test_torch.py
alsrgv: Added skip, but we should deprecate all Py2 tests :-)

tgaddair: Agreed.  Tentative plan is to drop Python 2 in 0.20.0.

horovod/torch/sync_batch_norm.py
```diff
@@ -0,0 +1,194 @@
...
+        size = input.numel() // input.size(1)
+        count = torch.tensor([size])
+
+        # calculate mean/invstd for input.
+        mean, invstd = torch.batch_norm_stats(input, eps)
+
+        count_handle = allgather_async(count.unsqueeze(0), name='sync_batch_norm.count')
+        mean_handle = allgather_async(mean.unsqueeze(0), name='sync_batch_norm.mean')
+        invstd_handle = allgather_async(invstd.unsqueeze(0), name='sync_batch_norm.invstd')
```
thuyen, line 107: We would have a lot of `BatchNorm` instances in a network. Would we need to add the instance `id` to the `name` arguments in `allreduce_async` calls here and below (line 157, 158) ?

alsrgv: Didn't seem necessary so far - current implementation of SyncBN requires each worker to execute model in the same sequence with respect to batch norms.

thuyen: Got you thanks!

horovod/torch/sync_batch_norm.py
```diff
@@ -0,0 +1,194 @@
...
+            sum_dy_xmu = synchronize(sum_dy_xmu_handle)
+
+            if _SYNC_BN_V2:
+                mean_dy = sum_dy / count_all.sum()
+                mean_dy_xmu = sum_dy_xmu / count_all.sum()
+            else:
+                # before 1.6.0, sum_dy was sum of means from every worker, so we just 
+                # need to divide it by number of workers
+                mean_dy = sum_dy / size()
+                mean_dy_xmu = sum_dy_xmu / size()
```
thuyen, line 171: I think we should divide by `count_all.sum()` for both pytorch 1.5 and 1.6.
 
For pytorch 1.5, right now if I comment out `_run_bn` code path (not using torch.nn.BatchNorm) and run the following on `v0.19.3` locally:
```
     torch.cuda.manual_seed(2020)
     bn = torch.nn.BatchNorm2d(10).cuda()
     bn_hvd = SyncBatchNorm(10).cuda()

     x = torch.rand(3, 10, 8, 8).cuda()
     x1 = x.clone().requires_grad_()
     x2 = x.clone().requires_grad_()
     y = bn(x1)
     y_hvd = bn_hvd(x2)

     y.sum().backward()
     y_hvd.sum().backward()
     print((x1.grad - x2.grad).abs().sum())
```
I got 
```
tensor(1298650., device='cuda:0')
```
But if  if we divide by `count_all.sum()` here, I got:
```
tensor(0.0004, device='cuda:0')
```

alsrgv: Great catch, I've raised #1980 to address this.  Apparently they made breaking changes in both 1.5.0 and 1.6.0.

eric-haibin-lin:
Does SyncBatchNorm support the use case where I have 2 nodes, each with 8 GPUs, and the sync only happens intra-node instead of inter-node (sync 2 groups of 8 workers instead of 16)

alsrgv:
@eric-haibin-lin, this implementation does not, but please feel free to extend it!

romerojosh:
@alsrgv @eric-haibin-lin I think the main limitation here is that Horovod still is currently designed around performing collectives across the global communicator, with no options currently to perform collectives on subsets of workers. @tgaddair, perhaps this is something we should begin to think about.

tgaddair:
Good point, @romerojosh.  Added #2139 to track.

PR #1933, opened 2020-05-03
Added optional initial_lr parameter to LearningRateScheduleCallback and updated docs

### LINKED ISSUE #1921: Keras LR callbacks have unintended behavior when resuming from checkpoint
**Environment:**
1. Framework: tf.keras
2. Framework version: TF 1.15.0
3. Horovod version: 0.19.0
4. MPI version: cray-mpich/7.7.10
5. CUDA version: n/a
6. NCCL version: n/a
7. Python version: 3.7.4
8. OS and version: Cray linux based on SLES 15
9. GCC version: 7.3.0

**Checklist:**
1. Did you search issues to find if somebody asked this question before? yes
2. If your question is about hang, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/running.rst)? n/a
3. If your question is about docker, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/docker.rst)? n/a
4. Did you check if you question is answered in the [troubleshooting guide](https://github.com/horovod/horovod/blob/master/docs/troubleshooting.rst)? yes

**Bug report:**
Please describe erroneous behavior you're observing and steps to reproduce it.

There is a problem with the Keras learning rate callbacks (inheriting from `LearningRateScheduleCallbackImpl`) as implemented when using checkpoints and resuming training. This class pulls the `initial_lr` from the model optimizer in `on_train_begin`:
https://github.com/horovod/horovod/blob/d1b13ec131af22b31d0ba999dae15a29991cfeae/horovod/_keras/callbacks.py#L137
All LR (and momentum) modifications are done with respect to that initial learning rate and the current epoch (or batch). However, if one is writing a checkpoint, the current modified learning rate and momentum are written to the checkpoint file. Then, upon loading that checkpoint and resuming training with the LR callback, it pulls the _modified_ LR as its new `initial_lr`. Unless the user takes care to reset the optimizer's LR (and momentum) after loading from checkpoint and before training, the original schedule applied will not produce the intended schedule. The Keras Imagenet resnet50 example is affected by this, for instance:
https://github.com/horovod/horovod/blob/master/examples/keras_imagenet_resnet50.py

In contrast, the LR scheduler in Keras (and tf.keras) is implemented such that modifications depend on the _current_ learning rate. This slightly different approach is therefore not affected by the checkpoint resume issue. It doesn't have any momentum correction, though.
https://github.com/tensorflow/tensorflow/blob/bab74a15d9ad6bb9066b3e31d601d6a45b1cb221/tensorflow/python/keras/callbacks.py#L1349

I think there are a couple of possible reasonable solutions. One is to change the multiplier logic to match that of the Keras LR scheduler so that the new LR is a result of the multiplier times the current LR. This logic change would likely break folks' code, though. Another possible solution is to allow (or require!) the user to set the initial LR in the callback constructor. This way I can ensure that it is always set to the correct, intended value. Finally, as I alluded to above, the user can reset their optimizer LR (and appropriately scale the momentum) after loading from checkpoint and before training. However, I consider this a workaround rather than a solution.

I'm hoping I explained it clearly enough that I don't need a MWE, but I can provide one if required.

tgaddair:
Fixes #1921.

Adds deprecation warnings with removed versions.

Fixes docs for SyncBatchNorm.

### DIFF
--- a/examples/keras_imagenet_resnet50.py
+++ b/examples/keras_imagenet_resnet50.py
@@ -144,7 +144,8 @@ callbacks = [
     # Horovod: using `lr = 1.0 * hvd.size()` from the very beginning leads to worse final
     # accuracy. Scale the learning rate `lr = 1.0` ---> `lr = 1.0 * hvd.size()` during
     # the first five epochs. See https://arxiv.org/abs/1706.02677 for details.
-    hvd.callbacks.LearningRateWarmupCallback(warmup_epochs=args.warmup_epochs, verbose=verbose),
+    hvd.callbacks.LearningRateWarmupCallback(warmup_epochs=args.warmup_epochs, initial_lr=initial_lr,
+                                             verbose=verbose),
 
     # Horovod: after the warmup reduce learning rate by 10 on the 30th, 60th and 80th epochs.
     hvd.callbacks.LearningRateScheduleCallback(start_epoch=args.warmup_epochs, end_epoch=30, multiplier=1.,
--- a/examples/keras_imagenet_resnet50.py
+++ b/examples/keras_imagenet_resnet50.py
@@ -96,6 +96,9 @@ model = keras.applications.resnet50.ResNet50(weights=None)
 # Horovod: (optional) compression algorithm.
 compression = hvd.Compression.fp16 if args.fp16_allreduce else hvd.Compression.none
 
+# Horovod: adjust learning rate based on number of GPUs.
+initial_lr = args.base_lr * hvd.size()
+
 # Restore from a previous checkpoint, if initial_epoch is specified.
 # Horovod: restore on the first worker which will broadcast both model and optimizer weights
 # to other workers.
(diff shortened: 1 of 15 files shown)


examples/keras_imagenet_resnet50.py
```diff
@@ -147,10 +147,11 @@
     hvd.callbacks.LearningRateWarmupCallback(warmup_epochs=args.warmup_epochs, verbose=verbose),
```
alsrgv, line 147: Pass initial_lr to `LearningRateWarmupCallback` as well?

tgaddair: Good catch!  Fixed.

horovod/keras/callbacks.py
```diff
@@ -99,9 +99,14 @@ def __init__(self, multiplier, start_epoch=0, end_epoch=None, staircase=True,
             steps_per_epoch: The callback will attempt to autodetect number of batches per
                              epoch with Keras >= 2.0.0. Provide this value if you have an older
                              version of Keras.
+            initial_lr: Initial learning rate at the start of training.
```
nvcastet, line 102: In the doc above for this callback, could we explicitly mention that when using a constant multiplier, it applies to the `initial_lr` and not the current learning rate? (because different behavior than TF Keras 2.2 callback which takes current lr)

nvcastet: same for TF Keras file

nvcastet: Nevermind, I missed the description for class at https://github.com/horovod/horovod/blob/2a3f43f35c840d7e8cfa9674a051ffa53be9918d/horovod/tensorflow/keras/callbacks.py#L75

PR #1946, opened 2020-05-12
Fix horovodrun LSF feature

nvcastet:
1- Remove _get_driver_ip from LSF code path (inserted by #1807)
2- Switch query from LSB_JOBID to CSM_ALLOCATION_ID because LSB_JOBID
may point to several allocations

Also, add unit test for bug 1.


EnricoMi:
@nvcastet build fails on `test_run_with_jsrun`, can you have a look?

nvcastet:
@tgaddair Great! Looking into the test error now.

Issue #1699, opened 2020-01-30
Average gradients before allreduce


szhengac:
When a large number of workers with half-precision are used, allreduce can possibly cause overflow. Thus, one solution is to divide the gradient by number of workers before performing allreduce while the averaging is performed after the allreduce in horovod.

tgaddair:
Hey @szhengac, have you observed overflow issues during training?  We experimented with a divide-before-sum approach in the past, and found that it was more likely to lead to underflow (as gradients tended to have smaller magnitudes on average, see [this](https://docs.nvidia.com/deeplearning/sdk/mixed-precision-training/index.html) page from NVIDIA).

But we can definitely consider providing both options as a config if necessary.

szhengac:
> Hey @szhengac, have you observed overflow issues during training? We experimented with a divide-before-sum approach in the past, and found that it was more likely to lead to underflow (as gradients tended to have smaller magnitudes on average, see [this](https://docs.nvidia.com/deeplearning/sdk/mixed-precision-training/index.html) page from NVIDIA).
> 
> But we can definitely consider providing both options as a config if necessary.

Yes. I observed overflow when I trained large bert with hundreds of gpus. Providing both options looks good to me.

tgaddair:
Sounds reasonable, should be pretty straightforward.  We already have the branch containing the divide-before-sum implementation on GitHub [here](https://github.com/horovod/horovod/compare/fp16_divide_before_sum), though it's from an old version of Horovod.

tgaddair:
cc @romerojosh

tgaddair:
@szhengac, can you try out the `gradient_predivide_factor` introduced in v0.20.0 and let us know if it works for your scenario?

szhengac:
I may try it when I have access to that many machines again someday :).

PR #1949, opened 2020-05-13
Add support for gradient_predivide_factor and averaging in Horovod backend.

romerojosh:
This PR implements support for `gradient_predivide_factor` in the `DistributedOptimizer`, similar to that available in [APEX DDP](https://nvidia.github.io/apex/parallel.html#apex.parallel.DistributedDataParallel). The purpose of this is to enable splitting the averaging before and after the allreduce, which can be useful in managing the numerical range for mixed precision computations (see #1699). The `gradient_predivide_factor` is applied as follows: 
```
        If op == Average, gradient_predivide_factor splits the averaging
        before and after the sum. Gradients are scaled by
        1.0 / gradient_predivide_factor before the sum and
        gradient_predivide_factor / size after the sum. 
```
To facilitate this, additional arguments, `prescale_factor` and `postscale_factor`, have been added to the basic `hvd.allreduce` functions, enabling the definition of multiplicative factors to scale the tensors before and after the allreduce respectively. For efficiency, the pre and post-scaling is implemented in the Horovod backend on the fused tensor buffer, rather than through framework level operations. For GPU, this required a CUDA kernel implementation to scale the GPU buffer which in turn, required adding compilation of CUDA code to the current build infrastructure. The changes to the build are a more general version of build code changes to enable CUDA compilation proposed in #1760. 

As an additional general benefit from these changes, gradient averaging in the optimizer can now be carried out within the Horovod backend on the fused tensor buffer using the `postscale_factor` argument, rather than on a tensor by tensor basis at the framework level. This should, in general, be more efficient.

Marking the current PR as WIP, but would appreciate feedback to finish up. The implementation is complete, but there are still a few details that need to be worked out (adding support for Keras, if/how to add this to the adasum path), as well as some additional testing.

### DIFF
--- a/examples/mxnet_imagenet_resnet50.py
+++ b/examples/mxnet_imagenet_resnet50.py
@@ -89,6 +89,8 @@ parser.add_argument('--log-interval', type=int, default=0,
                     help='number of batches to wait before logging (default: 0)')
 parser.add_argument('--save-frequency', type=int, default=0,
                     help='frequency of model saving (default: 0)')
+parser.add_argument('--gradient-predivide-factor', type=float, default=1.0,
+                    help='apply gradient predivide factor in optimizer (default: 1.0)')
 
 
 args = parser.parse_args()
@@ -322,7 +324,8 @@ def train_gluon():
     opt = mx.optimizer.create('sgd', **optimizer_params)
 
     # Horovod: create DistributedTrainer, a subclass of gluon.Trainer
-    trainer = hvd.DistributedTrainer(params, opt)
+    trainer = hvd.DistributedTrainer(params, opt,
+                                     gradient_predivide_factor=args.gradient_predivide_factor)
 
     # Create loss function and train metric
     loss_fn = gluon.loss.SoftmaxCrossEntropyLoss()
@@ -427,7 +430,8 @@ def train_module():
     opt = mx.optimizer.create('sgd', **optimizer_params)
 
     # Horovod: wrap optimizer with DistributedOptimizer
-    dist_opt = hvd.DistributedOptimizer(opt)
+    dist_opt = hvd.DistributedOptimizer(opt,
+                                        gradient_predivide_factor=args.gradient_predivide_factor)
 
     # Setup validation data and callback during training
     eval_data = None
--- a/horovod/common/controller.cc
+++ b/horovod/common/controller.cc
@@ -452,7 +452,7 @@ Response Controller::ConstructResponse(std::string& name, int joined_size) {
   }
 
   // If we are doing an allreduce, check that prescaling and postscaling factors
-  // are identical.
+  // are identical across ranks.
   double prescale_factor;
   double postscale_factor;
   if (message_type == Request::ALLREDUCE ||
--- a/horovod/common/controller.cc
+++ b/horovod/common/controller.cc
@@ -451,6 +451,36 @@ Response Controller::ConstructResponse(std::string& name, int joined_size) {
     }
   }
 
+  // If we are doing an allreduce, check that prescaling and postscaling factors
+  // are identical.
+  double prescale_factor;
+  double postscale_factor;
+  if (message_type == Request::ALLREDUCE ||
+      message_type == Request::ADASUM) {
+    prescale_factor = requests[0].prescale_factor();
+    postscale_factor = requests[0].postscale_factor();
+
+    for (unsigned int i = 1; i < requests.size(); ++i) {
+      if (error) {
+        break;
+      }
+      double request_prescale_factor = requests[i].prescale_factor();
+      double request_postscale_factor = requests[i].postscale_factor();
+
+      if (prescale_factor != request_prescale_factor ||
+          postscale_factor != request_postscale_factor) {
+        error = true;
+        error_message_stream
+            << "Mismatched prescale and/or postscale factors: "
+            << "One rank sent factors (" << prescale_factor
+            << ", " << postscale_factor << "), but another rank "
+            << "sent factors (" << request_prescale_factor
+            << ", " << request_postscale_factor << ").";
+        break;
+      }
+    }
+  }
+
   std::vector<int64_t> tensor_sizes;
   if (message_type == Request::ALLGATHER ||
       message_type == Request::ALLTOALL) {
@@ -601,6 +631,8 @@ Response Controller::ConstructResponse(std::string& name, int joined_size) {
       response.add_tensor_size(dim);
     }
     response.set_tensor_type(data_type);
+    response.set_prescale_factor(prescale_factor);
+    response.set_postscale_factor(postscale_factor);
   } else if (message_type == Request::BROADCAST) {
     response.set_response_type(Response::BROADCAST);
   } else if (message_type == Request::ALLTOALL) {
@@ -611,6 +643,8 @@ Response Controller::ConstructResponse(std::string& name, int joined_size) {
       response.add_tensor_size(dim);
     }
     response.set_tensor_type(data_type);
+    response.set_prescale_factor(prescale_factor);
+    response.set_postscale_factor(postscale_factor);
   }
   response.set_devices(devices);
 
@@ -675,7 +709,9 @@ ResponseList Controller::FuseResponses(std::deque<Response>& responses) {
         if (response.response_type() == new_response.response_type() &&
             response.devices() == new_response.devices() &&
             response.tensor_type() == new_response.tensor_type() &&
-            tensor_size + new_tensor_size <= TensorFusionThresholdBytes()) {
+            tensor_size + new_tensor_size <= TensorFusionThresholdBytes() &&
+            response.prescale_factor() == new_response.prescale_factor() &&
+            response.postscale_factor() == new_response.postscale_factor()) {
           // These tensors will fuse together well.
           tensor_size += new_tensor_size;
           response.add_tensor_name(std::move(new_response.tensor_names()[0]));
--- a/horovod/common/half.cc
+++ b/horovod/common/half.cc
@@ -39,6 +39,7 @@ bool is_avx_and_f16c() {
 }
 #endif
 
+#if HAVE_MPI
 // float16 custom data type summation operation.
 void float16_sum(void* invec, void* inoutvec, int* len,
                  MPI_Datatype* datatype) {
@@ -73,6 +74,7 @@ void float16_sum(void* invec, void* inoutvec, int* len,
     Float2HalfBits(&inout_float, inout + i);
   }
 }
+#endif
 
 } // namespace common
 } // namespace horovod
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -805,6 +805,22 @@ bool horovod_ccl_built() {
 #endif
 }
 
+bool horovod_cuda_built() {
+#if HAVE_CUDA
+  return true;
+#else
+  return false;
+#endif
+}
+
+bool horovod_rocm_built() {
+#if HAVE_ROCM
+  return true;
+#else
+  return false;
+#endif
+}
+
 int horovod_reduce_op_average() {
   return ReduceOp::AVERAGE;
 }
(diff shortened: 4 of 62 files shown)


horovod/tensorflow/__init__.py
```diff
@@ -66,14 +67,21 @@ def allreduce(tensor, average=None, device_dense='', device_sparse='',
...
         processes.
     """
     op = handle_average_backwards_compatibility(op, average)
-    # Averaging happens in framework code, so translate that to Sum for the actual call
+
     true_op = Sum if op == Average else op
+    if not rocm_built():
+      if (op == Average):
+        # Averaging happens via postscale_factor.
+        postscale_factor /= size()
```
tgaddair, line 83: For Elastic mode, having the size defined only during graph construction like this is causing problems when the world size scales up or down.  If possible, it would be good if we could move this into the C++ layer to avoid this problem.

One way we could do this: remove the "true op" concept and just pass the actual op enum to C++.  If it's AVERAGE, we can scale the `postscale_factor` dynamically within the TensorFlow custom op using the updated world size.  Then, we can still pass in SUM to the enqueue function so everything else works the same.

What do you think?

romerojosh: I see what you are saying. Perhaps we can move the modification of `postscale_factor` into `EnqueueTensorAllreduce` here, if op is average:
https://github.com/horovod/horovod/blob/b94807d20916a6ac5fd85922e593b2f35f62b4e0/horovod/common/operations.cc#L830-L837

But then this also leads to the question of how to handle the `gradient_predivide_factor` as well. For elastic, are we ok if the `postscale_factor` and `prescale_factor` remain constant w.r.t the world size? In general, the `gradient_predivide_factor` is set based the number of workers.

horovod/common/controller.cc
```diff
@@ -449,6 +449,36 @@ Response Controller::ConstructResponse(std::string& name, int joined_size) {
     }
   }
 
+  // If we are doing an allreduce, check that prescaling and postscaling factors
+  // are identical.
```
tgaddair, line 453: For clarity, can you change this to "are identical across ranks"?

examples/mxnet_imagenet_resnet50.py
```diff
@@ -322,7 +324,8 @@ def evaluate(epoch):
     opt = mx.optimizer.create('sgd', **optimizer_params)
 
     # Horovod: create DistributedTrainer, a subclass of gluon.Trainer
-    trainer = hvd.DistributedTrainer(params, opt)
+    trainer = hvd.DistributedTrainer(params, opt,
```
tgaddair, line 327: Regarding use in examples: I think we should try to balance presenting the available features to users with not overwhelming them with lots of parameters.  So for some examples (where we only use the defaults) I would suggest keeping them that way, while for the examples that show things like gradient compression and Adasum, I would add this parameter.  Does that sounds reasonable to you?

horovod/torch/optimizer.py
```diff
@@ -113,7 +116,18 @@ def _allreduce_grad_async(self, p):
         tensor = p.grad
         tensor_compressed, ctx = self._compression.compress(tensor)
 
-        handle = allreduce_async_(tensor_compressed, name=name, op=self.op)
+        if self.op == Average and self.gradient_predivide_factor != 1.0:
```
tgaddair, line 119: Since `allreduce_async_` is in the public API, I think we should consider how this affects the user's use this function.  Here we're defining an Average op as a Sum op with a post-scale factor scaled by 1 / size.  But if the user were to call `allreduce_async_` directly with `op=Average` and `postscale_factor=s` they would get a different result.

In the implementation of `_allreduce_async`, it seems that the divisor is only size when ROCm is used.  Which would seem to break `op=Average` when calling `allreduce_async_` directly.  Maybe we can pull this logic into `_allreduce_async` to simplify things?

setup.py
```diff
@@ -148,6 +152,15 @@ def get_cpp_flags(build_ext):
 
     raise DistutilsPlatformError(last_err)
 
+def get_nvcc_flags():
+    default_flags = ['--std=c++11', '-O3', '-Xcompiler', '-fPIC']
+    cc_list_env = os.environ.get('HOROVOD_BUILD_CUDA_CC_LIST')
```
tgaddair, line 157: Can you add docs about this flag and its defaults to https://github.com/horovod/horovod/blob/master/docs/install.rst#environment-variables ?

horovod/common/controller.cc
romerojosh: Yep, no problem. Done.

horovod/common/controller.cc
```diff
@@ -449,6 +449,36 @@ Response Controller::ConstructResponse(std::string& name, int joined_size) {
...
+  double prescale_factor;
+  double postscale_factor;
+  if (message_type == Request::ALLREDUCE ||
+      message_type == Request::ADASUM) {
+    prescale_factor = requests[0].prescale_factor();
+    postscale_factor = requests[0].postscale_factor();
+
+    for (unsigned int i = 1; i < requests.size(); ++i) {
+      if (error) {
```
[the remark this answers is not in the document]
romerojosh: Definitely agree. This part of the code has gotten pretty messy.

horovod/common/half.cc
```diff
@@ -39,6 +39,7 @@ bool is_avx_and_f16c() {
 }
 #endif
 
+#if HAVE_MPI
```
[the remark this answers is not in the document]
romerojosh: The reason is because `half.h` and `half.cc` used to only be used in MPI compilation paths. I added usage of these files to `collective_operations.h` in order to use functionality of the fp16 specialization for `ScaleBufferCPUImpl`, so this file is now compiled in all cases and requires the ifdef guard of the MPI specific `float16_sum` function.

horovod/common/operations.cc
```diff
@@ -813,22 +829,34 @@ Status EnqueueTensorAllreduce(std::shared_ptr<OpContext> context,
...
+    postscale_factor /= horovod_global.controller->GetSize();
+#else
     LOG(ERROR, horovod_global.controller->GetRank()) << "Enqueuing AVERAGE allreduce is not allowed.";
     return status.Aborted("AVERAGE not allowed.");
+#endif
+  } else if (reduce_op == ReduceOp::ADASUM) {
+#if HAVE_NCCL
+    if (device != CPU_DEVICE_ID) {
+      // Averaging by local size happens via postscale_factor
+      postscale_factor /= horovod_global.controller->GetLocalSize();
```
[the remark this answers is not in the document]
romerojosh: Nice catch. It does have that limitation. Added the additional `!HAVE_ROCM` to the if condition.

horovod/tensorflow/__init__.py
```diff
@@ -40,7 +40,8 @@
 
 
 def allreduce(tensor, average=None, device_dense='', device_sparse='',
-              compression=Compression.none, op=None):
+              compression=Compression.none, op=None,
+              prescale_factor=1.0, postscale_factor=1.0):
```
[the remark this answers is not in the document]
romerojosh: Yes. In the typical use case with `gradient_predivide_factor`, `prescale_factor = 1 / gradient_predivide_factor` and `postscale_factor = gradient_predivide_factor / size`.

horovod/tensorflow/__init__.py
```diff
@@ -65,14 +66,22 @@ def allreduce(tensor, average=None, device_dense='', device_sparse='',
...
         processes.
     """
     op = handle_average_backwards_compatibility(op, average)
-    # Averaging happens in framework code, so translate that to Sum for the actual call
-    true_op = Sum if op == Average else op
+
+    average_in_framework = False
+    if rocm_built():
+        # For ROCm, perform averaging at framework level
+        if op == Average or op == Adasum:
```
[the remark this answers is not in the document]
romerojosh: Works for me. Changed it.

horovod/tensorflow/__init__.py
```diff
@@ -202,7 +219,18 @@ def after_create_session(self, session, coord):
...
 def _make_allreduce_grads_fn(name, device_dense, device_sparse,
-                             compression, sparse_as_dense, op):
+                             compression, sparse_as_dense, op, gradient_predivide_factor):
+    if op == Average and gradient_predivide_factor != 1.0:
+        # Perform averaging via pre/postscaling factors.
+        # Split average operation across pre/postscale factors
+        prescale_factor = 1.0 / gradient_predivide_factor
+        postscale_factor = gradient_predivide_factor / size()
+        true_op = Sum
+    else:
```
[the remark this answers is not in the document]
romerojosh: The user isn't in direct control of the prescale and postscale factors in this part of the code, since it is within the `DistributedOptimizer`. Their only mechanism to change these here is through `gradient_predivide_factor`. I did add a new exception to ensure that `gradient_predivide_factor` is only used if `op==Average`. 

If a user wants to apply pre- and post-scale factors with other operations (sum, adasum), they can do so using the normal `allreduce` function.

horovod/tensorflow/__init__.py
```diff
@@ -420,7 +451,16 @@ def DistributedOptimizer(optimizer, name=None, use_locking=False, device_dense='
       op:
         The reduction operation to use when combining gradients across
         different ranks.
+      gradient_predivide_factor:
+        If op == Average, gradient_predivide_factor splits the averaging
```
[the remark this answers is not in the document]
romerojosh: I added exceptions to disallow cases where `gradient_predivide_factor != 1.0` and `op != Average`.

horovod/tensorflow/__init__.py
```diff
@@ -492,13 +533,22 @@ def DistributedGradientTape(gradtape, device_dense='', device_sparse='',
           op:
             The reduction operation to use when combining gradients across
             different ranks.
+          gradient_predivide_factor:
+            If op == Average, gradient_predivide_factor splits the averaging
+            before and after the sum. Gradients are scaled by
+            1.0 / gradient_predivide_factor before the sum and
+            gradient_predivide_factor / size after the sum.
         """
+        if gradient_predivide_factor != 1.0:
```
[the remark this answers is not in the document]
romerojosh: Since I added the new exception for `op != Average` here, I left it as is.

horovod/torch/optimizer.py
romerojosh: This is a similar to TF, in that in this section of the code, the user does not have explicit control over the pre- and post-scale factors but only indirect control through `gradient_predivide_factor`. 

For the `allreduce_async_` implementation, they could set `op=Average` and both scale factors and it should work fine (i.e. the tensors will be prescaled before allreduce by `prescale_factor`, postscaled after allreduce by `postscale_factor` and divided by `size` for the average. When `op==Average`, we do not use divisor, but instead apply the additional `size` factor into `postscale_factor` in `EnqueueTensorAllreduce` here: https://github.com/horovod/horovod/blob/b0e51db4e26ea5be153cf3f3489660ebd0e310e3/horovod/common/operations.cc#L840

I'm can't think of a situation where someone would want to mix pre- and post-scale factors with `op==Average` but I do think this is doing what would be expected in that case.

setup.py
romerojosh: I added documentation for this flag. Right now, the default compute capability is being detected by NVCC at build time. Do you think we should add a set of default compute capabilities to target a particular set of GPU architectures in all builds?

examples/mxnet_imagenet_resnet50.py
romerojosh: I went ahead and remove the `gradient_predivide_factor` argument from some examples to simplify things.

horovod/tensorflow/__init__.py
```diff
@@ -202,7 +219,18 @@ def after_create_session(self, session, coord):
 
 @_cache
 def _make_allreduce_grads_fn(name, device_dense, device_sparse,
-                             compression, sparse_as_dense, op):
+                             compression, sparse_as_dense, op, gradient_predivide_factor):
+    if op == Average and gradient_predivide_factor != 1.0:
```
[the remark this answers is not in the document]
romerojosh: If it's alright with you, my preference would be to leave the MXNet code as is (averaging in the C++ code, now for ROCm only) since that is how it was originally implemented and will be removed eventually anyway.

setup.py
tgaddair: I think the way you have it now is reasonable.  Thanks for updating the docs!

eric-haibin-lin:
@szhengac FYI

PR #1951, opened 2020-05-14
[WIP] Test Python3.8 with Spark3 (preview2)

EnricoMi:


EnricoMi:
re-opened from `horovod:test-spark-3` https://github.com/horovod/horovod/pull/1957

PR #1952, opened 2020-05-15
Add NCCL operation for allgather

zhouxianhao:
This PR is to improve the performance of allgather function when HOROVOD_GPU_ALLGATHER=NCCL. Using ncclAllgather when each rank has the same shape, otherwise using ncclBroadcast. Using ncclAllgather is faster than using ncclBroadcast, and using ncclBroadcast is faster then using MPI_Allgatherv.


horovod/common/ops/nccl_operations.cc
```diff
@@ -395,5 +395,116 @@ Status NCCLBroadcast::Execute(std::vector<TensorTableEntry>& entries,
...
+      }
+    }
+  }
+
+  // Do allgather.
+  if (same_shape) {
+    auto nccl_result = ncclAllGather(fused_input_data, buffer_data,
+                                     recvcounts[0] * element_size,
+                                     ncclChar,
+                                     *nccl_op_context_.nccl_comm_, *gpu_op_context_.stream);
```
romerojosh, line 466: Is there a reason for using `ncclChar` and other arguments in terms of bytes here rather than using the NCCL datatype associated with the tensor datatype and other arguments in terms of elements?

horovod/common/ops/nccl_operations.cc
```diff
@@ -395,5 +395,116 @@ Status NCCLBroadcast::Execute(std::vector<TensorTableEntry>& entries,
...
+  // Sizes of subcomponents of each entry from all ranks
+  auto** entry_component_sizes = new int64_t* [entries.size()];
+
+  // Offset of each subcomponent of every entry in the final buffer after
+  // allgatherv
+  auto** entry_component_offsets = new int64_t* [entries.size()];
+
+  int global_size = global_state_->controller->GetSize();
+  int global_rank = global_state_->controller->GetRank();
+  auto* recvcounts = new int[global_size]();
```
romerojosh, line 415: The allocations with new are not being deleted in this routine, but need to be to avoid memory leaks. This was fixed recently for the other allgather implementations in #1845.

horovod/common/operations.cc
```diff
@@ -148,6 +148,11 @@ OperationManager* CreateOperationManager(HorovodGlobalState& state) {
   std::vector<std::shared_ptr<BroadcastOp>> broadcast_ops;
   std::vector<std::shared_ptr<AllreduceOp>> adasum_ops;
 
+#if HAVE_NCCL && HOROVOD_GPU_ALLGATHER == 'N'
+  allgather_ops.push_back(std::shared_ptr<AllgatherOp>(
+      new NCCLAllgather(&nccl_context, &gpu_context, &state)));
+#endif
```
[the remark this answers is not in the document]
zhouxianhao: I put it here because I think the priority of NCCLAllgather should higher than MPIHierarchicalAllgather. What do you think about this?

horovod/common/ops/nccl_operations.cc
zhouxianhao: Because test_tensorflow.py will test tf.uint16, tf.int16 and tf.bool, but NCCL do not support it. I change it to ncclChar just like the NCCLBroadcast operation.

horovod/common/ops/nccl_operations.cc
zhouxianhao: I have fixed it.

horovod/common/operations.cc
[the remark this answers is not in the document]
romerojosh: The behavior should be that if a user sets `HOROVOD_HIERARCHICAL_ALLGATHER=1` that the hierarchical mode is used. Since there is no NCCL-based hierarchical allgather, this flag should trigger the use of `MPIHierarchicalAllgather`. For this to happen, the `NCCLAllgather` op has to be pushed to the vector after `MPIHierarchicalAllgather`. By default,  `HOROVOD_HIERARCHICAL_ALLGATHER=0` which disables `MPIHierarchicalAllgather`, so `NCCLAllgather` will be used in the default case.

horovod/common/ops/nccl_operations.cc
romerojosh: Makes sense to me. Thanks for clarifying.

horovod/common/operations.cc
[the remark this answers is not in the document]
zhouxianhao: Thanks for clarifying. I have change it.

PR #1954, opened 2020-05-15
Dropped support for Python 2

### LINKED ISSUE #1929: Proposal: Drop Python 2 support in v0.20.0
At this point, most frameworks in our ecosystem have dropped Python 2 support, so it makes sense for us to consider doing the same.  

Already, a number of our more recent features (elastic mode, synchronized batch normalization) are Python 3 only.

We still officially maintain support for Python 2 for all legacy code paths, but there are a number of reasons to consider dropping this:

1. Running Python 2 unit tests cost money to run in the cloud, and since we have nearly 1:1 Python 2 to Python 3 test coverage, this nearly doubles our cloud bill for CPU tests.
2. As new versions of Python continue to be released, it becomes a challenge to just maintain mutual compatibility between Python 3.6 (our earliest supported version of Python 3) and newer versions (python 3.8, which recently changes the default multiprocessing run mode, see #1891). 
3. Maintaining feature parity becomes a challenge, as a number of basic things supported natively in Python 3 require elaborate workarounds for Python 2 (for example, multiprocessing barrier).

Additionally, users who need to use Python 2 can continue to use older versions of Horovod.  As TensorFlow and PyTorch should not be releasing new Python 2 versions of their libraries, this shouldn't be an issue for users who are stuck using Python 2.

What does dropping Python 2 support mean?

1. No more Python 2 tests in Buildkite.
2. No more Python 2 Docker images or PyPi packages.
3. No more system version checks to perform different logic for Python 2 vs Python 3.
4. No more dependency on `six`, replace all usages with native Python 3 equivalents.
5. No more doc references to Python 2, or unit test exclusions based on Python 2.

TSC members, please review and provide your feedback:

@alsrgv @nvcastet @romerojosh @EnricoMi @karakusc @apeforest

tgaddair:
Fixes #1929.

- Removed Python 2 tests from Buildkite.
- Removed references to Python 2 in Docker images.
- Removed system version checks that perform different logic for Python 2 vs Python 3.
- Removed dependency on `six` and replaced all usages with native Python 3 equivalents.
- Removed doc references to Python 2 and added Python 3 to install requirements.
- Removed unit test exclusions for Python 2.
- Removed Python 2 from supported versions in setup.

### DIFF
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -8,14 +8,11 @@ repository=823773083436.dkr.ecr.us-east-1.amazonaws.com/buildkite
 
 # list of all the tests
 tests=( \
-       test-cpu-openmpi-py2_7-tf1_6_0-keras2_1_2-torch0_4_1-mxnet1_4_1-pyspark2_3_2 \
        test-cpu-openmpi-py3_6-tf1_6_0-keras2_1_2-torch0_4_1-mxnet1_4_1-pyspark2_3_2 \
-       test-cpu-gloo-py2_7-tf1_15_0-keras2_3_1-torch1_4_0-mxnet1_5_0-pyspark2_4_0 \
        test-cpu-gloo-py3_6-tf1_15_0-keras2_3_1-torch1_4_0-mxnet1_5_0-pyspark2_4_0 \
        test-cpu-gloo-py3_7-tf2_2_0-keras2_3_1-torch1_5_0-mxnet1_5_0-pyspark2_4_0 \
        test-cpu-gloo-py3_8-tf2_2_0-keras2_3_1-torch1_5_0-mxnet1_5_0-pyspark2_4_0 \
        test-cpu-openmpi-py3_6-tf1_14_0-keras2_2_4-torch1_2_0-mxnet1_4_1-pyspark2_4_0 \
-       test-cpu-openmpi-py2_7-tf2_0_0-keras2_3_1-torch1_3_0-mxnet1_5_0-pyspark2_4_0 \
        test-cpu-openmpi-py3_6-tf2_0_0-keras2_3_1-torch1_3_0-mxnet1_5_0-pyspark2_4_0 \
        test-cpu-openmpi-py3_6-tfhead-kerashead-torchhead-mxnethead-pyspark2_4_0 \
        test-cpu-mpich-py3_6-tf1_15_0-keras2_3_1-torch1_3_0-mxnet1_5_0-pyspark2_4_0 \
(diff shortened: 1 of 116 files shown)


EnricoMi:
Do we need to mention Python3 in docs somewhere?

.buildkite/gen-pipeline.sh
```diff
@@ -298,7 +285,7 @@ run_spark_integration() {
   # Horovod Spark Estimator tests
   if [[ ${test} != *"tf1_1_0"* && ${test} != *"tf1_6_0"* && ${test} != *"torch0_"* && ${test} != *"mpich"* && ${test} != *"oneccl"* ]]; then
     if [[ ${test} != *"tf2"* && ${test} != *"tfhead"* ]]; then
-      if [[ ! (  ${test} == *"py2"* && ${test} == *"gloo"* && ${test} != *"openmpi-gloo"* ) ]]; then
+      if [[ ! (  ${test} == *"gloo"* && ${test} != *"openmpi-gloo"* ) ]]; then
```
EnricoMi, line 288: now that there are only two conditions left, the overall negation should be applied to simplify the expression:

    ${test} != *"gloo"* || ${test} == *"openmpi-gloo"*

.buildkite/gen-pipeline.sh
```diff
@@ -319,7 +306,7 @@ run_spark_integration() {
       fi
     fi
 
-    if [[ ! (  ${test} == *"py2"* && ${test} == *"gloo"* && ${test} != *"openmpi-gloo"* ) ]]; then
+    if [[ ! (  ${test} == *"gloo"* && ${test} != *"openmpi-gloo"* ) ]]; then
```
EnricoMi, line 309: same here

tgaddair:
> Do we need to mention Python3 in docs somewhere?

I added a mention in the Install Guide under "requirements".  Might be worth it to put it somewhere on the main README, but couldn't think of a good place.

EnricoMi:
> > Do we need to mention Python3 in docs somewhere?
> 
> I added a mention in the Install Guide under "requirements". Might be worth it to put it somewhere on the main README, but couldn't think of a good place.

I think `install.rst` is good enough, haven't seen that before I send that comment.

EnricoMi:
Are these imports due to Python 2 compatibility?
```
try:
    from shlex import quote
except ImportError:
    from pipes import quote
```

```
try:
    from collections.abc import Iterable
except ImportError:
    from collections import Iterable
```

tgaddair:
> Are these imports due to Python 2 compatibility?
> 
> ```
> try:
>     from shlex import quote
> except ImportError:
>     from pipes import quote
> ```
> 
> ```
> try:
>     from collections.abc import Iterable
> except ImportError:
>     from collections import Iterable
> ```

Good catch, looks like these imports both work in Python 3.6 and above, so I'll update.

EnricoMi:
There is one more Python2 condition in .buildkite/gen-pipeline.sh: https://github.com/horovod/horovod/blob/master/.buildkite/gen-pipeline.sh#L261

PR #1956, opened 2020-05-16
Elastic Horovod on Spark auto-scale

EnricoMi:
Blocked by #2063.

### DIFF
--- a/horovod/run/common/service/task_service.py
+++ b/horovod/run/common/service/task_service.py
@@ -15,8 +15,8 @@
 
 import threading
 
-from horovod.run.common.util import network
-from horovod.run.common.util import safe_shell_exec
+from horovod.run.common import util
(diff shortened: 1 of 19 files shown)


horovod/run/common/service/task_service.py
```diff
@@ -194,14 +194,31 @@ def wait_for_initial_registration(self, timeout):
             self._wait_cond.release()
 
     def wait_for_command_start(self, timeout=None):
+        """Waits for a command to start.
```
tgaddair, line 197: The overloaded behavior is a little tricky to follow.  I would consider breaking this out into two methods: one that raises an exception and takes a Timeout object, and another that takes a numerical value and returns the state at the end of the timeout.

EnricoMi: You are right, this is horrible. I have split it up into two methods. Thanks!

PR #1960, opened 2020-05-17
Added HOROVOD_GPU_OPERATIONS installation variable

tgaddair:
This simplifies enabling NCCL support for all GPU operations.


setup.py
```diff
@@ -632,21 +632,29 @@ def get_common_options(build_ext):
...
+        if gpu_operations:
+            if gpu_operation:
+                raise DistutilsError(f'Cannot specify both HOROVOD_GPU_OPERATIONS and {op_variable_name} options. '
+                                     'Try unsetting one of these variables and reinstalling.')
+            return gpu_operations
+
+        if gpu_operations and gpu_operation not in options:
+            options_str = ', '.join(options)
+            raise DistutilsError(f'{op_variable_name}={gpu_operation} is invalid, '
+                                 f'supported values are: {{{options_str}}}.')
```
EnricoMi, line 651: why braces in `supported values are: {NCCL, MPI}`?

setup.py
```diff
@@ -632,21 +632,29 @@ def get_common_options(build_ext):
...
+
+    def get_gpu_op_variable(op_variable_name, options):
+        gpu_operation = os.environ.get(op_variable_name)
+        if gpu_operations:
+            if gpu_operation:
+                raise DistutilsError(f'Cannot specify both HOROVOD_GPU_OPERATIONS and {op_variable_name} options. '
+                                     'Try unsetting one of these variables and reinstalling.')
+            return gpu_operations
+
+        if gpu_operations and gpu_operation not in options:
```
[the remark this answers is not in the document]
tgaddair: Nice catch!  Fixed.

setup.py
tgaddair: In my mind I was thinking set notation, but removed since it's unclear.

setup.py
```diff
@@ -632,21 +632,29 @@ def get_common_options(build_ext):
...
+            return gpu_operations
+
+        if gpu_operation and gpu_operation not in options:
+            options_str = ', '.join(options)
+            raise DistutilsError(f'{op_variable_name}={gpu_operation} is invalid, '
+                                 f'supported values are: {options_str}.')
+
+        return gpu_operation
+
+    gpu_allreduce = get_gpu_op_variable('HOROVOD_GPU_ALLREDUCE', ['NCCL, MPI', 'DDL'])
```
nvcastet, line 655: typo here: 'NCCL', 'MPI', 'DDL' => caused the ppc64le builderror.
the 2 following lines have the same typos.

tgaddair: Good catch!  Fixed.

README.rst
```diff
@@ -52,8 +52,13 @@ about who's involved and how Horovod plays a role, read the LF AI `announcement
...
 .. contents::
 
+|
+
+Documentation
+-------------
 
-The full documentation and an API reference are published at https://horovod.readthedocs.io/en/latest.
+- `stable <https://horovod.readthedocs.io/en/stable>`_
+- `master <https://horovod.readthedocs.io/en/latest>`_
```
sblotner, line 61: Might want to call this "latest"

sblotner: Or just have the 1 link (not sure when someone would need to see the /stable version)

docs/summary.rst.patch
```diff
@@ -0,0 +1,10 @@
+55,58d54
```
sblotner, line 1: Should this be included?

README.rst
tgaddair: In practice, I think we actually want to direct users to /stable first (stable is the latest released version, correct?).  In a lot of cases, users get confused because they `pip install horovod`, but the docs do not match with the version they have installed (because latest is always the unreleased master branch).

Ideally, I think it would be good if we could have specific versioned sets of documentation like /0.19.2 to make it more clear for users.  For now, stable seems to be the closest proxy to that, but the name isn't very informative.  Maybe "released" or "production".

What do you think?

docs/summary.rst.patch
tgaddair: I don't think so, because the summary page is already part of Read The Docs, so it seems it would be kind of circular.  Do you feel it makes sense to include it there?

README.rst
sblotner: Ah, definitely want to point to the right version to reduce that confusion. Looking into how it works, I think /stable is the recommended version, as it's kept up to date with the highest release version (for tagged releases): https://docs.readthedocs.io/en/stable/versions.html

"Latest release" could work?

docs/summary.rst.patch
sblotner: I forgot that there is the automatic check to have the README and Summary page stay in sync (https://github.com/horovod/horovod/blob/master/docs/summary.rst.patch)

I agree that this content would be an exception and wouldn't need to be exactly the same between the two pages, but think the automatic sync check is useful in general, so maybe best to leave it in for now.

tgaddair: Agreed, definitely want to keep the check.  This just adds these lines as exceptions.

README.rst
tgaddair: Sounds good!  Changed to "Latest Release" instead.

EnricoMi: I agree, latest is a synonym for stable or latest release, master is a synonym for git head or latest source. Ideally, documentation of each released version should be available at horovod.readthedocs.io.

PR #1965, opened 2020-05-20
Blacklist hosts on failure, not _on_workers_recorded

EnricoMi:
Removes hosts from discovery as soon as worker fails. Allows for worker notification about scale-down through `HostsUpdatedInterrupt` in addition to existing `HorovodInternalError`.

### DIFF
--- a/horovod/run/elastic/discovery.py
+++ b/horovod/run/elastic/discovery.py
@@ -83,12 +83,13 @@ class HostManager(object):
         self._discovery = discovery
 
     def update_available_hosts(self):
-        # TODO(travis): also check for hosts removed from the blacklist in the future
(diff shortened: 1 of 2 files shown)


EnricoMi:
I presume the broadcast in `check_host_updates` in `horovod/common/elastic.py` will fail with an `HorovodInternalError` on a lost worker before it can handle the `HostsUpdatedInterrupt` anyway. Any benefit in this other than implementing the TODO?

EnricoMi:
This approach does not work for hosts with multiple slots as the first failed worker will blacklist the whole host and other worker's exit state won't be recorded: `host 127.0.0.1 has been blacklisted, ignoring exit from local_rank=0`, then `self._barrier.wait()` will never fire in  `registration._wait()`.

PR #1971, opened 2020-05-20
Fixing horovodrun test in gen-pipeline.sh

EnricoMi:
These three tests haven't been run for 10 months.

### DIFF
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -170,8 +170,7 @@ run_mpi_integration() {
         "horovodrun -np 2 -H localhost:2 python /horovod/examples/tensorflow_mnist.py"
       run_test "${test}" "${queue}" \
         ":terminal: Test Horovodrun (${test})" \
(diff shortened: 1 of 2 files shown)


EnricoMi:
There is no output from `mxnet_mnist.py`:
```
Creating network "buildkiteadbbbf4ed4314063a8f40591933ee2b9_default" with the default driver
localhost slots=2 > hostfile; horovodrun -np 2 -hostfile hostfile python /horovod/examples/mxnet_mnist.py
service                                                                         CONTAINER ID        STATUS
test-gpu-openmpi-py3_6-tf1_15_0-keras2_3_1-torch1_3_0-mxnet1_4_1-pyspark2_4_0   6bc937dd8b86        Exited (0) Less than a second ago
$ cd /var/lib/buildkite-agent/builds/buildkite-2x-gpu-g4-i-0e3d88377de9f6858-2/horovod/horovod
```
@tgaddair  Do you think this is right?

EnricoMi:
We need to use `bash -c "..."` here, fixing

Issue #1969, opened 2020-05-20
When network interface isn't specified, task processes fail to connect to driver.


aaron276h:
**Environment:**
1. Framework: (TensorFlow, Keras, PyTorch, MXNet): N/A
2. Framework version: N/A
3. Horovod version: 0.19.2
4. MPI version: N/A (Using Gloo Controller)
5. CUDA version: 10.0
6. NCCL version: 2.6.4
7. Python version: 3.6.10
8. OS and version: Ubuntu 18.04
9. GCC version: 4.8

**Bug report:**

When running multi-machine and not specifying a network interface, task processes fail to connect to driver. Note: using a GLOO controller.

Stack trace:

```
root@ip-172-31-37-52:/# horovodrun -np 2 -H localhost:1,172.31.35.37:1 -p 12345 --verbose ls
Filtering local host names.
Remote host found: 172.31.35.37
Checking ssh on all remote hosts.
SSH was successful into all the remote hosts.
Testing interfaces on all the hosts.
Launched horovod server.
Attempted to launch horovod task servers.
Waiting for the hosts to acknowledge.
----------------------------------------
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 37436)
Exception happened during processing of request from ('172.31.37.52', 52632)
----------------------------------------
Exception happened during processing of request from ('172.31.37.52', 52636)
Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/socketserver.py", line 654, in process_request_thread
    self.finish_request(request, client_address)
  File "/opt/conda/lib/python3.6/socketserver.py", line 364, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/opt/conda/lib/python3.6/socketserver.py", line 724, in __init__
    self.handle()
  File "/opt/conda/lib/python3.6/site-packages/horovod/run/common/util/network.py", line 105, in handle
    req = server._wire.read(self.rfile)
  File "/opt/conda/lib/python3.6/site-packages/horovod/run/common/util/network.py", line 79, in read
    message_len = struct.unpack('i', rfile.read(4))[0]
Traceback (most recent call last):
struct.error: unpack requires a buffer of 4 bytes
----------------------------------------
  File "/opt/conda/lib/python3.6/socketserver.py", line 654, in process_request_thread
    self.finish_request(request, client_address)
  File "/opt/conda/lib/python3.6/socketserver.py", line 364, in finish_request
    self.RequestHandlerClass(request, client_address, self)
[...]

PR #1977, opened 2020-05-22
Add integration tests for running the Horovod Cluster

EnricoMi:
In #1969 we learned that there are no integration tests with remote hosts. This adds integration tests that spin up Horovod clusters using localhost as remote hosts.

Fixes a bug introduced by elastic horovod #1849 (`task_fn`).


horovod/run/runner.py
```diff
@@ -105,7 +106,7 @@ def exec_command(command):
 
             ssh_successful_to_all_hosts = False
     if not ssh_successful_to_all_hosts:
-        exit(1)
```
EnricoMi, line 108: Any idea why there is this exit here? I hope the exception is equivalent.

PR #2009, opened 2020-06-04
Move native libraries build to cmake

nvcastet:
Some advantages of moving C++/C code to cmake:

- Incremental build (almost instantaneous build when developing and making small changes at a time)
- Parallel task build (much faster build)
- Separation of the build config phase with the build phase (less overhead for repeated builds)
- Reuse find_package modules provided by cmake for python, mpi, cuda, etc...
- Libraries can be built outside of the python build process (setup.py)
- Choice on the build system (make, ninja(should be faster), IDEs...)

Future work that could speed up more the build:
- Use of object libraries to reuse when possible object files between shared libraries (https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/Object-Library)


nvcastet:
@jeffdaily I have not ported the ROCm build code to cmake yet. I know AMD uses cmake for ROCm in other projects. What would you recommend to integrate easily here with cmake?

jeffdaily:
What is the rationale for moving to a CMake build?

jeffdaily:
Yes, AMD uses CMake for ROCm various projects.  Extensions to PyTorch such as horovod must invoke the hipify script prior to building, in order to make some CUDA-to-HIP name changes.  This is done in the existing horovod setup.py file, but could be invoked instead by a cmake command if ROCm PyTorch is detected.  The `is_cuda_available()` check you have now would report True for both CUDA and HIP (because for PyTorch, HIP "masquerades" as CUDA).  You would need an additional check to know whether CUDA was actually HIP.
`torch.version.cuda` is not None for CUDA, and 
`torch.version.hip` is not None for HIP.

jeffdaily:
I briefly reviewed your changes and you use `execute_process(COMMAND python -c ...` a lot.  Is there any concern that `python` might not point to the correct executable?

nvcastet:
@jeffdaily Thanks for the quick reply. I updated the PR top comment with the advantages.
Thanks for the tips, are there other cmake projects I could base myself on? Do you prefer to do the port yourself or you prefer that I do an initial version and have you review/test it?
For your python interpreter concern, i agree with you. I will use `${PYTHON_EXECUTABLE}` returned by the call to `find_package(PythonInterp "3.6" REQUIRED)`.

jeffdaily:
Perhaps a good, relatively small ROCm project would be the port of nccl to rccl.
https://github.com/ROCmSoftwarePlatform/rccl/blob/develop/CMakeLists.txt
This CMakeLists.txt demonstrates the use of building for our various gfx targets, as well as our recent transition from our hcc compiler to clang.

Otherwise, the logic of running the hipify script can be derived from what's currently in the setup.py for horovod.  It would need to be invoked using the python interpreter.

I'm okay if you make the first attempt at porting the ROCm stuff to CMake.  I anticipate helping, at the very least by building this on our hardware to verify since we don't have CI integrated with horovod at this time.

nvcastet:
@jeffdaily I made the ROCm changes. Could you check if it builds on your hardware?

jeffdaily:
There are some changes needed for ROCm.  Do I have permissions to push to this PR?

nvcastet:
@jeffdaily yes go ahead

jeffdaily:
I do not have permissions to push to your PR branch.  I suppose I could submit a PR against your fork+branch, but perhaps it's just easier to upload a diff file?  Let me know.
[hvd_add_cmake_rocm.diff.txt](https://github.com/horovod/horovod/files/5071727/hvd_add_cmake_rocm.diff.txt)

jeffdaily:
As a PR against your fork:
https://github.com/nvcastet/horovod/pull/27

jeffdaily:
In one of our tensorflow containers, I did try the tensorflow build, but it failed to find tensorflow.

jeffdaily:
I tried to build this PR with our ROCm TF container and it still failed to find tensorflow.

This is the output I see for the python executed code in FindTensorflow.cmake in our ROCm TF container.
```
2.2.0
/usr/local/lib/python3.6/dist-packages/tensorflow/include
-L/usr/local/lib/python3.6/dist-packages/tensorflow -l:libtensorflow_framework.so.2
-I/usr/local/lib/python3.6/dist-packages/tensorflow/include -D_GLIBCXX_USE_CXX11_ABI=1
```

nvcastet:
@jeffdaily It should work now for ROCM, could you retry?

jeffdaily:
@nvcastet can confirm.  Was able to build horovod for both TF and PT.  What is the make command for building the python packages?

CMakeLists.txt
```diff
@@ -0,0 +1,289 @@
...
+foreach(VAR in ITEMS HOROVOD_GPU_ALLREDUCE HOROVOD_GPU_ALLGATHER HOROVOD_GPU_BROADCAST HOROVOD_GPU_ALLTOALL)
+    if(DEFINED ${VAR})
+        string(SUBSTRING ${${VAR}} 0 1 ${VAR})
+        convert_to_ascii_dec(ASCII_DEC ${${VAR}})
+        add_definitions(-D${VAR}=${ASCII_DEC})
+    endif()
+endforeach()
+
+# PYTHON
+find_package(PythonInterp "3.6" REQUIRED)
```
tgaddair, line 100: It seems `PythonInterp` is deprecated.  With this one, it was unable to resolve virtualenv paths.  Instead, using `find_package(Python3 ...)` fixed the issue in more recent versions of CMake.

CMakeLists.txt
```diff
@@ -0,0 +1,289 @@
...
+# Get the latest version intalled
+execute_process(COMMAND bash -c "echo $PATH | tr ':' '\n' | xargs -I{} find {} -regex '.*/python[2-3]?\\.?[0-9]*' -printf '%f\n'  | sort -r | head -1"
+                OUTPUT_VARIABLE PY_EXE OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET)
+if (NOT PY_EXE)
+     set(PY_EXE ${PYTHON_EXECUTABLE})
```
tgaddair, line 105: If we go with the above suggestion:

```
set(PY_EXE ${Python3_EXECUTABLE})
```

CMakeLists.txt
```diff
@@ -0,0 +1,289 @@
...
+# Add architecture specific optimization flags
+set(ARCH_FLAGS "-mf16c" "-mavx" "-mfma")
+set_build_arch_flags("${ARCH_FLAGS}")
+# Specify Horovod exports
+if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-exported_symbols_list,${CMAKE_SOURCE_DIR}/horovod.exp")
+else()
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--version-script=${CMAKE_SOURCE_DIR}/horovod.lds")
+endif()
+set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-Bsymbolic-functions -Wl,-z,relro")
```
tgaddair, line 79: I cannot find these on macOS, but if I omit them I get a different error:

```
ld: warning: cannot export hidden symbol std::__1::thread::thread<void (horovod::common::ThreadPool::*)(), horovod::common::ThreadPool*, void>(void (horovod::common::ThreadPool::*&&)(), horovod::common::ThreadPool*&&) from CMakeFiles/pytorch.dir/__/common/thread_pool.cc.o
ld: warning: cannot export hidden symbol std::__1::thread::thread<void (horovod::common::TimelineWriter::*)(), horovod::common::TimelineWriter*, void>(void (horovod::common::TimelineWriter::*&&)(), horovod::common::TimelineWriter*&&) from CMakeFiles/pytorch.dir/__/common/timeline.cc.o
```

Any thoughts?

nvcastet:
@tgaddair Could you retry with latest?

tgaddair (approved):
LGTM!  Working in every scenario I could think to try.  Very nice work, this is a monumentous improvement!

kf-cuanschutz:
I am trying to recompile horovod for the architechture gfx908, what is the exact specification I should use?
It keeps compiling it for "gfx803". Also, how can I point it to my RCCL installation?
Here is my command so far:

```bash
__HIP_PLATFORM_AMD__=1  HOROVOD_CPU_OPERATIONS=MPI PYTORCH_ROCM_ARCH="gfx908" AMDGPU_TARGETS="gfx908" GFX_AR
CH="gfx908"  HOROVOD_WITH_PYTORCH=1 HOROVOD_GPU_OPERATIONS=NCCL HOROVOD_GPU=ROCM HOROVOD_ROCM_HOME=$ROCM_PATH HOROVOD_WITH_MPI=1 pip install --no-cache-dir  -
-target=$PIP_INSTALL_DIR  -v git+https://github.com/horovod/horovod.git@refs/pull/3588/head --upgrade >& make_horovod_update_V12.log &
```

PR #2057, opened 2020-06-24
Fixed PyTorch and MXNet tests to correctly compare tensor elements

### LINKED ISSUE #2055: PyTorch and MXNet unit tests are missing absolute values
Unit tests that compare "actual" and "expected" tensors use a pattern of subtracting one tensor from the other and taking the max values:

```
max_difference = actual.data.sub(expected).max()
```

This difference is then compared against a threshold (for floating point values):

```
assert max_difference <= threshold, 'hvd.allreduce produces incorrect results'
```

However, because this value does not take the absolute value, it is possible that all values are negative, and the "max value" reported is also negative, resulting in the test passing when it should fail.

cc @romerojosh

tgaddair:
Fixes #2055.

### DIFF
--- a/test/test_mxnet.py
+++ b/test/test_mxnet.py
@@ -20,7 +20,7 @@ import numpy as np
 import mxnet as mx
 
 from mxnet.base import MXNetError
-from mxnet.test_utils import same
+from mxnet.test_utils import almost_equal, same
 
 import horovod.mxnet as hvd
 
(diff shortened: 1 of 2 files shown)


romerojosh:
Great work @tgaddair, this error checking looks much better. Should we also use `torch.allclose` for the PyTorch sync BN tests as well?

tgaddair:
Good point, @romerojosh.  Fixed.

PR #2058, opened 2020-06-24
Added reset limit for elastic mode to prevent introducing excessive variance

### LINKED ISSUE #1947: Add reset limit when running in elastic mode
The user should be able to optionally specify a hard limit on the number of reset events that can occur, with an optional sliding window.

For example:

```
horovodrun --reset-limit 3
```

Or:

```
horovodrun --reset-limit 3 --reset-limit-window-duration 60
```

In the latter case, we will only fail of there are 3 reset events in a minute (60 seconds), while in the former, we will fail if we ever see 3 reset events over the lifetime of the job.

tgaddair:
Fixes #1947.

### DIFF
--- a/horovod/run/elastic/registration.py
+++ b/horovod/run/elastic/registration.py
@@ -159,7 +159,7 @@ class WorkerStateRegistry(object):
             return
 
         # Check that we have already reset the maximum number of allowed times
-        if self._reset_limit is not None and self._reset_count > self._reset_limit:
+        if self._reset_limit is not None and self._reset_count >= self._reset_limit:
             logging.error('reset count {} has exceeded limit {} -> stop running'
                           .format(self._reset_count, self._reset_limit))
             self._driver.stop(error_message=constants.RESET_LIMIT_EXCEEDED_MESSAGE.format(self._reset_limit))
--- a/horovod/run/elastic/registration.py
+++ b/horovod/run/elastic/registration.py
@@ -18,6 +18,8 @@ import threading
 
 from collections import defaultdict
 
+from horovod.run.elastic import constants
+
 READY = 'READY'
 SUCCESS = 'SUCCESS'
 FAILURE = 'FAILURE'
@@ -135,9 +137,9 @@ class WorkerStateRegistry(object):
 
         # Check that we have already reset the maximum number of allowed times
         if self._reset_limit is not None and self._reset_count >= self._reset_limit:
-            logging.info('reset count {} has reached limit {} -> stop running'
-                         .format(self._reset_count, self._reset_limit))
-            self._driver.stop()
+            logging.error('reset count {} has reached limit {} -> stop running'
+                          .format(self._reset_count, self._reset_limit))
+            self._driver.stop(error_message=constants.RESET_LIMIT_EXCEEDED_MESSAGE.format(self._reset_limit))
             return
 
         # Check for success state, if any process succeeded, shutdown all other processes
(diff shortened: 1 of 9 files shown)


horovod/run/elastic/registration.py
```diff
@@ -100,7 +104,8 @@ def _record_state(self, host, slot, state):
                 logging.info('record state: {}[{}] = {}'.format(host, slot, state))
                 self._states[key] = state
                 self._workers[state].add(key)
-                rendezvous_id = self._rendezvous_id
+
+            rendezvous_id = self._rendezvous_id
```
tgaddair, line 108: @EnricoMi note that this is a bugfix introduced in #1966.  Not sure if it's related to some of the errors we've been seeing, but worth being aware of.

horovod/run/elastic/registration.py
```diff
@@ -130,6 +135,13 @@ def _action(self):
     def _on_workers_recorded(self):
         logging.info('all {} workers recorded'.format(self.size()))
 
+        # Check that we have already reset the maximum number of allowed times
```
EnricoMi, line 138: I think we can reset once too often and then fail here even if all workers succeed if we check the limit here. We should check it right before we increment the counter further down, and then for `≥`.

horovod/spark/runner.py
```diff
@@ -307,6 +307,7 @@ def run_elastic(fn, args=(), kwargs={}, num_proc=None, min_np=None, max_np=None,
         elastic_timeout: Timeout for elastic initialisation after re-scaling the cluster.
                        If not set, falls back to `HOROVOD_ELASTIC_TIMEOUT` environment variable value.
                        If it is not set as well, defaults to 600 seconds.
+        reset_limit: Maximum number of resets after which the job is terminated.
```
EnricoMi, line 310: `Maximum number of resets before the job is terminated.`, 'after' sounds like right after the N-th reset the job is terminated, so we have N-1 retries / resets.

test/integration/elastic_common.py
```diff
@@ -235,3 +239,22 @@ def test_min_hosts_timeout(self, mock_get_min_start_hosts):
         message = 'Horovod detected that one or more processes exited with non-zero status'
         with pytest.raises(RuntimeError, match=message):
             self._run(discovery_schedule, exit_schedule=exit_schedule, np=4, min_np=4)
+
+    @mock.patch('horovod.run.elastic.driver.ELASTIC_TIMEOUT_SECS', 1)
+    @mock.patch('horovod.run.elastic.driver.DISCOVER_HOSTS_FREQUENCY_SECS', 0.01)
+    @mock.patch('horovod.run.gloo_run._get_min_start_hosts', return_value=1)
+    def test_reset_limit(self, mock_get_min_start_hosts):
+        discovery_schedule = [
+            (0, ['localhost:{}'.format(2)]),
```
EnricoMi, line 248: why the `format(2)`?

horovod/run/elastic/registration.py
tgaddair: Yes, good point.  Fixed.

horovod/spark/runner.py
tgaddair: In reality, there is a reset at the beginning when the job is started, but that is an implementation detail.  From the user perspective, we consider a reset to happen when a worker is added / removed after the initial registration.  So if I set `--reset-limit 0`, for example, then my job can never resize.  If I set it to `1`, it can resize exactly once, etc.  I'll update the docs for this param.

test/integration/elastic_common.py
tgaddair: Copy-paste ;)

PR #2059, opened 2020-06-25
Removed num_hosts from settings to fix network interface discovery for elastic mode

tgaddair:
This change addresses an issue when starting an elastic job with more than one initial host, and no specified NIC, which leads to an error during network interface discovery.


horovod/run/common/util/settings.py
```diff
@@ -59,13 +59,13 @@ def __init__(self, num_proc=None, verbose=0, ssh_port=None, extra_mpi_args=None,
 
 
 class Settings(BaseSettings):
-    def __init__(self, num_hosts=None, hosts=None, **kwargs):
+    def __init__(self, hosts=None, host_names=None, **kwargs):
```
EnricoMi, line 62: isn't `hosts` and `host_names` redundant? Can you extract the latter from the former? In which situation would you provide one but not the other?

The only spot where `host_names` is used is `mpi_run`. In `horovod.run.runner` we use `parse_hosts_and_slots` to get all host names from `hosts`. I think we should do the same in `mpi_run`.

tgaddair: Good point.  Fixed.

PR #2063, opened 2020-06-26
Fix test_run.py after removing num_hosts from settings

EnricoMi:
This fixes broken tests from removing `num_hosts` from settings (#2059) and `PYTHONPATH` injection into environment variables for MPI (#2038). Latter test were broken when run without `PYTHONPATH` set, which is true on buildkite. Adds specific tests for `PYTHONPATH` and `PATH` env injection and removes asserts from other more general tests.

Fixes silent failure of pytest so that above errors surfaces in buildkite unit tests again.

### DIFF
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -110,7 +110,7 @@ run_mpi_pytest() {
   # pytests have 4x GPU use cases and require a separate queue
   run_test "${test}" "${queue}" \
     ":pytest: Run PyTests (${test})" \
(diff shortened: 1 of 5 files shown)


EnricoMi:
Commit c9e7ba0 fixes `test_get_available_devices` by changing the `local-cluster` setup. It starts only a single worker which gets all GPUs (usually, in a Spark cluster there is only a single worker per host). Then, a single executor gets all CPUs and GPUs and each task only a single CPU and GPU on that executor. This should still sufficiently test the GPU script wiring into `get_available_devices` in `task_info.py`.

PR #2086, opened 2020-07-08
Updated min supported TensorFlow to 1.12, PyTorch to 1.2

tgaddair:


tgaddair:
Good catch, @romerojosh!  Fixed.

PR #2101, opened 2020-07-13
Use pthread_yield in torch WaitAndClear function.

romerojosh:

## Description
During some recent development, I noticed that the existing `WaitAndClear` function used in the PyTorch implementation introduces arbitrary 1 millisecond sleeps. This sleep can occur on any call to synchronize handles for PyTorch ops, adding a lot of additional latency. This can be especially problematic in cases where Horovod ops are included during the forward/backward compute graph (for example, when using sync BN). To remedy this, this PR replaces the sleep with a `pthread_yield()` as suggested by @tgaddair.

### DIFF
--- a/horovod/torch/mpi_ops_v2.cc
+++ b/horovod/torch/mpi_ops_v2.cc
@@ -16,6 +16,7 @@
 
 #include <chrono>
 #include <memory>
+#include <pthread.h>
 #include <thread>
 #include <torch/extension.h>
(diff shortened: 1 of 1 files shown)


tgaddair:
@alsrgv this may improve performance of SyncBatchNorm in PyTorch.

alsrgv:
I'll give it a try, thanks! :-)

PR #2102, opened 2020-07-13
Updated changelog and TSC members

tgaddair:
@apeforest is going to be stepping back from maintaining Horovod + MXNet integration, and @eric-haibin-lin will be taking on his role on the TSC going forward.


eric-haibin-lin:
Thanks!

PR #2124, opened 2020-07-21
Enable keras estimator to accept modelCheckPoint callback

abditag2:

## Description

Fixes # (issue).

## Review process to land 

### DIFF
--- a/horovod/keras/callbacks.py
+++ b/horovod/keras/callbacks.py
@@ -155,18 +155,13 @@ class LearningRateWarmupCallback(_impl.LearningRateWarmupCallbackImpl, keras.cal
 
 
 class ReturnedModelCheckpoint(keras.callbacks.ModelCheckpoint):
-    def __init__(self, filepath=None,
+    def __init__(self,
                  monitor='val_loss',
                  verbose=0,
                  save_best_only=False,
                  mode='auto',
                  save_freq='epoch'):
-        """
-        Note that `filepath` argument is not necessary. Set it only if you care about where the
-        model checkpoints are stored during the training. If filepath is not provided,
-        KerasEstimator will use a temporary directory for the checkpoints.
-        """
-        super(ReturnedModelCheckpoint, self).__init__(filepath=filepath,
+        super(ReturnedModelCheckpoint, self).__init__(filepath=None,
                                                       monitor=monitor,
                                                       verbose=verbose,
                                                       save_best_only=save_best_only,
--- a/horovod/keras/callbacks.py
+++ b/horovod/keras/callbacks.py
@@ -158,7 +158,7 @@ class ReturnedModelCheckpoint(keras.callbacks.ModelCheckpoint):
     def __init__(self, filepath=None,
                  monitor='val_loss',
                  verbose=0,
-                 save_best_only=True,
+                 save_best_only=False,
                  mode='auto',
                  save_freq='epoch'):
         """
--- a/horovod/keras/callbacks.py
+++ b/horovod/keras/callbacks.py
@@ -154,17 +154,16 @@ class LearningRateWarmupCallback(_impl.LearningRateWarmupCallbackImpl, keras.cal
                                                          steps_per_epoch, verbose, initial_lr)
 
 
-class ReturnedModelCheckpoint(keras.callbacks.ModelCheckpoint):
+class BestModelCheckpoint(keras.callbacks.ModelCheckpoint):
     def __init__(self,
                  monitor='val_loss',
                  verbose=0,
-                 save_best_only=False,
                  mode='auto',
                  save_freq='epoch'):
-        super(ReturnedModelCheckpoint, self).__init__(filepath=None,
-                                                      monitor=monitor,
-                                                      verbose=verbose,
-                                                      save_best_only=save_best_only,
-                                                      save_weights_only=False,
-                                                      mode=mode,
-                                                      save_freq=save_freq)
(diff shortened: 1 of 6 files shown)


horovod/spark/keras/estimator.py
```diff
@@ -160,6 +160,8 @@ class KerasEstimator(HorovodEstimator, KerasEstimatorParamsReadable,
 
     custom_objects = Param(Params._dummy(), 'custom_objects', 'custom objects')
     _keras_pkg_type = Param(Params._dummy(), '_keras_pkg_type', 'keras package type')
+    model_checkpoint_callback_fn = Param(Params._dummy(), 'model_checkpoint_callback_fn',
```
tgaddair, line 163: I think `checkpoint_callback` would be a better name, to be consistent with PyTorch Lightning.

horovod/spark/keras/remote.py
```diff
@@ -323,3 +329,10 @@ def fn(tf, keras):
         config.intra_op_parallelism_threads = 1
         keras.backend.set_session(tf.Session(config=config))
     return fn
+
+
+def _model_checkpoint_callback_fn():
```
tgaddair, line 334: I think we can simplify this a bit:

1. Change it from accepting a function to just accepting something that adheres to the `ModelCheckpoint` API.
2. Use the `checkpoint_callback.filepath` property to figure out where to read the saved model from (note that this filepath might be a template string: https://www.tensorflow.org/tutorials/keras/save_and_load#checkpoint_callback_options).
3. If not specified, create the default `ModelCheckpoint(ckpt_file)`.

abditag2: I don't think that the filepath should come from the user because they have to be responsible to figure out the appropriate temp folder with correct permissions. It also increases the chances of errors if users specify a fixed path, and the checkpoints from previous runs are present on that path. If a user wants copies of the checkpoint, they can add extra ModelCheckpoint callbacks using the callbacks parameter. 

If we agree that the user should not pass the filepath arg, I am not sure how to avoid not returning a function here as we want the user to specify the parameters of ModelCheckpoint callback.

tgaddair: There are many cases where the user will want to keep their checkpoints in a particular location.  Needing to provide two separate callbacks that do the same thing in different directories is a bad user experience.

If you want to support the user not having to specify the path (because they don't want to keep the checkpoint files around), then you can solve this by creating a subclass of `ModelCheckpoint` for this purpose:

```
KerasEstimator(checkpoint_callback=hvd.callbacks.ReturnedModelCheckpoint(
    monitor='val_acc',
    mode='max',
    save_best_only=True), ...)
```

Nice thing about this approach is (1) it's flexible if the user wants to provide an alternative location to write checkpoints, and (2) you can prevent them from setting unnecessary things like `filepath` and `save_weights_only`.

horovod/keras/callbacks.py
```diff
@@ -152,3 +152,24 @@ def __init__(self, warmup_epochs=5, momentum_correction=True, steps_per_epoch=No
         """
         super(LearningRateWarmupCallback, self).__init__(K, warmup_epochs, momentum_correction,
                                                          steps_per_epoch, verbose, initial_lr)
+
+
+class ReturnedModelCheckpoint(keras.callbacks.ModelCheckpoint):
+    def __init__(self, filepath=None,
+                 monitor='val_loss',
+                 verbose=0,
+                 save_best_only=True,
```
tgaddair, line 161: Is `save_best_only` needed here?  It seems that with a specific file path provided, it always saves the best only.

horovod/keras/callbacks.py
```diff
@@ -152,3 +152,24 @@ def __init__(self, warmup_epochs=5, momentum_correction=True, steps_per_epoch=No
         """
         super(LearningRateWarmupCallback, self).__init__(K, warmup_epochs, momentum_correction,
                                                          steps_per_epoch, verbose, initial_lr)
+
+
+class ReturnedModelCheckpoint(keras.callbacks.ModelCheckpoint):
+    def __init__(self, filepath=None,
```
tgaddair, line 158: I think we should not let the user specify `filepath`.  If they want a custom directory, they can just use the regular `ReturnedModelCheckpoint`.  Otherwise, there is no practical difference between this and `ModelCheckpoint`.

horovod/keras/callbacks.py
abditag2: Without `save_best_only`, in every epoch or whatever the monitoring interval is it overwrites the previous checkpoint. With it, it will only overwrite if the current checkpoint has better metric than the previously saved checkpoint. 

I think we need it.

horovod/keras/callbacks.py
abditag2: sure

horovod/keras/callbacks.py
tgaddair: Makes sense, but is there ever a case where the user would want to set `save_best_only=False` using this class?

abditag2: I cannot think of a use case off the top of my head. But I think, it is a better choice to conform with the defaults of the `ModelCheckpoint` args to avoid confusion for the users.

abditag2: I am open to change this if you have a strong preference.

tgaddair: My concern is that if we expose `save_best_only` to the user, but it doesn't do anything, that would create even more confusion for the user.  I think for now we may want to remove it, and always set `save_best_only=True` when initializing the parent constructor.

PR #2140, opened 2020-07-24
Add label_shapes parameter to KerasEstimator and TorchEstimator

abditag2:

## Description

Fixes # (issue).

## Review process to land 

### DIFF
--- a/horovod/spark/common/util.py
+++ b/horovod/spark/common/util.py
@@ -178,26 +178,31 @@ def check_shape_compatibility(metadata, feature_columns, label_columns,
                              'provided label shapes count {outputs}'
                              .format(labels=label_count, outputs=len(label_shapes)))
 
+    label_count = len(label_columns)
     if output_shapes is not None:
-        label_count = len(label_columns)
         if label_count != len(output_shapes):
             raise ValueError('Label column count {labels} must equal '
                              'model outputs count {outputs}'
                              .format(labels=label_count, outputs=len(output_shapes)))
 
-    if output_shapes is not None and label_shapes is None:
-        label_count = len(label_columns)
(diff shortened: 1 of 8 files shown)


horovod/spark/common/util.py
```diff
@@ -171,24 +171,38 @@ def check_shape_compatibility(metadata, feature_columns, label_columns,
...
+            target_size = abs(np.prod(target_shape))
+            if _col_size != target_size:
                 raise ValueError('Label column \'{col}\' with size {label} must equal that of the '
-                                 'model output at index {idx} with size {output}'
-                                 .format(col=col, label=col_size, idx=idx, output=output_size))
+                                 'model output and label shape at index {idx} with size {output}'
+                                 .format(col=_col, label=_col_size, idx=_idx, output=target_size))
+
+    if output_shapes is not None:
+        _check_label_cols_size(output_shapes)
```
tgaddair, line 202: Seems this should only be done if `label_shapes is None`, right?  Otherwise, we will be forcing the model output size to equal the label size, which may not be correct (for these custom loss functions).

horovod/spark/torch/remote.py
```diff
@@ -266,9 +267,16 @@ def transform_outputs(outputs, labels):
 
                         # reshape labels to match the output shape of the model
                         if hasattr(outputs[0], 'shape'):
-                            labels = [label.reshape(output.shape)
-                                      if output.shape.numel() == label.shape.numel() else label
-                                      for label, output in zip(labels, outputs)]
+                            # If label_shapes parameter is not provided, reshape the label columns
```
tgaddair, line 270: Can move this comment into the `else` branch so it's closer to the code it's describing.

PR #2143, opened 2020-07-28
Add alltoall support for TensorFlow, PyTorch, and MXNet.

romerojosh:

## Description
This PR adds support for the all-to-all collective in Horovod. The all-to-all collective can be described as a combination of a scatter and gather, where each worker will scatter a tensor to each worker, while also gathering scattered data from other workers. This type of collective communication can arise in model-parallel training strategies. 

The `hvd.alltoall` function introduced in this PR is `hvd.alltoall(tensor, splits=None)`,
where `tensor` is a multi-dimensional tensor of data to scattered and `splits` is an optional 1D tensor of integers with length equal to the number of workers, describing how to split and distribute `tensor`. `splits` is applied along the first dimension of tensor. If `splits` is not provided, an equal splitting is assumed, where the first dimension is divided by the number of workers. 

To better illustrate this, here is a diagram of three workers distributing a 2D tensor using `hvd.alltoall` with default (equal) splits:

a2a_equal

Alternatively, here is a diagram of three workers distributing a 2D tensor using `hvd.alltoall` with user-provided `splits` tensors:

a2a_unequal

The implementation supports TensorFlow, PyTorch, and MXNet using the MPI backend, the CUDA-aware MPI backend via `HOROVOD_GPU_ALLTOALL=MPI`, and the NCCL backend via `HOROVOD_GPU_ALLTOALL=NCCL`. 

Some limitations and points for discussion:
1. The implementation currently requires using the MPI-based controller. The reason for this limitation is that an all-to-all communication of the `splits` tensors is done before the MPI/NCCL all-to-all operation in order to determine the receive-side splits/displacements for the output (see `Controller::AlltoallGetRecvSplits` and usage in `AlltoallOp::PrepareOutputAndParams` for an example of this). The Gloo controller is currently unsupported since there is no all-to-all collective available in Gloo to be used in the controller for this purpose, however I think it is possible to implement one. 
An alternative implementation for this would be to have the coordinator gather all the `splits` arrays, similar to how for allgather the coordinator gathers all the first dimensions across the workers. However, I opted against this approach since unlike for allgather where only a single integer per worker is gathered, the coordinator would need to gather the full `splits` array for each worker for all-to-all, yielding a lot of unnecessary communication of metadata to the coordinator rank, growing with the square of the number of workers. 

2. For the NCCL backend, this implementation requires the new `ncclSend/ncclRecv` functions introduced in NCCL 2.7.0. Currently, Horovod testing uses builds with an older unsupported version of NCCL and the new tests I have implemented will be skipped in the CI in those cases. It would be good to enable testing with a supported version of NCCL to get coverage of this feature before landing this PR.

### DIFF
--- a/horovod/common/gloo/gloo_controller.cc
+++ b/horovod/common/gloo/gloo_controller.cc
@@ -266,6 +266,13 @@ void GlooController::Bcast(void* buffer, size_t size, int root_rank,
   gloo::broadcast(opts);
 }
 
+void GlooController::AlltoallGetRecvSplits(const std::vector<int32_t>& splits,
+                                           std::vector<int32_t>& recvsplits) {
+    throw std::runtime_error(
+        "GlooController::AlltoallGetRecvcounts not yet implemented. Use Horovod "
+        "with MPI to use this functionality.");
+}
+
 void GlooController::Barrier(Communicator communicator) {
   gloo::BarrierOptions opts(gloo_context_.GetGlooContext(communicator));
   gloo::barrier(opts);
--- a/horovod/common/gloo/gloo_controller.cc
+++ b/horovod/common/gloo/gloo_controller.cc
@@ -269,7 +269,7 @@ void GlooController::Bcast(void* buffer, size_t size, int root_rank,
 void GlooController::AlltoallGetRecvSplits(const std::vector<int32_t>& splits,
                                            std::vector<int32_t>& recvsplits) {
     throw std::runtime_error(
-        "GlooController::AlltoallGetRecvcounts not yet implemented. Use Horovod "
+        "GlooController::AlltoallGetRecvSplits not yet implemented. Use Horovod "
         "with MPI to use this functionality.");
 }
 
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -147,6 +147,7 @@ OperationManager* CreateOperationManager(HorovodGlobalState& state) {
   std::vector<std::shared_ptr<AllgatherOp>> allgather_ops;
   std::vector<std::shared_ptr<BroadcastOp>> broadcast_ops;
   std::vector<std::shared_ptr<AllreduceOp>> adasum_ops;
+  std::vector<std::shared_ptr<AlltoallOp>> alltoall_ops;
 
 #if HAVE_MPI && HAVE_GPU
   if (mpi_context.IsEnabled()) {
@@ -172,6 +173,11 @@ OperationManager* CreateOperationManager(HorovodGlobalState& state) {
 #endif
     allgather_ops.push_back(std::shared_ptr<AllgatherOp>(
         new MPIHierarchicalAllgather(&mpi_context, &state)));
+
+#if HOROVOD_GPU_ALLTOALL == 'M'
+    alltoall_ops.push_back(std::shared_ptr<AlltoallOp>(
+        new MPI_GPUAlltoall(&mpi_context, &gpu_context, &state)));
+#endif
   }
 #endif
 
@@ -188,6 +194,13 @@ OperationManager* CreateOperationManager(HorovodGlobalState& state) {
 #if HAVE_NCCL && HOROVOD_GPU_ALLGATHER == 'N'
   allgather_ops.push_back(std::shared_ptr<AllgatherOp>(
       new NCCLAllgather(&nccl_context, &gpu_context, &state)));
+#endif
+
+#if HAVE_NCCL && HOROVOD_GPU_ALLTOALL == 'N'
+#if HAVE_MPI
+  alltoall_ops.push_back(std::shared_ptr<AlltoallOp>(
+      new NCCLAlltoall(&nccl_context, &mpi_context, &gpu_context, &state)));
+#endif
 #endif
 
 #if HAVE_GLOO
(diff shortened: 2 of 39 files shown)


horovod/common/ops/nccl_operations.h
```diff
@@ -19,6 +20,9 @@
 
 #if HAVE_CUDA
 #include <nccl.h>
+#if NCCL_VERSION_CODE >= 2700
```
tgaddair, line 23: Should we use the [NCCL_VERSION](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/comms.html#c.ncclGetVersion) macro here?

```
if NCCL_VERSION_CODE >= NCCL_VERSION(2, 7, 0)
```

horovod/common/operations.cc
```diff
@@ -190,6 +196,11 @@ OperationManager* CreateOperationManager(HorovodGlobalState& state) {
       new NCCLAllgather(&nccl_context, &gpu_context, &state)));
 #endif
 
+#if HAVE_NCCL && HOROVOD_GPU_ALLTOALL == 'N'
```
tgaddair, line 199: An interesting case to consider: if the user sets `HOROVOD_GPU_OPERATIONS=NCCL`, then they get `HOROVOD_GPU_ALLTOALL=NCCL` for free.  But if their NCCL version < 2.7, this creates a runtime error.

Alternatively, we could fall back to using MPI All-to-All instead of raising the error, but then users might get poor performance without knowing why (particularly if they specified `HOROVOD_GPU_ALLTOALL=NCCL` explicitly).

I think that overall the way you have it implemented is the best solution, since there is a workaround for users who cannot upgrade NCCL for whatever reason: instead of using `HOROVOD_GPU_OPERATIONS`, set `HOROVOD_GPU_ALLREDUCE` and omit `HOROVOD_GPU_ALLTOALL`.

Maybe we can add some wording to that effect in the exception message when the NCCL version is too low for GPU All-to-All?

test/test_tensorflow.py
```diff
@@ -1374,6 +1376,10 @@ def test_horovod_broadcast_grad_gpu(self):
             # Skip if compiled with CUDA but without HOROVOD_GPU_OPERATIONS.
             self.skipTest("Not compiled with HOROVOD_GPU_OPERATIONS")
 
+        # This test does not apply if using gloo controller
+        if hvd.gloo_enabled():
```
tgaddair, line 1380: Allgather GPU does not work on Gloo?

horovod/common/gloo/gloo_controller.cc
```diff
@@ -266,6 +266,13 @@ void GlooController::Bcast(void* buffer, size_t size, int root_rank,
   gloo::broadcast(opts);
 }
 
+void GlooController::AlltoallGetRecvSplits(const std::vector<int32_t>& splits,
+                                           std::vector<int32_t>& recvsplits) {
+    throw std::runtime_error(
+        "GlooController::AlltoallGetRecvcounts not yet implemented. Use Horovod "
```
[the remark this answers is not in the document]
romerojosh: Good catch! This was leftover from an earlier version of the code. Fixed.

horovod/common/ops/gpu_operations.cc
```diff
@@ -143,5 +143,15 @@ bool GPUBroadcast::Enabled(const ParameterManager& param_manager,
   return entries[0].device != CPU_DEVICE_ID;
 }
 
+GPUAlltoall::GPUAlltoall(GPUContext* context,
+		         HorovodGlobalState* global_state)
+    : AlltoallOp(global_state), gpu_context_(context), gpu_op_context_(context, global_state) {}
+
+bool GPUAlltoall::Enabled(const ParameterManager& param_manager,
+		          const std::vector<TensorTableEntry>& entries,
```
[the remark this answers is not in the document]
romerojosh: Fixed.

horovod/common/ops/nccl_operations.h
romerojosh: Yes, this seems more fool-proof. Made the change. On a related note, I do still use `2700` explicitly for skipping tests based on the output of `nccl_built()`.

horovod/common/operations.cc
romerojosh: This aligns with my thinking as well. I thought about having it fallback to MPI but I'd rather have it explicitly break and inform users so they know to update NCCL, rather than silently operating with less performance than expected. 

I beefed up the error message here:
https://github.com/romerojosh/horovod/blob/08116280bb26d9a74bc33f28e90ec47fdece177b/horovod/common/ops/nccl_operations.cc#L589-L593

test/test_tensorflow.py
romerojosh: Nice catch. Got a little overzealous with the Gloo test skips it seems :). Removed.

romerojosh:
Thanks @tgaddair! Before landing, do we want to get a build with NCCL 2.7.x under CI first (and if so, could you help me figure out how we want to approach it)? At the moment, the NCCL backed all-to-all paths are not being tested.

Beyond this, I suggest we consider enabling Gloo controller support in a follow up PR once #2147 lands, and before cutting 0.20.0.

tgaddair:
Good point, let me see if I can get NCCL 2.7 in there for the 10.1 tests.

I'll create an issue to track the Gloo controller.

romerojosh:
Rebased to enable the NCCL all-to-all tests, but noticed that in the CI, most TensorFlow GPU tests are skipped. I just posted #2151 to fix this and will rebase again once it is merged to get a full set of working tests.

romerojosh:
Tests after rebasing with #2151 are passing so we are good to go. Will land this now.

PR #2147, opened 2020-07-28
Updated Gloo to support alltoall collective operation

### LINKED ISSUE #2148: Add Gloo support for All-to-All
#2143 adds MPI and NCCL support for all-to-all.  Gloo recently added an op for all-to-all, but our current submodule ref is behind.  We should update Gloo to the latest master, then add support for Gloo when using all-to-all.

tgaddair:
Fixes #2148.

### DIFF
--- a/test/test_mxnet.py
+++ b/test/test_mxnet.py
@@ -665,10 +665,6 @@ class MXTests(unittest.TestCase):
         rank = hvd.rank()
         size = hvd.size()
 
-        # This test does not apply if using gloo controller
-        if hvd.gloo_enabled():
-            self.skipTest("Alltoall currently does not support Gloo controller.")
-
         # This test does not apply if NCCL version < 2.7.0
         if hvd.nccl_built() and hvd.nccl_built() < 2700:
             self.skipTest("NCCL-based Alltoall requires NCCL version >= 2.7.0.")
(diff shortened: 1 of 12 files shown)


test/test_mxnet.py
```diff
@@ -634,10 +634,6 @@ def test_horovod_alltoall(self):
         rank = hvd.rank()
         size = hvd.size()
 
-        # This test does not apply if using gloo controller
-        if hvd.gloo_enabled():
```
romerojosh, line 638: There are several other alltoall tests below this that still contain this skip for gloo. Please remove from the remaning alltoall tests.

test/test_tensorflow.py
```diff
@@ -1423,10 +1423,6 @@ def test_horovod_alltoall_cpu(self):
         rank = hvd.rank()
         size = hvd.size()
 
-        # This test does not apply if using gloo controller
-        if hvd.gloo_enabled():
```
romerojosh, line 1427: There are several other alltoall tests below this that still contain this skip for gloo. Please remove from the remaining alltoall tests.

test/test_torch.py
```diff
@@ -891,10 +891,6 @@ def test_horovod_alltoall(self):
         rank = hvd.rank()
         size = hvd.size()
 
-        # This test does not apply if using gloo controller
-        if hvd.gloo_enabled():
```
romerojosh, line 895: There are several other alltoall tests below this that still contain this skip for gloo. Please remove from the remaining alltoall tests.

test/test_tensorflow.py
tgaddair: Good catch, not sure how I missed all those.  Fixed!

PR #2155, opened 2020-08-03
Fix mxnet 2 nightly build 

eric-haibin-lin:

## Description

MXNet nightly (version 2.0) made a few changes: 
- mxnet-mkl and mxnet-cuxxmkl binaries are no longer published. mkl is turned on by default. 
- `mx.gluon.parameter.ParameterDict` is dropped in favor of plain python `dict`
- `Parameter.name` is not longer unique globally unique. This means that we can no longer use `param.name` as the key for collective operations. Instead, this PR reverts back to using the index in the parameter list. 

## Review process to land 


horovod/mxnet/__init__.py
```diff
@@ -100,18 +100,21 @@ def __init__(self, params, optimizer, optimizer_params=None):
...
         if size() == 1: return
 
+        # In MXNet 2.0, param.name is no longer unique.
+        # Meanwhile, since horovod requires Python 3.6, there is no need to sort
+        # self._params as enumerating a python dict is always deterministic.
         for i, param in enumerate(self._params):
             if param.grad_req != 'null':
                 allreduce_(param.list_grad()[0], average=False,
-                           name=param.name, priority=-i)
+                           name=str(i), priority=-i)
```
romerojosh, line 109: Could you elaborate more on the `param.name` being no longer unique in MXNet 2.0? I'd hesitate to revert back to using `str(i)` as the allreduce request names, as it would reintroduce the problem described in #1679. Basically, in cases where users may have multiple `DistributedTrainer`s, this naming scheme will not differentiate between gradients being submitted by the different optimizers, instead producing multiple requests with the same name (i.e. multiple `allreduce.0`, `allreduce.1`, etc.).

I see in your changes to `broadcast_parameters` you use the dictionary key values, but I guess that isn't an option here because `self._params` is just a list of params, rather than the original dictionary. 

Maybe one option would be to add an optional `name` argument for the `DistributedTrainer` that gets prepended to the names of the allreduce operations being submitted? That way, users can provide unique trainer names if they need to disambiguate allreduce operations from multiple trainers.

eric-haibin-lin: I'm adding a parameter -> name mapping in the constructor (`self._param2name`). Maybe we can use that, too.

romerojosh: I think using `name = self._param2name[p._uuid]` after https://github.com/apache/incubator-mxnet/pull/18877 lands might be a cleaner option than adding a global name to the `DistributedTrainer`. Aligns more closely to the existing code and would be less disruptive to users who might be using multiple optimizers already. What do you think?

Dockerfile.test.gpu
```diff
@@ -116,7 +116,7 @@ RUN pip install "Pillow<7.0" --no-deps
 
 # Install MXNet.
 RUN if [[ ${MXNET_PACKAGE} == "mxnet-nightly" ]]; then \
-        pip install --pre mxnet-cu101mkl -f https://dist.mxnet.io/python/all; \
+        pip install --pre mxnet-cu101 -f https://dist.mxnet.io/python/all; \
```
szha, line 119: this is discontinued as NVIDIA only supports the latest two major versions and minor versions. can the horovod project update its dependencies?

horovod/mxnet/__init__.py
szha: cc @leezu to suggest on the choice of unique names for parameters.

leezu: The names in `self._param2name[p._uuid]` will not be globally unique if the user creates multiple Blocks with the same structure but different parameters. Actually I'm not convinced we should add `self._param2name[p._uuid]`

szha: @romerojosh @eric-haibin-lin what's the uniqueness and consistency requirement for this identifier? Does it need to be consistent across different workers?

romerojosh: The `name` of the allreduce operation needs to be consistent across all workers participating in the communication. This name identifier is what is used to match up tensors in the Horovod backend. Concerning uniqueness, names can be reused; however, all operations in flight must have unique names (i.e. a name can only be reused if the previous Horovod operation using that name has completed).

eric-haibin-lin: In fact after checking with @leezu, the following two blocks will have the same name for parameters in mxnet2:
```
net1 = nn.Dense(10)
net2 = nn.Dense(10)

assert net1.collect_params().keys() == net2.collect_params().keys()
```

So for horovod we probably need to go back to the approach where we add a `name` argument to DistributedTrainer.

szha:
@eric-haibin-lin do you know if the c_api_error.h is necessary here? I didn't seem to find any usage that's relevant to this header.
https://github.com/horovod/horovod/blob/a4d1a87fc19818b1f3d24db7225844bd504f5c03/horovod/mxnet/mpi_ops.h#L22

If not in use, it would be great to remove the dependency on it. In any case, I intend to merge the error entry with the new runtime API to reduce complexity in the thread local stores. The interfaces in c_api_error.h will be available in mxnet/runtime/c_runtime_api.h in near future.

eric-haibin-lin:
Merged in https://github.com/horovod/horovod/pull/2205

Issue #2141, opened 2020-07-26
Installing Horovod with pip install horovod[mxnet] errors out without building GPU extensions


davidrpugh:
**Environment:**
1. Framework: MXNet
2. Framework version: 1.6
3. Horovod version: 0.19.5
4. MPI version: 
5. CUDA version: 10.1
6. NCCL version: 2.4
7. Python version: 3.6+
8. Spark / PySpark version:
9. OS and version: CentOS
10. GCC version:

When building Horovod with `HOROVOD_WITH_MXNET=1` and installing Horovod with `pip install horovod[mxnet]`, GPU bindings are not built and an error is thrown.

Issue seems to be that `mxnet_require_list = ['mxnet>=1.4.1']` in the `setup.py` file does not install a GPU enabled builds of MXNet. GPU builds of MXNet have different naming conventions, for example `mxnet-cu101mkl` is the MKL and CUDA 10.1 version of GPU.

Pre-installing the desired version of MXNet and then installing Horovod using `pip install horovod` seems to work as expected.

tgaddair:
Good catch, @davidrpugh.  One option would be to add a separate extra for GPU-enabled MXNet, similar to what we do with TensorFlow.  Though it seems like we would need to do one for potentially every supported CUDA version, with or without MKL.  Is that right?

@eric-haibin-lin any thoughts?

davidrpugh:
For reference here is the list of versions available from PyPI.

```
This package supports Linux, Mac OSX, and Windows platforms. You may also want to check: - mxnet-cu92 with CUDA-9.2 support. - mxnet-cu92mkl with CUDA-9.2 support and MKLDNN support. - mxnet-cu91 with CUDA-9.1 support. - mxnet-cu91mkl with CUDA-9.1 support and MKLDNN support. - mxnet-cu90 with CUDA-9.0 support. - mxnet-cu90mkl with CUDA-9.0 support and MKLDNN support. - mxnet-cu80 with CUDA-8.0 support. - mxnet-cu80mkl with CUDA-8.0 support and MKLDNN support. - mxnet-cu75 with CUDA-7.5 support. - mxnet-cu75mkl with CUDA-7.5 support and MKLDNN support. - mxnet-mkl with MKLDNN support.
```

[...]

PR #2158, opened 2020-08-03
add `pip install horovod[mxnet-cu102]`

eric-haibin-lin:

## Description

Add a separate option for pip installation with mxnet cpu build (`pip install horovod[mxnet-cpu]`). By default `pip install horovod[mxnet]` requires the GPU build. https://github.com/horovod/horovod/issues/2141

## Review process to land 

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1508,7 +1508,8 @@ tensorflow_cpu_require_list = ['tensorflow-cpu']
 tensorflow_gpu_require_list = ['tensorflow-gpu']
 keras_require_list = ['keras>=2.0.8,!=2.0.9,!=2.1.0,!=2.1.1']
 pytorch_require_list = ['torch']
-mxnet_require_list = ['mxnet>=1.4.1']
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1508,7 +1508,8 @@ def build_extensions(self):
 tensorflow_gpu_require_list = ['tensorflow-gpu']
 keras_require_list = ['keras>=2.0.8,!=2.0.9,!=2.1.0,!=2.1.1']
 pytorch_require_list = ['torch']
-mxnet_require_list = ['mxnet>=1.4.1']
+mxnet_require_list = ['mxnet-cu102>=1.4.1']
```
tgaddair, line 1511: I wonder if this might confuse users, since the extra doesn't align with the actual package.

Maybe we can do this instead:

```
mxnet: mxnet>=1.4.1
mxnet-cu102: mxnet-cu102>=1.4.1
```

What do you think?

eric-haibin-lin: That makes sense. Just updated the code.

PR #2162, opened 2020-08-05
Fix for ROCm breakage - 200805

deven-amd:
The following commit breaks ROCm support for horovod-tensorflow

https://github.com/horovod/horovod/commit/7a1c761c663eb758536c1bff62d9d9c9a032dd38

horovod build fails because the `<cuda_runtime.h>` cannot be found on the ROCm platform

This PR/commit fixes the above breakage by including the HIP runtime file (instead of the CUDA one), when compiling with ROCm.

In addition to the above change, this commit also renames a couple of cuda* names to the corresponding gpu* versions instead.

/cc @jeffdaily

### DIFF
--- a/horovod/tensorflow/mpi_ops.cc
+++ b/horovod/tensorflow/mpi_ops.cc
@@ -28,12 +28,21 @@
 #define EIGEN_USE_THREADS
 
 #if HAVE_GPU
-#include <cuda_runtime.h>
 
-// Forward declaration of AsGpuStreamValue or AsCUDAStreamValue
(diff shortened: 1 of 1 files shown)


deven-amd:
@tgaddair Thank you for merging the PR

I want to figure out how to run the horovod unit tests locally.  I am attempting to create a cron job that periodically checks for any ROCm breakages in horovod (i.e. horovod built with `HOROVOD_GPU=ROCM`). 

Any advice / pointers on the proper/correct way to set this up will be very helpful.

thanks

deven

PR #2163, opened 2020-08-05
Updating TF syncbn for TF2/Keras compatibility. Updating code attribution.

romerojosh:
This PR updates the TF sync BN implementation to derive from `tf.keras.layers.BatchNormalization` for Keras/TF2 compatibility.

Additionally, original code source was from another repository, so updating attribution/copyright to reflect this: https://github.com/google/automl/blob/master/efficientdet/utils.py


DEKHTIARJonathan:
LGTM

PR #2178, opened 2020-08-13
Fixed gradient aggregation for Keras API changes in TensorFlow 2.4.0

tgaddair:
This PR addresses compatibility with changes in https://github.com/tensorflow/tensorflow/commit/0e529eb70debc7224b83d1b654f22324cd1a7d53:

- `_aggregate_gradients` now returns a list of `(grad, var)` tuples.
- Gradient aggregation now occurs in the `apply_gradients` method.

cc @omalleyt12 @reedwm @DEKHTIARJonathan

### DIFF
--- a/horovod/_keras/__init__.py
+++ b/horovod/_keras/__init__.py
@@ -13,8 +13,13 @@
 # limitations under the License.
 # ==============================================================================
 
+from distutils.version import LooseVersion
+
 import horovod.tensorflow as hvd
 import tensorflow as tf
+
+
+_PRE_TF_2_4_0 = LooseVersion(tf.__version__) < LooseVersion('2.4.0')
 
 
 def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sparse,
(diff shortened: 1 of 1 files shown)


horovod/_keras/__init__.py
```diff
@@ -13,10 +13,15 @@
 # limitations under the License.
 # ==============================================================================
 
+from distutils.version import LooseVersion
+
 import horovod.tensorflow as hvd
 import tensorflow as tf
 
 
+_PRE_TF_2_4_0 = LooseVersion(tf.__version__) < LooseVersion('2.4.0')
```
DEKHTIARJonathan, line 22: Why do you create a shortcut ? You use it once.

tgaddair: These LooseVersion checks tend to be pretty expensive, as they need to do string manipulation etc.  When this function gets called repeatedly (for example, eager execution), then it's better to call this only once.

omalleyt12:
Awesome! Please let me know if you run into any difficulties with the new Optimizer code!

I was also thinking that with the new `gradient_aggregator` argument in `Optimizer.__init__`, it would be possible for Horovod to be supported by extracting the logic in `_aggregate_gradients` into a function like `horovod.aggregate_gradients`, then users could use Horovod like:

```
optimizer = MyOptimizer(..., gradient_aggregator=horovod.aggregate_gradients)
```

The benefit of this approach is that bc it doesn't require wrapping the user's optimizer, the user has easier access to optimizer attrs like learning rate, etc

Of course, whether you want to pursue this approach is entirely up to you. The code in this PR will continue to work, and it might be too confusing to provide two different ways of supporting Optimizers (and the `gradient_aggregator` way would only work with >= TF 2.4)

PR #2212, opened 2020-08-26
Fixed issue with cross rank in heterogeneous Gloo jobs

tgaddair:
When running a Gloo job with heterogeneous workers (uneven distribution of processes / GPUs per worker), the driver would incorrectly assign the cross rank, resulting in runtime failures.

cc @richardliaw

### DIFF
--- a/horovod/runner/common/util/hosts.py
+++ b/horovod/runner/common/util/hosts.py
@@ -16,6 +16,8 @@
 import collections
 import re
 
+from dataclasses import dataclass
+
 
 class HostInfo:
     def __init__(self, hostname, slots):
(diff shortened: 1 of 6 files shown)


horovod/runner/common/util/hosts.py
```diff
@@ -28,29 +30,21 @@ def from_string(host_string):
         return HostInfo(hostname, int(slots))
 
 
+@dataclass
 class SlotInfo:
-    def __init__(self, hostname, rank, local_rank, cross_rank, size=None, local_size=None, cross_size=None):
```
EnricoMi, line 32: Before, `SlotInfo` creation required `hostname`, `rank`, `local_rank` and `cross_rank`. Now, everything is optional? Does the usage of `SlotInfo` reflect this change?

horovod/runner/common/util/hosts.py
```diff
@@ -28,29 +30,21 @@ def from_string(host_string):
...
-        self.local_rank = local_rank
-        self.local_size = local_size
-        self.cross_rank = cross_rank
-        self.cross_size = cross_size
+    hostname: str
+    rank: int
+    local_rank: int
+    cross_rank: int
+    size: int
+    local_size: int
```
EnricoMi, line 40: `Optional[int]` makes the optional parameters nullable, maybe that is equivalent to earlier `__init__`...

test/test_spark.py
```diff
@@ -526,7 +526,11 @@ def do_test_spark_run_does_not_default_env_to_os_env(self, use_mpi, use_gloo):
 
     """
     Test that horovod.spark.run fails with os.environ as env with mpi.
+    
+    TODO: figure out why this test sometimes causes hangs in CI
```
EnricoMi, line 530: yet another one of these ...

EnricoMi (approved):
minor comments to consider, approved anyway

horovod/runner/common/util/hosts.py
tgaddair: As a [dataclass](https://docs.python.org/3/library/dataclasses.html), because we do not specify defaults, these fields are all required.

horovod/runner/common/util/hosts.py
tgaddair: The new initialization has all fields defined at initialization, so no more nullable needed :)

test/test_spark.py
tgaddair: Yeah, I suspect there is something going on with processes not shutting down between tests that's leading to these errors.  It appears to be unique to our environment, as I cannot repro locally.  I will try and make it a priority to figure out what's going on after the 0.20 release.

PR #2218, opened 2020-08-30
[ray integration] Initial Ray Integration with RayExecutor API

richardliaw:

## Description

Initial PR to introduce a Ray runner for Horovod. The interface is currently specific to using Actors (stateful operators).

## Review process to land 


tgaddair:
Looks good so far!  A few of things to help with getting CI to pass:

1. We should add a `horovod[ray]` extra in setup.py similar to what we do for Spark [here](https://github.com/horovod/horovod/blob/master/setup.py#L1628).
2. Install using extras `[spark,ray]` for both [Dockerfile.test.cpu](https://github.com/horovod/horovod/blob/master/Dockerfile.test.cpu#L172) and [Dockerfile.test.gpu](https://github.com/horovod/horovod/blob/master/Dockerfile.test.gpu#L126).
3. Exclude Ray tests from MPI configs in Buildkite (as this feature is Gloo-only) by adding them to the [exclude_standalone_test](https://github.com/horovod/horovod/blob/master/.buildkite/gen-pipeline.sh#L107) variable.
4. Make sure these tests run in "standalone" mode for Gloo, meaning we don't launch them with `horovodrun`, as shown [here](https://github.com/horovod/horovod/blob/master/.buildkite/gen-pipeline.sh#L212).
5. Run `BUILDKITE_PIPELINE_SLUG=SLUG BUILDKITE_BRANCH=BRANCH .buildkite/gen-pipeline.sh > test/data/expected_buildkite_pipeline.yaml` to update expected Buildkite pipeline definition.

Hopefully, that should get everything working in CI.

test/test_cluster_ray.py
```diff
@@ -0,0 +1,155 @@
...
+
+This is currently not run on the Ray CI and is expected
+to land in Horovod repo soon. If not, then these
+should be run after every commit for anything that touches
+the Horovod-Ray integration.
+"""
+
+import os
+
+import ray
```
richardliaw, line 11: This file probably needs a little bit of clean-up; it's written in a way for multi-node setups but we can 1. get the pytest infrastructure up and 2. make it work on single-node setups first.

docs/ray.rst
```diff
@@ -0,0 +1,155 @@
...
+
+    # Launch the Ray actors on each machine.
+    # This will launch `num_slots` actors on each machine, each with
+    # 1 GPU allocated (set via CUDA VISIBLE DEVICES)
+    executor.start()
+
+
+All actors will be part of the Horovod ring, so ``RayExecutor`` invocations will be able to support arbitrary Horovod collective operations.
+
+Note that there is an implicit assumption on the cluster being homogenous in shape (i.e., all machines have the same number of slots available). This is simply
```
tgaddair, line 49: Should be addressed following #2212.  We could consider allowing `hosts` to be a list of `host:slot` in a follow-up PR.

examples/tensorflow2_mnist_ray.py
```diff
@@ -0,0 +1,93 @@
...
+    # Horovod: adjust number of steps based on number of GPUs.
+    mnist_model.fit(dataset, steps_per_epoch=500 // hvd.size(),
+                    callbacks=callbacks, epochs=num_epochs, verbose=verbose)
+
+
+ray.init()
+settings = RayExecutor.create_settings(timeout_s=30)
+executor = RayExecutor(settings, num_hosts=1, num_slots=2, use_gpu=False)
+executor.start()
+executor.execute(lambda worker: train(num_epochs=1))
```
tgaddair, line 92: Might be worth adding a convenience function here to address this scenario.  Something like:

```
def run(self, fn, *args, **kwargs):
    return self.execute(lambda worker: fn(*args, **kwargs))
```

When do you think?

examples/tensorflow2_mnist_ray.py
```diff
@@ -0,0 +1,93 @@
+import tensorflow as tf
+import horovod.tensorflow.keras as hvd
+import horovod.keras as hvd
+
+import ray
+from horovod.ray import RayExecutor
+
+
+def train(num_epochs):
```
tgaddair, line 9: What kind of function pickling does Ray use?  Is it similar to `cloudpickle` where we can capture variables / imports from outside the function definition?

horovod/ray/runner.py
```diff
@@ -0,0 +1,417 @@
...
+
+        # start global rendezvous server and get port that it is listening on
+        self.global_rendezv_port = self.rendezvous.start()
+        self.rendezvous.init(host_alloc_plan)
+        # remote_host_names = network.filter_local_addresses()
+        self.nics = driver_service.get_common_interfaces(
+            self.settings, list(self.hostnames_by_rank))
+
+        return {
+            # "HOROVOD_LOG_LEVEL": "DEBUG",
```
tgaddair, line 233: Can `HOROVOD_LOG_LEVEL` be configurable by the user?

test/data/expected_buildkite_pipeline.yaml
```diff
@@ -1175,7 +1175,7 @@ steps:
   agents:
     queue: cpu
 - label: ':pytest: Run PyTests (test-cpu-gloo-py3_8-tf2_3_0-keras2_3_1-torch1_6_0-mxnet1_5_0-pyspark3_0_0)'
-  command: bash -c "cd /horovod/test && (echo test_*.py | sed 's/test_keras.py//g' | sed 's/test_tensorflow_keras.py//g' | sed 's/test_interactiverun.py//g' | sed 's/test_spark_keras.py//g' | sed 's/test_spark_torch.py//g' | sed 's/test_spark.py//g' | sed 's/test_run.py//g' | xargs -n 1 horovodrun -np 2 -H localhost:2 --gloo pytest -v --capture=no) && pytest --forked -v --capture=fd test_spark.py test_run.py"
+  command: bash -c "cd /horovod/test && (echo test_*.py | sed 's/test_keras.py//g' | sed 's/test_tensorflow_keras.py//g' | sed 's/test_interactiverun.py//g' | sed 's/test_spark_keras.py//g' | sed 's/test_spark_torch.py//g' | sed 's/test_spark.py//g' | sed 's/test_run.py//g' | sed 's/test*_ray.py//g' | xargs -n 1 horovodrun -np 2 -H localhost:2 --gloo pytest -v --capture=no) && pytest --forked -v --capture=fd test_spark.py test_run.py test*_ray.p"
```
richardliaw, line 1178: hmm is this a typo?
```suggestion
  command: bash -c "cd /horovod/test && (echo test_*.py | sed 's/test_keras.py//g' | sed 's/test_tensorflow_keras.py//g' | sed 's/test_interactiverun.py//g' | sed 's/test_spark_keras.py//g' | sed 's/test_spark_torch.py//g' | sed 's/test_spark.py//g' | sed 's/test_run.py//g' | sed 's/test*_ray.py//g' | xargs -n 1 horovodrun -np 2 -H localhost:2 --gloo pytest -v --capture=no) && pytest --forked -v --capture=fd test_spark.py test_run.py"
```

tgaddair: Good catch!  Yes, that was a typo.  Fixed it in the `get-pipeline.sh` file.

docs/ray.rst
richardliaw: Sounds good!

examples/tensorflow2_mnist_ray.py
richardliaw: Yep; it is cloudpickle.

examples/tensorflow2_mnist_ray.py
richardliaw: Done!

horovod/ray/runner.py
richardliaw: Got it, now doable via `job.start(extra_env_vars)`

horovod/ray/runner.py
```diff
@@ -0,0 +1,445 @@
...
+        # allocation from being released, along with their children from
+        # going out of scope.
+        self.colocators = colocators
+        e_cls, e_args, e_kwargs = (executable_cls, executable_args,
+                                   executable_kwargs)
+
+        node_ids = map_blocking(
+            lambda a: a.create_workers.remote(e_cls, e_args, e_kwargs),
+            colocators)
+        assert len(set(node_ids)) == len(node_ids), (
```
tgaddair, line 316: Should this be a `RuntimeException`?  Seems dangerous to assert here.

horovod/ray/runner.py
```diff
@@ -0,0 +1,445 @@
...
+            remote_cls.remote(world_rank=rank, world_size=self.world_size)
+            for rank in range(world_rank_start, world_rank_start +
+                              self.num_slots)
+        ]
+
+        # Propogate cuda visible devices to the underlying
+        # colocated workers.
+        gpu_ids = ray.get_gpu_ids()
+        for worker, gpu_id in zip(self.workers, gpu_ids):
+            worker.update_env_vars.remote({"CUDA_VISIBLE_DEVICES": gpu_id})
```
tgaddair, line 152: `gpu_id` is only one GPU, is that correct?

For NCCL, we need to avoid setting `CUDA_VISIBLE_DEVICES`.  NCCL needs to have visibility of adjacent devices in order to use CUDA IPC.  Typically, we handle GPU isolation at the framework level, for example:

```
torch.cuda.set_device(hvd.local_rank())
```

If you need to set `CUDA_VISIBLE_DEVICES` because multiple Ray applications may be using the same nodes, then we can set `"CUDA_VISIBLE_DEVICES": gpu_ids` so every worker can see the devices in its local communicator.

horovod/ray/runner.py
```diff
@@ -0,0 +1,445 @@
...
+from horovod.runner.common.util import secret, timeout, hosts
+from horovod.runner.driver import driver_service
+from horovod.runner.http.http_server import RendezvousServer
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class MiniSettings:
+    """Mock object, only necessary for Tune.
```
tgaddair, line 18: Is this docstring still correct?  Seems this correctly encapsulates the settings needed for Ray.

denfromufa:
Super-exciting PR!

horovod/ray/runner.py
```diff
@@ -0,0 +1,449 @@
...
+        # hosts = parse_hosts(hosts_string="10.11.11.11:4,10.11.11.12:4")
+        parsed_hosts = hosts.parse_hosts(hosts_string=self.hoststring)
+        host_alloc_plan = hosts.get_host_assignments(parsed_hosts,
+                                                     self.world_size)
+
+        # start global rendezvous server and get port that it is listening on
+        self.global_rendezv_port = self.rendezvous.start()
+        self.rendezvous.init(host_alloc_plan)
+        # remote_host_names = network.filter_local_addresses()
+        self.nics = driver_service.get_common_interfaces(
```
yangw1234, line 233: Will this `driver_service` requires ssh acces to each ray nodes without prompt?

richardliaw: You can pass in an identity file now. Otherwise you need passwordless ssh access.

yangw1234: Thank you @richardliaw for your timely responses.  

However, is this "get_common_interfaces" strictly necessary? Also asking @tgaddair .  And is it possible to set the "HOROVOD_GLOO_IFACE" variable separately on each node using the nic associated the IP returned by `ray.service.get_ip_address`?

We have tried this at https://github.com/intel-analytics/analytics-zoo to avoid the ssh requirements and it works on our environment, but I am not sure this approach is strictly correct.

richardliaw: @yangw1234 can you make a new issue for this? It is hard to track comments on a closed PR! Thanks!

yangw1234: Sure. Added a new issue. #2271

PR #2220, opened 2020-08-31
Run these github workflows on push events

EnricoMi:
The `pull_request` is the wrong event, that triggers whenever a PR is modified (includes adding comments, reviews, ...), `push` is the right event.


EnricoMi:
Action are not tested when they originate from another repo.

PR #2221, opened 2020-08-31
Run these github workflows on push events

EnricoMi:
The `pull_request` is the wrong event, that triggers whenever a PR is modified (includes adding comments, reviews, ...), `push` is the right event.

Duplicates #2220.

### DIFF
--- a/.github/workflows/docs-summary.yml
+++ b/.github/workflows/docs-summary.yml
@@ -1,6 +1,6 @@
 name: Docs Summary Sync
 
-on: [pull_request]
+on: [push]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


tgaddair:
I noticed there are now two Buildkite runs per PR.  Is there a way to avoid this?  Would prefer not to double our AWS bill :)

EnricoMi:
This is unrelated to my change. This probably happened because I have first created a PR with this commit from G-Research organisation. But this did not execute the modified GitHub actions (for security reasons). So I created another PR from the same commit from horovod organisation. Both kicked off a buildkite build, one against the branch under G-Research, the other under horovod. Should be safe to merge.

Issue #2230, opened 2020-09-03
cudaGetLastError raises "invalid device ordinal" when running multiple containers on one host


tgaddair:
**Environment:**
1. Framework: any
2. Framework version: 
3. Horovod version: https://github.com/horovod/horovod/commit/e4554de96100f5a0e8686cd41cf99a6fe8a71e62
4. MPI version:
5. CUDA version: 10.0
6. NCCL version: 2.6.4
7. Python version:
8. Spark / PySpark version:
9. OS and version:
10. GCC version:

When running a Horovod job that spans multiple containers that reside on the same host (with isolated GPUs), the following error is raised:

Through debugging, it was determined this was introduced in #1949 with the line `ErrorCheck("ScaleBufferCudaImpl", cudaGetLastError());`.  Furthermore, it appears that any call to `cudaGetLastError()` in this scenario (even without Horovod) will result in such an error.  However, it is not known if this is a CUDA version specific issue.

As a workaround, we will remove this check until the underlying cause of this error is known.

cc @romerojosh

PR #2231, opened 2020-09-03
Disable cudaGetLastError to avoid errors when running on colocated containers

tgaddair:
See #2230.


romerojosh (approved):
Approving to unblock release for now.

PR #2233, opened 2020-09-03
Fix setting -D_GLIBCXX_USE_CXX11_ABI=0 when compiling CUDA kernels

maxhgerlach:

## Description

If I build the current Horovod master with TF 1.15.2 on Ubuntu 16.04 (cmake version 3.5.1), `import horovod.tensorflow` fails with
```
NotFoundError: .../lib/python3.7/site-packages/horovod/tensorflow/mpi_lib.cpython-37m-x86_64-linux-gnu.so:
undefined symbol: _ZN7horovod6common13DataType_NameB5cxx11ENS0_8DataTypeE
```
(demangled: `horovod::common::DataType_Name[abi:cxx11](horovod::common::DataType)`)

I noticed that `-D_GLIBCXX_USE_CXX11_ABI=0` had not been passed to the compiler when building `compatible_horovod_cuda_kernels`, but this would have been necessary with the Tensorflow package from PyPI. This patch fixes that and the import works fine on my system now.

## Review process to land 

### DIFF
--- a/horovod/common/ops/cuda/CMakeLists.txt
+++ b/horovod/common/ops/cuda/CMakeLists.txt
@@ -8,9 +8,7 @@ MESSAGE(STATUS "HVD_NVCC_COMPILE_FLAGS = ${HVD_NVCC_COMPILE_FLAGS}")
 
 list(APPEND CUDA_NVCC_FLAGS "${HVD_NVCC_COMPILE_FLAGS}")
 
-cuda_add_library(horovod_cuda_kernels cuda_kernels.cu)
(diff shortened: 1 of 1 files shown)


nvcastet (approved):
Thanks for catching that @maxhgerlach!
I don't know why `target_compile_definitions(...)` did not do the job.

tgaddair:
Going to merge this in now since we can't seem to get any spot instances for GPU tests right now.  We will run the full GPU suite in Buildkite before landing the release.

maxhgerlach:
Thanks for the review and merge!

PR #2234, opened 2020-09-03
Databricks supports horovod.spark.estimator

liangz1:

## Description

* Add `DBFSLocalStore` that allows horovod.spark Estimator to use Databricks Filesystem local files API to persist intermediate data and training artifacts.
* Update the ways to pin GPU: when GPU-aware scheduling is available, pin GPU to the assigned GPU index from spark; otherwise, pin GPU to `hvd.local_rank()`.

Fixes # (issue).

## Review process to land 

### DIFF
--- a/docs/spark.rst
+++ b/docs/spark.rst
@@ -300,4 +300,19 @@ and above. To run Horovod in Spark on Databricks, use `DBFSLocalStore` as the st
 
     store = DBFSLocalStore(prefix_path='/dbfs/...')
 
+The `DBFSLocalStore` uses `Databricks File System (DBFS) local file APIs <https://docs.databricks.com/data/databricks-file-system.html#local-file-apis>`__
+as a store of intermediate data and training artifacts.
+
+Databricks preconfigures GPU-aware scheduling for you on GPU clusters. See the
+`Databricks blog post <https://databricks.com/blog/2020/06/26/announcing-gpu-aware-scheduling-and-enhanced-deep-learning-capabilities.html>`__
+for details.
+
+With the Estimator API, horovod will launch # of tasks on each worker = # of GPUs on each worker, and each task will
+pin GPU to the assigned GPU from spark.
+
+With the Run API, the function `get_available_devices()` from `horovod.spark.task` will return a list of assigned GPUs
+for the spark task from which `get_available_devices()` is called.
+See `keras_spark3_rossmann.py <../examples/keras_spark3_rossmann.py>`__ for an example of using
+`get_available_devices()` with the Run API.
+
 .. inclusion-marker-end-do-not-remove
--- a/docs/spark.rst
+++ b/docs/spark.rst
@@ -290,21 +290,22 @@ Environment knobs
 -----------------
 * ``HOROVOD_SPARK_START_TIMEOUT`` - sets the default timeout for Spark tasks to spawn, register, and start running the code.  If executors for Spark tasks are scheduled on-demand and can take a long time to start, it may be useful to increase this timeout on a system level.
 
-Horovod in Spark on Databricks
+Horovod on Databricks
 ------------------------------
-Horovod in Spark in supported on
-`Databricks Runtime 7.4 ML <https://docs.databricks.com/release-notes/runtime/releases.html>`__
-and above. To run Horovod in Spark on Databricks, use `DBFSLocalStore` as the store object:
+To run Horovod in Spark on Databricks, use `DBFSLocalStore` as the store object:
 
 .. code-block:: python
 
     store = DBFSLocalStore(prefix_path='/dbfs/...')
 
-The `DBFSLocalStore` uses `Databricks File System (DBFS) local file APIs <https://docs.databricks.com/data/databricks-file-system.html#local-file-apis>`__
+The `DBFSLocalStore` uses Databricks File System (DBFS) local file APIs
+(`AWS <https://docs.databricks.com/data/databricks-file-system.html#local-file-apis>`__ |
+`Azure <https://docs.microsoft.com/en-us/azure/databricks/data/databricks-file-system#--local-file-apis>`__)
 as a store of intermediate data and training artifacts.
 
-Databricks preconfigures GPU-aware scheduling for you on GPU clusters. See the
-`Databricks blog post <https://databricks.com/blog/2020/06/26/announcing-gpu-aware-scheduling-and-enhanced-deep-learning-capabilities.html>`__
+Databricks pre-configures GPU-aware scheduling on Databricks Runtime 7.0 ML GPU and above. See GPU scheduling instructions
+(`AWS <https://docs.databricks.com/clusters/gpu.html#gpu-scheduling-1>`__ |
+`Azure <https://docs.microsoft.com/en-us/azure/databricks/clusters/gpu#gpu-scheduling>`__)
 for details.
 
 With the Estimator API, horovod will launch # of tasks on each worker = # of GPUs on each worker, and each task will
--- a/docs/spark.rst
+++ b/docs/spark.rst
@@ -290,5 +290,14 @@ Environment knobs
 -----------------
 * ``HOROVOD_SPARK_START_TIMEOUT`` - sets the default timeout for Spark tasks to spawn, register, and start running the code.  If executors for Spark tasks are scheduled on-demand and can take a long time to start, it may be useful to increase this timeout on a system level.
 
+Horovod in Spark on Databricks
+------------------------------
+Horovod in Spark in supported on
+`Databricks Runtime 7.4 ML <https://docs.databricks.com/release-notes/runtime/releases.html>`__
+and above. To run Horovod in Spark on Databricks, use `DBFSLocalStore` as the store object:
+
+.. code-block:: python
+
+    store = DBFSLocalStore(prefix_path='/dbfs/...')
 
 .. inclusion-marker-end-do-not-remove
--- a/horovod/spark/common/store.py
+++ b/horovod/spark/common/store.py
@@ -165,9 +165,13 @@ class FilesystemStore(Store):
         with self.get_filesystem().open(self.get_localized_path(path), 'rb') as f:
             return f.read()
 
-    def read_serialized_keras_model(self, ckpt_path, model=None):
-        """
-        This function will be overridden by DBFSLocalStore.
+    def read_serialized_keras_model(self, ckpt_path, model):
+        """Reads the checkpoint file of the keras model into model bytes and returns the base 64
+        encoded model bytes.
+        :param ckpt_path: A string of path to the checkpoint file.
+        :param model: A keras model. This parameter will be used in DBFSLocalStore\
+            .read_serialized_keras_model() when the ckpt_path only contains model weights.
+        :return: the base 64 encoded model bytes of the checkpoint model.
         """
         from horovod.runner.common.util import codec
 
@@ -444,7 +448,8 @@ class HDFSStore(FilesystemStore):
 
 
 class DBFSLocalStore(LocalStore):
-    """Uses DBFS as a store of intermediate data and training artifacts.
+    """Uses Databricks File System (DBFS) local file APIs as a store of intermediate data and
+    training artifacts.
 
     Initialized from a `prefix_path` starts with `/dbfs/...`, see
     https://docs.databricks.com/data/databricks-file-system.html#local-file-apis.
@@ -463,7 +468,7 @@ class DBFSLocalStore(LocalStore):
         # is used by providing `save_weights_only=True` to the ModelCheckpoint() callback.
         return 'checkpoint.tf'
 
-    def read_serialized_keras_model(self, ckpt_path, model=None):
+    def read_serialized_keras_model(self, ckpt_path, model):
         """
         Returns serialized keras model. On Databricks, only TFKeras is supported, not BareKeras.
         The parameter `model` is for providing the model structure when the checkpoint file only
--- a/horovod/spark/common/store.py
+++ b/horovod/spark/common/store.py
@@ -165,13 +165,13 @@ class FilesystemStore(Store):
         with self.get_filesystem().open(self.get_localized_path(path), 'rb') as f:
             return f.read()
 
-    def read_serialized_keras_model(self, path):
+    def read_serialized_keras_model(self, ckpt_path, model=None):
         """
         This function will be overridden by DBFSLocalStore.
         """
         from horovod.runner.common.util import codec
 
-        model_bytes = self.read(path)
+        model_bytes = self.read(ckpt_path)
         return codec.dumps_base64(model_bytes)
 
     def is_parquet_dataset(self, path):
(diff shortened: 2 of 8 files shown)


horovod/spark/keras/remote.py
```diff
@@ -301,15 +305,25 @@ def fn(hvd, tf, keras):
         for gpu in gpus:
             tf.config.experimental.set_memory_growth(gpu, True)
         if gpus:
-            tf.config.experimental.set_visible_devices(gpus[hvd.local_rank()], 'GPU')
+            if os.environ.get('CUDA_VISIBLE_DEVICES') is None:
```
liangz1, line 308: In the test(non-databricks environment), assert this branch is met.

WeichenXu123: What's the value of CUDA_VISIBLE_DEVICES set in non-databricks environment?

WeichenXu123:
Two issues: @mengxr 

* When driver fetch result model from rank 0 process, current impl is rank 0 serialize model to bytes array and sent back, but now the model already saved into DBFS, should we optimize it by "rank 0 process sent model saved dbfs path to driver, and driver load model from the dbfs path ?"

* Should we implement the DBFSLocalStore by:
  saving model to local file system temporary and then copy local file to dbfs by `sync_fn` interface ? This is the way HDFSStore used and it can support H5 model format.

liangz1:
1. For the first issue:
Current impl: 
Pros:
(1) Keep the return value type of the train function the same as LocalStore and HDFSStore. The code change is minimal.
Cons:
(1) Not taking the advantage of DBFS FUSE feature. But this is as good as the existing solution.

Return DBFS Path:
Pros:
(1) Replace "load & serialize model on worker + pickle to the driver" with "pickle path to driver + load model on the driver". Save the pickling cost.
Cons:
(1) Need to add if-else branches to handle different cases. More code change. 
(2) I'm not sure whether there is a consistency issue: After a file is written on worker, whether it is guaranteed to be available immediately on driver.

2. The second issue is discussed in the design doc "design decision 1: option 4". In short, the cons are more complex code change and unnecessary copy overhead.

horovod/spark/torch/remote.py
```diff
@@ -119,7 +121,14 @@ def train(serialized_model, optimizer_cls, model_opt_state_serialized,
...
         if cuda_available:
             # Horovod: pin GPU to local rank.
-            torch.cuda.set_device(hvd.local_rank())
+            if os.environ.get('CUDA_VISIBLE_DEVICES') is None:
+                assert not is_dbfs
+                torch.cuda.set_device(hvd.local_rank())
+            else:
+                # Databricks pyspark sets CUDA_VISIBLE_DEVICES for GPU scheduling.
+                # Pin the only visible GPU allocated to this task.
+                assert is_dbfs
```
liangz1, line 130: This will fail in CI. We should use is_dbfs to determine whether to use pin to GPU[0]

test/test_spark.py
```diff
@@ -1669,3 +1669,46 @@ def test_to_list(self):
...
+        dbfs_ckpt_path = dbfs_store.prefix_path + "/" + dbfs_ckpt_name
+        model.save(dbfs_ckpt_path)
+        serialized_model_dbfs = dbfs_store.read_serialized_keras_model(dbfs_ckpt_path)
+        reconstructed_model_dbfs = deserialize_keras_model(serialized_model_dbfs)
+        assert reconstructed_model_dbfs.get_config() == model.get_config()
+        assert reconstructed_model_dbfs.get_weights() == model.get_weights()
+
+        # test local_store.read_serialized_keras_model
+        local_ckpt_path = local_store.prefix_path + "/" + local_store.get_checkpoint_filename()
+        model.save(local_ckpt_path)
```
liangz1, line 1707: For TF 1.15.0, SavedModel format is not selected by adding the ".tf" suffix. https://github.com/tensorflow/docs/blob/920057b528427ac647b8dcb93d898fce76507d3d/site/en/r1/guide/saved_model.md#build-and-load-a-savedmodel The syntax for saving in SavedModel is quite different from the current one.
Manual test on databricks with tf downgraded to 1.15 also fails with the h5py error.

horovod/spark/torch/remote.py
```diff
@@ -119,7 +121,14 @@ def train(serialized_model, optimizer_cls, model_opt_state_serialized,
...
+                torch.cuda.set_device(0)
```
mengxr, line 129: We should do this:

~~~
if has GPU-aware scheduling:
  pin to the assigned GPU index no matter whether CUDA_VISIBLE_DEVICES is defined or not
else:
  pin to local rank (this might have issues)
~~~

test/test_spark.py
liangz1: https://dogfood.staging.cloud.databricks.com/#notebook/29305617/command/29305633

horovod/spark/common/store.py
```diff
@@ -164,6 +165,15 @@ def read(self, path):
         with self.get_filesystem().open(self.get_localized_path(path), 'rb') as f:
             return f.read()
 
+    def read_serialized_keras_model(self, ckpt_path, model=None):
+        """
+        This function will be overridden by DBFSLocalStore.
```
mengxr, line 170: * First sentence should describe what this method does.
* Describe the params and the return value.

horovod/spark/common/store.py
```diff
@@ -164,6 +165,15 @@ def read(self, path):
...
+        """
+        from horovod.runner.common.util import codec
+
+        model_bytes = self.read(ckpt_path)
+        return codec.dumps_base64(model_bytes)
```
mengxr, line 175: beyond this PR: base64 is not very efficient in encoding bytes. Why not transferring the model bytes directly? cc: @tgaddair

horovod/spark/common/store.py
```diff
@@ -431,3 +441,40 @@ def _check_url(self, url, prefix, path):
     @classmethod
     def filesystem_prefix(cls):
         return cls.FS_PREFIX
+
+
+class DBFSLocalStore(LocalStore):
+    """Uses DBFS as a store of intermediate data and training artifacts.
```
mengxr, line 447: Databricks File System (DBFS) local file APIs

horovod/spark/common/store.py
```diff
@@ -431,3 +441,40 @@ def _check_url(self, url, prefix, path):
...
+
+    Initialized from a `prefix_path` starts with `/dbfs/...`, see
+    https://docs.databricks.com/data/databricks-file-system.html#local-file-apis.
+
+    Note that DBFSLocalStore cannot be created via Store.create("/dbfs/..."), which will create a
```
mengxr, line 452: Maybe a follow-up: If user calls `Store.create("dbfs:/")`, shall we convert it into DBFSLocalStore?

horovod/spark/common/store.py
```diff
@@ -431,3 +441,40 @@ def _check_url(self, url, prefix, path):
...
+        Returns serialized keras model. On Databricks, only TFKeras is supported, not BareKeras.
+        The parameter `model` is for providing the model structure when the checkpoint file only
+        contains model weights.
+        """
+        import tensorflow
+        from tensorflow import keras
+        from horovod.spark.keras.util import TFKerasUtil
+
+        if LooseVersion(tensorflow.__version__) < LooseVersion("2.0.0"):
+            model.load_weights(ckpt_path)
```
mengxr, line 477: When `model=None`? If we never expect it to happen, we shouldn't leave the default to None.

horovod/spark/common/util.py
```diff
@@ -695,3 +695,13 @@ def to_list(var, length):
                              "label_cols")
 
     return var
+
+
+def _get_allocated_gpu(hvd):
```
mengxr, line 700:
~~~python
_get_assigned_gpu_or_default(default):
~~~

horovod/spark/keras/remote.py
```diff
@@ -154,7 +157,11 @@ def train(serialized_model, train_rows, val_rows, avg_row_size):
                 if _checkpoint_callback:
                     _checkpoint_callback.filepath = ckpt_file
                 else:
-                    _checkpoint_callback = k.callbacks.ModelCheckpoint(ckpt_file)
+                    if is_dbfs and LooseVersion(tf.__version__) < LooseVersion("2.0.0"):
+                        _checkpoint_callback = k.callbacks.ModelCheckpoint(ckpt_file,
```
mengxr, line 161: would be nice to have an inline comment to explain why (DBFS local APIs does not support random write).

mengxr (changes requested):

horovod/spark/common/store.py
liangz1: Added to backlog.

test/test_spark.py
```diff
@@ -1669,3 +1669,55 @@ def test_to_list(self):
...
+        with get_dbfs_output_dir() as run_output_dir:
+            dbfs_ckpt_path = run_output_dir + "/" + local_store.get_checkpoint_filename()
+            if LooseVersion(tensorflow.__version__) < LooseVersion("2.0.0"):
+                model.save_weights(dbfs_ckpt_path)
+            else:
+                model.save(dbfs_ckpt_path)
+            serialized_model_dbfs = dbfs_store.read_serialized_keras_model(dbfs_ckpt_path, model)
+            reconstructed_model_dbfs = deserialize_keras_model(serialized_model_dbfs)
+            if LooseVersion(tensorflow.__version__) >= LooseVersion("2.3.0"):
+                assert reconstructed_model_dbfs.get_config() == model.get_config()
```
liangz1, line 1710: In TF 2.2.0, pytest fails with
~~~
assertion error: {'build_input_shape': TensorShape([None, 32])} != {'build_input_shape': TensorShape([None, 32])}
~~~
In TF 2.1.0, 2.0.0, it has
~~~
{..., 'batch_input_shape': [None, 32]} != {..., 'batch_input_shape': (None, 32)}
~~~
but pytest did not fail with assertion error. 

Thus I disabled the model equality check for versions < 2.3.0.

docs/spark.rst
```diff
@@ -290,5 +290,29 @@ Environment knobs
 -----------------
 * ``HOROVOD_SPARK_START_TIMEOUT`` - sets the default timeout for Spark tasks to spawn, register, and start running the code.  If executors for Spark tasks are scheduled on-demand and can take a long time to start, it may be useful to increase this timeout on a system level.
 
+Horovod in Spark on Databricks
```
mengxr, line 293: Shall we just call it "Horovod on Databricks"?

docs/spark.rst
```diff
@@ -290,5 +290,29 @@ Environment knobs
...
+------------------------------
+Horovod in Spark in supported on
+`Databricks Runtime 7.4 ML <https://docs.databricks.com/release-notes/runtime/releases.html>`__
```
mengxr, line 296: Do not mention a future Runtime version here. Instead, mention the min Horovod version to have DBFSLocalStore or just don't mention it at all.

docs/spark.rst
```diff
@@ -290,5 +290,29 @@ Environment knobs
...
+and above. To run Horovod in Spark on Databricks, use `DBFSLocalStore` as the store object:
+
+.. code-block:: python
+
+    store = DBFSLocalStore(prefix_path='/dbfs/...')
+
+The `DBFSLocalStore` uses `Databricks File System (DBFS) local file APIs <https://docs.databricks.com/data/databricks-file-system.html#local-file-apis>`__
```
mengxr, line 303: Could you add two links one for AWS and one for Azure?

docs/spark.rst
```diff
@@ -290,5 +290,29 @@ Environment knobs
...
+as a store of intermediate data and training artifacts.
+
+Databricks preconfigures GPU-aware scheduling for you on GPU clusters. See the
```
mengxr, line 306: I don't think Databricks setup is any different from a Spark cluster with GPU scheduling enabled. We can just say Databricks pre-configures GPU-aware scheduling on Databricks Runtime 7.0 ML GPU and above, and link to GPU scheduling instructions.

horovod/spark/common/util.py
```diff
@@ -695,3 +695,13 @@ def to_list(var, length):
                              "label_cols")
 
     return var
+
+
+def _get_assigned_gpu_or_default(default):
+    try:
+        # if GPU-aware scheduling is available, pin to the assigned GPU index
+        from horovod.spark.task import get_available_devices
+        return int(get_available_devices()[0])
```
mengxr, line 704: What exception do you expect this call to raise? If it returns an empty list and `[0]` throws an index out of bound error, then it is better to do:

~~~python
from horovod.spark.task import get_available_devices
available_devices = get_available_devices()
if available_devices:
  return int(available_devices[0])
else:
  return default
~~~

The difference is that if the cluster has GPU-aware configured but GPU index was done wrongly, e.g., "gpu1" instead of "1". We should throw an exception instead of fallback.

mengxr (changes requested):
Some minor comments.

mengxr (approved):
LGTM. @tgaddair Could you take a look?

horovod/spark/common/store.py
tgaddair: The original intent was to be able to store as part of a JSON file that encoded Estimator params, so we would represent the model as a base64 encoded string.  I believe this was changed, however, so may not be needed anymore.

@abditag2 can confirm.

horovod/spark/common/store.py
```diff
@@ -431,3 +445,41 @@ def _check_url(self, url, prefix, path):
...
+        super(DBFSLocalStore, self).__init__(prefix_path, *args, **kwargs)
+
+    def get_checkpoint_filename(self):
+        # Use the default Tensorflow SavedModel format in TF 2.x. In TF 1.x, the SavedModel format
+        # is used by providing `save_weights_only=True` to the ModelCheckpoint() callback.
+        return 'checkpoint.tf'
+
+    def read_serialized_keras_model(self, ckpt_path, model):
+        """
+        Returns serialized keras model. On Databricks, only TFKeras is supported, not BareKeras.
```
tgaddair, line 473: Note: we will probably be dropping BareKeras support in the next release as well.

horovod/spark/common/store.py
```diff
@@ -431,3 +445,41 @@ def _check_url(self, url, prefix, path):
...
+        The parameter `model` is for providing the model structure when the checkpoint file only
+        contains model weights.
+        """
+        import tensorflow
```
tgaddair, line 477: Is the reason for having this method unique to this store that it will not work with BareKeras?  If so, we can move this out once we drop BareKeras support.

horovod/spark/keras/remote.py
```diff
@@ -154,7 +157,14 @@ def train(serialized_model, train_rows, val_rows, avg_row_size):
                 if _checkpoint_callback:
                     _checkpoint_callback.filepath = ckpt_file
                 else:
-                    _checkpoint_callback = k.callbacks.ModelCheckpoint(ckpt_file)
+                    if is_dbfs and LooseVersion(tf.__version__) < LooseVersion("2.0.0"):
```
tgaddair, line 160: Similar to previous comment: we should be able to adopt this pattern universally after we drop BareKeras, without needing to special-case DBFS.

tgaddair (commented):
Just to check my understanding: all of the special handling of model checkpointing for DBFS is specific to the fact that you are only supporting TensorFlow-native Keras, is that correct?  My assumption here is that we can remove this special-casing once we drop support for standalone Keras, but I want to make sure that's correct.

liangz1:
@tgaddair all of the special handling of model checkpointing for DBFS is to avoid using h5 format, which requires random writes, and DBFS does not support that.

tgaddair:
Ah, gotcha @liangz1.  Is there any reason not to use the SavedModel format over H5?  Are you aware of any performance differences?  I ask, because it would simplify things to converge on a single format.  If we need to support both, than it would be cleaner to have the store specify the model checkpoint format, then have separate handlers for each format (instead of special-casing on the store type).

liangz1:
One issue with SavedModel is we have to use save_weights_only in TF 1.x to enabled it, see [comment above](https://github.com/horovod/horovod/pull/2234#discussion_r483703692). That makes the code looks ugly.

Performance-wise: I searched online and didn't see discussions about performance regressions. I noticed that TF 2.x is default at SavedModel format and refers to h5 as "the old format". I would support unifying to SavedModel format in all cases, which could be done in follow-up PRs.

horovod/spark/common/store.py
abditag2: Correct. When we serialize the estimator, the model is stored as a b64 encoded string in Json along with the rest of the parameters of the SparkMLEstimator.

tgaddair: Thanks, so we may not need it here then since we're reading from disk, but this can be addressed in a follow-up PR.

tgaddair (approved):
Let's go ahead and merge to unblock you, and we can do some refactoring in a follow-up PR.

tgaddair:
Thanks for the PR, @liangz1!

horovod/spark/common/store.py
abditag2: Yes.

liangz1:
Thanks for the timely review! 🎉 @tgaddair @abditag2

PR #2239, opened 2020-09-04
Fixed usage of VERBOSE=1 when setting custom MAKEFLAGS

tgaddair:
When settings env variable `MAKEFLAGS`, the native args option `--` is omitted, leading to error message `Unknown argument VERBOSE=1`.


Dockerfile.cpu
```diff
@@ -3,7 +3,7 @@ FROM ubuntu:18.04
 ENV TENSORFLOW_VERSION=2.3.0
 ENV PYTORCH_VERSION=1.6.0
 ENV TORCHVISION_VERSION=0.7.0
-ENV MXNET_VERSION=1.6.0.post0
+ENV MXNET_VERSION=1.5.0
```
leezu, line 6: Should this be

```suggestion
ENV MXNET_VERSION=1.6.0
```

nvcastet (approved):
Thanks Travis!

PR #2242, opened 2020-09-05
Download all BuildKite artifacts

EnricoMi:
The Buildkite artifact download action did only download the first 100 artifacts, but Horovod has 483 (!). Move to the latest version of the action that fixes this.

Further, to avoid more disruptions, this moves to master so in case there is another bug in the actions I can fix them in actions' master and rerun the workflow without rerunning Buildkite (and waiting hours and raising the AWS bill).

With `continue-on-error: true`, those workflows won't fail on these actions. I keep an eye on them. Thanks for letting me test these actions here, if they work for Horovod, they work everywhere.

### DIFF
--- a/.github/workflows/buildkite_unit_tests.yml
+++ b/.github/workflows/buildkite_unit_tests.yml
@@ -19,7 +19,7 @@ jobs:
 
     steps:
     - name: Download Buildkite Artifacts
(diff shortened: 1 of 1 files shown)


tgaddair (approved):
Sounds good!  Thanks for the fix.

PR #2243, opened 2020-09-05
Don't run the Unit Tests workflow on master or tags

EnricoMi:
Buildkite does not run on master push event but on schedule, the download action will not be able to pick this up in a timely manner. Buildkite does not run on tags either. No need to run the workflow for those.


tgaddair (approved):
Good catch!

PR #2245, opened 2020-09-06
Do not fetch unit tests in forked horovods

EnricoMi:
The new Unit Tests github workflow does not work for PRs from forked horovod repos because they run in the context of the forked repo and thus do not have the BuildKite token.

Until I have a solution for this situation, lets not run the actions in forked repos.

### DIFF
--- a/.github/workflows/buildkite_unit_tests.yml
+++ b/.github/workflows/buildkite_unit_tests.yml
@@ -19,6 +19,7 @@ jobs:
 
     steps:
     - name: Download Buildkite Artifacts
+      if: github.repository_owner == 'horovod'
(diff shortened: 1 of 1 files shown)


tgaddair (approved):
Looks good, do you want to make an issue to track?

tgaddair:
Looks like there's a conflict in the workflow file at the moment.  Good to land once that's resolved.

EnricoMi:
> Looks good, do you want to make an issue to track?

good idea, there may be a few more iterations until this is all worked out

EnricoMi:
Look at that: `buildkite/horovod/pr — Build #3700 passed (1 minute, 12 seconds)`

tgaddair:
Nice! :)

PR #2247, opened 2020-09-07
Do not build and test commits with no code changes 

EnricoMi:
The `gen-pipeline.sh` script calls out to `get_commit_files.py` to determine all changed files of the commit (w.r.t. to base branch / default). If there are no code files in that set of modified files, then `gen-pipeline.sh` does not include any test builds and test runs in the buildkite pipeline.

A non-code buildkite pipeline looks like this: https://buildkite.com/horovod/horovod/builds/3660

This reduces round-trip time and AWS bill for trivial (i.e., non-code) changes.


tgaddair (approved):
This is great!  Since we have workflows to test that `.gep-pipeline` is working correctly, I think we can run checks like these with a lot more confidence.

I'm guessing the check for `BUILDLITE_PIPELINE_DEFAULT_BRANCH` on line 18 allows tests to run for the automatic nightly tests, is that correct?

EnricoMi:
> I'm guessing the check for `BUILDLITE_PIPELINE_DEFAULT_BRANCH` on line 18 allows tests to run for the automatic nightly tests, is that correct?

You are right, that bit is to enforce a full test pipeline on each nightly (well 2pm my time) build on master, no matter how minimal the changes.

tgaddair:
Sounds good, this will be an awesome way to save some AWS credits!

PR #2257, opened 2020-09-10
Fix workflow paths ignore

EnricoMi:


tgaddair (approved):
Is the issue that the quoting uses bash rules, and ignores special characters like * with single quotes?

EnricoMi:
> Is the issue that the quoting uses bash rules, and ignores special characters like * with single quotes?

The issue is the leading `^` in the last path, single quotes seem to work, just aligned quotes with double quotes in other places.

PR #2261, opened 2020-09-10
Fixed Gloo support by setting USE_LIBUV on macOS

tgaddair:
Setting `USE_LIBUV_DEFAULT` did not work for TensorFlow (likely, PyTorch pulls in libuv and Gloo separately).  By setting `USE_LIBUV` instead, we ensure that support is linked in.

Also added documentation clarifying that macOS users must install `libuv` to use Horovod with Gloo.


nvcastet (approved):
yes, `USE_LIBUV` is the right one, my mistake. Thank you for the PR!

PR #2263, opened 2020-09-10
Set compiler options based on cmake build type

leezu:
This fixes building Debug builds.

Further remove the ` set(CMAKE_CXX_FLAGS "$ENV{CPPFLAGS} ${CMAKE_CXX_FLAGS}")` line, as cmake already reads `CXXFLAGS` by default and there is no need to handle a non-standard `CPPFLAGS` environment variable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,10 +63,18 @@ string(REGEX REPLACE  "-std=[^ ]+" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 if(DEFINED ENV{AR})
     set(CMAKE_AR $ENV{AR})
 endif()
-# Make CMAKE read CPPFLAGS
-set(CMAKE_CXX_FLAGS "$ENV{CPPFLAGS} ${CMAKE_CXX_FLAGS}")
+
 # Add default project CXX flags
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread -fPIC -O3 -Wall -ftree-vectorize -g")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread -fPIC -Wall -ftree-vectorize")
+if(CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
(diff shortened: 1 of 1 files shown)


tgaddair (approved):
LGTM!  Thanks for the PR.

CMakeLists.txt
```diff
@@ -63,10 +63,18 @@ string(REGEX REPLACE  "-std=[^ ]+" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 if(DEFINED ENV{AR})
     set(CMAKE_AR $ENV{AR})
 endif()
-# Make CMAKE read CPPFLAGS
-set(CMAKE_CXX_FLAGS "$ENV{CPPFLAGS} ${CMAKE_CXX_FLAGS}")
+
 # Add default project CXX flags
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread -fPIC -O3 -Wall -ftree-vectorize -g")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread -fPIC -Wall -ftree-vectorize")
```
nvcastet, line 68: @leezu @tgaddair FYI, I think we used to always build with debug symbols "-g" before, even in release mode, which gives extra stacktrace info when program crashes. Downside of "-g" is bigger libraries.

leezu: Though `CMAKE_BUILD_TYPE=Release` generally does not include debug information. If the previous behaviour is desired, we can consider defaulting to the  `CMAKE_BUILD_TYPE=RelWithDebInfo` build type inside `setup.py`. WDYT? https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html

tgaddair: I think having the stack trace is a very useful feature, so `CMAKE_BUILD_TYPE=RelWithDebInfo` sounds good to me.  What do you think @nvcastet ?

nvcastet: yes I agree with RelWithDebInfo

tgaddair: Awesome, @leezu can you put together a PR?

leezu: @nvcastet I'm actually not sure how the `self.debug` attribute is set in setup.py:
https://github.com/horovod/horovod/blob/00dfae090777ba27d4ea768d7a92ba3760a247f5/setup.py#L64

Can you clarify? Essentially we just need to change the default value of this line to `RelWithDebInfo`

nvcastet: yes default value to `RelWithDebInfo`. To be honest i don't know how `self.debug` get set to `True`. I kept the code that was there before my PR.

PR #2264, opened 2020-09-10
Refactored tests by run commands and added macOS CI

tgaddair:


EnricoMi:
I was planning to refactor tests as well. Now that unit test stats are available through the publish action we have a good indication if number of tests changes. And we have the evidence (test result artifacts) to extract the exact list of test case names, to see which are missing / new.

Given that master is build on a schedule not on push, getting the current number of tests is a bit tricky. You would need to branch of the same head commit as this branch and trigger a full test build (touch a code file). Then you will get the stats for that master commit.

Dockerfile.test.gpu
```diff
@@ -55,6 +55,9 @@ RUN wget --progress=dot:mega https://bootstrap.pypa.io/get-pip.py && python get-
 RUN pip install -U --force pip setuptools requests pytest mock pytest-forked parameterized
 RUN echo pytest -v --capture=no --continue-on-collection-errors --junit-xml=/artifacts/junit.\$1.\${HOROVOD_RANK:-\${OMPI_COMM_WORLD_RANK:-\${PMI_RANK}}}.\$2.xml \${@:2} > /pytest.sh
 RUN cat /pytest.sh
+RUN echo pytest -v --capture=no --continue-on-collection-errors --junit-xml=/artifacts/junit.\$1.standalone.\$2.xml \${@:2} > /pytest_standalone.sh
+RUN chmod a+x /pytest_standalone.sh
+RUN cat /pytest_standalone.sh
```
EnricoMi, line 60: the cat line (here and three lines above) can go away once we are happy with the content of the script file, which we already are I think

Dockerfile.test.gpu
```diff
@@ -55,6 +55,9 @@ RUN wget --progress=dot:mega https://bootstrap.pypa.io/get-pip.py && python get-
 RUN pip install -U --force pip setuptools requests pytest mock pytest-forked parameterized
 RUN echo pytest -v --capture=no --continue-on-collection-errors --junit-xml=/artifacts/junit.\$1.\${HOROVOD_RANK:-\${OMPI_COMM_WORLD_RANK:-\${PMI_RANK}}}.\$2.xml \${@:2} > /pytest.sh
 RUN cat /pytest.sh
+RUN echo pytest -v --capture=no --continue-on-collection-errors --junit-xml=/artifacts/junit.\$1.standalone.\$2.xml \${@:2} > /pytest_standalone.sh
+RUN chmod a+x /pytest_standalone.sh
```
EnricoMi, line 59: this `chmod` could be added to `/pytest.sh` as well

.travis.yml
```diff
@@ -0,0 +1,51 @@
+os:
```
EnricoMi, line 1: can we use buildkite for macos? this introduces the third pipeline (github workflows, buildkite and travis ci)

EnricoMi (commented):
I like moving the tests into `parallel` and `single` subdirs, tells you which to run through mpi and which directly through pytest, also simplifies `gen-pipeline.sh`.

.travis.yml
tgaddair: It's tricky with Buildkite, because our approach is based around Docker containers, but we cannot run macOS in Docker.  Travis CI has some benefits here because it's free and provides macOS VMs out of the box.

tgaddair:
@EnricoMi @abditag2 this is ready for review.

tgaddair:
Looks like Spark hangs are popping up again.  I think I may have found a solution, however.

abditag2 (approved):
LGTM

EnricoMi:
Spark PyTests might need a bump in the timeout, maybe 10m.

tgaddair:
@EnricoMi looks like one of the Spark tests was terminating in Spark 3 with a strange caught exception, causing the next test to hang.  Running that test in a subprocess helped, and may be a solution for some of the other hung tests.

EnricoMi:
> @EnricoMi looks like one of the Spark tests was terminating in Spark 3 with a strange caught exception, causing the next test to hang. Running that test in a subprocess helped, and may be a solution for some of the other hung tests.

I saw your commits, very delicate issues, well done. Good starting point for other issues, I agree.

EnricoMi:
@richardliaw this might also conflict with #2291

PR #2275, opened 2020-09-15
[ray] Better NIC detection

### LINKED ISSUE #2271: Make horovod on ray not requiring ssh identity file or passwordless access
**Is your feature request related to a problem? Please describe.**

Horovod on Ray still requires users to have ssh passwordless access (or through identity file) from the driver to each ray node.  

Not all ray clusters meet this. It would be nice to run horovod on ray without this requirement.

**Describe the solution you'd like**

Be able to run horovod on ray without ssh passwordless access or identity file.

**Describe alternatives you've considered**
Currently no alternatives I've considered.

**Additional context**
No

richardliaw:

## Description

Fixes #2271

## Review process to land 


tgaddair (commented):
Thanks for putting this together, @richardliaw!  This is a good start.

I think there are a couple things we can do to address this SSH issue, one short-term and one long-term.

For an immediate workaround, we can allow the user to specify the `nics` and avoid using SSH to find the common interface.  This should already be handled by `get_common_interfaces` [here](https://github.com/horovod/horovod/blob/master/horovod/runner/driver/driver_service.py#L214), just need to test that it works.

The more general solution is to find the common interfaces without requiring passwordless SSH.  This should also be doable, but will require more work.  Finding the common interfaces essentially consists of three parts:

1. Launch task servers on every worker host as shown [here](https://github.com/horovod/horovod/blob/master/horovod/runner/driver/driver_service.py#L49).
2. Have every worker probe the next worker using its set of interfaces to determine which are "routable". This is shown [here](https://github.com/horovod/horovod/blob/master/horovod/runner/common/util/network.py#L193). This is necessary because some interfaces will show up on all hosts but will not actually be capable of routing traffic on the network (e.g., `lo`).
3. Intersect the routable interfaces from each worker to find a common set, then we can choose any among these and be sure it will work.

My thinking is that we should be able to use the existing `get_common_interfaces` code, but swap out the part where we instantiate the task servers (see [here](https://github.com/horovod/horovod/blob/master/horovod/runner/driver/driver_service.py#L105)). Instead of using SSH, we can use Ray primitives to create the servers via dependency injection, or via environment similar to how we handle Kubeflow in `get_remote_command`.  

What do you think?

richardliaw:
Thanks for the response!

> For an immediate workaround, we can allow the user to specify the nics and avoid using SSH to find the common interface. This should already be handled by get_common_interfaces here, just need to test that it works.

Yep, saw this. If this is the case, a quick solution would also be to raise an error message if `get_common_interfaces` fails, saying to the user to specify a `settings.nics` parameter.


> The more general solution is to find the common interfaces without requiring passwordless SSH. This should also be doable, but will require more work. Finding the common interfaces essentially consists of three parts...

Thanks for the thorough explanation!

Ray should be able to easily start the task servers without requiring ssh. As you mentioned, this could be easily done by exposing the underlying logic for creating servers on each Ray worker (perhaps via dependency injection).

A quick clarifying question - you mentioned that the `worker needs to probe the next worker using its set of interfaces to determine which are "routable".`. So if I understand correctly, this means that each worker is just determining if its own interfaces are routable, and hypothetically can use any other worker (i.e., the driver) as a probe target right?

horovod/ray/runner.py
```diff
@@ -47,6 +48,21 @@ def __init__(self, world_rank=0, world_size=1):
...
+                        snicaddr.family == socket.AF_INET):
+                    chosen_nic = nic
+                    break
+        if chosen_nic is None:
+            logger.info(
+                "Unable to identify nic corresponding "
+                f"to ip address: {self.hostname}")
+        else:
+            os.environ["HOROVOD_GLOO_IFACE"] = str(chosen_nic)
+            os.environ["NCCL_SOCKET_IFNAME"] = str(chosen_nic)
```
richardliaw, line 64: @tgaddair should we still keep this?

tgaddair: I don't think we need this if we go with the dependency injection route I suggested.

tgaddair:
> A quick clarifying question - you mentioned that the `worker needs to probe the next worker using its set of interfaces to determine which are "routable".`. So if I understand correctly, this means that each worker is just determining if its own interfaces are routable, and hypothetically can use any other worker (i.e., the driver) as a probe target right?

I think the issue with using the driver is that some nics may be routable to the driver, but not to the other workers (if they're on separate subnets).  So using the ring probing, we ensure that we find the nic(s) that route between workers not not just the driver.  

There is a transitivity assumption here: if nic x routes A to B and x routes B to C, then nic x should also route A to C.  By doing the intersection, we are guaranteed to find at least one such nic, or to fail the job.  If we only go through the driver, we know we can route A to driver and B to driver, but not necessarily A to B, if that makes sense.

examples/tensorflow2_mnist_ray.py
```diff
@@ -43,13 +48,15 @@ def train(num_epochs):
 
     # Horovod: add Horovod DistributedOptimizer.
     opt = hvd.DistributedOptimizer(opt)
+    print("compiling")
```
richardliaw, line 51:
```suggestion
```

richardliaw:
@yangw1234 can you try out this PR to see if it is again what you are looking for?

tgaddair (approved):
Looks great!  Very clean reuse of the probe logic.

yangw1234:
> @yangw1234 can you try out this PR to see if it is again what you are looking for?

Yeah, I think this works. Thanks!

PR #2286, opened 2020-09-17
Always test master

EnricoMi:
My earlier change #2284 didn't consider the always-test-on-master-feature. You can see that the merge commit in master did kick off a full test run on Buildkite but the test workflow did not run. Not sure if this change fixes it. We can only test this in master.

With this merged into master, the test workflow should run on the merge commit though it is a non-code change.


tgaddair:
Related to #2287.

tgaddair (approved):

PR #2289, opened 2020-09-17
Fixed Elastic tf.keras and added additional integration tests

### LINKED ISSUE #2285: tf-keras example is not working well when scale worker up in elastic mode
**Environment:**
1. Framework: tf-keras
2. Framework version: 1.15.0
3. Horovod version: v0.20.0

**Bug report:**
Just run this example ```examples/elastic/tensorflow_keras_mnist_elastic.py``` in elastic mode. It will be blocked when try to scale worker up.

**FYI**
I am trying to resolve this issue. Unfortunately, I can't find out how to reset the uniq id of op name. 
For example, sync state in first time, the name is ```dict.sz```. But the second time is ```dict.sz_1``` and so on. In this case the new worker's name is ```dict.sz``` in first time. It is mismatch between new and old worker.
Any idea for this issue?

tgaddair:
Fixes #2285.

### DIFF
--- a/horovod/_keras/__init__.py
+++ b/horovod/_keras/__init__.py
@@ -29,12 +29,15 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
 
         def __init__(self, **kwargs):
             self._name = name or "Distributed%s" % self.__class__.__base__.__name__
-            self._device_dense = device_dense
-            self._device_sparse = device_sparse
-            self._compression = compression
-            self._sparse_as_dense = sparse_as_dense
             self._aggregated_gradients = False
-            self._gradient_predivide_factor = gradient_predivide_factor
+            self._allreduce_grads = hvd._make_allreduce_grads_fn(
+                self._name,
+                device_dense,
+                device_sparse,
+                compression,
+                sparse_as_dense,
+                hvd.Average,
+                gradient_predivide_factor)
             super(self.__class__, self).__init__(**kwargs)
 
         def get_gradients(self, loss, params):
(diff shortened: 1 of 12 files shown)


horovod/_keras/__init__.py
```diff
@@ -58,40 +61,9 @@ def _aggregate_gradients(self, grads_and_vars):
...
-                                                     compression=self._compression,
-                                                     prescale_factor=prescale_factor,
-                                                     postscale_factor=postscale_factor)
-                            averaged_gradients.append(avg_grad)
-                        else:
-                            averaged_gradients.append(None)
-                    return averaged_gradients
-            else:
-                return gradients
+            return self._allreduce_grads(grads)
```
tgaddair, line 66: @romerojosh this part may be relevant to you since you recently did some work on this part of the code.

horovod/_keras/elastic.py
```diff
@@ -54,6 +60,10 @@ def __init__(self, backend, state, *args):
...
         self.backend = backend
         self.state = state
+        self.initial_epoch = self.state.epoch
+
+    def on_train_begin(self, logs=None):
+        self.initial_epoch = self.state.epoch
 
     def on_epoch_end(self, epoch, logs=None):
-        self.state.epoch = epoch
+        self.state.epoch = self.initial_epoch + epoch + 1
```
EnricoMi, line 69: is `epoch` the epoch that ended? why `self.state.epoch = self.initial_epoch + epoch + 1`

tgaddair: Yes, `epoch` is the epoch that just ended.  The idea is that at the end of the epoch, we want to `commit` to save our progress.  We could avoid having to do the `+ 1` here if we called this `on_epoch_begin`, but if we commit before incrementing the epoch number (but after resetting the batch number), then we would end up repeating the last epoch.  So this way we ensure that when we commit, the state is fully up to date in case an error occurs.

The `initial_epoch` is because the `epoch` provided by the callback is a relative epoch (we cannot offset it at an initial epoch).  So if we complete some number of epochs and then reset, we do not want to lose our progress.

I will add comments to explain this in the code.

horovod/_keras/__init__.py
romerojosh: Excellent! Glad to see this duplicate code being removed.

romerojosh (approved):
LGTM 👍

PR #2291, opened 2020-09-18
[ray] elastic executor

richardliaw:

## Description

Allows Elastic jobs to be executed via Ray.

## Review process to land 


test/test_ray_elastic.py
```diff
@@ -0,0 +1,297 @@
+"""Ray-Horovod Elastic training unit tests.
```
tgaddair, line 1: We should add this as a standalone test in `gen-pipeline.sh`.  I am working on a refactor so we won't need to do this for every test that doesn't get executed by `horovodrun` (#2264), but for now we need to follow the same process that we used in the last Ray PR.

horovod/runner/gloo_run.py
```diff
@@ -74,22 +88,9 @@ def slot_info_to_command(slot_info):
...
-            'HOROVOD_CROSS_SIZE={cross_size} '
-                .format(hostname=host_name,
-                        rank=slot_info.rank,
-                        size=slot_info.size,
-                        local_rank=slot_info.local_rank,
-                        local_size=slot_info.local_size,
-                        cross_rank=slot_info.cross_rank,
-                        cross_size=slot_info.cross_size))
+        env_vars = create_slot_env_vars(slot_info)
+        horovod_rendez_env = " ".join(
```
tgaddair, line 92: This is a much cleaner implementation, but one side effect is that it has apparently broke a few of our more exceptionally fragile set of tests in `test_spark.py`.  Basically, this version removes a space that is expected here: https://github.com/horovod/horovod/blob/master/test/test_spark.py#L776

If you remove the extra space in the string on that line of the test, those tests should pass.

tgaddair (commented):
Great work!  Just a few things around handling the return values.

horovod/runner/gloo_run.py
richardliaw: wow.......

richardliaw: that's wild, thanks for the pointer!

horovod/ray/elastic.py
```diff
@@ -0,0 +1,265 @@
...
+        worker_env_vars.update(self.run_env_vars.copy())
+        worker_env_vars.update(self.env_vars.copy())
+        worker_env_vars.update({"PYTHONUNBUFFERED": "1"})
+
+        def worker_loop(slot_info, events):
+            worker = self._create_remote_worker(slot_info, worker_env_vars)
+            future = worker.execute.remote(lambda _: worker_fn())
+
+            result = None
+            while result is None:
```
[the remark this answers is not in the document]
richardliaw: Yeah it should be the case

horovod/ray/elastic.py
```diff
@@ -0,0 +1,265 @@
...
+                    self.return_values.append(retval)
+                    # Success
+                    result = 0, time.time()
+                except ray.exceptions.GetTimeoutError:
+                    # Timeout
+                    if any(e.is_set() for e in events):
+                        ray.kill(worker)
+                        result = 1, time.time()
+                    else:
+                        result = None
```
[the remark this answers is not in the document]
richardliaw: yeah I wanted to make it explicit (since it's not easy to follow), but I can remove it.

horovod/ray/elastic.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+class RayHostDiscovery(HostDiscovery):
+    """Uses Ray global state to obtain host mapping.
+
+    Assumes that the whole global state is available for usage."""
+
+    def __init__(self, use_gpu=False, cpus_per_slot=1):
+        self.use_gpu = use_gpu
+        self.cpus_per_slot = cpus_per_slot
```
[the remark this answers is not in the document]
richardliaw: yep! added to both elastic and regular ray executors.

horovod/ray/elastic.py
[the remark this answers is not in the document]
richardliaw: added a todo

horovod/ray/elastic.py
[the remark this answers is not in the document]
richardliaw: done!

test/test_ray_elastic.py
richardliaw: Done. Take a look, I think I did it right? (but not sure).

horovod/runner/gloo_run.py
richardliaw: Fixed!

horovod/ray/elastic.py
```diff
@@ -0,0 +1,265 @@
...
+            exit_code, timestamp = value
+            if exit_code != 0:
+                raise RuntimeError(
+                    'Horovod detected that one or more processes '
+                    'exited with non-zero '
+                    'status, thus causing the job to be terminated. '
+                    'The first process '
+                    'to do so was:\nProcess name: {name}\nExit code: {code}\n'
+                    .format(name=name, code=exit_code))
+        return self.return_values
```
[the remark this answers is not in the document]
richardliaw: Yeah that sounds good... I just did the local scoping suggestion.

horovod/ray/elastic.py
```diff
@@ -0,0 +1,265 @@
...
+
+        def worker_loop(slot_info, events):
+            worker = self._create_remote_worker(slot_info, worker_env_vars)
+            future = worker.execute.remote(lambda _: worker_fn())
+
+            result = None
+            while result is None:
+                try:
+                    retval = ray.get(future, timeout=0.1)
+                    self.return_values.append(retval)
```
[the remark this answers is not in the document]
richardliaw: Got it; pushed a fix.

horovod/ray/runner.py
```diff
@@ -284,12 +285,22 @@ def __init__(self,
                  num_hosts: int = 1,
                  num_slots: int = 1,
                  cpus_per_slot: int = 1,
-                 use_gpu: bool = False):
+                 use_gpu: bool = False,
```
tgaddair, line 288: What do you think about making the default for `use_gpu=None`?

The logic would then become:

```
RayExecutor(..., use_gpu=True)  -> 1 GPU per slot
RayExecutor(..., gpus_per_slot=2)  -> 2 GPUs per slot
RayExecutor(..., use_gpu=True, gpus_per_slot=2)  -> 2 GPUs per slot
RayExecutor(...)  -> 0 GPUs per slot
RayExecutor(..., use_gpu=False, gpus_per_slot=1)  -> ValueError
```

Feels intuitive to me, but not sure if it makes things more confusing.  Feel free to leave as is if you think the current approach makes more sense.

test/test_ray_elastic.py
tgaddair: Looks good!  Only additional step is to run this from the root of the repo:

```
BUILDKITE_PIPELINE_SLUG=SLUG BUILDKITE_BRANCH=BRANCH .buildkite/gen-pipeline.sh > test/data/expected_buildkite_pipeline.yaml
```

We do this to make sure that the generated pipeline is not broken.

horovod/ray/elastic.py
```diff
@@ -0,0 +1,289 @@
...
+                cpus_per_slot=cpus_per_slot,
+                gpus_per_slot=gpus_per_slot)
+        self.cpus_per_slot = cpus_per_slot
+        self.gpus_per_slot = gpus_per_slot
+        self.use_gpu = use_gpu
+        self.settings = settings
+        self.driver = None
+        self.rendezvous = None
+        self.env_vars = env_vars or {}
+        self.return_values = []
```
tgaddair, line 168: I believe this can now be removed.

test/test_ray_elastic.py
richardliaw: It printed `WARNING:root:no commit (None) or default branch (None) given` - should be OK right?

tgaddair: Yeah, that's fine.

horovod/ray/elastic.py
richardliaw: done!

horovod/ray/runner.py
richardliaw: Hmm interesting. I think I will keep it as is to minimize the number of ways to do the most common case (1 gpu per worker) - the proposal would give you 3 ways:
 
* use_gpu=True + gpus_per_slot=1, 
* use_gpu=None + gpus_per_slot=1, 
* use_gpu=True + gpus_per_slot=None

while the status quo is 2 ways:
* use_gpu=True + gpus_per_slot=1, 
* use_gpu=True + gpus_per_slot=None

tgaddair: Sounds good!

tgaddair:
@richardliaw looks like you have some docs issues.  Should be as simple as adding a line for `ray.exceptions` here:

https://github.com/horovod/horovod/blob/master/docs/mocks.py#L57

horovod/ray/elastic.py
```diff
@@ -0,0 +1,300 @@
...
+from horovod.ray.runner import BaseHorovodWorker
+
+logger = logging.getLogger(__name__)
+
+if hasattr(ray.exceptions, "GetTimeoutError"):
+    GetTimeoutError = ray.exceptions.GetTimeoutError
+elif hasattr(ray.exceptions, "RayTimeoutError"):
+    GetTimeoutError = ray.exceptions.RayTimeoutError
+else:
+    raise ImportError("Unable to find Ray Timeout Error class "
```
tgaddair, line 30: Is there a min version of Ray we should set in `setup.py`?

docs/ray.rst
```diff
@@ -108,6 +108,60 @@ A unique feature of Ray is its support for `stateful Actors <https://docs.ray.io
...
+
+    import ray
+    ray.init(address="auto")  # attach to the Ray cluster
+    settings = ElasticRayExecutor.create_settings(verbose=True)
+    executor = ElasticRayExecutor(
+        settings, use_gpu=True, cpus_per_slot=2)
+    executor.start()
+    executor.run(training_fn)
+
+Ray will automatically start remote actors which execute ``training_fn`` on nodes as they become available. Note that ``executor.run`` call will terminate whenever any one of the training functions terminates.
```
tgaddair, line 164: I think it should be "one of the training functions terminates successfully, or all workers fail".

tgaddair (approved):
Incredible work!  Very excited to see this landed.

horovod/ray/elastic.py
richardliaw: well this specific handling allows compat between >1.0 and <1.0; the overall integration should be compatible with releases since 0.8.0 (released a long while ago)

docs/ray.rst
richardliaw: Ah good catch, will patch.

Issue #2287, opened 2020-09-17
Present unit test results on commits and pull requests


EnricoMi:
With #2213, tests in buildkite produce JUnit xml result files and a new GitHub workflow downloads these files via the [download-buildkite-artifact-action](https://github.com/EnricoMi/download-buildkite-artifact-action) and summarizes them via the [publish-unit-test-result-action](https://github.com/EnricoMi/publish-unit-test-result-action).

Given the different situations when Buildkite runs, we have to cater for the following cases:

- Buildkite runs only on `push` to the horovod repo, not on `push` to forks (#2245)
- Buildkite runs on pushes to branches that have a pull request in the horovod repo, not on pull requests in forks (#2245)
- Buildkite does not run unit tests on branches that have no code changes (#2247)
- Buildkite always runs unit tests on master (#2284, #2286, #2288)
- Actions running in one repo can only comment on pull reuqest in the same repo (#2333)
  - The publish-unit-test-result-action on a pull request in horovod must run in the horovod repo
  - If that pull request merges a horovod branch into horovod master, everything is fine (`push` and `pull_request` events work)
  - If that pull request merges from a fork, than the action has to run in the horovod repo (only `pull_request_target` works)
  - Pull requests that merge horovod master into forks should run in the fork repo to comment on that fork pull request (`pull_request_target`)

Further incremental fixes:
#2242, #2243, #2257, #2270, #2293, #2357

EnricoMi:
Finished with #2293

PR #2293, opened 2020-09-18
Rework non-code change detection, recombine master and branch unit test workflow

EnricoMi:
* removed code that gets changed files from pull requests, always get changed files for commit against default branch
* moved regexp patterns into Python script to have a single place with regexps
* allows to reuse the Python script in github workflow
* combines Master and Branch unit test workflow into a single one
* fixes different behaviour between buildkite and workflow: bildkite only runs tests when diff against master contains code changes, workflow only downloads tests when current commit has code changes (which is a different set of commits than buildkite tests)
* reduces complexity


EnricoMi:
@tgaddair this is ready to go in, #2287 links here already

test/test_buildkite.py
```diff
@@ -13,15 +13,19 @@
...
+import sys
 import unittest
 import warnings
 from shutil import copy
 
+from common import tempdir
 from horovod.runner.common.util import safe_shell_exec
 
-from common import tempdir
+sys.path.append('../.buildkite')
```
tgaddair, line 26: This is assuming we're running out of the `test` folder.  To make it more generic, we could change it to:

```
sys.path.append(os.path.join(os.path.dirname(__file__), os.pardir, '.buildkite'))
```

tgaddair (commented):
Looks good! Just one comment.

test/test_buildkite.py
EnricoMi: excellent, fixed and decorated with a comment

tgaddair (approved):
LGTM!  Thanks!

PR #2296, opened 2020-09-18
Remove dist.mxnet.io link from setup.py

leezu:
This page contains nightly releases that are not intended for the use of the general public. They are provided as is for the purpose of testing for MXNet and ecosystem developers.


tgaddair (approved):

EnricoMi:
@leezu  Is it ok to use that URL in our `Dockerfile.test.*`? There we test against the nightly builds. I agree this link is not needed in that `setup.py`.

leezu:
@EncricoMi I think Horovod development falls under "testing by ecosystem developers", thus yes

PR #2235, opened 2020-09-03
Make sure mkldnn is included with mxnet before enabling it

nvcastet:

PR #2304, opened 2020-09-21
Revert "Make sure mkldnn is included with mxnet before enabling it"

leezu:
Reverts horovod/horovod#2235

If Mxnet_USE_MKLDNN is true, we must build Horovod with MKLDNN. We must not create a mismatch between Horovod and MXNet, as this can cause Segfaults


tgaddair:
cc @nvcastet 

I remember that this was causing install issues if CPU MXNet was installed, but mkldnn support was not available.  @leezu, do you know it can happen that `Mxnet_USE_MKLDNN` is True but the installed mxnet package does not support mkldnn?

nvcastet:
I talked to @leezu offline.
We need to update this PR with this logic:
If MXNET MKLDNN is present and there is no MKLDNN header, print warning message and return MXNET NOT FOUND (cmake will then decide if mxnet build was required by the user)
If it is required, the build config will fail therefore pip install will fail.
if it is not required, it will continue with the other frameworks and will not build anything for MXnet.

leezu:
> do you know it can happen that Mxnet_USE_MKLDNN is True but the installed mxnet package does not support mkldnn?

@tgaddair means the installed mxnet package is inconsistent / broken and can't be used to build horovod

tgaddair:
I see, this makes sense.  Thanks for putting together the fix!

cmake/Modules/FindMxnet.cmake
```diff
@@ -28,6 +28,10 @@ if (LEN EQUAL "5")
     string(TOUPPER ${Mxnet_USE_MKLDNN} Mxnet_USE_MKLDNN)
     string(TOUPPER ${Mxnet_USE_CUDA} Mxnet_USE_CUDA)
     if (Mxnet_USE_MKLDNN)
+        if (NOT EXISTS ${Mxnet_INCLUDE_DIRS}/mkldnn)
+            set(MXNET_FOUND FALSE)
```
nvcastet, line 33: You want something like that:
```
set(MXNET_FOUND FALSE)
if (MXNET_FIND_REQUIRED)
    set(MSG_LEVEL "FATAL_ERROR")
else()
    set(MSG_LEVEL "WARNING")
endif()
message(${MSG_LEVEL} "MXNet was found with mkl-dnn support but mkldnn header files are missing. Please, install MXNet with mkldnn header files.")
return()
```

leezu: Why is that not handled by find_package? According to doc: 

> The package configuration file may set <PackageName>_FOUND to false to tell find_package that component requirements are not satisfied.

https://cmake.org/cmake/help/latest/command/find_package.html

cmake/Modules/FindMxnet.cmake
```diff
@@ -27,7 +27,17 @@ if (LEN EQUAL "5")
     list(GET Mxnet_OUTPUT 4 Mxnet_USE_CUDA)
     string(TOUPPER ${Mxnet_USE_MKLDNN} Mxnet_USE_MKLDNN)
     string(TOUPPER ${Mxnet_USE_CUDA} Mxnet_USE_CUDA)
-    if (Mxnet_USE_MKLDNN AND EXISTS ${Mxnet_INCLUDE_DIRS}/mkldnn)
+    if (Mxnet_USE_MKLDNN)
+        if (NOT EXISTS ${Mxnet_INCLUDE_DIRS}/mkldnn)
+            if (MXNET_FIND_REQUIRED)
```
nvcastet, line 32: Sorry the variable is `Mxnet_FIND_REQUIRED` instead.

nvcastet (approved):

leezu:
Thank you @nvcastet and @karan6181!

PR #2305, opened 2020-09-21
Fix RelWithDebInfo build, make it the default with -03 optimizations

leezu:
As per discussion in https://github.com/horovod/horovod/pull/2263

Thanks @nvcastet


tgaddair:
Very strange, looks like Horovod builds successfully but fails at runtime with this setting.

leezu:
Previously we had `-pthread -fPIC -O3 -Wall -ftree-vectorize -g` now we have `-pthread -fPIC -Wall -ftree-vectorize -g -O3`. I'm not sure why there is any effect on the runtime.

tgaddair:
Hey @leezu, based on this change, is your conclusion that we're best off reverting #2263?

nvcastet:
@leezu @tgaddair ok I think i fixed it. RelWithDebInfo was missing from the Utilities cmake script. I also use -03 in RelWithDebInfo because RelWithDebInfo uses -02 by default, and i remove our logic around cmake build type because cmake adds the good flags by default.

leezu (commented):
Thank you @nvcastet

tgaddair (approved):
Thanks @nvcastet!  Can confirm this works on macOS as well.

PR #2325, opened 2020-09-24
Expose name argument to TF allreduce operation.

romerojosh:
## Description
For some reason, the `name` argument is not exposed in the public `allreduce` API for TF and is only available in the private `_allreduce` version. This PR adds the `name` argument to the TF `allreduce` to be more consistent with the other framework implementations.


tgaddair (approved):
Nice catch!

PR #2328, opened 2020-09-24
Add option to strip outer name scope from Horovod ops in TF.

romerojosh:
## Description
Currently for TensorFlow, Horovod will use the op node name from TensorFlow to match up collective requests. In model parallel scenarios, it is possible that tensors will be produced from graphs that are constructed differently across workers, leading to inconsistent op names, and mismatched collectives. While there is the option to assign a custom `name` to an op, this does not fully get around this issue as the TF op node name can still include an outer name scope that is prepended to the custom name (e.g. you can get collectives named `graph1/my_custom_name` and `graph2/my_custom_name`, where the `graph*/` scope is applied by TF and currently cannot be removed by the user).

This PR introduces an optional `ignore_name_scope` option to Horovod ops in TF, which will strip the outer name scope applied by TF to the node name, allowing the custom name mechanism to work in this scenario.


tgaddair (approved):
LGTM!

PR #2329, opened 2020-09-24
Moved examples into framework-specific subfolders

tgaddair:


tgaddair:
@EnricoMi what do you think about this organization for examples?

.gitignore
```diff
@@ -14,9 +14,9 @@ docs/_build
 env
 venv/
 
-examples/checkpoint*
+examples/**/checkpoint*
+examples/**/checkpoints/
```
EnricoMi, line 18: doesn't `examples/**/checkpoint*` match `examples/**/checkpoints/?

EnricoMi (commented):
I like the re-org. I think under `spark` and `elastic` we should see similar framework subfolders for each of the ML frameworks.

EnricoMi:
>    619 tests +     10     575 heavy_check_mark +     10       44 zzz ±    0  0 heavy_multiplication_x ±0 

I am wondering, why there are 10 new tests. I will look into the artifacts for that.

EnricoMi:
> > 619 tests +     10     575 heavy_check_mark +     10       44 zzz ±    0  0 heavy_multiplication_x ±0
> 
> I am wondering, why there are 10 new tests. I will look into the artifacts for that.

Reason is that the reference master build failed (cancelled): https://buildkite.com/horovod/horovod/builds/3951 So some test did not run in the base reference.

.gitignore
tgaddair: Yes, it should, good point.  That was a holdover from the previous version.  Will fix that.

tgaddair:
Thanks for the review, @EnricoMi!  Added additional subdirectories under `spark` and `elastic`.

EnricoMi (approved):
I love this reorg!

EnricoMi:
Heads up: this will conflict with #2291

tgaddair:
Good point @EnricoMi.

cc @richardliaw heads up that the Ray examples will need to move into `examples/ray` when you rebase the Elastic Ray branch.

richardliaw:
got it, thanks for the heads up!

PR #2364, opened 2020-10-10
Fix broken links in benchmarks.rst

jgonzalezab:
Fixed broken links of imagenet TFRecord preprocessing script and Tensorflow v1, Tensorflow v2 and PyTorch synthetic benchmarks


EnricoMi:
@jgonzalezab ignore the empty unit test results, they should have been reported here since Buildkite did not run any, since there were no code changes (only doc). I will fix this unnecessary unit test result report soon.

tgaddair (approved):
Thanks for the fix!

PR #2366, opened 2020-10-12
Check for changes step has to be run on pull request events

EnricoMi:


EnricoMi:
Fixes situations like https://github.com/horovod/horovod/pull/2364#issuecomment-706510777.

tgaddair (approved):

PR #2373, opened 2020-10-14
Df schema

abditag2:

## Description

Fixes # (issue).

## Review process to land 


tgaddair (approved):

OscarDPan:
Hi @tgaddair @abditag2 may I know what's the schema inference for? I found it really degrades the performance during inference. Downgrading Horovod to 0.20.3 drastically improved the speed.

thuningxu:
The reason we do this, is that sometimes Spark has trouble inferring the schema of the resulted dataframe. Why Spark has to infer the schema? This is because df.rdd strips the schema.

I think a better fix is to preserve the schema first and then re-apply it to the prediction result RDD to create the proper DF.

PR #2383, opened 2020-10-19
Auto-detect ccache during build configuration

leezu:


leezu:
There is a test failure in `Tf2KerasTests.test_gradient_aggregation`. Is there a way to retrigger the buildkite tests only?

tgaddair:
Hey @leezu, this is a known issue, I'm putting together a fix now to workaround.  We can rebase this PR once that's ready.

nvcastet:
@leezu Do you see any speedup during build from scratch? Does it speedup only incremental builds, if yes by how much?

leezu:
There is no speedup as long as the cache is empty. If the cache is filled, the speedup is proportional to the amount of code changes since between the prior builds and the current built. If all objects are in the cache, the "build" time of objects is essentially instant.

Build time with empty cache

```
337.15user 24.35system 1:20.94elapsed 446%CPU (0avgtext+0avgdata 832648maxresident)k
488104inputs+1592200outputs (1463major+3779506minor)pagefaults 0swaps
```

Build with cache and after  `git clean -ffxd`

```
16.31user 5.60system 0:16.39elapsed 133%CPU (0avgtext+0avgdata 180976maxresident)k
278240inputs+663992outputs (867major+303567minor)pagefaults 0swaps
```

The majority of the 16s for build with cache are actually `setup.py` related calls unrelated to the compilation.

Numbers obtained via

```
% HOROVOD_WITH_GLOO=1 \
    HOROVOD_WITHOUT_MPI=1 HOROVOD_WITH_MXNET=1 HOROVOD_WITHOUT_PYTORCH=1 \
    HOROVOD_WITHOUT_TENSORFLOW=1 /usr/bin/time python3.8 setup.py install --user
```

leezu:
To obtain the numbers, I further deleted 

https://github.com/horovod/horovod/blob/342c18c6b82a0358c03b60e3c213f9f2edda1b3e/setup.py#L183-L186

as it adds a lot of overhead to the `setup.py` call unrelated to the compilation.

tgaddair (approved):
This is awesome @leezu, thanks for putting this together!

EnricoMi:
Is Buildkite in any way using this caching?

nvcastet:
you would need to mount the same ccache directory in all the build containers to be useful for the CI. I don't know if it is possible with buildkite.

I was asking about if ccache was benefiting the end-user also (when cache starts empty) because we re-build the same objects again when installing horovod for more than one framework. But we don't pass exactly the same compiler flags for each framework so `ccache` will probably rightly not use its cache.

EnricoMi:
I just wanted to make sure it is not used on Buildkite, otherwise we would have to disable it for some situations, which we don't need to. Good work, btw.

leezu:
> I just wanted to make sure it is not used on Buildkite, otherwise we would have to disable it for some situations, which we don't need to. Good work, btw.

@EnricoMi which situations do you have in mind here? Generally there should be no downside to enabling ccache in the CI system.

EnricoMi:
Caching is stateful, and CI should build things without initial state. If caching does not work correctly, CI might build but Horovod does not build from scratch. Everyone has seen this before, code build locally but fails on CI because it is built there from scratch.

So I would have deactivated ccache at least for (scheduled or all) master builds to have a chance to spot that issue then.

leezu:
@EnricoMi from https://ccache.dev/

> Is it safe?
> 
> The most important aspect of a compiler cache is to always produce exactly the same output that the real compiler would produce. This includes providing exactly the same object files and exactly the same compiler warnings that would be produced if you use the real compiler. The only way you should be able to tell that you are using ccache is the speed.
> 
> Ccache of course tries to provide these guarantees. However:
> 
>     Compilers are moving targets. Newer compiler versions may, and often do, introduce features that ccache can't foresee. In some cases it's also very hard for ccache to handle the compiler behaviors in a good way, especially in a way that is backward compatible with legacy compilers.
>     The fastest mode (the "direct mode") has a corner case which can result in false positive cache hits. This and other minor limitations are listed under caveats in the manual.
> 
> If you experience any bugs or undocumented limitations, please report them.

Where the caveat is

> The direct mode fails to pick up new header files in some rare scenarios. See THE DIRECT MODE above.

I haven't experienced such issue in practice, but if it were a concern for CI one could disable the direct mode it.

For context, MXNet uses ccache on CI ([with a shared filesystem](https://aws.amazon.com/efs/)) since several years without issues.

PR #2388, opened 2020-10-20
Fixed --log-hide-timestamp to apply to driver logs with Gloo

### LINKED ISSUE #2382: Output is annoying
**Environment:**
1. Framework: (TensorFlow)
2. Framework version: 2.0
3. Horovod version: up-to-date
4. MPI version:
5. CUDA version: 10.1
6. NCCL version: 2.4
7. Python version: 3.8
8. OS and version: ubuntu 18.04
9. GCC version:

**Checklist:**

**Your question:**
Is it possible to remove the leading timestamp in the output?
**Mon Oct 19 20:44:09 2020[0]<stderr>**:I1019 20:44:09.594008 139669443786560 hvd_controller.py:30] Step: 1400/329817, steps_per_second: 4.89, {train_loss: 50.83835, lr: 0.00246, accuracy: 0.00000}

I tried set environment variable HOROVOD_LOG_HIDE_TIME="1", but have no effect.
Thanks.

tgaddair:
Fixes #2382.

### DIFF
--- a/horovod/runner/common/util/safe_shell_exec.py
+++ b/horovod/runner/common/util/safe_shell_exec.py
@@ -78,9 +78,9 @@ def terminate_executor_shell_and_children(pid):
             pass
 
 
-def forward_stream(src_stream, dst_stream, prefix, index):
+def forward_stream(src_stream, dst_stream, prefix, index, hide_timestamp):
     def prepend_context(line, rank, prefix):
-        localtime = time.asctime(time.localtime(time.time()))
+        localtime = time.asctime(time.localtime(time.time())) if not hide_timestamp else ''
         return '{time}[{rank}]<{prefix}>:{line}'.format(
             time=localtime,
             rank=str(rank),
@@ -159,7 +159,7 @@ def _create_event(ctx):
     return ctx.Event()
 
 
-def execute(command, env=None, stdout=None, stderr=None, index=None, events=None):
+def execute(command, env=None, stdout=None, stderr=None, index=None, events=None, hide_timestamp=False):
     ctx = multiprocessing.get_context('spawn')
 
     # When this event is set, signal to middleman to terminate its children and exit.
@@ -195,8 +195,8 @@ def execute(command, env=None, stdout=None, stderr=None, index=None, events=None
     if stderr is None:
         stderr = sys.stderr
 
-    stdout_fwd = in_thread(target=forward_stream, args=(stdout_r, stdout, 'stdout', index))
-    stderr_fwd = in_thread(target=forward_stream, args=(stderr_r, stderr, 'stderr', index))
+    stdout_fwd = in_thread(target=forward_stream, args=(stdout_r, stdout, 'stdout', index, hide_timestamp))
+    stderr_fwd = in_thread(target=forward_stream, args=(stderr_r, stderr, 'stderr', index, hide_timestamp))
 
     # TODO: Currently this requires explicitly declaration of the events and signal handler to set
     #  the event (gloo_run.py:_launch_jobs()). Need to figure out a generalized way to hide this behind
--- a/horovod/runner/common/util/safe_shell_exec.py
+++ b/horovod/runner/common/util/safe_shell_exec.py
@@ -78,9 +78,9 @@ def terminate_executor_shell_and_children(pid):
             pass
 
 
-def forward_stream(src_stream, dst_stream, prefix, index, timestamp_output):
+def forward_stream(src_stream, dst_stream, prefix, index, prefix_output_with_timestamp):
     def prepend_context(line, rank, prefix):
-        localtime = time.asctime(time.localtime(time.time())) if timestamp_output else ''
+        localtime = time.asctime(time.localtime(time.time())) if prefix_output_with_timestamp else ''
         return '{time}[{rank}]<{prefix}>:{line}'.format(
             time=localtime,
             rank=str(rank),
@@ -159,7 +159,7 @@ def _create_event(ctx):
     return ctx.Event()
 
 
-def execute(command, env=None, stdout=None, stderr=None, index=None, events=None, timestamp_output=False):
+def execute(command, env=None, stdout=None, stderr=None, index=None, events=None, prefix_output_with_timestamp=False):
     ctx = multiprocessing.get_context('spawn')
 
     # When this event is set, signal to middleman to terminate its children and exit.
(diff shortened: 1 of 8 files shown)


horovod/runner/common/util/safe_shell_exec.py
```diff
@@ -78,9 +78,9 @@ def terminate_executor_shell_and_children(pid):
             pass
 
 
-def forward_stream(src_stream, dst_stream, prefix, index):
+def forward_stream(src_stream, dst_stream, prefix, index, hide_timestamp):
```
EnricoMi, line 81: I wouldn't call this `hide_timestamp` but `with_timestamp=True`

EnricoMi (changes requested):
LGTM otherwise

horovod/runner/launch.py
```diff
@@ -610,7 +611,8 @@ def _run_elastic(args):
                                                 start_timeout=tmout,
                                                 output_filename=args.output_filename,
                                                 run_func_mode=args.run_func is not None,
-                                                nics=args.nics)
+                                                nics=args.nics,
+                                                log_hide_timestamp=args.log_hide_timestamp)
```
EnricoMi, line 615: the command line flag could be named `--log-without-timestamp` and then `log_with_timestamp = ! args.log_without_timestamp`, if it does not have to be a flag it could be `--log-with-timestamp=false`

tgaddair: Changed to `--timestamp-output` for consistency with `mpirun`.

horovod/runner/common/util/safe_shell_exec.py
tgaddair: Good point.  Looking into this further, it looks like we have an inconsistency here between how `horovodrun` handles MPI vs Gloo.  With MPI, we need to set `--timestamp-output` to get this behavior, which we don't do by default.  So I think the best thing to do is change the default to Gloo to also hide the timestamp, then only enable (for both) when the user passes in `--timestamp-output`.  What do you think?

horovod/runner/launch.py
tgaddair: Also, retain the old arg since it's also used for something else (Horovod internal logging).

horovod/runner/launch.py
```diff
@@ -427,9 +427,11 @@ def parse_args():
                                help='Minimum level to log to stderr from the Horovod backend. (default: WARNING).')
     group_logging_timestamp = group_logging.add_mutually_exclusive_group()
     group_logging_timestamp.add_argument('--log-hide-timestamp', action=make_override_true_action(override_args),
-                                         help='Hide the timestamp from Horovod log messages.')
```
EnricoMi, line 430: I presume there is no way to rename this `--log-hide-timestamp`, or deprecate it? Better name would be `--log-without-timestamp`. Even worse: `--no-log-hide-timestamp`. Are they different to `--timestamp-output`?

horovod/runner/launch.py
EnricoMi: Do we want to hook into MPI naming convention that much? It is not the best flag name.

horovod/runner/launch.py
EnricoMi: Just read your comment above... Still think they should be marked deprecated and a better named flag put in its place. With the `--timestamp-output`: That is been created independent to `--log-hide-timestamp` on purpose?

horovod/runner/common/util/safe_shell_exec.py
EnricoMi: I am happy with this being opt-in behaviour, for MPI and Gloo. Still, the `horovodrun` command line argument does not need to be called `--timestamp-output` just because we forward it to that flag when using MPI. Something like `--log-remote-shell-with-timestamp` would be meaningful.

horovod/runner/launch.py
tgaddair: We can rename and deprecate the old flag, I will do that.

Yes, I believe it is correct to have separate flags here, since they are serving different purposes.

horovod/runner/launch.py
tgaddair: I agree it's not the best name, but since we need to keep a separate `--log-with-timestamp` flag, I think having a distinct name that MPI users will immediately understand (to help with telling them apart) is beneficial.  What do you think?

horovod/runner/common/util/safe_shell_exec.py
tgaddair: What about `--prepend-output-with-timestamp` / `-prepend-timestamp`?  I think we should avoid "log" here since this applies to all stdout / stderr, not only logs (which have a different meaning for other args).

EnricoMi: What about `--prefix-output-with-timestamp`? Otherwise `--prepend-output-with-timestamp` is fine.

tgaddair: `prefix` is good, too.  Wasn't sure which to go with, so sounds like prefix it is.

horovod/runner/launch.py
```diff
@@ -426,10 +453,23 @@ def parse_args():
...
+                                         action=make_override_true_action(override_args),
+                                         help=argparse.SUPPRESS)
+    group_logging_timestamp.add_argument('--log-without-timestamp', dest='log_with_timestamp',
+                                         action=make_override_false_action(override_args), 
+                                         help='Hide the timestamp from Horovod internal log messages.')
+    group_logging_timestamp.add_argument('-prefix-timestamp', '--prefix-output-with-timestamp', action='store_true',
+                                         dest='prefix_output_with_timestamp',
+                                         help='Timestamp each line of output to stdout, stderr, and stddiag.')
+    group_logging_timestamp.add_argument('--log-hide-timestamp', 
+                                         dest='log_with_timestamp',
```
EnricoMi, line 466: I think this should read `log_without_timestamp`

horovod/runner/launch.py
```diff
@@ -426,10 +453,23 @@ def parse_args():
...
-                                         help='Hide the timestamp from Horovod log messages.')
-    group_logging_timestamp.add_argument('--no-log-hide-timestamp', dest='log_hide_timestamp',
-                                         action=make_override_false_action(override_args), help=argparse.SUPPRESS)
+    group_logging_timestamp.add_argument('--log-with-timestamp', 
+                                         action=make_override_true_action(override_args),
+                                         help=argparse.SUPPRESS)
+    group_logging_timestamp.add_argument('--log-without-timestamp', dest='log_with_timestamp',
+                                         action=make_override_false_action(override_args), 
+                                         help='Hide the timestamp from Horovod internal log messages.')
+    group_logging_timestamp.add_argument('-prefix-timestamp', '--prefix-output-with-timestamp', action='store_true',
```
EnricoMi, line 462: `-prefix-timestamp` with a single `-` because it is the short version? or because one `-` is missing?

tgaddair: Yes, this is the short version for people who want to avoid keystrokes.

horovod/runner/launch.py
tgaddair: Good catch!

tgaddair: Ah, actually this one is correct.  This is for the deprecated arg, so we avoid breaking existing users.  When they use this flag, they will get a deprecation warning.  We will remove this in the next major release.

EnricoMi: it is confusing though

tgaddair: The args are suppressed from `horovodrun --help`, so only users who are already using them will be aware of them.

EnricoMi (approved):
much clearer and great deprecation

PR #2401, opened 2020-10-27
Add support for backward_passes_per_step > 1 for LegacyOptimizers (TF) in Graph Mode.

aaron276h:

## Description
This PR is a follow up #2346. This PR adds support for backward_passes_per_step > 1 for TF legacy optimizers (tf.train.Optimizer) executing in graph(non-eager) mode. This is one of the features that we have built into[ Determined AI's](https://determined.ai/) fork of Horovod that we would like to upstream.

## Review process to land 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Added support for backward_passes_per_step > 1 for TF Keras eager execution. ([#2371](https://github.com/horovod/horovod/pull/2371))
 
+- Added support for backward_passes_per_step > 1 for TF LegacyOptimizer in graph mode. ([#2401](https://github.com/horovod/horovod/pull/2401))
+
 ### Changed
 
 ### Deprecated
--- a/examples/tensorflow/tensorflow_mnist_estimator.py
+++ b/examples/tensorflow/tensorflow_mnist_estimator.py
@@ -111,7 +111,7 @@ def cnn_model_fn(features, labels, mode):
             learning_rate=0.001 * hvd.size(), momentum=0.9)
 
         # Horovod: add Horovod Distributed Optimizer.
-        optimizer = hvd.DistributedOptimizer(optimizer)
+        optimizer = hvd.DistributedOptimizer(optimizer, backward_passes_per_step=1)
 
         train_op = optimizer.minimize(
             loss=loss,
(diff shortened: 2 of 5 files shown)


tgaddair (approved):
Nice work!  Awesome to see feature parity among the different optimizers.

Richie-yan:
Hi, @aaron276h 
When I ran tensorflow_mnist_estimator.py using gradient accumulation, I set the backward_passes_per_step to 3 and encountered the following problems:
屏幕快照 2020-11-02 上午11 55 29
When the broadcast is shown in the above picture, the counter names of the two ranks are inconsistent, which leads to hang. I think it is related to the following code:
屏幕快照 2020-11-02 下午12 06 48
Does this code need to distinguish rank on the counter variable?

aaron276h:
@Richie-yan thanks for flagging this issue. Could you take a look at #2415, that should address the issue you are running into.

Richie-yan:
Hi, @aaron276h @tgaddair 
I tried to run TF's Roberta large model using gradient accumulation, and found a problem:
Set batch_size to 10：

- Do not use gradient accumulation, TF's bfc memory usage is about 23.3GB;

- Using gradient accumulation, TF's bfc memory usage is about 29.2GB.

According to my statistics, the gradient of the Roberta large model occupies about 1.17GB of GPU memory.
However, using gradient accumulation will take up an additional 5.9GB of GPU memory. 
When using gradient accumulation, the use of GPU memory does not meet expectations. What may be the cause?

tgaddair:
Good catch @Richie-yan.  I'm not sure it's possible to avoid this, as the gradients need to be stored into a separate variable in order to perform the local aggregation.  Do you have some thoughts on how this additional copy can be avoided?

Richie-yan:
@tgaddair 
thank you for your reply
I haven't thought of other methods that can replace this kind of copy.
However, I tried to modify the variable initialization method, and it seems that the use of video memory has returned to normal, from the original additional 5.9GB to about 1.3GB. My method is as follows:
屏幕快照 2020-11-17 上午10 22 48

aaron276h:
@Richie-yan that's really interesting, this should give the correct performance and if you are observing that this provides better memory performance we should definitely make this change. 

Seems to be potentially related to this old thread but it's not clear why we see 5x memory usage rather than 2x from using `tf.zeroes_initializer()`.

Richie-yan:
@aaron276h 
Yes, changing to the above method, I can further increase the batch_size of the model without OOM, thereby improving the throughput of the model.
I think we can further investigate why 5x memory usage is occupied using `tf.zeroes_initializer().`

PR #2405, opened 2020-10-29
Add debug log in operation.cc

CaptainDuke:
I encountered a Nan problem during training when I modified part of the source code. However, the try-cache here makes everything go well, no warnings or errors raised. PerformOperation() just makes callback like normal.

I found it difficult to locate the error so I added logs line by line. Finally I found my problem occured in InitializeBuffer due to incorrect CUDA driver API useage, which led to failure in ExecuteOperation. Then I fixed it and continued my development.

I believe it would be a useful infomation for other developers.

Thx!


tgaddair (approved):
LGTM!  Thanks for the PR.

PR #2413, opened 2020-11-01
Pin h5py < 3 to fix incompatibility with TensorFlow model serialization

tgaddair:
See: https://github.com/h5py/h5py/issues/1732


EnricoMi (approved):
LGTM

PR #2415, opened 2020-11-02
Make counter for local gradient aggregation a local variable.

aaron276h:
## Description
When using a an Estimator.RunHook to broadcast variables we broadcast all 
global variables. This PR marks the counter used for local gradient aggregation
as a local variable to avoid having it broadcasted as part of this callback.

Fixes issue raised in comments of #2401.

## Review process to land 


tgaddair (approved):
LGTM!

PR #2426, opened 2020-11-09
Use mxnet==1.5.0 for CPU Dockerfile until issues with 1.6.0 and 1.7.0 are resolved

tgaddair:
Fixes #2425.

Was able to reproduce locally.  Using v1.7.0.post1 had a separate error, but 1.5.0 worked fine.

PR #2427, opened 2020-11-09
Test MXNet 1.6 on CI

leezu:
Test if CI repro https://github.com/horovod/horovod/pull/2426


leezu:
@tgaddair do you know where
 ```
ERROR: The Compose file is invalid because:
Service test-cpu-openmpi-py3_6-tf2_1_0-keras2_3_1-torch1_4_0-mxnet1_5_0-pyspark2_4_7 has neither an image nor a build context specified. At least one must be provided.
```
comes from? I think I replaced all occurrences of `1_5_0` and `1.5.0` in the context of mxnet with 1.6 and am surprised that the 1_5 persists in the build environment.

PR #2435, opened 2020-11-10
Adding support for batched D2D memcopy kernel on GPU.

romerojosh:
This PR introduces a batched device-to-device (D2D) memcopy via a CUDA kernel to replace the individual calls to `cudaMemcpyAsync` when packing the fusion buffer. As with any type of kernel fusion, this reduces latency by limiting the number of small kernels calls and instead launching fewer, larger kernels. While these packing D2D memcopies are generally hidden behind compute, in high performance cases requiring low latency from Horovod, fusing these D2D memcopies improves performance. Even in more typical cases, this can still improve the performance by reducing any exposed Horovod processing time.

For performance and simplicity, this implementation will pad the destination fusion buffer address for each input tensor to a 16 byte aligned address to allow for using the 16 byte loads/stores in the CUDA kernel, assuming the input tensors are also 16 byte aligned. This should be true in most cases, but the kernel will adjust the load/store size if the input is not 16 byte aligned. 

Currently, the feature is toggle-able via environment variable `HOROVOD_BATCH_D2D_MEMCOPIES` but that can be removed. I've left it in for now to enable simpler performance testing/comparison.


horovod/common/ops/cuda/cuda_kernels.h
```diff
@@ -16,11 +16,26 @@
 #ifndef CUDA_KERNELS_H
 #define CUDA_KERNELS_H
 
+#include <cuda_runtime.h>
+
 #include "../../message.h"
 
+#define BATCHED_D2D_CAPACITY 160
```
tgaddair, line 23: I understand that the padding value of 16 corresponds to the load/store size in the CUDA kernel.  Is there a similar motivation for setting the capacity to 160, or is it an empirical choice?  Is it possible / desirable to make it configurable?

horovod/common/global_state.h
```diff
@@ -110,6 +110,9 @@ struct HorovodGlobalState {
   // benefit from a smaller chunk size.
   int64_t adasum_mpi_chunk_size = 1<<30;
 
+  // Enable use of batched d2d memcopy kernel on GPU
+  bool batch_d2d_memcopies = false;
```
tgaddair, line 114: Is the plan to change the default to true after more testing is done to verify performance improvements?  Is there a chance that performance could degrade with this setting?

tgaddair (approved):
LGTM!  What are your thoughts on making this the default behavior?

horovod/common/ops/cuda/cuda_kernels.h
romerojosh: Setting the capacity to 160 is required to keep the `BatchedD2DParams` struct size within the limits allowed by CUDA for arguments to kernels of 4KB. 160 is near the maximum capacity allowed.

horovod/common/global_state.h
romerojosh: I think we should enable this by default, though it might be good to allow it to be disabled via the env var just in case people see performance degradations. I haven't seen any performance degradations in my testing on V100 cards. I will update the PR to default this to true. This will also make sure testing is run using the new kernel, since right now it is not picking it up.

horovod/common/operations.cc
```diff
@@ -504,6 +504,14 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
            "allgather and hierarchical allreduce.";
   }
 
+  // Set flag to enable use of batched memcopy kernel on GPU
+  auto horovod_batch_d2d_memcopies =
+      std::getenv(HOROVOD_BATCH_D2D_MEMCOPIES);
+  if (horovod_batch_d2d_memcopies != nullptr &&
+      std::strtol(horovod_batch_d2d_memcopies, nullptr, 10) > 0) {
+    state.batch_d2d_memcopies = true;
```
tgaddair, line 512: Since the default is now `true`, we should also change this behavior such that if the user specifies `HOROVOD_BATCH_D2D_MEMCOPIES=0` it sets the value to `false`.

romerojosh: Yep, good catch. Just fixed this.

Issue #2419, opened 2020-11-05
horovod on Spark crash using Gloo when executors more than 12


qs-zh:
**Environment:**
1. Framework: TensorFlow
2. Framework version: 1.5.0
3. Horovod version: v0.20.3
4. MPI version: NA
5. CUDA version: NA
6. NCCL version: NA
7. Python version: 3.6
8. Spark / PySpark version: 2.4.6
9. OS and version: Ubuntu 16.04.6
10. GCC version: 5.4.0
11. CMake version: 3.17.1

**Bug report:**
Hi I'm doing model training on cpu machines using horovod on spark with Gloo.
horovod is installed using the following command:
```
HOROVOD_WITH_TENSORFLOW=1 HOROVOD_WITHOUT_PYTORCH=1 HOROVOD_WITHOUT_MXNET=1 HOROVOD_WITH_GLOO=1 HOROVOD_WITHOUT_MPI=1 pip install horovod
```
Model training is working fine when the number of executors are <=12. But when I increase the number of executors, I'll receive the following error:
at driver node:

```
Checking whether extension tensorflow was built with Gloo.
Extension tensorflow was built with Gloo.
Exception in thread Thread-3:
Traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File "/data2/yarn/nm/usercache/ld-dev/appcache/application_1603883037311_28420/container_e1322_1603883037311_28420_01_000008/Python/lib/python3.6/site-packages/horovod/spark/runner.py", line 158, in run_spark
    result = procs.mapPartitionsWithIndex(mapper).collect()
  File "/data2/yarn/nm/usercache/ld-dev/appcache/application_1603883037311_28420/container_e1322_1603883037311_28420_01_000008/Python/lib/python3.6/site-packages/pyspark/rdd.py", line 816, in collect
    sock_info = self.ctx._jvm.PythonRDD.collectAndServe(self._jrdd.rdd())
  File "/data2/yarn/nm/usercache/ld-dev/appcache/application_1603883037311_28420/container_e1322_1603883037311_28420_01_000008/Python/lib/python3.6/site-packages/py4j/java_gateway.py", line 1257, in __call__
    answer, self.gateway_client, self.target_id, self.name)
[...]

PR #2436, opened 2020-11-11
Provide the timeout value in useful places

EnricoMi:
Providing the timeout value in logging and exceptions helps to identify which timeout is actually being used and causes the exception.

Also adds `spark_require_list` to `dev_require_list` so that `pip install -e .[dev]` also installs spark packages.
Further sets `h5py<3` in `spark_require_list`.

Improves resolving issues like #2419.


EnricoMi:
The failing `test_colocator` test is handled in #2438.

horovod/common/gloo/http_store.cc
```diff
@@ -49,8 +49,9 @@ void HTTPStore::wait(const std::vector<std::string>& keys,
     const auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(
         std::chrono::steady_clock::now() - start);
     if (timeout != gloo::kNoTimeout && elapsed > timeout) {
-      GLOO_THROW_IO_EXCEPTION(GLOO_ERROR_MSG("Wait timeout for key(s): ",
-                                             ::gloo::MakeString(keys)));
+      auto timeout_seconds = std::chrono::duration_cast<std::chrono::seconds>(timeout);
```
tgaddair, line 52: Nice!  Can you also add a suggestion here to try increasing `HOROVOD_GLOO_TIMEOUT_SECONDS`?

EnricoMi: Ideally, that should be in the exception message, but this part of the code does not know (and should not make this assumption) where the timeout value is coming from, it is just an argument to this method. The env var is defined and read in `gloo_context.cc`.

Ideally, the timeout exception should be caught there and re-thrown with an enhanced message saying where to set the timeout. For that, the exception must be of a very specific type, like gloo timeout exception for that to work.

My hope is seeing the timeout at all in the logs helps to realize that whatever timeout knob you used is the wrong one. Not ideal but better than nothing.

Any better idea?

tgaddair: We could move the reading of `HOROVOD_GLOO_TIMEOUT_SECONDS` into this class.  It's never set to anything else in practice.

EnricoMi: There is no gloo:TimeoutException, that would have been too good... Then I would rather add the env var to the error message, though it is bad practise, but moving the env var in this code would be worse I think.

tgaddair: I think having the env variable in the error is very useful.  Otherwise users will not know how to resolve the issue, and will ask questions on GitHub that could be answered by the error message (this has happened a few times).  I agree it's making an assumption that isn't guaranteed by the interface, but I think it's worth it.  What do you think?

EnricoMi: Yes, the information is more useful for users than clean code separation for developers in this case, I will go for it.

EnricoMi: I had to move the constants into `gloo_context.h`: https://github.com/horovod/horovod/pull/2436/commits/6cd9213d634397a9ac67c9472c66a4df4d6d4b3a

tgaddair (approved):
LGTM! Thanks for putting this together.

Issue #2438, opened 2020-11-11
Unit Test: Ray Colocator test gives flaky resource availability results


tgaddair:
Starting from nightly build 2020/11/10, the `test_ray.py::test_colocator` has been failing sporadically with the following error:

```
assert resources.get(f"node:{ip_address}", 0) == 1 - 4 * 0.01
assert 1.0 == 0.96
  +1.0
  -0.96
```

See: https://buildkite.com/horovod/horovod/builds/4274#3f5eddde-efbd-4078-861a-8d79205d8478

cc @richardliaw

richardliaw:
this slipped my mind - let me take a look today.

EnricoMi:
This is still enabled and still flaky: https://github.com/horovod/horovod/runs/1659627755

richardliaw:
@EnricoMi thanks for the ping! I've opened a PR to disable in #2628.

PR #2439, opened 2020-11-11
Disable flaky checks in Ray colocator test

tgaddair:
See #2438.


EnricoMi (approved):
Properly disabled and documented, thanks!

PR #2453, opened 2020-11-16
Add multi-tensor hvd.grouped_allreduce API.

romerojosh:

## Description

This PR introduces a new API to Horovod, `hvd.grouped_allreduce`. The purpose for this API is to give users explicit control over how Horovod fuses (or "groups") tensors for allreduce. Specifically, a list of tensors provided to `hvd.grouped_allreduce` will be treated logically as a single request, and will only be processed by the backend when all tensors in the list are available. This in contrast to Horovod's normal process, which will greedily fuse any available tensors during a cycle. While this greedy fusing is appropriate in many situations, a number of circumstances can arise where users may want greater control over how this fusion is done. 

One situation is where a user wants to reduce the latency of Horovod coordination/negotiation via reducing the `HOROVOD_CYCLE_TIME`, but also wants to ensure that fused allreduce message do not become too small. This is not currently possible as the fusion message sizes and cycle time are tightly coupled. By defining explicit groups, the user is free to reduce the cycle time to as low a value as required for faster negotiation/coordination, while maintaining reasonable message sizes for network efficiency.

A second situation is when a user wants deterministic operation from Horovod. As it has been established previously, the dynamic packing of the fusion buffer can cause allreduce results to be non-deterministic, as the location of tensors in the fusion buffer can impact summation order. The only method to get deterministic results from Horovod right now is to disable fusion completely (via setting `HOROVOD_FUSION_THRESHOLD=0`), with an associated loss in performance. By explicitly defining the fusion groups via this API, deterministic fusion can be achieved, as we have a mechanism to guarantee that the fusion buffers will be packed with a deterministic ordering (both iteration to iteration, and run to run). It should be noted that by default, this feature will allow groups to fuse into larger groups, which reintroduces non-determinism. To disable this and run with deterministic groups, the environment variable `HOROVOD_DISABLE_GROUP_FUSION` has been added. 

This feature is available both as a direct Horovod operation `hvd.grouped_allreduce` and through an additional argument (`num_groups`) to `hvd.DistributedOptimizer/hvd.DistributedTrainer`. By setting `num_groups`, Horovod will split the list of gradient tensors into the requested number of groups and use appropriate `horovod.grouped_allreduce` call to perform the gradient averaging operation. 

This PR supersedes #1130 and updates to a cleaner API while also completing support for MXNet and PyTorch along with TensorFlow.


horovod/torch/functions.py
```diff
@@ -68,6 +67,7 @@ def broadcast_optimizer_state(optimizer, root_rank):
         root_rank: The rank of the process from which the optimizer will be
                    broadcasted to all other processes.
     """
+    from horovod.torch.optimizer import DistributedOptimizer
```
tgaddair, line 70: Why is this here?

horovod/torch/optimizer.py
```diff
@@ -101,6 +106,29 @@ def set_backward_passes_per_step(self, passes):
...
+            # Get list of parameters with grads
+            for param_group in self.param_groups:
+                for p in param_group['params']:
+                    if p.requires_grad:
+                        p_list.append(p)
+
+            # To ensure parameter order and group formation is consistent, broadcast p_list order
+            # from rank 0 and use for every worker
+            p_list_names = [self._parameter_names.get(p) for p in p_list]
+            p_list_names = broadcast_object(p_list_names, root_rank=0)
```
tgaddair, line 121: This helper function has turned out to be really useful.

horovod/torch/optimizer.py
```diff
@@ -101,6 +106,29 @@ def set_backward_passes_per_step(self, passes):
...
+            p_list = sorted(p_list, key=lambda p : p_list_names.index(self._parameter_names.get(p)))
+
+            # Form groups
+            d, r = divmod(len(p_list), self._num_groups)
+            p_groups = [tuple(p_list[n * d + min(n, r):(n + 1) * d + min(n + 1, r)]) for n in range(self._num_groups)]
```
tgaddair, line 126: I noticed this code in a few other places, would it make sense to abstract into a utility function in `common`?

horovod/common/group_table.cc
```diff
@@ -0,0 +1,113 @@
...
+  std::lock_guard<std::mutex> guard(mutex_);
+  return id_to_tensor_names_.at(group_id);
+}
+
+bool GroupTable::empty(void) const {
+  std::lock_guard<std::mutex> guard(mutex_);
+  return tensor_name_to_id_.empty();
+}
+
+int32_t GroupTable::RegisterGroup(const std::vector<std::string>& tensor_names) {
```
tgaddair, line 43: Not a huge deal, but would be nice to consolidate a lot of duplication between these two functions.

horovod/_keras/__init__.py
```diff
@@ -28,7 +28,8 @@
 def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sparse,
                                  compression, sparse_as_dense, gradient_predivide_factor,
                                  op, backward_passes_per_step=1,
-                                 average_aggregated_gradients=False):
+                                 average_aggregated_gradients=False,
+                                 num_groups=0):
```
tgaddair, line 32: I'm wondering if instead of `num_groups` we might want to name it `groups` in case we want to overload the semantics in the future.  For instance, allowing the user to provide explicits groups, or some other grouping criteria.  What do you think?  We can always cross the bridge if we come to it.

tgaddair (commented):
LGTM!  No blockers from me, just a few nits.  Feel free to land when ready.

tgaddair (approved):

horovod/torch/functions.py
romerojosh: I had issues when I tried importing `broadcast_object` in `optimizer.py` due to a circular dependency on `DistributedOptimizer`. To fix this, I limited the import of `DistributedOptimizer` in this file to the scope of the function that needs it only.

horovod/_keras/__init__.py
romerojosh: I'd prefer to leave it `num_groups` for now for simplicity. I think we may want to use a target message size as an alternative grouping criteria in the future, but the argument name `groups` doesn't make a lot of sense in that context either, but rather `group_size_mb` or something like that. As far as explicitly defining groups, the list of gradient tensors within the optimizer are not easily accessible to the user at the time the `DistributedOptimizer` is constructed to form groups. However, more explicit control can still be achieved if desired through manual use of Horovod primitives instead of the `DistributedOptimizer` wrapper.

horovod/common/group_table.cc
romerojosh: Actually, I can just delete the duplicate functions as only one variant of the register and deregister group functions are being used anyway.

horovod/torch/optimizer.py
romerojosh: Yes, I agree. 👍

horovod/torch/optimizer.py
romerojosh: Sure, I can do that.

horovod/_keras/__init__.py
tgaddair: Sounds good, that sounds reasonable to me.

horovod/torch/functions.py
tgaddair: Ah, gotcha.  We can probably move things around in a follow-up PR to address that, but this is good for now.

PR #2455, opened 2020-11-17
Reduce memory usage for TF1 gradient aggregation.

aaron276h:
## Description
As pointed out by @Richie-yan in https://github.com/horovod/horovod/pull/2401, using `tf.zeros_initializer()` introduces additional memory overhead.


tgaddair (approved):
Thanks for the quick fix!

PR #2458, opened 2020-11-17
Remove deprecated

tgaddair:
Removes the following deprecated features:

- DDL allreduce implementation
- Parameter `average` for `allreduce` (replaced with `op=Average`)
- `--log-hide-timestamp` arg to `horovodrun`


horovod/runner/launch.py
```diff
@@ -213,32 +211,6 @@ def make_override_false_action(override_args):
     return make_override_bool_action(override_args, False)
 
 
-def make_deprecated_bool_action(override_args, bool_value, replacement_option):
```
EnricoMi, line 216: Why are you removing this method? We may want to deprecate another bool arg in the future and then have to come up with this method again. Worth keeping this un-used code?

tgaddair: That's a good point.  It's unused for now, but we can keep it for future params.

PR #2459, opened 2020-11-17
Require initial_lr parameter to LearningRateScheduleCallback

tgaddair:


romerojosh (approved):
LGTM. We should just make sure to emphasize this change in the release notes.

PR #2468, opened 2020-11-19
Changed default cycle time from 5ms to 1ms and fusion threshold from 64MB to 128MB

tgaddair:
It's been observed empirically that the default cycle time of 5ms is too long, resulting in excessively long pauses between allreduce calls.

Similarly, the fusion buffer threshold of 64MB if often too small for models such as ResNet50 with fp32 gradients.


romerojosh (approved):
Awesome. LGTM!

PR #2483, opened 2020-11-26
provide a way for ever-blacklisted hosts be available again

woodlgz:

## Description

 Currently blacklisted nodes are blocked forever.
 Howerver, such a node may just be out of service temporarily (due to accident like a quick blackout) and may recover soon. 
 This issue provide a way to bring such recoverable nodes back.

related to issue https://github.com/horovod/horovod/issues/1926

## Review process to land 


tgaddair:
Thanks for the PR @woodlgz!  I like the idea of introducing the health check.  I will take a look at this soon.

zw0610:
I would really have this feature introduced into horovod elastic. Will there be any further actions?

zw0610:
@tgaddair do you think we can still move forward with this feature?

PR #2433, opened 2020-11-10
Updated oneCCL to use the new API

shirosankaku:
Signed-off-by: Yana Shchyokotova <yana.shchyokotova@intel.com>

Updated oneCCL related files with oneCCL new API. No need to update docs and write extra tests.

oneCCL v1.0 specification
https://oneapi-src.github.io/oneAPI-spec/elements/oneCCL/source/index.html

Upd: oneCCL repo in Dockerfile.test.cpu will be changed soon to the official one.

PR #2487, opened 2020-11-27
Temporarily disable oneCCL tests

tgaddair:
CI is failing due to failure to import oneCCL.  We need to disable for now until #2433 lands.


EnricoMi (approved):
LGTM!

PR #2498, opened 2020-12-03
Jump out early to reduce unnecessary loop overhead in NCCLAllgather

firejq:


tgaddair (approved):
Thanks for the fix, @firejq!

horovod/common/ops/nccl_operations.cc
```diff
@@ -498,6 +498,7 @@ Status NCCLAllgather::Execute(std::vector<TensorTableEntry>& entries,
     for (int rc = 1; rc < global_size; ++rc) {
       if (tensor_sizes[ec * global_size + rc] != tensor_sizes[ec * global_size]) {
         same_shape = false;
+        break;
```
romerojosh, line 501: Putting the `break` in this inner loop is saving some overhead, but it still leaving a lot there. There is still the outer loop over `entries.size()` that runs in entirety along with some partial run of the inner loop over `global_size` up until the `break` condition is hit. It would be better to put the break on the outer loop with condition `if (same_shape == false) break;`.

firejq: Thanks! It's awesome! 
Since this pr has been merged, I will put this code to another one.

PR #2501, opened 2020-12-03
Add breaking for outer loop in NCCLAllgather

firejq:
Thanks for @romerojosh 's [suggestion](https://github.com/horovod/horovod/pull/2498#discussion_r535452791) in #2498 !

/CC  @tgaddair


romerojosh:
LGTM but could you rebase and sign the DCO to pass checks?

firejq:
Sorry that I will close this because It seems that some of my misoperations caused pollution of commit logs.

PR #2503, opened 2020-12-04
Add breaking for outer loop in NCCLAllgather

firejq:
Apply according to @romerojosh 's [suggestion](https://github.com/horovod/horovod/pull/2498#discussion_r535452791) in #2498 .


romerojosh (approved):
LGTM!

PR #2522, opened 2020-12-14
fixup: CMake header search for flatbuffer

chongxiaoc:
Only need to include flatbuffer header path if it is not provided in
tensorflow path.

## Description

Fixes # (issue).

## Review process to land 


chongxiaoc:
@tgaddair By looking at Cmake, I found I missed the same condition of adding flatbuffer headers as Eigen.
The reason is that as today, flatbuffer is not provided by tensorflow, so local test always adds additional path.
The missed condition is added in case in future tensorflow would provide flatbuffer, so horovod doesn't have to include redundant headers path.

tgaddair (approved):

Issue #2132, opened 2020-07-23
Add unit tests for DistributedOptimizer


tgaddair:
DistributedOptimizer has become an increasingly complicated class that does things like convert sparse tensors to dense, branching code paths for eager vs graph execution, etc.  However, none of this is under direct unit testing.

We should add unit tests that explicitly capture and verify all major code paths in the DistributedOptimizer.

Young768:
Is there any update on this issue? I have modified my own DistributedOptimizer and want to verify it via unit-test.

maxhgerlach:
@Young768, I don't think there is complete test coverage yet, but if you just search for `DistributedOptimizer` in the files in `test/parallel`, you will see that several unit tests have been added in the mean time.

PR #2523, opened 2020-12-15
Add group_lists parameter in DistributedOptimizer for customized grouping

zhuzilin:

## Description

This PR adds a `group_lists` parameter to `DistributedOptimizer` and `DistributedGradientTape` for pytorch and tensorflow. The `group_lists` is an extension to `num_groups`, which was introduced in #2453 .

The current `num_groups` parameter allows users to statically control the gradient fusion group and will evenly split the number of parameters by the number the user specified. It is a great api, because we do need to statically group certain grads in some optimization. However, we often need to group them by size or by the semantic of the model (e.g. group every resnet block). Therefore, I believe it would better if horovod could give user more freedom in the strategy of grouping.

This PR introduces a `group_lists` parameter to `DistributedOptimizer` and `DistributedGradientTape`, which will takes a list of list of parameters, the parameters in the inner list will be put into the same group and the unmentioned ones will themselves form a group each. For instance, if we have 3 dense layers:
```
layer 1: dense_1/weight, bias_1/bias
layer 2: dense_2/weight, bias_2/bias
layer 3: dense_3/weight, bias_3/bias
```
The following group will results in:
```python
group_lists:
[[dense_1/weight, bias_1/bias],  # list 0
 [dense_2/weight, bias_2/bias]]  # list 1
resulting groups:
[[dense_1/weight, bias_1/bias],  # group 0
 [dense_2/weight, bias_2/bias],  # group 1
 [dense_3/weight],               # group 2
 [bias_3/bias]]                  # group 3
```
With this parameter, users could decide how they would like to group the grad on their own and in python.

I think it would be simpler to disallow setting `num_groups` and `group_lists` at the same time, otherwise we may need the user to avoid the possible conflict.

As for the implementation detail, to make the pytorch and tf api coherent, for torch, the argument would be a list of list of `torch.Tensor` and for tf it would be a list of list of `tf.Variable`. And I have to removed the `@_cache` decorator on `_make_allreduce_grads_fn` because neither list nor `tf.Variable` is hashable. 

I believe this PR would introduce more possiblities to horovod and any discussion on this is welcomed :).

Thank you for your time on reviewing this PR.

cc @romerojosh @tgaddair 

## Review process to land 


romerojosh (commented):
@zhuzilin I think this PR looks great and is a welcome extension to the grouping functionality! 

@tgaddair, I remembered a comment you made on the original grouping PR here about the naming of the `num_groups` to something more generic to facilitate this type of enhancement (https://github.com/horovod/horovod/pull/2453#discussion_r526412585). Perhaps we should go ahead and start the process of deprecating `num_groups` in favor of a single argument in this PR?

tgaddair:
Nice work @zhuzilin!  I like @romerojosh's suggestion about consolidating the params.  What does everyone think about this change:

1. Rename `group_lists` to `groups`
2. If `groups` is an `int`, we treat it as `num_groups`
3. Make `num_groups as deprecated with the intent of having both options use the `groups` param

zhuzilin:
@tgaddair I agree with these change. There is one more thing about `@_cache`. We cannot use a list in the key of a dictionary, therefore I've simply remove the `@_cache` on `_make_allreduce_grads_fn` at the moment. I wonder if there will be any unneglectable negative impact to performance and if there are any better solution.

horovod/tensorflow/__init__.py
```diff
@@ -330,10 +330,9 @@ def after_create_session(self, session, coord):
             session.run(self.bcast_op)
 
 
-@_cache
 def _make_allreduce_grads_fn(name, device_dense, device_sparse,
                              compression, sparse_as_dense, op, gradient_predivide_factor,
-                             num_groups):
+                             num_groups, group_lists=None):
```
tgaddair, line 335: Hey @zhuzilin, instead of removing the `@_cache`, you should be able to make the `group_lists` hashable using a simple trick.  First we can define these helper functions in `horovod/tensorflow/util.py`:

```
def vars_to_refs(vars):
    return tuple(v.ref() for v in vars)

def refs_to_vars(refs):
    return [r.deref() for r in refs]
```

Then we can use these functions to wrap the `_make_allreduce_grads_fn`:

```
@_cache
def _make_cached_allreduce_grads_fn(..., groups):
    groups = refs_to_vars(group_refs) if isinstance(groups, tuple) else groups
    ...

def _make_allreduce_grads_fn(...):
    groups = vars_to_refs(groups) if isinstance(groups, list) else None
    return _make_allreduce_grads_fn(..., groups)
```

Does that make sense?

zhuzilin:
@tgaddair @romerojosh I've upated the code and merged the `num_groups` and `group_lists` into `groups`. Could you take another look?

zhuzilin:
FYI, I updated the code because the output of `allreduce_grads` in tensorflow did not contain those `None`-grads. Notice that this is also a bug in the current release.

horovod/tensorflow/__init__.py
```diff
@@ -401,15 +432,15 @@ def __init__(self, optimizer, name=None, use_locking=False, device_dense='',
                     device_sparse='', compression=Compression.none,
                     sparse_as_dense=False, op=Average, gradient_predivide_factor=1.0,
                     backward_passes_per_step=1, average_aggregated_gradients=False,
-                    num_groups=0):
+                    groups=None):
```
romerojosh, line 435: We shouldn't completely remove the `num_groups` argument in this PR. Instead, assign `num_groups` to `groups` at the beginning of the `__init__` function and print a deprecation warning that `num_groups` will be removed in a future Horovod version and users should use `groups` instead. For example:
```
if num_groups != 0:
    warnings.warn('Parameter `num_groups` has been replaced by `groups` and will be removed in v0.23.0.', DeprecationWarning)
```

horovod/tensorflow/__init__.py
```diff
@@ -343,26 +344,47 @@ def _make_allreduce_grads_fn(name, device_dense, device_sparse,
...
+                    index_group, grad_group = [list(t) for t in zip(*group)]
+                    reduce_ops_group = _grouped_allreduce_cond(grad_group,
+                                                               device_dense=device_dense,
+                                                               device_sparse=device_sparse,
+                                                               compression=compression,
+                                                               op=op,
+                                                               prescale_factor=prescale_factor,
+                                                               postscale_factor=postscale_factor)
+                    for i in range(len(index_group)):
+                        reduce_ops[index_group[i]] = reduce_ops_group[i]
```
romerojosh, line 387: The `reduce_ops` list does not need to be submitted to TF in a sorted order. The TF scheduler should handle operation ordering automatically per worker and the Horovod runtime can handle out of order allreduce requests between workers.

zhuzilin: @romerojosh 
We will zip the variables and gradients with `list(zip(avg_grads, vars))` in `compute_gradients`, where `avg_grads` is the return value of `_allreduce_grads`. Therefore, we need to make sure the reduced grad is in the same order of the input vars.

horovod/tensorflow/__init__.py
zhuzilin: I have updated the code for the public APIs. Shall I retain the `num_groups` in constructor of inner classes like `_DistributedOptimizer`?

romerojosh: We should remove `num_groups` from inner classes at this point. You may need to modify some files in the TF Keras implementation, like this here: https://github.com/horovod/horovod/blob/2a775b2e059817093e476a5a44433426dda35ae5/horovod/_keras/__init__.py#L40-L48
I realize that the grouping functionality wasn't actually added to the Keras `DistributedOptimizer` constructor, but you don't need to worry about that here. We can fix that in a follow up PR.

horovod/tensorflow/__init__.py
romerojosh: Ok, now I see the issue you mentioned about missing the `None` grad tensors in the returned list in the groups path. Thanks for catching that!

horovod/tensorflow/__init__.py
zhuzilin: @romerojosh 
I just tried to add the support for keras. As for mxnet, I have no experience with it, so there is nothing I can do...

BTW, I've also made some changes in `LocalGradientAggregationHelper` and `LocalGradientAggregationHelperEager`, because we'll need variable lists as inputs to `allreduce_grads`. I wonder if there is any unitest I can add for the new functionality, because for now I am just running code in the example folder with different settings and that is not comprehensive enough...

zhuzilin:
@romerojosh @tgaddair Could you have another look at this PR? Thank you :)

horovod/tensorflow/__init__.py
romerojosh: @zhuzilin No worries on MXNet and thanks for covering TF and PyTorch. Indeed, there is a big gap in unit testing the `DistributedOptimizer`, especially the with any optional arguments. We have an issue (#2132) open for it, but we still haven't added more testing. Your changes here look good to me so this is not a blocker for me.

@tgaddair I think we should make an effort to resolve #2132 and get more extensive tests put together before the next major release.

romerojosh (commented):
@zhuzilin Sorry for the delay in reviewing. Thanks for making updates! The changes look good to me, but it looks like there are a couple of conflicts that need to be resolved. Can you rebase your branch off the current master? This should pick up CI build/test fixes as well.

zhuzilin:
@romerojosh I've rebased the current branch. And the only test fail is `test_fault_tolerance_hosts_remove_and_add`:

```
ray_8_cpus = None

    @pytest.mark.skipif(
        not gloo_built(), reason='Gloo is required for Ray integration')
    def test_fault_tolerance_hosts_remove_and_add(ray_8_cpus):
        with fault_tolerance_patches():
            discovery_schedule = [
                (10, ['host-1:2', 'host-2:1', 'host-3:2']),
                (10, ['host-1:2']),
                (None, ['host-1:2', 'host-4:1', 'host-5:1']),
            ]
            nics = list(psutil.net_if_addrs().keys())[0]
    
            settings = ElasticRayExecutor.create_settings(min_np=1, nics={nics})
            settings.discovery = SimpleTestDiscovery(discovery_schedule)
            executor = ElasticRayExecutor(
                settings, cpus_per_slot=1, override_discovery=False)
    
            training_fn = _create_training_function(iterations=30)
            executor.start()
            trace = StatusCallback()
            results = executor.run(training_fn, callbacks=[trace])
            assert len(results) == 4
    
            events = trace.fetch()
>           assert sum(int("started" in e) for e in events) == 7, events
E           AssertionError: [{'pid': 3793, 'started': True}, {'pid': 3791, 'started': True}, {'pid': 3798, 'started': True}, {'pid': 3797, 'started': True}, {'pid': 3791, 'training': True}, {'pid': 3797, 'training': True}, ...]
E           assert 4 == 7
E             +4
E             -7

test_ray_elastic.py:266: AssertionError
```

I'm not sure if this error has anything to do with the changes in this PR. Could you give me some help? Thank you!

romerojosh (approved):
LGTM @zhuzilin! Thanks for the contribution.

I think the `test_fault_tolerance_hosts_remove_and_add` failure is unrelated to your changes. Could you verify this @tgaddair?

tgaddair:
It looks like it was a flakiness issue.  The overall Buildkite tests ran successfully, indicating the test was retried and passed.

tgaddair (approved):

PR #2376, opened 2020-10-14
Support Store.create("dbfs:/") mapping to DBFSLocalStore("/dbfs/...")

liangz1:
## Checklist before submitting

- [x] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)?
- [x] Did you update the docs?
- [x] Did you write any tests to validate this change?  
- [x] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users?

## Description

This PR allows `Store.create()` to create `DBFSLocalStore` when both of the following conditions are met:
* The code is running in Databricks Runtime
* The `prefix_path` matches any of the following patterns:
  * `/dbfs/...`
  * `file:///dbfs/...`
  * `dbfs:/...`

Fixes # (issue).

## Review process to land 

1. All tests and other checks must succeed.
2. At least one member of the [technical steering committee](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md) must review and approve.
3. If any member of the technical steering committee requests changes, they must be addressed.

PR #2526, opened 2020-12-15
Avoid overwriting methods in the super class of DBFSLocalStore

liangz1:

## Description

The issue is that on Databricks, DBFSLocalStore cannot correctly write intermediate files due to the previous PR #2376. I updated DBFSLocalStore to avoid overwriting the methods in the superclass: `matches` and `get_localized_path`, because those are needed for the underlying LocalStore to work correctly. The dbfs parsing logic should be defined in dedicated methods.

I manually tested the change on Databricks and it works well.

## Review process to land 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Remove ssh reliance for Ray elastic training ([#2528](https://github.com/horovod/horovod/pull/2528))
 
+- Fixed "Intermediate path does not exist" error with DBFSLocalStore. ([#2526](https://github.com/horovod/horovod/pull/2526))
+
 ## [v0.21.0] - 2020-11-23
 
 ### Added
--- a/horovod/spark/common/store.py
+++ b/horovod/spark/common/store.py
@@ -144,7 +144,7 @@ class Store(object):
     def create(prefix_path, *args, **kwargs):
         if HDFSStore.matches(prefix_path):
             return HDFSStore(prefix_path, *args, **kwargs)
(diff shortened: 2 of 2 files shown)


tgaddair (approved):
LGTM!  @liangz1, can you sign the DCO for this commit?  Also, we will need to rerun tests once #2530 lands.

liangz1:
The buildkite tests fail to build with torchvision errors, seems #2531 is fixing it.

@tgaddair Could you suggest the estimate:
1. When will the next maintenance version including this patch be released?
2. When will the horovod version supporting TF 2.4 with CUDA 11, and PyTorch 1.7.1 released?

Thanks!

tgaddair:
> The buildkite tests fail to build with torchvision errors, seems #2531 is fixing it.
> 
> @tgaddair Could you suggest the estimate:
> 
> 1. When will the next maintenance version including this patch be released?
> 2. When will the horovod version supporting TF 2.4 with CUDA 11, and PyTorch 1.7.1 released?
> 
> Thanks!

Hey @liangz1,

I'd like to release v0.21.1 early next week once this PR and a few others are landed.

Horovod is already compatible with TF 2.4, CUDA 11, and PyTorch 1.7.1 to my knowledge.  Have you experienced issues with any of these versions?

liangz1:
Thanks for sharing the info! I just want to confirm the supported version before upgrading.

liangz1:
@tgaddair There seems to be an error with the unit test report generation. Can you double-check and suggest whether this PR is ready for merging?

tgaddair (approved):
LGTM!  There was an update to the workflow code after this PR was created, so I suspect that was the issue.

PR #2529, opened 2020-12-16
add support to check supported frameworks versions (#2518)

### LINKED ISSUE #2518: Improve error handling when user upgrades framework without reinstalling Horovod
When the user installs Horovod with a certain version of a framework, then upgrades the framework without reinstalling Horoovd, an error will commonly occur, for example: https://github.com/horovod/horovod/issues/236

We should track the version of the framework at install time within a metadata file stored in the Horovod site-package folder.  At runtime, we can check this before loading the native library, e.g., here:

https://github.com/horovod/horovod/blob/master/horovod/tensorflow/mpi_ops.py#L46

If an error occurs here, we can catch it and re-raise, telling the user to reinstall Horovod.  If there is a version mismatch but no error, we should issue a warning to the user.

chongxiaoc:
Check if the versions of supported frameworks match with horovod
installation. Throw out a warning to let users know that version is
mismatched, which could result in a possible undefined-symbol error
when using the framework.

## Description

Fixes #2518 .

## Review process to land 

### DIFF
--- a/horovod/common/exceptions.py
+++ b/horovod/common/exceptions.py
@@ -29,3 +29,20 @@ class HostsUpdatedInterrupt(RuntimeError):
     In elastic mode, this will result in a reset event without a restore to committed state.
     """
     pass
+
+
+def get_version_mismatch_message(name, version, installed_version):
+    return f'Framework {name} installed with version {installed_version} but found version {version}.\n\
+             This can result in unexpected behavior including runtime errors.\n\
+             Reinstall Horovod using `pip install --no-cache-dir` to build with the new version.'
+
+
+class HorovodVersionMismatchError(Exception):
+    """Internal error raised when the runtime version of a framework mismatches its version at
+    Horovod installation time.
+    """
+    def __init__(self, name, version, installed_version):
+        super().__init__(get_version_mismatch_message(name, version, installed_version))
+        self.name = name
+        self.version = version
+        self.installed_version = installed_version
--- a/horovod/common/util.py
+++ b/horovod/common/util.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 # =============================================================================
 
+import json
 import multiprocessing
 import os
 import sys
@@ -22,6 +23,8 @@ import sysconfig
 import warnings
 
 from contextlib import contextmanager
+
+from horovod.common.exceptions import get_version_mismatch_message, HorovodVersionMismatchError
 
 
 EXTENSIONS = ['tensorflow', 'torch', 'mxnet']
@@ -251,3 +254,15 @@ def split_list(l, n):
     """
     d, r = divmod(len(l), n)
     return [l[i * d + min(i, r):(i + 1) * d + min(i + 1, r)] for i in range(n)]
+
+
+def check_installed_version(name, version, exception=None):
+    file_path = os.path.abspath(os.path.join(os.path.dirname(os.path.abspath(__file__)),\
+        os.pardir, "metadata.json"))
+    with open(file_path) as f:
+        installed_version = json.load(f).get(name)
+        if installed_version != version:
+            if exception is None:
+                warnings.warn(get_version_mismatch_message(name, version, installed_version))
+            else:
+                raise HorovodVersionMismatchError(name, version, installed_version) from exception
--- a/horovod/tensorflow/mpi_ops.py
+++ b/horovod/tensorflow/mpi_ops.py
@@ -23,8 +23,8 @@ from tensorflow.python.framework import load_library
 from tensorflow.python.framework import ops
 from tensorflow.python.platform import resource_loader
 
-from horovod.common.util import get_ext_suffix, get_average_backwards_compatibility_fun, gpu_available, \
-    num_rank_is_power_2
+from horovod.common.util import check_installed_version, get_ext_suffix, \
+    get_average_backwards_compatibility_fun, gpu_available, num_rank_is_power_2
 from horovod.common.basics import HorovodBasics as _HorovodBasics
 from horovod.tensorflow.util import _executing_eagerly
 
@@ -42,8 +42,14 @@ def _load_library(name):
     library = load_library.load_op_library(filename)
     return library
 
-
-MPI_LIB = _load_library('mpi_lib' + get_ext_suffix())
+# Check possible symbol not found error from tensorflow version mismatch
+try:
+    MPI_LIB = _load_library('mpi_lib' + get_ext_suffix())
+except Exception as e:
+    check_installed_version('tensorflow', tf.__version__, e)
+    raise e
+else:
+    check_installed_version('tensorflow', tf.__version__)
 
 _basics = _HorovodBasics(__file__, 'mpi_lib')
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Remove ssh reliance for Ray elastic training ([#2528](https://github.com/horovod/horovod/pull/2528))
 
+- Fixed error handling for changing framework without reinstalling horovod ([#2529](https://github.com/horovod/horovod/pull/2529))
+
 ## [v0.21.0] - 2020-11-23
 
 ### Added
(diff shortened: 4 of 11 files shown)


chongxiaoc:
@tgaddair 
The metadata file is generated, though cmake file write is a bit of messy.
To do:
- Standardize the part of raising error, print of stderr might not be a standard way.
- Add additional unit tests.
- If possible, generating metadata file in setup.py, not cmake, which can be more neat.

chongxiaoc:
It seems that some compilcation error was introduced by missing intel oneapi header, for example:
```
[0m[91m  /usr/local/lib/python3.6/dist-packages/mxnet/include/mkldnn/oneapi/dnnl/dnnl.hpp:23:10: fatal error: oneapi/dnnl/dnnl_config.h: No such file or directory
[0m[91m   #include "oneapi/dnnl/dnnl_config.h"
[0m[91m            ^~~~~~~~~~~~~~~~~~~~~~~~~~~
[0m[91m  compilation terminated.
[0m[91m  horovod/mxnet/CMakeFiles/mxnet.dir/build.make:758: recipe for target 'horovod/mxnet/CMakeFiles/mxnet.dir/mpi_ops.cc.o' failed
  make[2]: *** [horovod/mxnet/CMakeFiles/mxnet.dir/mpi_ops.cc.o] Error 1
```

tgaddair:
@chongxiaoc please see #2530 regarding the MXNet issue.

horovod/common/util.py
```diff
@@ -251,3 +253,11 @@ def split_list(l, n):
...
     d, r = divmod(len(l), n)
     return [l[i * d + min(i, r):(i + 1) * d + min(i + 1, r)] for i in range(n)]
+
+
+def check_framework_version(name, version):
+    file_path = os.path.abspath(os.path.dirname(os.path.abspath(__file__)))+"/../metadata.json"
+    with open(file_path) as f:
+        install_version = json.load(f)
+        if not install_version[name] == version:
+            raise HorovodVersionMismatch(install_version[name])
```
tgaddair, line 263: We should avoid raising a`RuntimeWarning`.  Instead, you can do something like:

```
warnings.warn(f'Framework {name} installed with version {installed_version} but found version {version}.  This can result in unexpected behavior including runtime errors.  Reinstall Horovod using `pip install --no-cache-dir` to build with the new version.')
```

Something like that.

horovod/tensorflow/mpi_ops.py
```diff
@@ -17,6 +17,7 @@
 # =============================================================================
 """Inter-process communication using MPI."""
 
+import sys
```
tgaddair, line 20: Nit: try to maintain alphabetical order.  Standard import order is:

```
import system_import
from system_import import something

import third_party_import
from third_party import something

import horovod_imports
from horovod import something
```

Within each of those, we use alphabetical order.

horovod/tensorflow/mpi_ops.py
```diff
@@ -42,6 +46,11 @@ def _load_library(name):
     library = load_library.load_op_library(filename)
     return library
 
+try:
+    util.check_framework_version("tensorflow", tf.__version__)
+except HorovodVersionMismatch as e:
+    print("tensorflow version mismatches: \n Expected: {}. Found: {}"\
+        .format(e.installed_version, tf.__version__), file=sys.stderr)
```
tgaddair, line 53: Here's my suggestion for how we sequence these things:

```
frameworks_equal, installed_version = check_installed_version('tensorflow', tf.__version__)
try:
    MPI_LIB = _load_library('mpi_lib' + get_ext_suffix())
except Exception as e:
    check_installed_version('tensorflow', tf.__version__, e)
    raise e
else:
    check_installed_version('tensorflow', tf.__version__)
```

So basically we have 4 cases:

1. Exception was raised and version mismatch -> raise a HorovodVersionMismatchError
2. Exception was raised but versions equal -> re-raise original exception
3. No exception raised and version mismatch -> issue HorovodVersionMismatchWarning
4. No exception raised but versions equal -> do nothing

horovod/common/util.py
```diff
@@ -251,3 +253,11 @@ def split_list(l, n):
     """
     d, r = divmod(len(l), n)
     return [l[i * d + min(i, r):(i + 1) * d + min(i + 1, r)] for i in range(n)]
+
+
+def check_framework_version(name, version):
```
tgaddair, line 258: See comment in TensorFlow call site for more details, but basically I think we can add an additional optional parameter here:

```
def check_framework_version(name, version, exception):
```

Then we handle a version mismatch differently depending on if we're handling an exception or not.  If we're handling an exception, we want to wrap and re-raise the exception.  If we're not handling an exception, we want to issue a warning.

So:

```
...
if installed_version != version:
    if exception is not None:
        raise HorovodVersionMismatchError(name, version, installed_version) from exception
    else:
        warnings.warn(..., HorovodVersionMismatchWarning)
```

What do you think?

tgaddair (commented):
Nice work!  Few suggestions for how we can convey this info the user, but nice job getting the CMake stuff figured out!

horovod/common/util.py
```diff
@@ -251,3 +254,17 @@ def split_list(l, n):
...
     d, r = divmod(len(l), n)
     return [l[i * d + min(i, r):(i + 1) * d + min(i + 1, r)] for i in range(n)]
+
+
+def check_installed_version(name, version, exception=None):
+    file_path = os.path.abspath(os.path.join(os.path.dirname(os.path.abspath(__file__)),\
+        os.pardir, "metadata.json"))
+    with open(file_path) as f:
+        installed_version = json.load(f)[name]
+        if not installed_version == version:
```
tgaddair, line 264: Nit: `if installed_version != version:`

horovod/common/util.py
```diff
@@ -251,3 +254,17 @@ def split_list(l, n):
     """
     d, r = divmod(len(l), n)
     return [l[i * d + min(i, r):(i + 1) * d + min(i + 1, r)] for i in range(n)]
+
+
+def check_installed_version(name, version, exception=None):
+    file_path = os.path.abspath(os.path.join(os.path.dirname(os.path.abspath(__file__)),\
+        os.pardir, "metadata.json"))
+    with open(file_path) as f:
+        installed_version = json.load(f)[name]
```
tgaddair, line 263: What if the user didn't build with this framework at all?

We should use `json.load(f).get(name)` to gracefully return `None`, otherwise the error message will be confusing.

horovod/common/exceptions.py
```diff
@@ -29,3 +29,12 @@ class HostsUpdatedInterrupt(RuntimeError):
     In elastic mode, this will result in a reset event without a restore to committed state.
     """
     pass
+
+class HorovodVersionMismatchError(RuntimeError):
+    """Internal error raised when the runtime version of a framework mismatches its version at
+    Horovod installation time.
+    """
+    def __init__(self, name, version, installed_version):
+        self.name = name
```
tgaddair, line 38: We should customize the error message by passing the message to the parent constructor:

```
super().__init__(message)
```

horovod/common/exceptions.py
```diff
@@ -29,3 +29,12 @@ class HostsUpdatedInterrupt(RuntimeError):
     In elastic mode, this will result in a reset event without a restore to committed state.
     """
     pass
+
+class HorovodVersionMismatchError(RuntimeError):
```
tgaddair, line 33: Let's replace `RuntimeError` -> `Exception`.

horovod/common/util.py
```diff
@@ -251,3 +254,17 @@ def split_list(l, n):
...
+        if not installed_version == version:
+            if exception is None:
+                warnings.warn(f'Framework {name} installed with version {installed_version} but found version {version}.\n\
```
tgaddair, line 266: Let's move this into `common/exceptions.py` so we can reuse it as the exception message as well:

```
def get_version_mismatch_message(name, version, installed_version):
    return ...
```

tgaddair (commented):
Looks good!  I see the MXNet handling had to be moved to basics, which makes sense, since it comes first.

Only minor comments, then we can land.

chongxiaoc:
fixed

tgaddair (approved):
Nice!  LGTM.

PR #2537, opened 2020-12-18
Updated CI to use latest Buildkite stack

tgaddair:
New Buildkite queue allows us to do two things:

1. Updates Nvidia drivers to support CUDA 11.1.
2. Allocates 10% of the GPU instances to a dedicated pool to increase liveness.

### DIFF
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -420,12 +420,12 @@ for test in ${tests[@]-}; do
   if [[ ${test} == *-gpu-* ]] || [[ ${test} == *-mixed-* ]]; then
     # if gloo is specified, run gloo gpu unit tests
     if [[ ${test} == *-gloo* ]]; then
-      run_gloo_pytest ${test} "4x-gpu-g4"
+      run_gloo_pytest ${test} "4x-gpu-v510"
     fi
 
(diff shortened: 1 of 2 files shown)


.buildkite/gen-pipeline.sh
```diff
@@ -438,14 +438,14 @@ for test in ${tests[@]-}; do
   if [[ ${test} == *-gpu-* ]] || [[ ${test} == *-mixed-* ]]; then
     # if gloo is specified, run gloo gpu integration tests
     if [[ ${test} == *-gloo* ]]; then
-      run_gloo_integration ${test} "2x-gpu-g4"
+      run_gloo_integration ${test} "2x-gpu-v510"
```
chongxiaoc, line 441: What is the format here? 2x means 2 GPUs? g4 or v510 is different GPU card?

chongxiaoc (approved):
just have a question for the format since I'm not familiar.

.buildkite/gen-pipeline.sh
tgaddair: It's arbitrary.  Before, it was <# gpus per container>-<cpu or gpu>-<sku>.  Now, since we always use g4 for the sku, I replaced it with the version of the Buildkite stack.

PR #2539, opened 2020-12-19
Move sphinx to pypi

EnricoMi:
Moves to PyPi package for sphinx, also moves to latest version (3.x rather than 2.x), This also saves 40MB download per build.


tgaddair (approved):
Nice fix!

PR #2569, opened 2021-01-04
Move to g++-7

EnricoMi:
Use Ubuntu 18.04 as default, use g++-7 only for tfhead (see #1270).


tgaddair (commented):
I think we can try dropping GCC4 and only use GCC7 from here on.

At least for TF, all versions we support are compiled with GCC7:

https://www.tensorflow.org/install/source#gpu

I believe all versions of PyTorch we support should be the same.

EnricoMi:
Alright, moved to g++-7 entirely, also with `Docker.test.gpu` and the `Dockerfile.cpu` and `Dockerfile.gpu` files. Kept the GPP_VERSION parameter for the test docker files so we can easily test different versions in the future.

tgaddair (approved):
This looks good!  Thanks for making these changes.

PR #2580, opened 2021-01-07
Added support for sparse tensors to PyTorch DistributedOptimizer

tgaddair:
By default, sparse gradients will be aggregated using `allgather` similar to how this feature works for TensorFlow.

Alternatively, the user can set `hvd.DistributedOptimizer(..., sparse_as_dense=True)`, converting sparse gradients to dense to force usage of allreduce instead of allgather.  This can improve performance if gradients do not have many missing values.

### DIFF
--- a/horovod/torch/optimizer.py
+++ b/horovod/torch/optimizer.py
@@ -37,7 +37,7 @@ class _DistributedOptimizer(torch.optim.Optimizer):
                  backward_passes_per_step=1, op=Average,
                  gradient_predivide_factor=1.0,
                  num_groups=0,
-                 sparse=False):
+                 sparse_as_dense=False):
         super(self.__class__, self).__init__(params)
         self._compression = compression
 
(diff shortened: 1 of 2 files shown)


horovod/torch/optimizer.py
```diff
@@ -219,9 +241,13 @@ def synchronize(self):
                     self._allreduce_delay[gp] = self.backward_passes_per_step
                     gp.grad.set_(self._compression.decompress(output, gctx))
             else:
-                output = synchronize(handle)
+                # When handle is a callable function, it returns the aggregated tensor result
+                output = synchronize(handle) if not callable(handle) else handle()
                 self._allreduce_delay[p] = self.backward_passes_per_step
-                p.grad.set_(self._compression.decompress(output, ctx))
+                if p.grad.is_sparse:
+                    p.grad.zero_().add_(self._compression.decompress(output, ctx))
```
chongxiaoc: I see, put sparse result to original dense zero matrix.

tgaddair: For some reason you can't `set_` a sparse tensor, so we need to use this hack.  It's kind of ugly.

chongxiaoc (approved):
look good to me

romerojosh (approved):
LGTM!

PR #2596, opened 2021-01-14
[ray] fix local_rank issue

### LINKED ISSUE #2583: Horovod on Ray has all local_rank as 0
I am currently exploring using Horovod on Ray, and have a dual-gpu machine for testing purpose with RayExecutor settings as `num_hosts=1, num_slots=2, use_gpu=True`.  What I noticed is that, instead of each GPU got one process to run, actually each gpu got two processes and two of the actual running processes are both running on the same GPU (as shown below)
image

Then I dig deeper into this and noticed that, local_rank is 0 on both processes. This is reproduced by
```python
executor.start()
def fn():
    hvd.init()
    print(hvd.size(), hvd.rank(), hvd.local_rank())
executor.run(fn)
```
result:
image

So looks like hvd.size() and hvd.rank() are correct while hvd.local_rank() is 0 for both processes.

And I confirm this only happens on Horovod with Ray, not standalone Horovod:
```python
import horovod.tensorflow.keras as hvd
hvd.init()
print(hvd.size(), hvd.rank(), hvd.local_rank())
```
result:
image

above shows standalone Horovod got correct values.

Two additional information:
1. I searched related issues https://github.com/horovod/horovod/issues/307, https://github.com/horovod/horovod/issues/515, and https://github.com/horovod/horovod/issues/287.  Didn't solve the problem. Plus, my standalone horovod is running perfectly fine, and issue only exist when using RayExecutor.
2. I assume I did the correct thing by running horovod-ray script with plain `python script.py` instead of using `horovodrun`(as shown in the above two result screenshots), as my understanding is that horovod-ray's running schedule is controlled by RayExecutor() so there is no need to use `horovodrun`.

richardliaw:

## Description

Fixes #2583

## Review process to land 

### DIFF
--- a/horovod/ray/runner.py
+++ b/horovod/ray/runner.py
@@ -45,11 +45,9 @@ def map_blocking(fn, collection):
 class BaseHorovodWorker:
     executable = None
 
-    def __init__(self, local_rank=0, local_size=1, world_rank=0, world_size=1):
+    def __init__(self, world_rank=0, world_size=1):
         os.environ["HOROVOD_HOSTNAME"] = self.hostname()
         os.environ["HOROVOD_RANK"] = str(world_rank)
-        os.environ["HOROVOD_LOCAL_RANK"] = str(local_rank)
-        os.environ["HOROVOD_LOCAL_SIZE"] = str(local_size)
         os.environ["HOROVOD_SIZE"] = str(world_size)
 
     def hostname(self) -> str:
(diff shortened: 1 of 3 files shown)


richardliaw:
cc @tgaddair

horovod/ray/runner.py
```diff
@@ -45,9 +45,11 @@ def map_blocking(fn, collection):
 class BaseHorovodWorker:
     executable = None
 
-    def __init__(self, world_rank=0, world_size=1):
+    def __init__(self, local_rank=0, local_size=1, world_rank=0, world_size=1):
         os.environ["HOROVOD_HOSTNAME"] = self.hostname()
         os.environ["HOROVOD_RANK"] = str(world_rank)
+        os.environ["HOROVOD_LOCAL_RANK"] = str(local_rank)
+        os.environ["HOROVOD_LOCAL_SIZE"] = str(local_size)
```
tgaddair, line 52: You may also need to set cross rank and cross size.

richardliaw:
should be ready for another review!

tgaddair (approved):
LGTM!

PR #2603, opened 2021-01-19
Make 'pip install -e .[dev]' work with Python3.8

EnricoMi:
Versions of some packages in [dev] do not exist for Python 3.8, moving to the first version that supports 3.8 for those.


tgaddair (commented):
Since dev versions are somewhat arbitrary (any version we support officially), what do you think about using 3.8 versions for all Python versions?

EnricoMi:
Good point, this complexity is not needed. Will move to the py3.8 versions then.

tgaddair (approved):
LGTM!

PR #2606, opened 2021-01-19
Mention g++ above 4.9 is also fine

EnricoMi:
Now that we even moved to g++-7 we should mention in the docs that g++-4.9 and above is fine.


tgaddair (approved):
Thanks for the update!

PR #2611, opened 2021-01-21
Fix DL estimators for getting the output df schema

### LINKED ISSUE #2536: Performance degredation with Spark Estimator during schema inference
#2373 solves the problem of Spark being unable to infer the schema following an operation on the underlying RDD.  However, the new schema inference process is expensive, as it requires sampling directly from the data.

Instead, we should preserve the original schema when we reconstruct the DataFrame.

cc @OscarDPan @thuningxu

irasit:

- [ Y] Did you write any tests to validate this change?  

## Description
#2373 introduced a big delay when generating the output schema. Instead we can get the schema from the input df schema and label columns.

Fixes #2536.

## Review process to land 

### DIFF
--- a/horovod/spark/common/util.py
+++ b/horovod/spark/common/util.py
@@ -712,6 +712,11 @@ def is_databricks():
     return "DATABRICKS_RUNTIME_VERSION" in os.environ
 
 
+def get_output_cols(input_df_schema, output_cols):
+    final_output_cols = [f.name for f in input_df_schema.fields]
+    return final_output_cols + output_cols
+
+
 def get_spark_df_output_schema(input_df_schema, label_cols, output_cols, metadata):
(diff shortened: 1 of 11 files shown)


horovod/spark/common/util.py
```diff
@@ -710,3 +710,33 @@ def _get_assigned_gpu_or_default(default):
 
 def is_databricks():
     return "DATABRICKS_RUNTIME_VERSION" in os.environ
+
+
+def get_spark_df_output_schema(input_df_schema, label_cols, output_cols):
+    if len(label_cols) != len(output_cols):
```
irasit, line 716: @tgaddair Please check here. Is it OK to always assume label_cols and output_cols are 1:1 matching?

tgaddair (approved):
LGTM!

PR #2613, opened 2021-01-22
Make tests that use commons work on their own

EnricoMi:
Some tests import `test/utils/commons.py` but are missing the required `sys.path` modification. Those tests only work when run after other test that perform this modification. This makes them work on their own.


tgaddair (approved):
Nice fix!  A number of people have run into issues with this.

Issue #2570, opened 2021-01-04
Reworking test space combinations


EnricoMi:
The buildkite pipeline currently tests the following parameter combinations:

|processor|mpi kind|python|tensorflow|keras|pytorch|mxnet|pyspark|
|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
|cpu|openmpi_gloo|py3_6|tf1_15_0|keras2_2_4|torch1_2_0|mxnet1_4_1|pyspark2_3_2|
|cpu|gloo|py3_6|tf2_0_0|keras2_2_4|torch1_3_0|mxnet1_4_1|pyspark2_3_2|
|cpu|openmpi|py3_6|tf2_1_0|keras2_2_4|torch1_4_0|mxnet1_4_1|pyspark2_4_7|
|cpu|openmpi|py3_6|tf2_2_0|keras2_3_1|torch1_5_0|mxnet1_5_0|pyspark2_4_7|
|cpu|gloo|py3_7|tf2_3_1|keras2_3_1|torch1_6_0|mxnet1_5_0|pyspark2_4_7|
|cpu|gloo|py3_8|tf2_4_0|keras2_3_1|torch1_7_1|mxnet1_5_0|pyspark3_0_1|
|cpu|openmpi|py3_6|tfhead|kerashead|torchhead|mxnethead|pyspark2_4_7|
|cpu|mpich|py3_6|tf1_15_0|keras2_3_1|torch1_4_0|mxnet1_5_0|pyspark2_4_7|
|cpu|oneccl|py3_6|tf1_15_0|keras2_3_1|torch1_4_0|mxnet1_5_0|pyspark2_4_7|
|cpu|oneccl_ofi|py3_6|tf1_15_0|keras2_3_1|torch1_4_0|mxnet1_5_0|pyspark2_4_7|
|gpu|openmpi|py3_6|tf1_15_0|keras2_2_4|torch1_3_0|mxnet1_4_1|pyspark2_4_7|
|gpu|gloo|py3_6|tf2_0_0|keras2_3_1|torch1_4_0|mxnet1_4_1|pyspark2_4_7|
|gpu|openmpi|py3_6|tf2_3_1|keras2_3_1|torch1_6_0|mxnet1_6_0|pyspark2_4_7|
|gpu|openmpi_gloo|py3_6|tf2_4_0|keras2_3_1|torch1_7_1|mxnethead|pyspark2_4_7|
[...]

PR #2614, opened 2021-01-22
Rework test space combinations

EnricoMi:
Implements #2570.
- Removes oneccl-ofi test image, as this is identical to the oneccl image, ofi tests use the latter image now.
- Removes Gloo support from images that do not have `gloo` in their name (to avoid falling back to Gloo if the MPI kind is not available).
- Removes Elastic Horovod on Spark integration tests from test images other than the baseline, baseline with tf1, and PySpark variations (strips off 5 CPU-hours of testing, 30 minutes of buildkite time)


EnricoMi:
@tgaddair Compared against [last full master build](https://github.com/horovod/horovod/pull/2577#issuecomment-755788564), this has 17% more test runs, but a bit less CPU-time and [buildkite-time](https://buildkite.com/horovod/horovod/builds/4651).

tgaddair:
Nice work!  Removing Gloo support to prevent fallback is a good catch.

tgaddair (approved):

PR #2618, opened 2021-01-22
Use docker buildkit caching

EnricoMi:
Trying docker buildkit caching as described [here](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/256).


EnricoMi:
The buildkite agent images have docker version 18.06.1-ce installed, 18.09 is minimum requirement.

PR #2620, opened 2021-01-24
Install horovod for all frameworks in Dockerfile, incl. Spark

EnricoMi:
Installing horovod does not install all dependencies. Also installs pyspark and Spark.

What is the purpose of these docker images? Are those pushed to docker hub? What do you think of adding spark to them?

### DIFF
--- a/Dockerfile.cpu
+++ b/Dockerfile.cpu
@@ -5,6 +5,9 @@ ENV PYTORCH_VERSION=1.6.0
 ENV TORCHVISION_VERSION=0.7.0
 ENV MXNET_VERSION=1.5.0
 
+ENV PYSPARK_PACKAGE=pyspark==2.4.7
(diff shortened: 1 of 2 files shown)


tgaddair:
Good idea @EnricoMi.  The goal of these images should be to run Horovod in your chosen environment without needing to figure out which packages you need.  So adding Spark, Ray, etc. sounds like a good idea to me.

tgaddair (approved):
I like this approach, could also add Ray in the future, maybe as different Dockerfiles for Spark and Ray.

EnricoMi:
> could also add Ray in the future

Maybe ray should be added to `all-frameworks`.

tgaddair:
I think it is, but there are some dependencies that are needed to make it work in Docker. I have them locally, but will make a note to add them to the examples.

PR #2631, opened 2021-01-29
Let hvd.alltoall return the received splits if non-uniform splits are sent (TensorFlow, PyTorch, MXNet) 

maxhgerlach:

## Description

Often alltoall graph nodes occur in pairs (as in `input->alltoall->[processing]->alltoall->output`). With non-uniform splits, one currently needs to add an extra hvd.alltoall() op just to know how much data was received from which workers in the first hvd.alltoall() so those can be targeted properly again in the second hvd.alltoall(). Since this information is already available internally, I propose to expose the received_splits as an extra output of the op.

## Review process to land 

### DIFF
--- a/horovod/common/common.h
+++ b/horovod/common/common.h
@@ -267,6 +267,7 @@ struct TensorTableEntry {
   // storage complexity of collecting all worker split arrays
   // on coordinator rank.
   std::vector<int32_t> splits;
+  std::shared_ptr<Tensor> received_splits;
 };
 using TensorTable = std::unordered_map<std::string, TensorTableEntry>;
 
--- a/horovod/common/common.h
+++ b/horovod/common/common.h
@@ -223,6 +223,14 @@ public:
                      std::shared_ptr<PersistentBuffer>* tensor) = 0;
   virtual Status AllocateOutput(TensorShape shape,
                                 std::shared_ptr<Tensor>* tensor) = 0;
+  virtual Status AllocateOutput(int output_index, TensorShape shape,
+                                std::shared_ptr<Tensor>* tensor) {
+    if (output_index == 0) {
+      return AllocateOutput(shape, tensor);
+    } else {
+      throw std::logic_error("output_index != 0 not supported");
+    }
+  }
   virtual Status AllocateZeros(int64_t num_elements, DataType dtype,
                                 std::shared_ptr<Tensor>* tensor) = 0;
   virtual Framework framework() const = 0;
--- a/horovod/common/ops/collective_operations.h
+++ b/horovod/common/ops/collective_operations.h
@@ -253,19 +253,16 @@ protected:
     }
 
     // Allocate and fill received_splits output
-    if (e.context->framework() == Framework::TENSORFLOW ||
-        e.context->framework() == Framework::PYTORCH) {
-      TensorShape received_splits_shape;
-      received_splits_shape.AddDim(recvsplits.size());
-      Status rstatus = e.context->AllocateOutput(1, received_splits_shape,
-                                                 &e.received_splits);
-      if (!rstatus.ok()) {
-        return rstatus;
-      }
-      auto* target_pointer = reinterpret_cast<int32_t*>(
-          const_cast<void*>(e.received_splits->data()));
-      std::copy(recvsplits.cbegin(), recvsplits.cend(), target_pointer);
+    TensorShape received_splits_shape;
+    received_splits_shape.AddDim(recvsplits.size());
+    Status rstatus = e.context->AllocateOutput(1, received_splits_shape,
+                                               &e.received_splits);
+    if (!rstatus.ok()) {
+      return rstatus;
     }
+    auto* target_pointer = reinterpret_cast<int32_t*>(
+        const_cast<void*>(e.received_splits->data()));
+    std::copy(recvsplits.cbegin(), recvsplits.cend(), target_pointer);
 
     return Status::OK();
   }
--- a/horovod/common/ops/collective_operations.h
+++ b/horovod/common/ops/collective_operations.h
@@ -213,7 +213,7 @@ protected:
 
     const auto& splits = e.splits;
     std::vector<int32_t> recvsplits;
-    // Perform alltoall of splits to get expeceted receive splits
+    // Perform alltoall of splits to get expected receive splits
     global_state_->controller->AlltoallGetRecvSplits(splits, recvsplits);
 
     // Every tensor participating in Alltoall operation may have different
@@ -250,6 +250,20 @@ protected:
     Status status = e.context->AllocateOutput(output_shape, &e.output);
     if (!status.ok()) {
       return status;
+    }
+
+    // Allocate and fill received_splits output
+    if (e.context->framework() == Framework::TENSORFLOW) {
+      TensorShape received_splits_shape;
+      received_splits_shape.AddDim(recvsplits.size());
+      Status rstatus = e.context->AllocateOutput(1, received_splits_shape,
+                                                 &e.received_splits);
+      if (!rstatus.ok()) {
+        return rstatus;
+      }
+      auto* target_pointer = reinterpret_cast<int32_t*>(
+          const_cast<void*>(e.received_splits->data()));
+      std::copy(recvsplits.cbegin(), recvsplits.cend(), target_pointer);
     }
 
     return Status::OK();
--- a/horovod/common/ops/collective_operations.h
+++ b/horovod/common/ops/collective_operations.h
@@ -253,7 +253,8 @@ protected:
     }
 
     // Allocate and fill received_splits output
-    if (e.context->framework() == Framework::TENSORFLOW) {
+    if (e.context->framework() == Framework::TENSORFLOW ||
+        e.context->framework() == Framework::PYTORCH) {
       TensorShape received_splits_shape;
       received_splits_shape.AddDim(recvsplits.size());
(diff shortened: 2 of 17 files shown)


horovod/tensorflow/mpi_ops.py
```diff
@@ -322,9 +327,12 @@ def _alltoall_grad(op, grad):
     splits = tf.cond(tf.equal(tf.size(splits), 0),
                      lambda : tf.ones([size()], dtype=tf.int32) * (tf.shape(tensor)[0] // size()),
                      lambda : splits)
-    recvsplits = alltoall(splits, splits=[1 for _ in range(size())],
-                          ignore_name_scope=ignore_name_scope)
-    return [alltoall(grad, splits=recvsplits, ignore_name_scope=ignore_name_scope), None]
+    recvsplits = op.outputs[1]
```
romerojosh, line 330: Great! Reusing the `recvsplits` should make the alltoall gradient computation more efficient.

horovod/tensorflow/mpi_ops.cc
```diff
@@ -163,6 +163,9 @@ class TFOpContext : public common::OpContext {
   AllocateOutput(common::TensorShape shape,
                  std::shared_ptr<common::Tensor>* tensor) override;
   virtual common::Status
+  AllocateOutput(int output_index, common::TensorShape shape,
+                 std::shared_ptr<common::Tensor>* tensor) override;
+  virtual common::Status
```
romerojosh, line 168: Instead of defining an overloaded `AllocateOutput` with the additional `output_index` argument, it might be cleaner to just add an `output_index` to the existing method with a default value of 0.

horovod/common/ops/collective_operations.h
```diff
@@ -252,6 +252,20 @@ class AlltoallOp : public HorovodOp {
       return status;
     }
 
+    // Allocate and fill received_splits output
+    if (e.context->framework() == Framework::TENSORFLOW) {
```
romerojosh, line 256: Have you considered extending support for this to the other supported frameworks?

horovod/common/common.h
```diff
@@ -259,6 +267,7 @@ struct TensorTableEntry {
   // storage complexity of collecting all worker split arrays
   // on coordinator rank.
   std::vector<int32_t> splits;
+  std::shared_ptr<Tensor> received_splits;
```
romerojosh, line 270: Since `received_splits` is an output, I think it would more general to make `output` a vector with multiple entries rather than add a separate entry. Since this is the only multi-output operation currently, might not be worth the more extensive code changes for this right now, but something to keep in mind.

romerojosh (commented):
Thanks for the contribution @maxhgerlach! This is a great improvement to the alltoall implementation. Just left a few comments.

horovod/tensorflow/mpi_ops.py
maxhgerlach: Thanks for the review @romerojosh!

horovod/tensorflow/mpi_ops.cc
maxhgerlach: That was my first idea. But default arguments on virtual and override functions seem to be frowned upon in C++, which I learnt from the Clang warnings I was prompted with (see e.g. https://google.github.io/styleguide/cppguide.html#Default_Arguments). Realistically, there probably is only the potential of issues if any derived class were to specify a different default value from the base class, which is quite unlikely in this case. So if you prefer, I could combine these to a single definition with a default argument (and ignore warnings).

horovod/common/ops/collective_operations.h
maxhgerlach: I personally don't have any experience implementing ops in the other frameworks. So this would delay things a bit.

horovod/common/common.h
maxhgerlach: One potential caveat that came to my mind: There are a lot of instances in the code that look like `(void*)e.output->data()`. These would have to be changed to something like `(void*)e.outputs[0]->data()`, but it would then be quite easy to introduce a typo like `(void*)e.outputs->data()` and the compiler would not notice because both std::vector and horovod::common::Tensor have a member function data(). So maybe it would then be worth to change data() and size() in Tensor to less error prone names?

romerojosh: That is a good point. If you did want to go down the route of converting `output` to a vector of tensors, did you have better names in mind?

horovod/common/ops/collective_operations.h
romerojosh: No, this isn't blocking, though it is nice to try and keep features across the framework implementations more in sync.

horovod/tensorflow/mpi_ops.cc
romerojosh: Ok, sounds like a good enough reason to me to do things the way you did. Thanks for being thorough!

horovod/common/ops/collective_operations.h
maxhgerlach: OK, I just pushed my go at an implementation for Pytorch.

Giving it some effort, I unfortunately did not manage to install a version of MXNet in my dev environment that would be compatible with the Horovod build. I probably won't find time very soon to focus on this.

horovod/torch/mpi_ops_v2.cc
```diff
@@ -372,13 +372,17 @@ int DoBroadcastCudaOnCPU(::torch::Tensor tensor, ::torch::Tensor output, int roo
...
+int DoAlltoall(::torch::Tensor tensor, ::torch::Tensor splits,
+               ::torch::Tensor output, ::torch::Tensor output_received_splits,
+               const std::string& name) {
   ThrowIfError(common::CheckInitialized());
 
   auto device = GetDeviceID(tensor);
   auto ready_event = RecordReadyEvent(device);
   auto hvd_tensor = std::make_shared<TorchTensor>(tensor);
   auto hvd_context = std::make_shared<TorchOpContext>(device, output);
+  assert(GetDeviceID(output_received_splits) == CPU_DEVICE_ID);
```
maxhgerlach, line 384: @romerojosh, here I wasn't sure how to best deal with the possibility that output_received_splits might be a GPU tensor (which could be caused by a bug on the Python side).

horovod/common/ops/collective_operations.h
maxhgerlach: All right, I managed to sort out my MXNet installation problems and pushed an implementation for that framework, too.

horovod/common/common.h
maxhgerlach: I'd rather postpone the naming debate to some future pull request, actually... If we decide to go forward with turning `output` into a vector, should we also turn `input` into a vector and treat `splits` like an input tensor?

As long as there are no other operations with multiple outputs and as long as there are not more than two different "types" of output, I'd personally prefer the code to be a bit simpler all over.

romerojosh: Yes, I agree. We can postpone these changes until there are more multi input/output operations.

horovod/torch/mpi_ops_v2.cc
romerojosh: I think in the case where `output_received_splits` is a GPU tensor, we should make a temporary CPU output tensor, use that in the backend, and then copy the CPU output tensor to the GPU output tensor during the callback, similar to what is done inthe `*CudaOnCPU`routines. What do you think?

romerojosh (commented):
Thanks a lot for the updates @maxhgerlach and extending this feature to the other supported frameworks! Left one comment on the PyTorch implementation to resolve, but otherwise, changes look good to me.

horovod/torch/mpi_ops_v2.cc
maxhgerlach: Thanks, that sounds reasonable. I'll look into.

maxhgerlach: OK, I have changed the PyTorch API such that if the incoming `splits` happen to be on GPU, the `received_splits` will be put on GPU, too. That's probably the least surprising behavior.

maxhgerlach:
Thanks for the review and advice, @romerojosh! I hope to have addressed all concerns. It looks like the CI doesn't run GPU tests right now, so an edge case in some build configuration might still be off, of course.

horovod/torch/mpi_ops_v2.cc
romerojosh: I agree. Thanks for updating the PR!

romerojosh (approved):
This PR looks good to me! Concerning GPU tests, the ppcle64 runner should be testing GPUs, at least for TensorFlow and PyTorch, though those tests aren't as extensive as the normal CI GPU tests. I also tested this PR myself in the NVIDIA NGC 20.12 MXNet container and that seems to be working as well. 

@tgaddair Do we have an ETA on when we will reenable GPU tests in the CI? This PR does make modifications to the C++ backend, so it'd be nice to get a full round of GPU testing to catch any edge cases.

tgaddair:
Hey @romerojosh @maxhgerlach, GPU tests have been reenabled now.  You may need to rebase from master to get them to trigger.

maxhgerlach:
That sound's great, thanks @tgaddair! I rebased the branch on top of master.

maxhgerlach:
OK, only this flaky unrelated test failed:
> 1 out of 23 runs failed: test_fault_tolerance_hosts_remove_and_add (test.single.test_ray_elastic)

PR #2644, opened 2021-02-03
Pin tf-nightly-gpu to fix unit tests

tgaddair:


tgaddair:
This is no longer needed as tf-nightly-gpu has been fixed.

PR #2647, opened 2021-02-05
[TF2.4 - HVD Keras Fix] - Horovod worker desynchronization - Potentially deadlock training TF 2.4

DEKHTIARJonathan:
CC: @romerojosh @tgaddair @nluehr 

This PR aims to fix a bug with TF2.4 already reported here: https://github.com/tensorflow/tensorflow/issues/46863

Basically this PR does the following:
- Rollback the `allreduce` in `_aggregate_gradients` and make sure to execute the allreduce during gradient computation
- Yes it means that in the eventuality the Grad Step is skipped (by mixed precision) we have allreduced for no reason, but this event is supposed to be very rare. And it's still better than having worker to desynchronize or deadlock.
- `self._aggregated_gradients = True` has been moved to gradient computation instead of `allreduce` itself. This is done to work around that gradient tape is used to compute gradient with a CTL

@tgaddair : would be nice to push a new release" to address this bug, that way we can unblock TF2.4 users ;) 

Thanks everyone

### DIFF
--- a/horovod/_keras/__init__.py
+++ b/horovod/_keras/__init__.py
@@ -68,6 +68,32 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
 
             super(self.__class__, self).__init__(**kwargs)
 
+        def _compute_gradients(self, loss, var_list, grad_loss=None, tape=None):
+            """
+            Compute gradients of all trainable variables.
+
+            See Optimizer.get_gradients() for more info.
+
+            In DistributedOptimizer, get_gradients() is overriden to also
+            allreduce the gradients before returning them.
+            """
+            if _PRE_TF_2_4_0:
+                return super(self.__class__, self)._compute_gradients(
+                    loss, var_list, grad_loss, tape)
+
+            tape = backprop.GradientTape() if tape is None else tape
+            grads_and_vars = super(self.__class__, self)._compute_gradients(
+                # pylint: disable=protected-access
+                loss,
+                var_list,
+                grad_loss,
+                tape=tape)
+            grads = [g for g, _ in grads_and_vars]
+            weights = [v for _, v in grads_and_vars]
+
+            allreduced_grads = self._allreduce(grads)
+            return list(zip(allreduced_grads, weights))
+
         def get_gradients(self, loss, params):
             """
             Compute gradients of all trainable variables.
@@ -81,13 +107,13 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
             return self._allreduce(gradients)
 
         def _aggregate_gradients(self, grads_and_vars):
-            grads, vars = list(zip(*grads_and_vars))
-            aggregated_grads = self._allreduce(grads)
             if _PRE_TF_2_4_0:
-                # Prior to TF 2.4.0, this function was expected to return only a list of
-                # grads, not a list of (grad, var) tuples.
+                grads, vars = list(zip(*grads_and_vars))
+                aggregated_grads = self._allreduce(grads)
                 return aggregated_grads
-            return list(zip(aggregated_grads, vars))
+            else:
+                return super(self.__class__, self)._aggregate_gradients(
+                    grads_and_vars)
 
         def _allreduce(self, grads):
             self._aggregated_gradients = True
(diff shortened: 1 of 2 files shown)


tgaddair:
@DEKHTIARJonathan thanks for the PR!  Seems this is breaking some tests, can you look into this?

```
[1]<stdout>:test_tensorflow2_keras.py::Tf2KerasTests::test_gradient_aggregation FAILED
--
  | [1]<stdout>:test_tensorflow2_keras.py::Tf2KerasTests::test_session SKIPPED (Not ...)
  | [0]<stderr>:[2021-02-05 20:33:38. 76607: W /tmp/pip-req-build-a5836dcu/horovod/common/stall_inspector.cc:105] One or more tensors were submitted to be reduced, gathered or broadcasted by subset of ranks and are waiting for remainder of ranks for more than 60 seconds. This may indicate that different ranks are trying to submit different tensors or that only subset of ranks is submitting tensors, which will cause deadlock.
  | [0]<stderr>:Missing ranks:
  | [0]<stderr>:0: [PartitionedCall/DistributedRMSprop_Allreduce/cond/then/_5/DistributedRMSprop_Allreduce/cond/HorovodAllreduce_DistributedRMSprop_Allreduce_UnsortedSegmentSum_0]
  | [0]<stderr>:1: [StatefulPartitionedCall/cond/then/_16/cond/PartitionedCall/DistributedTestingOptimizer_Allreduce/cond/then/_56/DistributedTestingOptimizer_Allreduce/cond/HorovodAllreduce_grads_0]


```

https://buildkite.com/horovod/horovod/builds/4887#34ba55ce-1e9b-4e4e-a824-efb41fbd7084/216-766

DEKHTIARJonathan:
@tgaddair @romerojosh I remove some bugs in my PR that was leading to duplicated allreduces.

I also had to remove the test: `test_gradient_aggregation(self):` from `test/parallel/test_tensorflow2_keras.py` which obviously doesn't make sense anymore. This test was checking that allreduce was working as expected during `_aggregate_gradients` call, which doesn't happen anymore. Soooo => deleted
@tgaddair : let me know if this test is useful for previous versions of TF (which I kinda have doubt), I can re-enable it if TF_VERSION<2.4.

DEKHTIARJonathan:
@romerojosh @tgaddair I have completely reworked the PR, we found duplicated calls to `allreduce()` that should address the fews bugs.

In addition, I had to skip with TF 2.4+ the `test_gradient_aggregation()` from `test/parallel/test_tensorflow2_keras` which naturally doesn't make any sense anymore since we don't allreduce anymore during `_aggregate_gradients`, so this failing test was totally expected.

Let's see how the CI comes out

tgaddair:
Hey @DEKHTIARJonathan, can you explain the skipping of the gradient aggregation test?  We still need to support this functionality, even if we don't call `_aggregate_gradients` explicitly.

DEKHTIARJonathan:
This PR shifts the gradient allreduce when the gradient is computed (both through the DistributedOptimizer or through the DistributedGradientTape). Which means that this unittest, by design in TF2.4, doesn't call allreduce in `_aggregate_gradients()`.

Since this unittest provide a gradient as constant (without effectively computing it), the gradient will never be allreduced.
The safest bet for this PR (which is to return to a working state) was to skip this unittest that conceptually doesn't make sense in the current direction for TF2.4.

Though if the usecase is still important, I recommend pushing a following PR doing the followings:
- fix `test_gradient_aggregation()` not to use `tf.constant()` for the gradient value and actually compute it for real.
- remove the `pytest.skip`

horovod/_keras/__init__.py
```diff
@@ -68,6 +68,32 @@ def __init__(self, **kwargs):
...
+                    loss, var_list, grad_loss, tape)
+
+            tape = backprop.GradientTape() if tape is None else tape
+            grads_and_vars = super(self.__class__, self)._compute_gradients(
+                # pylint: disable=protected-access
+                loss,
+                var_list,
+                grad_loss,
+                tape=tape)
+            grads = [g for g, _ in grads_and_vars]
```
tgaddair, line 91: Can simplify: `grads, vars = list(zip(*grads_and_vars))`

DEKHTIARJonathan: Fixed

tgaddair (approved):
LGTM!

PR #2671, opened 2021-02-17
Adding Horovod License to the metadata

DEKHTIARJonathan:
Not a big change, though I noticed that Horovod wasn't reporting the license in the metadata:

```
License: UNKNOWN
```

Now this will be fixed ;)


EnricoMi (approved):
LGTM!

PR #2673, opened 2021-02-17
[WIP] Elastic Spark example

EnricoMi:
Provides some Elastic Spark examples.


examples/spark/keras/keras_spark_rossmann_estimator.py
```diff
@@ -305,13 +314,18 @@ def fn(v):
...
-    train_rows = train_df.filter(~train_df.Validation).count()
-    val_rows = train_df.filter(train_df.Validation).count()
-    test_rows = test_df.count()
+    train_rows, val_rows, test_rows = train_df.count(), val_df.count(), test_df.count()
     print('Training: %d' % train_rows)
     print('Validation: %d' % val_rows)
     print('Test: %d' % test_rows)
 
+    # Save data frames as Parquet files.
+    train_df.write.parquet('%s/train_df.parquet' % args.data_dir, mode='overwrite')
```
tgaddair, line 323: We shouldn't need to write the DataFrames / recreate the Spark sessions manually. The major benefit of the Estimators is that they will do this for you automatically.

I suspect the reason this test is failing is because we're calling `spark.stop()` here, followed by `keras_estimator.fit(train_df)` (line 415), however we're not reloading `train_df` from disk in the new session. The DF is still pointing to the old session, and is therefore invalid in the new session.

So my suggestion would be to remove these lines and just use the same Spark session throughout (in Spark 3.1, we should also be able to dynamically switch between CPU and GPU stages in the same session, but we can do this in a follow-up PR).

PR #2676, opened 2021-02-18
Deprecate mxnet==1.4.1 tests

EnricoMi:
There is a conflict between numpy dependency required by tensorflow 1.15.x and mxnet 1.4.1:

    The conflict is caused by:
        mxnet 1.4.1 depends on numpy<1.15.0 and >=1.8.2
        tensorflow 1.15.5 depends on numpy<1.19.0 and >=1.16.0

### DIFF
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -7,9 +7,9 @@ set -eu
 repository=823773083436.dkr.ecr.us-east-1.amazonaws.com/buildkite
 
 # our baseline test is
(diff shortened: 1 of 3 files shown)


EnricoMi:
We could resolve the conflict by testing mxnet==1.4.1 with tensorflow==1.14.1, which would require GCC 4.8 thus complicating things.

tgaddair (approved):

leezu (approved):
It's fine to switch to newer mxnet

PR #2678, opened 2021-02-18
Update sparse torch

chongxiaoc:

## Description

Fixes # (issue).

## Review process to land 


chongxiaoc:
@tgaddair 
Tests all pass now.
Please see the last commit, I have to move the sparse allreduce test to an earlier place (by renaming it), otherwise it is hanging in unittest in pytorch 1.7.1. However, if you run the test standalone, it will not hang. I'm not sure what's the root behind that.

This happens on local test with pytorch 1.7.1 cpu tests as well.

Issue #2654, opened 2021-02-08
Wrong scaling in the gradient of hvd.allreduce(), hvd.allgather(), hvd.broadcast()


loic-ehrhardt:
**Environment:**
1. Framework: TensorFlow
2. Framework version: 2.4.1
3. Horovod version: 0.21.1
4. MPI version: 4.0.2
5. ~~CUDA version:~~
6. ~~NCCL version:~~
7. Python version: 3.6.8
8. ~~Spark / PySpark version:~~
9. OS and version: Red Hat Enterprise Linux release 8.3 (Ootpa)
10. GCC version: 9.2.1 20191120
11. CMake version: 3.11.4

**Bug report:**

Unless there is something I don't quite understand, the gradients implemented for `hvd.allreduce()`, `hvd.allgather()`, and `hvd.broadcast()` have a wrong scaling. The correct gradients seem to be multiplied by `hvd.size()`.

Minimal example script:
```python
import tensorflow as tf
import horovod.tensorflow.keras as hvd
hvd.init()

x = tf.constant([1 + hvd.rank()], dtype=tf.float32)
print("x =", x)

with tf.GradientTape(persistent=True) as tape:
    tape.watch(x)

    x_mean = hvd.allreduce(x)
    print("x_mean =", x_mean)
    print("d(x_mean)/dx =", tape.gradient(x_mean, x))

    x_gather = hvd.allgather(x)
    print("x_gather =", x_gather)
    print("d(x_gather)/dx =", tape.gradient(x_gather, x))

    x_bcast = hvd.broadcast(x, 0)
    print("x_bcast =", x_bcast)
    print("d(x_bcast)/dx =", tape.gradient(x_bcast, x))
```

Which produces, when called with `horovodrun -np 3 python3 main.py`
```python
[0]<stdout>:x = tf.Tensor([1.], shape=(1,), dtype=float32)
[1]<stdout>:x = tf.Tensor([2.], shape=(1,), dtype=float32)
[2]<stdout>:x = tf.Tensor([3.], shape=(1,), dtype=float32)
[0]<stdout>:x_mean = tf.Tensor([2.], shape=(1,), dtype=float32)
[1]<stdout>:x_mean = tf.Tensor([2.], shape=(1,), dtype=float32)
[2]<stdout>:x_mean = tf.Tensor([2.], shape=(1,), dtype=float32)
[...]

PR #2687, opened 2021-02-23
Fix gradients for allgather() and broadcast() (#2654).

loic-ehrhardt:
The original code was broken because _allreduce() does Sum by default.

### DIFF
--- a/horovod/tensorflow/mpi_ops.py
+++ b/horovod/tensorflow/mpi_ops.py
@@ -219,7 +219,7 @@ def _allgather_grad(op, grad):
       The gradient with respect to the input of the op.
     """
     ignore_name_scope = op.get_attr('ignore_name_scope')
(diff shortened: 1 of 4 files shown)


tgaddair:
Thanks for the PR @loic-ehrhardt!  Can you also make the same changes for PyTorch and MXNet gradients?

loic-ehrhardt:
> Thanks for the PR @loic-ehrhardt! Can you also make the same changes for PyTorch and MXNet gradients?

Hi @tgaddair. Sure, I added the fix for PyTorch. However, I did not find the definition of the gradient for MXNet. I also could not find a corresponding test in `test/parallel`. Are there gradients defined `allgather()` and `broadcast()` for MXNet?

tgaddair:
Thanks @loic-ehrhardt. Looks like we don't have grads implemented for MXNet, so no worries there.

tgaddair (approved):

PR #2707, opened 2021-03-04
Fixed PyTorch MNIST dataset

tgaddair:
See https://github.com/pytorch/vision/issues/3500

### DIFF
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -178,7 +178,9 @@ run_mpi_integration() {
     run_test "${test}" "${queue}" \
       ":tensorflow: MPI Keras MNIST (${test})" \
       "bash -c \"${oneccl_env} \\\$(cat /mpirun_command) python /horovod/examples/keras/keras_mnist_advanced.py\""
+  fi
 
+  if [[ ${test} != *"torch1_2"* ]]; then
     run_test "${test}" "${queue}" \
       ":fire: MPI PyTorch MNIST (${test})" \
       "bash -c \"${oneccl_env} \\\$(cat /mpirun_command) python /horovod/examples/pytorch/pytorch_mnist.py --data-dir /data/pytorch_datasets\""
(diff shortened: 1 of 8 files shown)


.buildkite/gen-pipeline.sh
```diff
@@ -178,10 +178,12 @@ run_mpi_integration() {
     run_test "${test}" "${queue}" \
       ":tensorflow: MPI Keras MNIST (${test})" \
       "bash -c \"${oneccl_env} \\\$(cat /mpirun_command) python /horovod/examples/keras/keras_mnist_advanced.py\""
+  fi
 
+  if [[ ${test} != *"torch1_2"* ]]; then
```
EnricoMi, line 183: We are adding a few more test runs here as we did not run it for tf2 before, now we do. This is expected?

test/single/data/expected_buildkite_pipeline.yaml
```diff
@@ -1000,6 +1000,22 @@ steps:
     automatic: true
   agents:
     queue: cpu
+- label: ':fire: MPI PyTorch MNIST (test-cpu-mpich-py3_8-tf2_4_1-keras2_4_3-torch1_7_1-mxnet1_7_0_p2-pyspark_3_0_1)'
```
EnricoMi, line 1003: here are the new tf2 runs of this example

EnricoMi (approved):
LGTM! Are you expecting the example script to run for tf2 now?

test/single/data/expected_buildkite_pipeline.yaml
tgaddair: I think this is okay, we should make the decision based on the version of PyTorch, not TF.

.buildkite/gen-pipeline.sh
tgaddair: Yeah, in my mind if the Docker has TF 2 and PyTorch 1.3, we should still run it, based off the version of PyTorch.

PR #2709, opened 2021-03-04
Fixed race condition in PyTorch MNIST example

tgaddair:


examples/elastic/pytorch/pytorch_mnist_elastic.py
```diff
@@ -48,13 +51,14 @@
...
 kwargs = {'num_workers': 1, 'pin_memory': True} if args.cuda else {}
-data_dir = args.data_dir or f'data-{hvd.rank()}'
-train_dataset = \
-    datasets.MNIST(data_dir, train=True, download=True,
-                   transform=transforms.Compose([
-                       transforms.ToTensor(),
-                       transforms.Normalize((0.1307,), (0.3081,))
-                   ]))
+data_dir = args.data_dir or './data'
+with FileLock(os.path.expanduser("~/.datalock")):
```
EnricoMi, line 55: Can we put it relative to the datadir, e.g. as `.lock` inside the datadir?

EnricoMi: isn't there a download MNIST call, and that should go into the lock, not this load MNIST call?

tgaddair: This won't work if the directory does not exist yet.

tgaddair: Renamed to `.horovod_lock` for clarity, let me know if that makes more sense. I believe this is the only place we're downloading the MNIST data.

EnricoMi (approved):
You could `os.makedirs(data_dir, exist_ok=True)` and then put the lock into `data_dir`. But as you like, `.horovod_lock` is already giving the lock file some contenxt

tgaddair:
It's a good point, I thought about maybe adding a wrapper around `FileLock` that does something like this. Could be useful as a follow-up.

PR #2718, opened 2021-03-15
Added cross rank and size helper functions

### LINKED ISSUE #2717: Diffrenet Data in different machine
**Environment:**
1. Framework: (TensorFlow, Keras, PyTorch, MXNet) pytorch
2. Framework version: 1.6.0
3. Horovod version: v0.21.0
4. MPI version: 4.0.5
5. CUDA version: 10.0
6. NCCL version: 2.7.6
7. Python version: 3.7
8. OS and version: ubuntu 16.04
9. GCC version: 5.4.0

**Checklist:**

**Your question:**
Please ask your question here.

I have a  relatively large dataset (about 3T), however, I have 2 machines and each only have 2T disk，so I cannot put the dataset in one machine.
I split the dataset into 2 parts, each machine stores 1.5T data, and I decide to use horovod for distributed training.
The traditional tutorial does not hold this situation in dataloader. I cannot get data1 on machine2.
So I have some questions on dataloader.
`train_sampler = torch.utils.data.distributed.DistributedSampler(
    train_dataset, num_replicas=hvd.size(), rank=hvd.rank())`
here num_replicas=hvd.size(), however, in this situation I think it should be num_replicas=hvd.size()/2, because I have 2 machines, and the rank=hvd.rank()%num_gpus_per_node.
I'm not sure whether it's right and I also want to know if there is a tutorial for this situation or what should I write this data sampler.

Thanks a lot!

tgaddair:
Fixes #2717.


chongxiaoc (approved):
Look good to me.
It seems that those MPI level functions are already there, we just forgot to export them into horovod function interfaces.

PR #2723, opened 2021-03-18
Add NVTX tracing hooks for profiling with Nsight Systems

maxhgerlach:

## Description

This adds two types of hooks to NVTX to make it easier to profile Horovod models with NVIDIA Nsight Systems:

1. Very low-overhead annotations containing just the type of communication op, timestamps for its beginning and end, as well as the tensor size in bytes. These ranges start when tensors are enqueued and end when the op has finished. When no profiler is attached, the performance impact should be absolutely negligible. I find it easier to work with these than to manually insert annotation ops into my model code.

2. More detailed annotations that are only active when a Horovod timeline is being written. Users may prefer to work with the Nsight application rather than the tracing UI in Chrome. Also it is interesting to correlate the Horovod timeline data with actual GPU kernel utilization and other profiling data available there. These include the full name of each source tensor, which necessitates some string copies, so there is a slight overhead. 

Building Horovod with `HOROVOD_WITHOUT_NVTX=1` disables all of this.

If Horovod is built with NVTX support, users can still set the environment variable `HOROVOD_DISABLE_NVTX_RANGES=1` to skip generating any ranges even when a profiler is attached.

Screen Shot Nsight with Horovod NVTX


tgaddair:
This is awesome, thanks for the contribution @maxhgerlach! 

@romerojosh can you help review?

tgaddair (approved):

horovod/tensorflow/mpi_ops.cc
```diff
@@ -412,10 +504,13 @@ class HorovodAllreduceOp : public AsyncOpKernel {
     auto hvd_context = std::make_shared<TFOpContext>(context);
     auto hvd_tensor = std::make_shared<TFTensor>(tensor);
     auto hvd_output = std::make_shared<TFTensor>(*output);
+    TensorflowNvtxRange nvtx_range(RegisteredNvtxOp::HorovodAllreduce,
+                                   hvd_tensor->size());
     auto enqueue_result = EnqueueTensorAllreduce(
         hvd_context, hvd_tensor, hvd_output, ready_event, node_name, device,
-        [context, done](const common::Status& status) {
+        [context, done, nvtx_range](const common::Status& status) mutable {
```
romerojosh, line 511: Would it be possible to move the creation of these ranges down into the Horovod backend, rather than create them within framework specific source? For example, moving the creation of these ranges into the `EnqueueTensor*` functions in `operations.cc`, and then adding the range end call to the code in `GPUOpContext::FinalizeGPUQueue` in `gpu_operations.cc`?  The benefit of this approach is that it would not require adding this NVTX code to every framework implementation individually.

romerojosh (commented):
Hey @maxhgerlach, this is an excellent contribution! I often will add temporary NVTX ranges in my own development to get more insight from profiles generated from nsys, so having this functionality available to users is great. I will try and run a few workloads using this PR in the next couple of days to comment on the usability of the generated markers.

I think the implementation is solid in general, but did leave a comment concerning handling ranges within the framework specific code.

maxhgerlach:
Thanks for reviewing this, @tgaddair and @romerojosh!

In my experience the detailed annotations are most helpful to understand the impact of communication ops inserted into the forward pass of a model. Making sense of dozens of (potentially fused) gradient allreduces is trickier with the Nsight UI.

horovod/tensorflow/mpi_ops.cc
maxhgerlach: I agree that it would be nicer to not artificially couple this to the TF framework source. However, if I move the creation of these ranges somewhere else, I can no longer easily end these ranges from the callbacks because these callbacks are only defined in framework code. I would have to explicitly track the nvtx_range object somewhere (maybe in the tensor table entry?) and make sure to explicitly end it in every possible place (basically everywhere the callback is called in `operations.cc` and `gpu_operations.cc`).

How would you feel about pulling the (appropriately renamed) TensorflowNvtxRange into a common header file? I could then add those few lines to start and end ranges to `torch/mpi_ops_v2.cc` and `mxnet/mpi_ops.cc`.

romerojosh: A header is one way to make it more generic, but still requires anyone who adds a new op to also make sure to add code to add this range handling to their callbacks which isn't ideal. I agree that that `nvtx_range` objects will need to be stored, and I think it is ok to add them to the tensor table entries. To deal with ending ranges where callbacks are called, maybe we can add a method to the tensor table entries like `TensorTableEntry::ExecuteCallback(Status& s)` that will end the nvtx range in the entry and then call the callback? Then, we would just replace direct calls like `e.callback(status)` to `e.ExecuteCallback(status)` to the callback with this method. 
There is one place in the code where callbacks are extracted from the entries and placed into a separate vector here: https://github.com/horovod/horovod/blob/612fe6b21aa51f7683682950a6a07d4bc6935713/horovod/common/operations.cc#L563-L569
However, the callbacks can just be called within `FinalizeTensorQueue`, since there doesn't seem to be a reason they are gathered and called outside.

maxhgerlach: OK, you convinced me to go with your suggestion. Thanks for the pointers! In the updated version the NVTX ranges are independent of any framework-specific code. That should indeed make it easier to contribute new ops in the future.

Now I've touched more parts of the codebase than I originally set out to. Therefore I am all the more open to advice on how to name and structure things.

horovod/common/operations.cc
```diff
@@ -1010,7 +1001,20 @@ Status EnqueueTensorAllreduces(std::vector<std::shared_ptr<OpContext>>& contexts
...
+  if (tensors.size() == 1) {
+    auto& e = entries[0];
+    e.nvtx_op_range.Start(RegisteredNvtxOp::HorovodAllreduce, e.tensor->size());
+  } else {
+    auto total_size =
+        std::accumulate(entries.begin(), entries.end(), 0ll,
+                        [](int64_t size_sum, const TensorTableEntry& e) {
+                          return size_sum + e.tensor->size();
+                        });
+    auto& e = entries[tensors.size() - 1];
```
romerojosh, line 1016: There is a chance that this range for the grouped allreduce might be slightly inaccurate in some cases. There is a guarantee that the tensors submitted in a group will be involved in collectives executed on the same cycle, but there is no guarantee that they will all be a part of the same call to `PerformOperation`. For example, if the group contains a set of entries with a combined size that exceeds `HOROVOD_FUSION_THRESHOLD`, that will result in multiple calls to `PerformOperation` on the separate fused buffers. Subsequently, the NVTX range will be ended by whichever of those `PerformOperation` calls contains `entries[tensor.size() - 1]`, which isn't necessarily the last one executed. 

Perhaps there is a way to get around this by having the entries of the group share ownership of a range (maybe store a `shared_ptr` to the range in the tensor table entry), and have the range end on object destruction?

romerojosh (commented):
Thanks for updating the PR @maxhgerlach to this more framework general approach! 

I left a comment concerning how the ranges are handled with the grouped allreduce. 

Besides this comment, would you be willing to add an environment variable option to disable the NVTX range generation, for example `HOROVOD_DISABLE_NVTX_RANGES`? There are still many NVTX ranges created in the path outside the timeline (e.g. one per every gradient allreduced in a typical case) so it would be nice to have a mechanism to disable them in an NVTX enabled build.

horovod/common/operations.cc
maxhgerlach: Thanks for catching that mistake! 

I ended up with an opaque wrapper around a `shared_ptr` to a range object. This way the compiler can optimize it away if we build with `HOROVOD_WITHOUT_NVTX=1`.

maxhgerlach:
Hey @romerojosh, @tgaddair, I've added an environment variable `HOROVOD_DISABLE_NVTX_RANGES` to disable generating any NVTX ranges from Horovod and I believe to have fixed the issue with grouped allreduces. Happy to address further comments!

There seems to be a general problem with the buildkite pipeline right now. It [fails to build](https://buildkite.com/horovod/horovod/builds/5178#c2848301-c3b4-400c-9671-f0088a6c7360) images based on `tfhead` because of an error with Keras. Apparently tests aren't run in the other containers either.

tgaddair:
Hey @maxhgerlach, tests should be fixed on HEAD now, thanks for rebasing. @romerojosh is there anything else you wanted to address before merging?

romerojosh (approved):
LGTM!

whatdhack:
@maxhgerlach ,  thanks for a great add. A questions.  Does the output get added to the Horovod TImeline json file or a new separate file is created ?

maxhgerlach:
Hi @whatdhack, 

thanks for your interest. You will need to run your Horovod script through `nsys` to gather profiling data, which will be collected in a separate file unrelated to the Horovod timeline json. This data can then be viewed and processed in the Nsight Systems UI.

You might want to check out the documentation for Nvidia Nsight to see the various options how to instrument the profiler. Often I would use `nsys launch` from a shell script, limiting it to rank 0 only, then enable the profiling for a few training steps via `nsys start`.

whatdhack:
@maxhgerlach , thanks for the quick reply and clarifying that ehe  NVTX data  is collected per rank , and is limited to a small number of ranks, as of course if there are 1000s of ranks , then it is not viable to look at the NVTX output for each rank.   Does Nsight  do any post processing to create an aggregate profile/report at this point ?  Also, is the NVTX output able to capture information on  the  negotiation mpi calls ?

maxhgerlach:
Hey @whatdhack, 

Nvidia has quite a thorough [User Guide](https://docs.nvidia.com/nsight-systems/UserGuide/index.html) that should help you get started. 

If you plug in the profiler similarly to `nsys {launch|profile} [nsys options] mpirun [mpi options] python horovod_script.py`, you will get an aggregate report encompassing multiple ranks (only tried this for small local jobs myself). It will also automatically insert NVTX ranges for collective MPI calls so you should see those negotiations, there is a section in the guide about that.

The functionality from this PR helps you correlate what Horovod is doing with all the rich information the nsys profiler is giving you on its own.

Issue #2726, opened 2021-03-19
Support MXNet with oneDNN replacing MKLDNN


EnricoMi:
Horovod supports MXNet only with MKLDNN headers. Version 2.0.0 renames [MKLDNN to oneDNN](https://github.com/apache/incubator-mxnet/commit/aa92db02201799841bf52db5a0b8507719f04e61) ([mxnet issue #20054](https://github.com/apache/incubator-mxnet/issues/20054)).

An approach for Horovod to work with MXNet pre 2.0.0 (MKLDNN) and post 2.0.0 (oneDNN) is to duplicate how Horovod builds against MKLDNN, including [CMake detection code](https://github.com/horovod/horovod/blob/master/cmake/Modules/FindMxnet.cmake). Depending on the MXNet version, we can than build it with MKLDNN or oneDNN support.

EnricoMi:
@tgaddair @leezu what are your thoughts?

tgaddair:
Sounds good to me! I think some minor duplication in CMake is fine.

EnricoMi:
#2728 fixes this clean enough, once mxnet fully renames the feature to onednn, there will be some more work to be done. Closing for now.

PR #2728, opened 2021-03-20
Make horovod work with mxnet and onednn headers

EnricoMi:
Work towards #2726. This is a quick fix, but I think oneDNN should become a full feature as MKLDNN. So it should not say MKLDNN where it actually uses oneDNN headers.


EnricoMi:
@leezu is the feature MKLDNN still called MKLDNN in MXNet, though it is renamed to oneDNN?

https://github.com/apache/incubator-mxnet/commit/aa92db02201799841bf52db5a0b8507719f04e61#diff-a026da7a53c6578d97815a6cc5803d74f4f702beedde22047648aead2e81b57e

https://github.com/apache/incubator-mxnet/blob/master/tools/pip/setup.py#L143-L151

leezu (approved):

tgaddair (approved):

EnricoMi:
I have merged this so master works with mxnet nightly. Will further fix in case mxnet head renames the feature to to 'ONEDNN'.

PR #2730, opened 2021-03-21
Add PyTorch 1.8.1 to test space, deprecate 1.2.0

EnricoMi:
PyTorch 1.8.1 has been released earlier this month. Adding this to test space while deprecating (test) support for 1.2.0.

### DIFF
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -7,7 +7,7 @@ set -eu
 repository=823773083436.dkr.ecr.us-east-1.amazonaws.com/buildkite
 
 # our baseline test is
-baseline="test-cpu-gloo-py3_8-tf2_4_1-keras2_4_3-torch1_8_0-mxnet1_7_0_p2-pyspark_3_0_1"
+baseline="test-cpu-gloo-py3_8-tf2_4_1-keras2_4_3-torch1_8_1-mxnet1_7_0_p2-pyspark_3_0_1"
 # in run_gloo_integration we run 'Elastic Spark * Tests' for this baseline
 # so it has to have Gloo mpi kind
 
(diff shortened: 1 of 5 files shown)


EnricoMi:
@leezu is there a chance MXNet releases mxnet-cu111==1.8.0

Horovod tests PyTorch and MXNet framework together in one Docker image:
- latest PyTorch (1.8.0) is available for CUDA 10.1, 10.2, and 11.1
- latest MXNet (1.8.0) is available for CUDA 11.0 and 11.2

So there is currently no way to include MXNet into Horovods test space right now.

EnricoMi:
> @EnricoMi CUDA 11.1 introduces CUDA Enhanced Compatibility which provides two benefits:
> 
> > By leveraging semantic versioning across components in the CUDA Toolkit, an application can be built for one CUDA minor release (such as 11.1) and work across all future minor releases within the major family (such as 11.x).
> > CUDA has relaxed the minimum driver version check and thus no longer requires a driver upgrade with minor releases of the CUDA Toolkit.
> 
> So I suggest you use Cuda 11.2 in the container and Pytorch for Cuda 11.1 should work without problems. In my experience there were speed regressions in Cuda 11.1 compared to 11.0, so we did not create a Cuda 11.1 release.

Thanks for the background info, this will require some Dockerfile-rework but it sounds reasonable to utilize that major version compatibility. I'll give it a try.

EnricoMi:
@leezu btw., mxnet-cu112 has not been released to [PyPi](https://pypi.org/project/mxnet-cu112/) yet, is there an ETA?

leezu:
> @leezu btw., mxnet-cu112 has not been released to PyPi yet, is there an ETA?

It's pending https://github.com/pypa/pypi-support/issues/970
For now, have you tried mxnet-cu110 inside the 11.2 container?

EnricoMi:
Blocked by fixing torch tests broken with torch 1.8.0 (#2731).

tgaddair (approved):

EnricoMi:
@tgaddair thanks!

PR #2731, opened 2021-03-21
Docker should not cache nightly packages

EnricoMi:
The tensorflow, torch, torchvision and mxnet dependencies are installed in `Dockerfile.test.cpu` and `Dockerfile.test.gpu` before Horovod sources are copied, thus they are cached by Docker as long as the RUN command does not change. This should not happen for nightly packages.

This PR moves the installation of nightly packages after copying Horovod sources. This forces their download and installation on every run. Release versions, however, are still installed before the copy and thus subject for caching.

See all the RUN commands before `Step 50/63 : COPY . /horovod` in https://buildkite.com/organizations/horovod/pipelines/horovod/builds/5125/jobs/b3a90cef-21b8-41b9-8595-b5a25c16bd51/raw_log. **They are all cached, which is a great cache hit rate if it wouldn't be *test-gpu-gloo-py3_8-tfhead-keras2_4_3-torchhead-mxnethead-pyspark_3_0_1*, which is bad for testing nightly versions.**

This is probably the reason why we haven't seen `mxnet-nightly-cu110` fail earlier when they have moved from MKLDNN headers to oneDNN headers.

Also fixes this warning that does not make the build fail, so it is burried in the logs:
```
pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
tf-nightly-gpu 2.6.0.dev20210331 requires h5py~=3.1.0, but you have h5py 2.10.0 which is incompatible.
```

Further, `Pillow` had been fixed to `<7.0` though the respective bug has been fixed in `torchvision>=0.5.0`. To avoid issues with future torch vision version, this constraint has been relaxed.

Installation of nightly torch packages is done with `-v`, which pollutes the logs with thousands of lines. Enable when really needed to debug something.

Depends on #2814 being merged into master first.

### DIFF
--- a/Dockerfile.test.cpu
+++ b/Dockerfile.test.cpu
@@ -125,12 +125,15 @@ RUN if [[ ${MPI_KIND} != "None" ]]; then \
     fi
 
 # Install TensorFlow and Keras (releases).
-# Pin h5py: https://github.com/h5py/h5py/issues/1732
+# Pin h5py only for tensorflow<2.5: https://github.com/h5py/h5py/issues/1732
 # Pin scipy!=1.4.0: https://github.com/scipy/scipy/issues/11237
 RUN if [[ ${TENSORFLOW_PACKAGE} != "tf-nightly" ]]; then \
         pip install ${TENSORFLOW_PACKAGE}; \
         if [[ ${KERAS_PACKAGE} != "None" ]]; then \
-            pip install ${KERAS_PACKAGE} "h5py<3" "scipy!=1.4.0" "pandas<1.1.0"; \
+            if [[ ${TENSORFLOW_PACKAGE} == tensorflow*==1.* ]] || [[ ${TENSORFLOW_PACKAGE} == tensorflow*==2.[01234].* ]]; then \
+                h5py="h5py<3"; \
+            fi; \
+            pip install ${KERAS_PACKAGE} ${h5py:-} "scipy!=1.4.0" "pandas<1.1.0"; \
         fi; \
         mkdir -p ~/.keras; \
         python -c "import tensorflow as tf; tf.keras.datasets.mnist.load_data()"; \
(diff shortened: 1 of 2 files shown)


tgaddair (approved):
Good catch!

EnricoMi:
@tgaddair this is blocked by the tests in `test_torch.py` that break for `torch==1.8.0` and `torchhead`. I would prefer those tests to be fix here.

EnricoMi:
Looks like [tf 2.5 looks for CUDA 11.0](https://buildkite.com/horovod/horovod/builds/5174#526ec731-9ab2-4aad-8e11-af211817afeb/215-247) where we use11.2 because other frameworks needs it:

    2021-03-26 12:50:07.503321: W tensorflow/stream_executor/platform/default/dso_loader.cc:60] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory

Is this related to https://github.com/tensorflow/tensorflow/issues/46093?

@romerojosh @DEKHTIARJonathan any ideas / pointers?

DEKHTIARJonathan:
> Looks like [tf 2.5 looks for CUDA 11.0](https://buildkite.com/horovod/horovod/builds/5174#526ec731-9ab2-4aad-8e11-af211817afeb/215-247) where we use 11.2 because other frameworks needs it:
> 
> ```
> 2021-03-26 12:50:07.503321: W tensorflow/stream_executor/platform/default/dso_loader.cc:60] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory
> ```
> 
> Is this related to [tensorflow/tensorflow#46093](https://github.com/tensorflow/tensorflow/issues/46093)?
> 
> @romerojosh @DEKHTIARJonathan any ideas / pointers?

You can rebuild TF manually (probably not the answer you look for) with TF 11.2.
Otherwise I'm not sure if there's a way around that

EnricoMi:
@DEKHTIARJonathan thanks for the info, looks like that error was a red herring. I have found another issue in the docker-compose.test.yml, now all builds are green.

PR #2733, opened 2021-03-22
Make horovod work with mxnet onednn feature (#2726)

EnricoMi:
MXNet is going to rename the MKLDNN feature to ONEDNN in mxnet head. This should handle that "new" feature appropriately.

https://github.com/apache/incubator-mxnet/pull/20070

Contains #2830.


cmake/Modules/FindMxnet.cmake
```diff
@@ -71,6 +77,4 @@ else()
   set(Mxnet_COMPILE_FLAGS "${Mxnet_COMPILE_FLAGS} -D_GLIBCXX_USE_CXX11_ABI=0")
 endif()
 
-mark_as_advanced(Mxnet_INCLUDE_DIRS Mxnet_LIBRARIES Mxnet_COMPILE_FLAGS Mxnet_USE_MKLDNN Mxnet_USE_CUDA Mxnet_VERSION)
-
-mark_as_advanced(Mxnet_INCLUDE_DIRS Mxnet_LIBRARIES Mxnet_COMPILE_FLAGS Mxnet_USE_MKLDNN Mxnet_USE_CUDA Mxnet_VERSION)
+mark_as_advanced(Mxnet_INCLUDE_DIRS Mxnet_LIBRARIES Mxnet_COMPILE_FLAGS Mxnet_USE_MKLDNN Mxnet_USE_ONEDNN Mxnet_USE_CUDA Mxnet_VERSION)
```
EnricoMi, line 80: @leezu you have duplicated this line in https://github.com/horovod/horovod/commit/bec40a7d9ccad52bb0dc580aeac09b2b7427fc2a#diff-bd892e8ebe35ec5332e410bd9d22dd664bfb6d92ab50b7938aa4dcf3e180bd4eR66:

https://github.com/horovod/horovod/blob/0acb1a53cc832c8c580e03c1a29c0e14d95234a1/cmake/Modules/FindMxnet.cmake#L74-L76

Was that intentional?

leezu: No, I don't think so

EnricoMi: Thanks, I will remove it then

EnricoMi:
Blocked by fixing torch tests broken with torchhead (#2731).

leezu:
Thank you @EnricoMi for the fix. Is the PR still blocked?

EnricoMi:
It is not blocked. Either mxnet has a bug or horovod's CMake code. The last time ran the tests the mxnet nightly should have had the onednn feature  but the build failed. I did not find time to check what feature the CMake compile step is seeing, whether it is MLKDNN or ONEDNN.

EnricoMi:
@leezu @tgaddair Buildkite is happy with mxnet head, but TravisCI does not like this fix. Can't really see why TravisCI does not see the mxnet that is being installed. Maybe this is macOS specific. Any ideas?

leezu:
Thanks @EnricoMi. Running the installation step on a Mac locally invokes `/usr/local/opt/python@3.7/bin/python3.7 -c "import os; import mxnet as mx; import build_utils; print(mx.__version__); print(mx.libinfo.find_include_path()); print(' '.join(mx.libinfo.find_lib_path())); print(build_utils.is_mx_mkldnn()); print(build_utils.is_mx_onednn()); print(build_utils.is_mx_cuda())"`

Running that manually provides

```
1.5.0
/usr/local/lib/python3.7/site-packages/mxnet/include/
/usr/local/lib/python3.7/site-packages/mxnet/libmxnet.so
False
INFO: Cannot detect if ONEDNN is enabled in MXNet. Please set MXNET_USE_ONEDNN=1 if ONEDNN is enabled in your MXNet build.
False
False
```

So the problem here may be that the feature-check for onednn raises an exception in MXNet 1.5 and the horovod build_utils.py then prints the INFO line, confusing the cmake script

leezu (approved):
Thank you!

PR #2747, opened 2021-03-26
TF 2.5 requires h5py~=3.1 and keras-nightly

EnricoMi:
Master build is broken: https://buildkite.com/horovod/horovod/builds/5177#2cb0a574-204e-4040-9953-ee06a5d1495d/213-1775

The tf-nightly requires h5py~=3.1 where we used to downgrade it to h5py<3:

    tf-nightly 2.5.0.dev20210326 requires h5py~=3.1.0, but you have h5py 2.10.0 which is incompatible.

Further, tf-nightly requires keras-nightly:

    AttributeError: module 'keras.utils.generic_utils' has no attribute 'populate_dict_with_module_objects'


EnricoMi:
@tgaddair master has been broken for some days, where it does not get past building. This will fix it partly, but at least we some test coverage back. I will look into this over the weekend or on Monday. Please merge it into master if you are happy.

EnricoMi:
Now we end up with a segfault: https://buildkite.com/horovod/horovod/builds/5186#c90f0044-5ec1-49b5-a9cd-d97c8e9193f1/214-236

Maybe we should disable nightlies for now.

EnricoMi:
Fixed by #2750 and #2731.

PR #2749, opened 2021-03-27
Split pipeline into more stages, fan out slowly, nightlies last

EnricoMi:
This splits the Buildkite pipeline into more stages. Each stage builds only the images tested in the stage, then runs cpu, then gpu tests and waits for completion.

1. build all images except with nightly head versions
2. baseline (CPU)
3. MPI kind variations (CPU)
4. ML framework variations (CPU)
4. PySpark variations (CPU
6. baseline (mixed)
7. ML framework variations (GPU)
9. build all images with nightly head versions
10. ML framework head versions (CPU+GPU)

This should reduce the costs for builds that have failures while increasing wall-clock time of successful builds. **Further this moves nightly versions to the end of the pipeline so we see all released versions pass before nightlies break our pipeline.**

This does not change the tests that run, only the order.

### DIFF
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -14,15 +14,15 @@ baseline="test-cpu-gloo-py3_8-tf2_4_1-keras2_4_3-torch1_7_1-mxnet1_7_0_p2-pyspar
 # skip tests when there are no code changes
 dir="$(dirname "$0")"
 code_files=$(python "$dir/get_changed_code_files.py" || echo failure)
-tests=$(if [[ "${BUILDKITE_BRANCH:-}" == "${BUILDKITE_PIPELINE_DEFAULT_BRANCH:-}" ]] || [[ -n "$code_files" ]]; then
-  # we vary the baseline along the Python dimension and PySpark together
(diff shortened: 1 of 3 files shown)


EnricoMi:
@tgaddair I think with this we can keep the nightlies in the pipeline, they fail after all other tests pass. What do you think?

EnricoMi:
Buildkite time for this pipeline goes up by ~1 hour to 2.5 hours.

tgaddair:
Nice work on this @EnricoMi!  I think it makes sense to make sure nothing runs after the nightly tests, though we may also be able to speed things up by running them in parallel with the previous set of tests. That way they don't prevent other tests from running, but they also don't slow down the pipeline as significantly. What do you think?

EnricoMi:
Good point, but then we need to build the images before that block where tests run. Then, broken builds will prevent all tests from running.

tgaddair:
Ah, yeah I see what you mean. It's probably okay to do them in sequence, the more I think about it, since many of the CPU tests will be moving to GitHub Actions.

EnricoMi:
Obsoleted by #2798 which might be near to land.

PR #2750, opened 2021-03-27
Removed keras from tf-nightly

tgaddair:


EnricoMi (approved):
tfhead seems to work with keras-nightly, see https://github.com/horovod/horovod/pull/2747/commits/a8998cfe57a8e1669b18b4f7ca2e9fbc2e776c59 in #2747.

But I am also fine with this if it works.

PR #2814, opened 2021-04-01
Pin tf-nightly to 2.6.0.dev20210331

EnricoMi:
TF-Head seems to break our tests in `test-cpu-gloo-py3_8-tfhead-keras_none-torchhead-mxnethead-pyspark_3_0_1`:

```
test_keras.py:23: in <module>
    import keras
/usr/local/lib/python3.8/dist-packages/keras/__init__.py:25: in <module>
    from keras import models
/usr/local/lib/python3.8/dist-packages/keras/models.py:20: in <module>
    from keras import metrics as metrics_module
/usr/local/lib/python3.8/dist-packages/keras/metrics.py:27: in <module>
    from keras import activations
/usr/local/lib/python3.8/dist-packages/keras/activations.py:20: in <module>
    from keras.layers import advanced_activations
/usr/local/lib/python3.8/dist-packages/keras/layers/__init__.py:24: in <module>
    from keras.engine.input_layer import Input
/usr/local/lib/python3.8/dist-packages/keras/engine/input_layer.py:21: in <module>
    from keras.engine import base_layer
/usr/local/lib/python3.8/dist-packages/keras/engine/base_layer.py:41: in <module>
    from keras.mixed_precision import loss_scale_optimizer
/usr/local/lib/python3.8/dist-packages/keras/mixed_precision/loss_scale_optimizer.py:1180: in <module>
    mixed_precision._register_wrapper_optimizer_cls(optimizer_v2.OptimizerV2,
E   AttributeError: module 'tensorflow.python.training.experimental.mixed_precision' has no attribute '_register_wrapper_optimizer_cls'
```

PR #2816, opened 2021-04-01
Move GPU torch nightlies to cu111

EnricoMi:
We pull GPU torch nightlies from cu110, but nightlies moved to cu111.

Depends on #2814 being merged into master first.


EnricoMi:
Done in # #2730.

PR #2830, opened 2021-04-08
Un-pin mxnet-nightly

EnricoMi:
In #2750 we have pinned mxnet nightly, this is to unpin it again once this is fixed upstream.


EnricoMi:
Fixed in #2733.

PR #2798, opened 2021-03-28
Add GitHub Workflow CI

EnricoMi:
A GitHub Workflow for running the CPU test pipeline.
[grafik](https://github.com/horovod/horovod/actions/runs/1003273503)

CPU tests finish in 1:30 on GitHub, GPU tests finish in 1:15 on Buildkite.

The workflow file is generated by Python script `.github/gen-workflow-ci.py` based on the output of `.buildkite/gen-pipeline.sh`, so that we implement test logic only in one place.

These are the implemented features:
- builds all CPU images and runs appropriate tests
- builds GPU images but does not run any tests
- tests for GPU images are run on Buildkite after all above builds and tests succeed
- tests three images on macOS (replaces TravisCI)
- images where all frameworks are nightly versions do not prevent the Buildkite tests
- PRs without code changes (e.g. docs only) to not run the full build and test cycle (that is what Buildkite pipeline does)
- test results are downloaded from Buildkite and together with GitHub test results posted to the pull request
- build and test steps have timeouts and are [retried three times](https://github.com/horovod/horovod/runs/2255478598?check_suite_focus=true#step:15:910)
- workflow detects when its yaml file is not up-to-date
- earlier workflow runs on the same branch are cancelled
- triggering the Buildkite builds and publishing test results for fork PRs (can only be tested after merging to master)
- daily master build schedule
- Docker images are build on PR (not pushed), as well as master and version tags (and pushed to Dockerhub), but only if the build and test jobs succeed
- Docs and example sync workflows integrated into the CI workflow

Limitations:
- Buildkite build always has [author "Travis Addair" and build title "GPU Tests triggered by GitHub"](https://buildkite.com/horovod/horovod/builds/5298): we may use a service account here.
- ONECCL image does not build and is excluded (af2f3bbf): https://github.com/horovod/horovod/runs/2221724507?check_suite_focus=true (#2846)
- Workflow runs on one branch block other workflow runs, so that GitHub workflows can quickly saturate

Missing:
- disable Buildkite triggered by git pushs (with this merged into master the buildkite pipeline is empty)

PR #2839, opened 2021-04-12
Concurrently running collective operations on process subsets [TensorFlow]

### LINKED ISSUE #105: allreduce on a subset of processes
Hello, 
Is it possible to call `allreduce` on a subset of Horovod processes? This is useful when your doing distributed batch normalization, i.e. , you estimate the batch statistics for a single node then `allreduce` them across a subset of nodes... this way the batch size used in the batch normalization can be independent from the batch size in the overall gradient computation.

### LINKED ISSUE #2139: Extend API to allow running collective ops across a subset of ranks (cross, local, etc.)
Currently, Horovod's API is restricted to global communicators, but there are some use cases (e.g., Sync Batch Norm, #1923), where it can be useful to only perform ops within a node or similar.

We could imagine extending the API with an additional parameter:

```
hvd.allreduce(t, comm=hvd.Local)
```

maxhgerlach:

## Description

This WIP pull request aims to resolve the limitation of Horovod that multiple subsets of the world of Horovod processes cannot run distinct collective operations in parallel. For this draft I followed suggestions in comments by @romerojosh and @tgaddair. It would also be great to unify this work with @EiffL's related pull request that appeared shortly before this draft (I already took some inspiration in intermediate refactorings).

At this time, this includes three ways to, e.g., build a Horovod model to be trained by four worker processes with two concurrent allreduce operations on the "even" or "odd" subset:

### 1) Static process sets
```
# on all ranks
even_set = hvd.ProcessSet([0,2])
odd_set = hvd.ProcessSet([1,3])
hvd.init(process_sets=[even_set, odd_set])

for p in [hvd.global_process_set, even_set, odd_set]:
  print(p)
# ProcessSet(process_set_id=0, ranks=[0, 1, 2, 3], mpi_comm=None)
# ProcessSet(process_set_id=1, ranks=[0, 2], mpi_comm=None)
# ProcessSet(process_set_id=2, ranks=[1, 3], mpi_comm=None)

# on ranks 0 and 2
result = hvd.allreduce(tensor_for_even_ranks, process_set=even_set)

# on ranks 1 and 3
result = hvd.allreduce(tensor_for_odd_ranks, process_set=odd_set)
```
Having initialized Horovod like this, the configuration of process sets cannot be changed without restarting the program. If one only uses the default global process set 0, there is no impact on performance.

### 2) Static process sets from MPI communicators
```
# on all ranks
from mpi4py import MPI
comm = MPI.COMM_WORLD
subcomm = MPI.COMM_WORLD.Split(color=MPI.COMM_WORLD.rank % 2,
                               key=MPI.COMM_WORLD.rank)

split_process_set = hvd.ProcessSet(subcomm)

hvd.init(comm, process_sets=[split_process_set])

for p in [hvd.global_process_set, split_process_set]:
    print(p)
# ProcessSet(process_set_id=0, ranks=[0, 1, 2, 3], mpi_comm=<mpi4py.MPI.Intracomm object at 0x7fb817323dd0>)
# ProcessSet(process_set_id=1, ranks=[0, 2], mpi_comm=<mpi4py.MPI.Intracomm object at 0x7fb87e2ddfb0>)
## (split_process_set differs by rank)

# on ranks 0 and 2
result = hvd.allreduce(tensor_for_even_ranks, process_set=split_process_set)

# on ranks 1 and 3
result = hvd.allreduce(tensor_for_odd_ranks, process_set=split_process_set)
```
@EiffL, here I incorporated some of your work from https://github.com/DifferentiableUniverseInitiative/horovod/pull/2 -- I hope that's fine with you. This allows for a very similar API to your PR with the slight limitation that there must absolutely be a global communicator for all Horovod processes. 

### 3) Dynamic process sets
```
# on all ranks
hvd.init(process_sets="dynamic")  # alternatively set HOROVOD_DYNAMIC_PROCESS_SETS=1
even_set = hvd.add_process_set([0,2])
odd_set = hvd.add_process_set([1,3])

for p in [hvd.global_process_set, even_set, odd_set]:
  print(p)
# ProcessSet(process_set_id=0, ranks=[0, 1, 2, 3], mpi_comm=None)
# ProcessSet(process_set_id=1, ranks=[0, 2], mpi_comm=None)
# ProcessSet(process_set_id=2, ranks=[1, 3], mpi_comm=None)

# on ranks 0 and 2
result = hvd.allreduce(tensor_for_even_ranks, process_set=even_set)

# on ranks 1 and 3
result = hvd.allreduce(tensor_for_odd_ranks, process_set=odd_set)
```
Process sets can be registered and deregistered dynamically via `hvd.add_process_set()` and `hvd.remove_process_set()`. Calls to these functions must be made identically and in the same order by all processes. This comes with some synchronization overhead in `RunLoopOnce()`.

### Op support

On TensorFlow the `process_set` argument works for: allgather, allreduce, alltoall, broadcast, grouped_allreduce, DistributedGradientTape, DistributedOptimizer

### Refactoring
This entails some refactoring of the Horovod codebase:
- Data structures that are specific to one set of processes are moved from `HorovodGlobalState` to separate `ProcessSet` structs. There is one global `ProcessSetTable`.
- There is one `Controller` for each `ProcessSet`. Some timeline-specific concerns have been moved out of the `Controller` class.
- Each process set holds its own `MPIContext` or `GlooContext` with communicators limited to that subset of processes. The other context objects that are not related to the control plane like `NCCLContext` and (eventually) `CCLContext` can remain global if they do some minor extra internal bookkeeping.

### State
The MPI, Gloo, and NCCL backends as well as the TensorFlow bindings are quite feature complete now and there is reasonable test coverage. We have started to use this branch actively, so it would be much appreciated if this could be merged into Horovod eventually. Once there is some consensus on the way to proceed here, I should be able to find some time to look into supporting PyTorch and MXNet (OneCCL might be a bit trickier).

Incomplete list of things still to do:
- [x] enable remaining ops besides `hvd.allreduce` + tests
- [x] consistency checks
- [x] `StallInspector` should be process-set specific
- [x] gradients
- [x] fix builds with Gloo, CCL
- [x] fix potential hangs at shutdown with static process sets
- [ ] Pytorch bindings
- [ ] Mxnet bindings
- [x] support process sets with Gloo controller and operations
- [ ] support process sets with CCL operations
- [ ] initialization code is a bit spread out now, maybe this could be cleaned up somewhat
- [ ] autotuning only supports the global process set for now, should that be extended?
- [ ] Adasum only supports the global process set for now; extending this looks a bit more involved
- [x] nicer Python API
- [x] DistributedOptimizer
- [x] DistributedGradientTape

Fixes #105 (eventually)
Fixes #2139 (eventually)


maxhgerlach:
rebased to master, fixing small conflicts with #2874

maxhgerlach:
rebased to master and removed some flakiness

tgaddair:
Awesome! @romerojosh can you help review?

romerojosh:
@maxhgerlach Excellent work, glad to see this PR moving forward! I will try and get a first pass review done this week.

horovod/common/response_cache.cc
```diff
@@ -360,8 +360,8 @@ const std::set<uint32_t>& CacheCoordinator::timeline_bits() const {
   return timeline_bits_;
 }
 
-bool CacheCoordinator::should_shut_down() const {
-  assert(synced_);
+bool CacheCoordinator::should_shut_down(bool dont_worry_about_synced) const {
```
romerojosh, line 363: Rather than modify the API here, we can just remove the assert I think.

horovod/tensorflow/__init__.py
```diff
@@ -34,6 +34,7 @@
 from horovod.tensorflow.mpi_ops import mpi_threads_supported, mpi_enabled, mpi_built
```
romerojosh, line 34: We might want to add the ability to specify a process set to the `DistributedOptimizer` and `DistributedGradientTape` wrappers.

horovod/common/operations.cc
```diff
@@ -603,86 +661,136 @@ bool RunLoopOnce(HorovodGlobalState& state) {
...
+#endif // HAVE_MPI
 
-    LOG(TRACE, rank) << "Performing " << response.tensor_names_string();
-    LOG(TRACE, rank) << "Processing " << response.tensor_names().size()
-                     << " tensors";
-    PerformOperation(response, horovod_global);
-    LOG(TRACE, rank) << "Finished performing "
-                     << response.tensor_names_string();
+  std::unordered_map<int32_t, ResponseList> process_set_response_lists;
+  for (auto process_set_id : state.process_set_table.Ids()) {
```
romerojosh, line 680: Would it be faster to interleave the `ComputeResponseList` calls and `PerformOperation` calls here, rather than doing them in two separate loops? This would allow more rapid scheduling of collective ops and potentially less GPU idle time, since you aren't waiting for coordination of every process group in a big sequential chunk before launching collectives. 

Caveat: There is the WAIT_FOR_DATA spin loop that will hold back the background thread on every `PerformOperation` call, however, I am planning on posting a PR soon to modify that to use a more asynchronous `cudaStreamWaitEvent` to manage the data dependency. Seems like this feature could benefit from that.

romerojosh (commented):
Hi @maxhgerlach, this PR is looking excellent! I have done a first brief pass over the code and left a few initial comments. 

I think this is very well designed, which shows given the relatively minimal changes required to the op implementations once you set up the process sets.

I'm kind of questioning whether the dynamic addition/subtraction of process sets is worth the added complexity of this PR. Trying to think of real use cases that would require this (other than providing a more convenient API for lazier users :) ).  The static initialization is nice since it better enforces the requirement that all process sets are defined on all processes, in the same order. However, you've already implemented it, so there is that too.

maxhgerlach:
Hi @romerojosh, thanks for taking the time for such a thorough first review and raising some very good points!

Regarding the dynamic process sets: When starting out with this branch, the API with "hvd.add_process_set()` and `hvd.remove_process_set()` felt most natural to me. As I proceeded then, I gained a deeper understanding of the internal architecture of Horovod and learnt about these intricate synchronization issues. 

I agree that most serious users will be able to define their process sets at initialization time although it is a bit less convenient (for instance `hvd.size()` isn't available before `hvd.init()` so that information must be pulled from somewhere else). 

Anyway, there could be some training schemes where it is practical to define process sets relatively late. For instance, you might start out with a very sparse model (certain parameters not shared between workers) and then dynamically converge towards a somewhat denser model (starting to share some subsets of these parameters between subsets of Horovod workers). That decision would depend on the training process for these sparse sets of parameters and would be different from experiment to experiment so it might not be possible to foresee the most fitting process sets at initialization time. Then again, users interested in such a fine grained training regime could also be asked to dump their parameters and restart their training jobs.

Personally I'd leave that feature in because the code is already written and seems to work well by now, but I'd also totally understand the reasoning behind cutting some complexity.

I will try to address your other comments this week!

horovod/common/basics.py
```diff
@@ -30,39 +32,79 @@ def __init__(self, pkg_path, *args):
...
         """
         if comm is None:
             comm = []
+        elif not isinstance(comm, list):
+            comm = [comm]
+        if process_sets is None:
+            process_sets = []
+        elif isinstance(process_sets, str) and process_sets.lower() == "dynamic":
+            process_sets = []
+            os.environ["HOROVOD_DYNAMIC_PROCESS_SETS"] = "1"
```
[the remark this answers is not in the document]
maxhgerlach: There is no deep reason to have the env var, but it makes things easier in the unit test script: https://github.com/maxhgerlach/horovod/blob/995fc21d5af269e8bea83fbd2cc415408b9854a2/test/parallel/test_tensorflow.py#L59

Since only the first test to be executed can effectively pass parameters to the background thread via `hvd.init()`, every single test function would need to be adapted to make sure Horovod is initialized properly for the process set tests. That felt a bit invasive to me. Alternatively, all process set tests would need to be moved into a separate file (making sure to share all framework-specific setup logic) and a similar split would be required for the PyTorch and MXNet tests in due time.

horovod/tensorflow/__init__.py
maxhgerlach: Definitely. How should the API look like?

With TF1-style legacy optimizers one currently makes calls like:
```
avg_grad_var_pairs = opt.compute_gradients([(grad0, var0), (grad1, var1), ...])
```
Should we extend this to support something like the following if we want to have grad1 averaged over a non-global process set?
```
avg_grad_var_tuples = opt.compute_gradients([(grad0, var0), (grad1, var1, process_set_1), ...]))
```
Maybe it's not a good idea to clandestinely change the signature of the base class method.

Alternatively we could support handing in a callback `get_process_set_for_variable(var)` or even a dictionary `{var: process_set, ...}` listing all variables that should live in non-global process sets.

horovod/common/basics.py
```diff
@@ -288,3 +330,112 @@ def rocm_built(self):
...
+            if ps_size < 0:
+                raise RuntimeError("Process set table was modified outside of get_process_sets()")
+            ranks_array = (ctypes.c_int * ps_size)()
+            res = int(self.MPI_LIB_CTYPES.horovod_get_process_set_ranks(ctypes.c_int(ps_id), ranks_array))
+            if res < 0:
+                raise RuntimeError("Process set table was modified outside of get_process_sets()")
+            ret[ps_id] = list(ranks_array)
+        return ret
+
+    def comm_process_set(self, comm) -> int:
```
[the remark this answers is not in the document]
maxhgerlach: I like it! Sometimes numeric ids are convenient, e.g., if the process set id is encoded in a variable name string. But having a proper `ProcessSet` type on the Python side as the main API definitely feels safer to use.

horovod/common/response_cache.cc
maxhgerlach: done

horovod/common/basics.py
```diff
@@ -30,39 +32,79 @@ def __init__(self, pkg_path, *args):
...
             comm = []
+        elif not isinstance(comm, list):
+            comm = [comm]
+        if process_sets is None:
+            process_sets = []
+        elif isinstance(process_sets, str) and process_sets.lower() == "dynamic":
+            process_sets = []
+            os.environ["HOROVOD_DYNAMIC_PROCESS_SETS"] = "1"
+
+        process_set_sizes = [len(ps) for ps in process_sets]
```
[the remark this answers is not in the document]
maxhgerlach: Good point, thanks. 

I added two commits to ensure that each set of processes can only be added as a process set a single time, regardless of permutations. Also error handling in case there are mismatches in process set arguments to `hvd.init()` between calling processes is more thorough and friendlier now.

horovod/common/operations.cc
maxhgerlach: Yes, I think you are right that interleaving those two loops has the potential to reduce latency. Also I believe there is nothing left in `ComputeResponseList` (except for process set 0) that would make use of the global `Controller`, so there is reason to believe that this should not cause any dead locks. I've made the change in my last commit for today.

> Caveat: There is the WAIT_FOR_DATA spin loop that will hold back the background thread on every PerformOperation call

I see. Would that matter in typical situations where there is one global process set (e.g., `[0,1,2,3]`) and the other process sets are non-overlapping (e.g., `[0,1]` and `[2,3]`)? If the background thread is blocked in processes 0 and 1 because of a `PerformOperation` call for process set `[0,1]`, the processes in `[2,3]` should not be affected and can call `ComputeResponseList` right away.

> however, I am planning on posting a PR soon to modify that to use a more asynchronous cudaStreamWaitEvent to manage the data dependency. Seems like this feature could benefit from that.

That sounds nice!

horovod/tensorflow/__init__.py
romerojosh: I was thinking of something even more simple than what you are proposing here: just adding a process set argument to the `DistributedOptimizer`/`DistributedGradientTape` constructors to limit the action of those wrappers to a provided process set. Do you think there are common scenarios where a user would want/need to have per gradient control over the process set within the same optimizer?

horovod/common/operations.cc
romerojosh: > If the background thread is blocked in processes 0 and 1 because of a PerformOperation call for process set [0,1], the processes in [2,3] should not be affected and can call ComputeResponseList right away.

Yes, that is right. The background thread for `[2,3]` should not be impacted by a block on `[0,1]` if those are the only sets in use (non-overlapping). I was thinking more about the cases where there is overlap in the process sets, which I think will also be a common use case, since at some point you probably need to aggregate results/do some computation across your disjoint process sets using the global process set or another overlapping process set. 

Thanks for making the change to interleave the operations.

horovod/tensorflow/__init__.py
maxhgerlach: Ah, excellent! Most people will indeed be able to build separate optimizers for their special variables that live in specific process sets. It's a very good idea to limit the extra complexity in `DistributedOptimizer`. Otherwise users can always wrap / subclass / hack it themselves.

I am going to add such an argument to the constructors once we have the nicer Python API with `ProcessSet` objects.

maxhgerlach:
rebased to master to work around a torchhead test failure and squashed my last four fixup commits

horovod/common/basics.py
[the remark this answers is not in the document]
maxhgerlach: OK, I've changed the Python API to use lightweight `hvd.ProcessSet` objects as you proposed. The interface should feel more streamlined now.

horovod/tensorflow/__init__.py
maxhgerlach: done, and fixed some details on the way

test/parallel/test_process_sets_static.py
```diff
@@ -0,0 +1,53 @@
...
+            self.skipTest("Only one worker available")
+
+        try:
+            import mpi4py
+            mpi4py.rc.initialize = False
+        except ImportError:
+            self.skipTest("This test requires mpi4py.")
+
+        if mpi_rank == 0:
+            hvd.init(process_sets=[hvd.ProcessSet([0]),
```
romerojosh, line 32: I think this test is a good place to discuss where I think the ProcessSet API can be improved a bit. 
I think we should treat the `ProcessSet` objects as somewhat analogous to a distributed group in the PyTorch distributed API (https://pytorch.org/docs/stable/distributed.html#torch.distributed.new_group) where the specific `ProcessSet` object that is either passed into `hvd.init` or returned from `hvd.add_process_set` acts as the handle for later Horovod operations on the process group. 
As an example, I think a call like this should be designed to be incorrect usage:
```
hvd.init(process_sets=[hvd.ProcessSet([2,3])])
```
as a temporary `ProcessSet` object is passed into init and after the call, the user no longer can access the handle (or should not be able to).

I think correct usage should look something like this:
```
ps = hvd.ProcessSet([2,3]) # Created but not initialized. Usage in any collective op will raise an error
hvd.init(process_sets=[ps]) # Initializes Horovod and sets up passed in ProcessSet ps
hvd.allreduce(tensor, process_set=ps)
```
This also then removes the need to have an API to extract existing process sets from Horovod (along with process set IDs). Essentially, it would be the responsibility of the user to maintain ownership of the ProcessSet objects if they intend to use them.

What do you think?

romerojosh (commented):
Hi @maxhgerlach. Thanks for working on changes from the first review!
I like the `ProcessSet` changes you've made so far and I think it really cleans up the API. As noted in my comments, I think we can encapsulate the details even more. In particular, I'd really like to see process set id integers be made a completely internal detail.

Otherwise, I think everything is coming together quite nicely.

horovod/common/basics.py
```diff
@@ -288,3 +353,127 @@ def rocm_built(self):
...
+        elif result == self.HOROVOD_PROCESS_SET_ERROR_DYNAMIC:
+            raise ValueError(
+                "Set HOROVOD_DYNAMIC_PROCESS_SETS=1 to allow removing process sets after Horovod initialization.")
+        elif result == self.HOROVOD_PROCESS_SET_ERROR_UNKNOWN_SET:
+            return None
+        elif result == self.HOROVOD_PROCESS_SET_ERROR_GLOO:
+            raise ValueError("Multiple process sets are only supported with MPI controllers, not Gloo.")
+        return result
+
+    def process_set_rank(self, process_set_id):
```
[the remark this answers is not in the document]
maxhgerlach: Those are useful or needed for "Horovod internal" usage, e.g., for implementing the gradient of a process-set specific allgather. Most are not exported to the user-facing module `horovod.tensorflow`.

Probably I should add leading underscores to the names of thse to make this clearer and align better with Horovod naming conventions.

For now I've kept all functions that directly access the ctypes interface to the C++ library in basics.py and out of process_sets.py.

test/parallel/test_process_sets_static.py
maxhgerlach: Hey @romerojosh,

thanks for the second review and making sure we will end up with a really well rounded interface! :)

I kind of liked having both options, i.e., letting users either use the `ProcessSet` objects or those IDs (mainly because in the training code we run on this branch we convert tensor or variable names with encoded IDs into process set arguments all the time). _However_, having given this some more thought, I think you are right that this exposed implementation detail will probably feel more confusing than useful to users. They can just have their own `ProcessSet` object storages that will generally be better adapted to their actual needs.

I will give this a second pass of clean ups. Maybe we can completely get rid of the Horovod-internal dict `_id_to_process_sets` as well.

horovod/common/process_sets.py
```diff
@@ -0,0 +1,171 @@
...
+        return _basics.rank() in self.ranks
+
+    def __str__(self) -> str:
+        return f"ProcessSet(process_set_id={self.process_set_id}, ranks={self.ranks}, mpi_comm={self.mpi_comm})"
+
+
+global_process_set = ProcessSet([])
+global_process_set.process_set_id = 0
+
+_id_to_process_sets: Dict[int, List[ProcessSet]] = {0: [global_process_set]}
```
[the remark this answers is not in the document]
maxhgerlach: My internal resasoning was basically that if a user managed to create lots of equivalent `ProcessSet` instances, then removing one of them should either invalidate all of them (current implementation) or pop that handle of the list until it's empty and the internal process set should actually be removed. The simpler third alternative is, of course, to disallow the creation of lots of equivalent process set objects.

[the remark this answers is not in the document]
romerojosh: Should multiple process sets containing the same set of processes be considered "equivalent" or should they be distinct? For example, if I were to run something like:
```
ps1 = hvd.add_process_set([0,1,2,3])
ps2 = hvd.add_process_set([0,1,2,3])
hvd.remove_process_set(ps1)
```
should `ps2` still work? I think it should, but it sounds like in the current implementation that isn't the case. 

Is there an issue with allowing a user to create multiple distinct process sets that contain the same set of processes (similar to create a bunch of MPI subcommunicators containing the same ranks)? I understand from a performance standpoint, you may want to back the "equivalent" process sets with the same communicator in the backend, but depending on the complexity of this code to allow for this kind of aliasing, it might be easier to just create multiple backend process sets, even if they contain the same sets of ranks. You can always print a warning if this occurs to tell users that performance may be degraded if they do this.

horovod/common/basics.py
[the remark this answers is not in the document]
maxhgerlach: > Probably I should add leading underscores to the names of these to make this distinction clearer and to align better with Horovod naming conventions.

done that

horovod/common/process_sets.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def setup(basics):
+    # type: (Optional[HorovodBasics]) -> None
+    """" Horovod internal, to be called after the Horovod C++ module has been loaded. """
+    global _basics
+    _basics = basics
+
+
+def update_process_sets(process_sets: List[ProcessSet]):
```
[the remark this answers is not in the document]
maxhgerlach: done

horovod/common/process_sets.py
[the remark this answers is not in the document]
maxhgerlach: > disallow the creation of lots of equivalent process set objects

I went with this and removed the global dict.

horovod/common/process_sets.py
```diff
@@ -0,0 +1,171 @@
...
+
+    if process_set.process_set_id in _id_to_process_sets:
+        for some_process_set in _id_to_process_sets.pop(process_set_id):
+            some_process_set.invalidate()
+
+    returned_id = _basics.remove_process_set_impl(process_set_id)
+    return returned_id is not None
+
+
+def process_set_by_id(process_set_id: int) -> ProcessSet:
```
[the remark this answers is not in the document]
maxhgerlach: done

test/parallel/test_process_sets_static.py
maxhgerlach: OK, things should be a bit simpler now. 

The biggest caveat I can think of is that duplicate process sets aren't handled transparently anymore. So if one tries to have `[ProcessSet(range(0, size, 2)), ProcessSet([0])]` this would now raise an exception when running with two or three processes, but be fine with four or more processes. But people should take a moment to think about their process set setups anyway.

(Allowing these duplicates without Horovod-internal object tracking is dangerous in combination with `remove_process_set()`)

horovod/common/process_sets.py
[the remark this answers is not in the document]
maxhgerlach: Sorry, I missed your message in this thread.

Do you think it's really important to allow for that flexibility? I wouldn't want users to accidentally add the same backend process set again and again in their training loops. 

(As of my latest commit the second add_process_set call would now raise an exception.)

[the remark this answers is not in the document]
romerojosh: Nope, not important to allow for that, and disallowing it works (and is simpler). Important point was just to make sure individual process set objects (even if containing the same ranks), do not interact with each other. Since you now cannot have multiple process sets containing the same ranks, this skirts past the issue altogether. 👍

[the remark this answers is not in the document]
maxhgerlach: Cool, I'm growing to appreciate simplicity more and more. I guess just having extra ids for duplicate process sets in the backend would be the best way to go after all if we ever decide to go that way. Trying to track an additional object-id correspondence in the frontend comes with nasty  surprises as you pointed out.

horovod/common/operations.cc
```diff
@@ -603,86 +686,125 @@ bool RunLoopOnce(HorovodGlobalState& state) {
...
-  std::vector<std::string> tensor_names;
-  if (state.parameter_manager.IsAutoTuning()) {
-    total_tensor_size = horovod_global.tensor_queue.GetTensorDataForAutotuner(
-        response_list, tensor_names);
+#if HAVE_MPI
+  if (state.dynamic_process_sets) {
+    // Initialize any newly added process set that has been registered by all
+    // Horovod processes.
+    if (global_mpi_context.IsEnabled()) {
+      state.process_set_table.InitializeRegisteredIfReady(global_mpi_context);
```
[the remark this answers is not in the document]
maxhgerlach: I merged the two functions to use just one allgather together.

maxhgerlach:
@romerojosh, I went ahead and added support for Gloo instead of MPI. Unfortunately on [Buildkite](https://buildkite.com/horovod/horovod/builds/6059#075b1e6d-e5f4-4a60-ba02-4ca18f07acd4) there is one occasional failure in `test_horovod_reinit` from `parallel/test_torch.py`, which I haven't been able to reproduce locally so far:

```
#...
[0]<stdout>:test_torch.py::TorchTests::test_horovod_rank PASSED
[1]<stdout>:test_torch.py::TorchTests::test_horovod_rank PASSED
[0]<stderr>:terminate called after throwing an instance of 'gloo::IoException'
[0]<stderr>:  what():  [/pytorch/third_party/gloo/gloo/transport/tcp/pair.cc:898] Connect timeout [127.0.0.1]:51096
[0]<stderr>:Fatal Python error: Aborted
[0]<stderr>:
[0]<stderr>:Thread 0x00007f95857fb700 (most recent call first):
[0]<stderr>:<no Python frame>
[0]<stderr>:
[0]<stderr>:Thread 0x00007f95861fc700 (most recent call first):
[0]<stderr>:<no Python frame>
[0]<stderr>:
[0]<stderr>:Thread 0x00007f9586bfd700 (most recent call first):
[0]<stderr>:<no Python frame>
[0]<stderr>:
[0]<stderr>:Thread 0x00007f95875fe700 (most recent call first):
[0]<stderr>:<no Python frame>
[0]<stderr>:
[0]<stderr>:Thread 0x00007f9699005740 (most recent call first):
[0]<stderr>:  File "/usr/local/lib/python3.8/dist-packages/horovod/common/basics.py", line 102 in init
[0]<stderr>:  File "/horovod/test/parallel/test_torch.py", line 102 in test_horovod_reinit
#...
```

Do you see anything suspicious in my changes?

horovod/runner/http/http_server.py
```diff
@@ -122,13 +122,29 @@ def do_DELETE(self):
...
-            if scope in self.server.scope_size:
-                # scope is not an "ad-hoc" scope added by a client
+        if scope in self.server.scope_size:
+            # scope defined by host_alloc_plan at server initialization
+            with self.server.finished_list_lock:
+                self.server.finished_list[scope].append(key)
                 if self.server.scope_size[scope] == len(self.server.finished_list[scope]):
                     with self.server.cache_lock:
                         self.server.cache.get(scope, {}).clear()
+                        self.server.finished_list.clear()
```
maxhgerlach, line 132: `finished_list` was never cleared before this change. 

The list is supposed to track all clients that requested their key to be deleted (which happens at the end of the Rendezvous process). Once the list is full, the cache for this scope is cleared. This worked fine for the first Rendezvous process, but if Horovod was shut down and reinitialized, the list would grow too long and the cache would never be cleared anymore.

I am not sure if this detail might be related to the occasional Gloo timeouts I've seen in `test_horovod_reinit()`. 

Also I haven't understood yet why a separate finished_list is tracked here and key-value pairs aren't simply popped off the cache right away (it seems that there may be situations where a GET after a supposedly successful DELETE for the same key would still return a value).

maxhgerlach: > Also I haven't understood yet why a separate finished_list is tracked here and key-value pairs aren't simply popped off the cache right away (it seems that there may be situations where a GET after a supposedly successful DELETE for the same key would still return a value).

OK, I've figured out this bit. `Rendezvous()` in gloo_context.cc calls `Finalize()` on the HTTPStore potentially before all clients have connected. So a GET after a DELETE is actually supposed to return a value, else there will be problems:
```
[1]<stdout>:[2021-06-22 16:19:12.630567: T /mnt/data/max/horovod/horovod/common/gloo/http_store.cc:132] Send DELETE request to http://127.0.0.1:43444/local_localhost_process_set_hash_1e9d548172549547/1
[0]<stdout>:[2021-06-22 16:19:12.630843: T /mnt/data/max/horovod/horovod/common/gloo/http_store.cc:103] Send GET request to http://127.0.0.1:43444/local_localhost_process_set_hash_1e9d548172549547/1
Scope local_localhost_process_set_hash_1e9d548172549547: Deleted key 1
scope local_localhost_process_set_hash_1e9d548172549547: get 1 -> None
[0]<stderr>:terminate called after throwing an instance of 'gloo::EnforceNotMet'
```
I need to be more careful to defer the actual deletion.

maxhgerlach: Rather than introduce more complexity to the `RendezvousServer` in Python, I just put a Gloo barrier into `Rendezvous()`, which generally feels like a sane thing to have there to me. 

This seems to have cleared up the weird issues with the tests!

romerojosh:
Hi @maxhgerlach, sorry for the delay in getting in another review pass of your current updates. Will aim to get that done in the next couple of days.

@tgaddair and I were discussing that it might be a good plan forward with this PR to aim to land this initial PR with TF support only with subsequent PRs to add PyTorch and MXNet (and any other missing features), just to split up the review load. What do you think about that? It seems this PR might already be in a state to land (pending review) in this case.

maxhgerlach:
Hi @romerojosh, 

no worries, these are quite a lot of changes. I am looking forward to any comments once you're ready!

Merging this soonish in a non-framework-complete state sounds excellent to me. This should not only make the reviews easier to handle for you, but it would also reduce the effort on my side to keep this PR up to date with any major changes in master.

I think I should be able to provide follow-up PRs for PyTorch and MXNet not much later then.

But I'd be much obliged if anybody with more experience with CCL could look into supporting that API. Similarly, support in the Adasum op might be something to leave for later.

maxhgerlach:
Rebased to master and added three commits since my last comment
- Avoid race condition at shutdown 7e65d5f
- Add public method ProcessSet.rank(), also use that internally 8b87628
- Add proper error messages for oneCCL, update changelog a13928f

romerojosh (approved):
Hi @maxhgerlach, excellent work (and thanks for handling the shutdown race condition I brought up on Slack). 

I left a minor comment about error messages. Since process set IDs aren't supposed to be a public detail, might be nice to change the wording, but not a big deal. Otherwise, I think we are good to land this one!

@tgaddair Before merging, you might to have someone look over any gloo changes, since I am less familiar with that part of the code.

maxhgerlach:
Thanks for this concluding review @romerojosh!

I updated those error messages so they don't talk about IDs anymore. In the same go I corrected those checks for CCL that I had just introduced...

Regarding the changes related to Gloo I wasn't 100% certain that I went in the right direction, so those could indeed warrant an extra look.

maxhgerlach:
Oof, there is another potential race condition at shutdown in the function `HorovodBasics._get_process_set_ids_and_ranks()`. This leads to a spurious failure in `test_process_sets_static.py`. By now that function is really only meant for internal use, but out of convenience I've kept it in the tests.

I believe that this one is ultimately caused because that Python function does not consistently lock access to the Horovod-internal `ProcessSetTable`. One way to go about this would be to add a C interface to the mutex, which may not be wise. Another way would be to move the whole function into the C++ backend. That could be a bit tricky, because I would need to figure out how to allocate memory for a Python data structure from C++.

For the sake of simplicity and because this function is not meant for general usage, I'd rather sidestep the issue for now and tack a synchronization barrier before shutting down in those test scripts (https://github.com/horovod/horovod/pull/2839/commits/3526fad8efde7eb75c337d2db0cd53ed5e56591c).

EnricoMi:
@maxhgerlach #3020 will fix the CI workflow (we have moved from Buildkite to GitHub in #2798) and you can rerun the jobs: https://github.com/horovod/horovod/actions/runs/1008824316

EnricoMi:
@maxhgerlach maybe you need to rebase with master to pick up the fix.

maxhgerlach:
@EnricoMi, thanks for the pointer. I rebased to master once more. (but I didn't see any relevant errors in the last buildkite results either)

maxhgerlach:
@EnricoMi, I think no tests on GPU have been run after my last rebase & push: https://github.com/maxhgerlach/horovod/runs/3013380827?check_suite_focus=true
> You must set the BUILDKITE_API_ACCESS_TOKEN environment variable (e.g. BUILDKITE_API_ACCESS_TOKEN = "xyz")

However, as far as the changes in this PR are concerned, the last Buildkite-only test run didn't unveil any failures as far as I can tell.

EnricoMi:
That job shouldn't be running for PRs from forks, because they do not have that secret (as it could be easily stolen). PR #3025 should fix that.

For forks we have the "CI (Fork)" workflow (e.g. https://github.com/horovod/horovod/actions/runs/1009885052) which kicks in once the CI workflow finishes. It then triggers the GPU Buildkite tests and pulls the results.

We have just moved to this Github CI workflow and supporting forks is tricky and can only be tested from master and with fork PRs. Please bear with me.

maxhgerlach:
Thanks for all the work; I just wanted to notify you in case that was unexpected. :)

horovod/runner/http/http_server.py
maxhgerlach: @tgaddair, this synchronization barrier that I stuck into `Rendezvous()` and clearing the `finished_list` in the rendezvous server are the biggest changes to the general Gloo code coming with this PR.

tgaddair (approved):
LGTM! This PR is amazing.

PR #2859, opened 2021-04-22
Skip TensorFlowTests::test_horovod_allreduce_multi_gpu if there are less than 2 GPUs per worker

maxhgerlach:

## Description

The test requires at least four GPUs if it is run with two workers. Sometimes one of the Buildkite CI jobs seem to be run with only three active GPUs although four GPUs were requested. This causes quite confusing failure messages "ncclCommInitRank failed: invalid usage" if this test is not skipped. Example: https://github.com/horovod/horovod/pull/2839/checks?check_run_id=2402052960


romerojosh (approved):
Thanks @maxhgerlach! Yes, this 4 GPU requirement issue has come up before, so thanks for submitting the fix.

maxhgerlach:
Thanks @romerojosh!

PR #2863, opened 2021-04-23
Pin Ray to 1.2.0 in tests

tgaddair:


Dockerfile.test.cpu
```diff
@@ -75,6 +75,9 @@ RUN if [[ ${SPARK_PACKAGE} != *"-preview"* ]]; then \
         (cd /spark/python && python setup.py sdist && pip install --no-cache-dir dist/pyspark-*.tar.gz && rm dist/pyspark-*); \
     fi
 
+# Install Ray.
+RUN pip install --no-cache-dir ray==1.2.0
```
chongxiaoc, line 79: Is it possible to use `1.3.0`

Some feedback from @yuduber  about bug in `1.2.0` 
```
For now, we will pin on a ray version of 1.3.0, because ray 1.2.0 has port conflict issue and can not be used in prod.
```

chongxiaoc (commented):
Leave a question.

Dockerfile.test.cpu
tgaddair: This pin is just for the tests (because of the flaky colocator test), so it shouldn't affect users who wish to use a different version themselves.

chongxiaoc (approved):

PR #2872, opened 2021-04-27
add allgather cache

Agoniii:
Signed-off-by: Agoniii <815244047@qq.com>

## Description

Add response_cache for Allgather

## Review process to land 

### DIFF
--- a/horovod/common/controller.cc
+++ b/horovod/common/controller.cc
@@ -434,6 +434,7 @@ ResponseList Controller::ComputeResponseList(std::atomic_bool& shut_down,
     // order consistently across workers.
     for (auto& response : response_list.responses()) {
       if ((response.response_type() == Response::ResponseType::ALLREDUCE ||
+           response.response_type() == Response::ResponseType::ALLGATHER ||
            response.response_type() == Response::ResponseType::ADASUM ||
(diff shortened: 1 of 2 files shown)


romerojosh:
Hi @Agoniii, thanks for extending the response cache to cover allgather collectives!

The unit test failure of `test_broadcast_object` does uncover an existing bug in the response cache implementation. The issue is that the checks to see if messages or responses are cached (i.e. the two `ResponseCache::cached` methods) do not check that the `RequestType` and `ResponseType` are the same. Hence, if two Horovod operations are scheduled with the same name, but do different collective operations and the first operation is cached, the execution of the second operation is incorrect. 

In the unit test failure, `test_allgather_object` performs an allgather using the name `dict.sz` and caches it. Then the following `test_broadcast_object` performs a broadcast using the same name `dict.sz`; however, it sees the existing entry in the cache from the allgather, and proceeds to perform the wrong operation. I will push a quick PR to fix this issue to unblock this one.

romerojosh:
Posted #2873 that should fix the issue and resolve these test failures. Will try to get that approved and merged soon to unblock you.

Agoniii:
@romerojosh Thanks. I cherry-picked the #2873 to test whether it can solve these failures.

romerojosh (approved):
LGTM!

Issue #2496, opened 2020-12-02
Dynamic timeline API tests fails on macOS


tgaddair:
```
[0]<stdout>:parallel/test_torch.py:2279: 
[0]<stdout>:_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[0]<stdout>:
[0]<stdout>:fname = '/var/folders/z3/_825pg0s3jvf0hb_q8kzmg5h0000gn/T/tmpya35o_iw'
[0]<stdout>:check_cycle = True
[0]<stdout>:
[0]<stdout>:    def check_file(fname, check_cycle=True):
[0]<stdout>:        if hvd.rank() == 0:
[0]<stdout>:            with open(fname, 'r') as timeline_file:
[0]<stdout>:                timeline_text = timeline_file.read()
[0]<stdout>:                assert 'allreduce.test_allreduce' in timeline_text, timeline_text
[0]<stdout>:                assert 'start_time_since_epoch_in_micros' in timeline_text, timeline_text
[0]<stdout>:>               assert 'NEGOTIATE_ALLREDUCE' in timeline_text, timeline_text
[0]<stdout>:E               AssertionError: [
[0]<stdout>:E                 {"name": "process_name", "ph": "M", "pid": 0, "args": {"start_time_since_epoch_in_micros":1606864241652573}},
```

Removing the NEGOTIATE_ALLREDUCE check:

```
[1]<stdout>:parallel/test_torch.py::TorchTests::test_horovod_join_broadcast PASSED   [ 86%]
[1]<stdout>:parallel/test_torch.py::TorchTests::test_horovod_rank PASSED             [ 87%]
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

PR #2874, opened 2021-04-28
Require setting HOROVOD_TIMELINE=DYNAMIC to enable dynamic timeline API usage

### LINKED ISSUE #2804: Unexpected performance decrease after hvd.init on CPU nodes
**Environment:**
1. Framework: TensorFlow
2. Framework version: 2.3.1
3. Horovod version: 0.21.0
4. MPI version: OpenMPI-4.0.3
5. CUDA version: 11.0.2
6. NCCL version: 2.7.8
7. Python version: 3.8.2
8. Spark / PySpark version: N/A
9. Ray version: N/A
10. OS and version: Red Hat Enterprise Linux Server release 7.9 (Maipo) 
11. GCC version: 9.3.0
12. CMake version: 3.16.4

**Bug report:**

I see an unexpected performance decrease when running Horovod (using a single MPI rank) compared to just native TensorFlow. Essentially, I have a slightly altered version of the `tensorflow2_synthetic_benchmark.py` (see [here](https://github.com/casparvl/software-layer/blob/tensorflow/tests/reframe/eessi-checks/applications/src/tensorflow2_synthetic_benchmark.py)) to which I've added an extra argument `--use-horovod`. With that argument, it runs the exact same example, but with native tensorflow (i.e. it skips the hvd.init(), skips the wrapping in the Horovod distributed gradient tape, etc).

Now, I performed 4 different runs all using a single rank. Run 1 & 2 were performed on a GPU node, run 3 & 4 on a CPU node.

1. `srun -n 1 python tensorflow2_synthetic_benchmark.py --model ResNet50 --batch-size 32 --num-iters 5 --num-batches-per-iter 5 --num-warmup-batches 5`
2. `srun -n 1 python tensorflow2_synthetic_benchmark.py --model ResNet50 --batch-size 32 --num-iters 5 --num-batches-per-iter 5 --num-warmup-batches 5 --use-horovod`
3. `srun -n 1 python tensorflow2_synthetic_benchmark.py --model ResNet50 --batch-size 32 --num-iters 5 --num-batches-per-iter 5 --num-warmup-batches 5 --no-cuda`
4. `srun -n 1 python tensorflow2_synthetic_benchmark.py --model ResNet50 --batch-size 32 --num-iters 5 --num-batches-per-iter 5 --num-warmup-batches 5 --no-cuda --use-horovod`

And the the respective performance numbers are:

1.  51.7 img/s
2. 50.0 img/s
3. 28.7 img/s
4. 6.9 img/s

Clearly, I'm wondering about that last result. 

I've checked a horovod timeline, there's virtually nothing there (I see the broadcast operations, but as expected a broadcast with one worker only takes a few microseconds).

Finally, I nailed it down to the hvd.init(): if I repeat run 3, but only add the lines
```
hvd.init()
```
That's when my performance drops (note: only importing the horovod.tensorflow module doesn't change the performance, it's really the `hvd.init()` that causes the performance drop).

Would you have any idea what is causing the performance difference between run 3 and 4, as defined above? I saw that `horovodrun` has an option to disable mpi-thread support and mentions this sometimes 'can slow down other components'. Is there an equivalent way of disabling mpi-thread support when launching with other mpi wrappers such as `mpirun` or `srun`? (I thought I remember an environment variable that could be set, but can't seem to find it... )

romerojosh:
## Description
Fixes #2804. 

PR #2215 changed the Horovod initialization code to initialize the timeline unconditionally, rather than only in cases where the user sets `HOROVOD_TIMELINE`. This is a problem because the timeline spawns an additional writer thread that can consume CPU resources. Additionally, there are still places in the code that check whether the timeline is initialized in order to skip unnecessary operations, for example this code block here:
https://github.com/horovod/horovod/blob/2436af9d1c0681090919e314e0cfbd9bc13d7a53/horovod/common/ops/nccl_operations.cc#L165-L167
Since that PR, these branches/operations are executed every time, whether or not a user is using the timeline functionality. 

Another place this occurs is in the response cache coordination logic, where additional bits to communicate information only relevant to the timeline are communicated all the time now, due to `timeline_enabled` being set to true (unless `HOROVOD_TIMELINE=DISABLE`): https://github.com/horovod/horovod/blob/2436af9d1c0681090919e314e0cfbd9bc13d7a53/horovod/common/response_cache.cc#L356-L360
  
This PR fixes this by reverting to the original behavior where the timeline is initialized during Horovod initialization only if `HOROVOD_TIMELINE` is set. For the dynamic timeline API, users can set `HOROVOD_TIMELINE=DYNAMIC` to initialize the timeline for that API to work. 

While the existing implementation has an option to set `HOROVOD_TIMELINE=DISABLE`, this option does not keep the timeline from initializing. Also, given the performance implications and that more users are more often not using the timeline or dynamic timeline API, I think it is better for this feature to be an opt-in, rather than opt-out as it currently is.

I also went ahead and reenabled the `test_timeline_api` test to make sure this change doesn't break the dynamic timeline functionality. It was fully disabled in #2495 due to #2496, but I've limited the skip only to systems running macOS.

### DIFF
--- a/horovod/common/basics.py
+++ b/horovod/common/basics.py
@@ -96,7 +96,7 @@ class HorovodBasics(object):
         Raises a `ValueError` if Horovod is not initialized.
         """
         result = self.MPI_LIB_CTYPES.horovod_stop_timeline()
-        if not result:
+        if result == -1:
             raise ValueError('Horovod has not been initialized; use hvd.init().')
 
     def size(self):
--- a/horovod/common/basics.py
+++ b/horovod/common/basics.py
@@ -85,8 +85,10 @@ class HorovodBasics(object):
         result = self.MPI_LIB_CTYPES.horovod_start_timeline(
             ctypes.c_char_p(file_path.encode('utf-8')),
             ctypes.c_bool(mark_cycles))
-        if not result:
+        if result == -1:
             raise ValueError('Horovod has not been initialized; use hvd.init().')
+        elif result == -2:
+            raise ValueError('Set HOROVOD_TIMELINE=DYNAMIC to enable dynamic timeline usage.');
 
     def stop_timeline(self):
         """Stops the active timeline recording and closes the file.
--- a/horovod/common/controller.cc
+++ b/horovod/common/controller.cc
@@ -985,7 +985,7 @@ void Controller::SynchronizeTimelineEnabled() {
   timeline_enabled_ = timeline_enabled_pending_;
 }
 
-bool Controller::TimeLineEnabled() {
+bool Controller::TimelineEnabled() {
   std::lock_guard<std::recursive_mutex> guard(timeline_mutex_);
   return timeline_enabled_;
 }
--- a/horovod/common/controller.h
+++ b/horovod/common/controller.h
@@ -127,7 +127,7 @@ public:
   bool IsCoordinator() const { return is_coordinator_; };
   bool IsHomogeneous() const { return is_homogeneous_; };
   void SetTimelineEnabled(bool value);
-  bool TimeLineEnabled();
+  bool TimelineEnabled();
   void SetTimelineEnabledPending(bool value);
   bool TimelineEnabledPending();
   void SetMarkCyclesInTimelinePending(bool value);
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -420,19 +420,21 @@ void BackgroundThreadLoop(HorovodGlobalState& state) {
 #endif // HAVE_NVTX
 
   // Open the timeline file on coordinator.
-  auto timeline_env = std::getenv(HOROVOD_TIMELINE);
-  auto horovod_timeline = timeline_env != nullptr ? std::string(timeline_env) : std::string("");
   bool should_enable_timeline = false;
-  if (is_coordinator) {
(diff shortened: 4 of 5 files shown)


Vikas-kum:
Thanks @romerojosh . 
I think this is happening because we are checking initialized instead of active. Since, every emit event is synchronized in timeline.cc, which can be cause of perf impact. So, instead we may need to check timeline.active() instead of timeline.initialized() before emitting events. For example , in below code, it should be timeline.active() which will only be true if timeline is started dynamically or timeline was enabled with environment variable. 


```
horovod/horovod/common/ops/nccl_operations.cc

Lines 165 to 167 in 2436af9
 if (global_state_->timeline.Initialized()) { 
   gpu_context_->RecordEvent(gpu_op_context_.event_queue, NCCL_ALLREDUCE, *gpu_op_context_.stream); 
 } 

```
The problem can be fixed by checking if timeline is active. https://github.com/horovod/horovod/blob/master/horovod/common/timeline.cc#L172 . 

In the new pattern timeline is always initialized but inactive unless start_timeline is called. 
Set HOROVOD_TIMELINE = DYNAMIC approach may cause expose the regression to users who set it.

tgaddair (approved):
LGTM! Awesome catch @romerojosh!

PR #2882, opened 2021-04-30
[Ray] Support client

amogkam:
## Checklist before submitting

- [x] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)?
- [x] Did you update the docs?
- [x] Did you write any tests to validate this change?  
- [x] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users?

## Description

Placement groups do not work with Ray Client in Ray 1.3.

This PR introduces a layer that transparently allows Horovod to work with Ray Client.

We will likely remove this layer in Ray 1.4.

## Review process to land 

1. All tests and other checks must succeed.
2. At least one member of the [technical steering committee](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md) must review and approve.
3. If any member of the technical steering committee requests changes, they must be addressed.

PR #2883, opened 2021-05-01
[ray] use ray node_id as unique id

richardliaw:

## Description

Implements two features:

* use ray node ID instead of hostname as unique identifier. This allows for multi-container -- singlehost setups.
* share cuda visible devices among workers of the same node when using the PackStrategy.

~Wait for #2882 to merge first.~

## Review process to land 

### DIFF
--- a/horovod/ray/strategy.py
+++ b/horovod/ray/strategy.py
@@ -1,10 +1,8 @@
-import socket
 import logging
-import os
 import ray
-from typing import Dict, List
 from horovod.ray.utils import map_blocking
 from horovod.ray.worker import BaseHorovodWorker
+
 logger = logging.getLogger(__name__)
 
 
@@ -115,7 +113,8 @@ class ColocatedStrategy(StrategyInterface):
 
 
 class PackStrategy(StrategyInterface):
-    def __init__(self, *, settings, num_workers, use_gpu, cpus_per_worker, gpus_per_worker):
+    def __init__(self, *, settings, num_workers, use_gpu, cpus_per_worker,
+                 gpus_per_worker):
         self.settings = settings
         self._num_workers = num_workers
         self.cpus_per_worker = cpus_per_worker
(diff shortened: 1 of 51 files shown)


horovod/ray/strategy.py
```diff
@@ -175,4 +176,27 @@ def create_workers(self):
...
+
+            futures = []
+            for node_id, gpu_ids in node_id_to_gpus.items():
+                all_ids = ",".join([str(gpu_id) for gpu_id in gpu_ids])
+
+                for worker in node_workers[node_id]:
+                    futures.append(
+                        worker.update_env_vars.remote({
+                            "CUDA_VISIBLE_DEVICES":
+                            all_ids
```
tgaddair, line 199: As of recent versions of NCCL, we no longer need to make sure that all visible devices on the node are visible to all local ranks, so it might make sense to remove this. Particularly because recent versions of TF and PyTorch will initialize memory on every GPU at startup, which can result in OOM errors if all devices are visible to all local ranks.

test/single/test_ray.py
```diff
@@ -176,6 +183,33 @@ def test_gpu_ids(ray_start_4_cpus_4_gpus):
...
+
+    def _test(worker):
+        import horovod.torch as hvd
+        hvd.init()
+        local_rank = str(hvd.local_rank())
+        return local_rank in os.environ["CUDA_VISIBLE_DEVICES"]
+
+    all_valid_local_rank = hjob.execute(_test)
+    assert all(all_valid_local_rank)
+    hjob.shutdown()
```
tgaddair, line 209: Nit: should probably be wrapped in a `try-finally` block in case an exception is raised or assertion fails.

tgaddair (approved):

horovod/ray/strategy.py
richardliaw: ah yes, as mentioned offline: 

>  this is because Ray will set CUDA_VISIBLE_DEVICES to a single GPU on the node by default, and with Horovod we typically run with all visible GPUs

>  code that runs with all visible devices seems to be often used in examples and other estimator code

PR #2896, opened 2021-05-05
Keras estimator: set reader's epoch = 1 to avoid sample duplication and drop-out in a single epoch

chongxiaoc:

## Description
We set infinite epochs for petastorm reader in horovod, due to the known issue that data partition is not guarantee to be equivalent size. TF's dataset shuffle function will automatically refill the shuffle buffer after retrieving a single element.
Code in horovod:
```
            if shuffle:
                dataset = dataset.shuffle(shuffle_buffer_size)
```
Mixing infinite iterations from reader and TF's shuffle function, will introduce sample duplication and drop-out in a single epoch. For example, training dataset is [`0,1,2,3,4,5,6,7]`, and shuffle size is 8.
Shuffling buffer is initialized as `[0,1,2,3,4,5,6,7]`
It randomly pick 7, and refill with 0 again: `[0,1,2,3,4,5,6,0]`
Now it is highly possible 0 will be selected twice in next few iterations, even before other values are selected once.

This PR fixes above issue by setting reader's epoch to be 1, while looping infinitely inside dataloader.
Also we introduced cache() support from `tf.data.Dataset` since it speeds up throughput.

## Review process to land 

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -112,7 +112,7 @@ mxnet_require_list = ['mxnet>=1.4.1']
 pyspark_require_list = ['pyspark>=2.3.2;python_version<"3.8"',
                         'pyspark>=3.0.0;python_version>="3.8"']
 # Pin h5py: https://github.com/h5py/h5py/issues/1732
-spark_require_list = ['h5py<3', 'numpy', 'petastorm==0.11.0rc6', 'pyarrow>=0.15.0']
+spark_require_list = ['h5py<3', 'numpy', 'petastorm>=0.11.0', 'pyarrow>=0.15.0']
 ray_require_list = ['ray']
 pytorch_spark_require_list = pytorch_require_list + \
                              spark_require_list + \
--- a/setup.py
+++ b/setup.py
@@ -112,7 +112,7 @@ mxnet_require_list = ['mxnet>=1.4.1']
 pyspark_require_list = ['pyspark>=2.3.2;python_version<"3.8"',
                         'pyspark>=3.0.0;python_version>="3.8"']
 # Pin h5py: https://github.com/h5py/h5py/issues/1732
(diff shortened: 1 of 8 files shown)


chongxiaoc:
Rebased and change pytorch inmem caching usage from petastorm `v0.11.0rc6`

tgaddair (approved):
LGTM!

chongxiaoc:
Also fixes #2909

setup.py
```diff
@@ -112,7 +112,7 @@ def build_extensions(self):
 pyspark_require_list = ['pyspark>=2.3.2;python_version<"3.8"',
                         'pyspark>=3.0.0;python_version>="3.8"']
 # Pin h5py: https://github.com/h5py/h5py/issues/1732
-spark_require_list = ['h5py<3', 'numpy', 'petastorm>=0.9.8,<0.11.0', 'pyarrow>=0.15.0']
+spark_require_list = ['h5py<3', 'numpy', 'petastorm>=0.11.0', 'pyarrow>=0.15.0']
```
EnricoMi, line 115: @tgaddair do you think requiring latest petastorm for Horovod on Spark is too much of a constraint? Should we support petastorm <0.11 for a transition period?

If we go forward with this, that "limitation" must be recorded in the CHANGELOG and release notes.

chongxiaoc: The only difference of `petastorm v0.11.0` is refactoring `pytorch inmem dataloader`, which I believe is not used frequently. Yep, I agree to add `CHANGELOG` and release notes.

chongxiaoc: updated CHANGLOG to bring `pytorch inmem dataloder` change.

tgaddair: I think this is fine. Petastorm is pretty lightweight with few overlapping dependencies, so shouldn't be an issue. Unfortunately, it's frequently the case that Horovod needs to be updated to the latest Petastorm to take advantage of new features.

PR #2915, opened 2021-05-18
FP16 support for GPU tensors in mxnet

Adnios:

## Description

The previous pr https://github.com/horovod/horovod/pull/1908 is not finished, I have improved it on the basis of it

### DIFF
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -18,6 +18,7 @@ from horovod.common.util import check_extension, split_list
 check_extension('horovod.mxnet', 'HOROVOD_WITH_MXNET',
                 __file__, 'mpi_lib')
 
+from horovod.mxnet.compression import Compression
 from horovod.mxnet.functions import allgather_object, broadcast_object
 from horovod.mxnet.mpi_ops import allgather
 from horovod.mxnet.mpi_ops import allreduce, allreduce_, grouped_allreduce, grouped_allreduce_
@@ -38,7 +39,8 @@ import warnings
 
 # This is where Horovod's DistributedOptimizer wrapper for MXNet goes
 class DistributedOptimizer(mx.optimizer.Optimizer):
-    def __init__(self, optimizer, gradient_predivide_factor=1.0, num_groups=0):
+    def __init__(self, optimizer, compression=Compression.none, gradient_predivide_factor=1.0, num_groups=0):
+        self._compression = compression
         if gradient_predivide_factor != 1.0 and rocm_built():
             raise ValueError('gradient_predivide_factor not supported yet with ROCm')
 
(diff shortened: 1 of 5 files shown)


tgaddair (approved):
Awesome PR! Looks good to me.

horovod/mxnet/__init__.py
```diff
@@ -64,16 +66,23 @@ def _do_allreduce(self, index, grad):
                 index_split = split_list(index, self._num_groups)
 
                 for i, (grads, indices) in enumerate(zip(grad_split, index_split)):
-                    grouped_allreduce_(tensors=grads, average=False, name="{}:{}".format(indices[0], indices[-1]), priority=-i,
+                    tensors_compressed, ctxs = zip(*[self._compression.compress(grad) for grad in grads])
+                    grouped_allreduce_(tensors=tensors_compressed, average=False,
+                                       name="{}:{}".format(names[0], names[-1]), priority=-i,
                                        prescale_factor=1.0 / self._gradient_predivide_factor)
+                    grads = [self._compression.decompress(t, ctx) for t, ctx in zip(tensors_compressed, ctxs)]
```
romerojosh, line 73: This is not quite right. This leaves the final allreduced and uncompressed results in the local list `grads`, but doesn't actually update the real parameter gradients.

horovod/mxnet/__init__.py
```diff
@@ -64,16 +66,23 @@ def _do_allreduce(self, index, grad):
...
             else:
               for i in range(len(index)):
-                  allreduce_(grad[i], average=False,
+                  tensor_compressed, ctx = self._compression.compress(grad[i])
+                  allreduce_(tensor_compressed, average=False,
                              name=str(index[i]), priority=-i,
                              prescale_factor=1.0 / self._gradient_predivide_factor)
+                  grad[i] = self._compression.decompress(tensor_compressed, ctx)
```
romerojosh, line 80: This line errors out as `grad` is a tuple so you can't assign the resulting tensor to `grad[i]`. `grad[i][:]` might work but the copy is not ideal.

horovod/mxnet/__init__.py
```diff
@@ -167,17 +182,22 @@ def _allreduce_grads(self):
 
                 for entries in entries_by_dtype.values():
                     grads, names = zip(*entries)
-                    grouped_allreduce_(tensors=grads, average=False, name="{}:{}".format(names[0], names[-1]), priority=-i,
+                    tensors_compressed, ctxs = zip(*[self._compression.compress(grad) for grad in grads])
+                    grouped_allreduce_(tensors=tensors_compressed, average=False, name="{}:{}".format(names[0], names[-1]), priority=-i,
                                        prescale_factor=1.0 / self._gradient_predivide_factor)
+                    grads = [self._compression.decompress(t, ctx) for t, ctx in zip(tensors_compressed, ctxs)]
```
romerojosh, line 188: Same issue here. Pretty sure this is just updating a local list but not actually modifying the real parameter gradients.

horovod/mxnet/__init__.py
```diff
@@ -167,17 +182,22 @@ def _allreduce_grads(self):
...
             # Meanwhile, since horovod requires Python 3.6, there is no need to sort
             # self._params as enumerating a python dict is always deterministic.
             for i, param in enumerate(self._params):
                 if param.grad_req != 'null':
-                    allreduce_(param.list_grad()[0], average=False,
+                    tensor_compressed, ctx = self._compression.compress(param.list_grad()[0])
+                    allreduce_(tensor_compressed, average=False,
                                name=self._prefix + str(i), priority=-i,
                                prescale_factor=1.0 / self._gradient_predivide_factor)
+                    param._grad[0][:] = self._compression.decompress(tensor_compressed, ctx)
```
romerojosh, line 199: This is a place where it looks like you are actually modifying the parameter gradients. However, this is using a copy which isn't ideal, especially since this code is called even if someone isn't using compression.

romerojosh (commented):
I think that this merged PR has some issues that need to be addressed as noted in my comments. 

@tgaddair, we need to revert this one and remerged when fixed.

PR #2921, opened 2021-05-20
Added workflow to publish Docker images to DockerHub automatically

tgaddair:


tgaddair:
Builds are failing because GitHub runners only have 14GB of disk space. We may need to make use of [this](https://github.com/machulav/ec2-github-runner#customize-hardware-configuration) plugin to use a beefier instance in EC2.

tgaddair:
Only limitation at the moment is that the Docker builds don't run in parallel. This looks to be addressed by https://github.com/machulav/ec2-github-runner/issues/8, but not ETA is given at the moment.

build-docker-images.sh
```diff
@@ -1,41 +0,0 @@
-#!/bin/bash
```
richardliaw, line 1: hmm any reason to remove this file? this could still help users if they wanted to build themselves right?

tgaddair: Let me add a README to help with this.

tgaddair: Added README, let me know what you think.

richardliaw (approved):
Looks fine to me!

build-docker-images.sh
richardliaw: nice!

.github/workflows/docker.yml
```diff
@@ -0,0 +1,115 @@
...
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+
+  stop-runner:
+    name: Stop self-hosted EC2 runner
+    needs:
+      - start-runner # required to get output from the start-runner job
+      - docker # required to wait when the main job is done
+    runs-on: ubuntu-latest
+    if: ${{ always() }} # required to stop the runner even if the error happened in the previous jobs
```
EnricoMi, line 101: The `${{` and `}}` are not required in `if:` conditions.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,115 @@
...
+  schedule:
+    - cron: '0 10 * * *' # everyday at 10am
+  push:
+    branches: [ master ]
+    tags: [ 'v*.*.*' ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  start-runner:
```
EnricoMi, line 13: Why is this running on a self-hosted runner? Because GH runner have too small disks?

docker/horovod-ray/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+USER ray
+RUN sudo mkdir -p /horovod && sudo chown ray:users /horovod
+WORKDIR /horovod
+
+# Install Horovod, temporarily using CUDA stubs
+COPY --chown=ray:users . .
+RUN python setup.py sdist && \
+    sudo ldconfig /usr/local/cuda/targets/x86_64-linux/lib/stubs && \
+    HOROVOD_GPU_OPERATIONS=NCCL HOROVOD_WITH_TENSORFLOW=1 HOROVOD_WITH_PYTORCH=1 pip install --no-cache-dir -v $(ls /horovod/dist/horovod-*.tar.gz)[ray] && \
```
EnricoMi, line 46: why not `[all-frameworks,ray]`?

docs/contributors.rst
```diff
@@ -339,31 +339,4 @@ Verify that the latest version of Horovod is now available:
 
     $ pip install --upgrade horovod
 
-
```
EnricoMi, line 342: no replacement here? The content of `docker/README.md`?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,115 @@
+name: docker
+
+on:
+  schedule:
+    - cron: '0 10 * * *' # everyday at 10am
+  push:
+    branches: [ master ]
+    tags: [ 'v*.*.*' ]
+  pull_request:
```
EnricoMi, line 9: Why on `pull_request` on master? Merging PR into master is covered by `push`.

.github/workflows/docker.yml
EnricoMi: @tgaddair saw your comment above.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,115 @@
+name: docker
+
+on:
+  schedule:
+    - cron: '0 10 * * *' # everyday at 10am
```
EnricoMi, line 5: since we are publishing with fixed framework versions it would be sufficient to publish when merging into master, not every day, so coupling this to our unit test workflow

.github/workflows/docker.yml
```diff
@@ -0,0 +1,115 @@
...
+      -
+        name: Docker meta
+        id: meta
+        uses: crazy-max/ghaction-docker-meta@v2
+        with:
+          # list of Docker images to use as base name for tags
+          images: |
+            horovod/${{ matrix.docker-image }}
+          # generate Docker tags based on the following events/attributes
+          tags: |
```
EnricoMi, line 64: Which tags will scheduled master builds get? `nightly` or a date?

.github/workflows/docker.yml
tgaddair: Yeah, they run out disk pretty quickly, as the image is over 10GB.

docker/horovod-ray/Dockerfile
tgaddair: No need for Spark and MXNet for Ray, and PyTorch / TensorFlow should already be installed.

docs/contributors.rst
tgaddair: Yeah, because this is done automatically through a workflow, we no longer need to perform these steps manually.

.github/workflows/docker.yml
tgaddair: The `pull_request` allows us to test that build runs successfully (without pushing to DockerHub).

docs/contributors.rst
EnricoMi: ah, you mean this were the docker release steps, I was always wondering where I could find them...

.github/workflows/docker.yml
tgaddair: That's mostly true, except for the Ray image, which is updated nightly.  Might be worth finding a way to exclude the others.

.github/workflows/docker.yml
tgaddair: The cron builds will get the `nightly` tag.

EnricoMi:
The docker meta action adds the following label: `org.opencontainers.image.licenses=NOASSERTION`, happy with that? No idea what that means, though.

.github/workflows/docker.yml
EnricoMi: good

.github/workflows/docker.yml
tgaddair: Nice, fixed!

tgaddair:
Hmm, me neither, but seems to be okay.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,115 @@
...
+        with:
+          mode: start
+          github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
+          ec2-image-id: ami-0759580dedc953d1f
+          ec2-instance-type: c5.9xlarge
+          subnet-id: subnet-0983be43
+          security-group-id: sg-4cba0d08
+
+  docker:
+    if: ${{ github.repository_owner == 'horovod' }}
```
EnricoMi, line 38: a few more `if:` clauses in the file

EnricoMi (approved):
Looks good, nice workflow!

docs/contributors.rst
tgaddair: :)

EnricoMi:
> Only limitation at the moment is that the Docker builds don't run in parallel.

You could spin up one self-hosted instance per matrix value to build in parallel. Easiest way would be to move start-runner and stop-runner steps into docker job. Though it does not improve readability ...

tgaddair:
> > Only limitation at the moment is that the Docker builds don't run in parallel.
> 
> You could spin up one self-hosted instance per matrix value to build in parallel. Easiest way would be to move start-runner and stop-runner steps into docker job. Though it does not improve readability ...

I thought about that, but I don't think it will work, because the two jobs need to execute on different machines. For the `start EC2` job it will run on a normal GitHub machine, while the `docker` job needs to execute on the already initialized EC2 instance.

EnricoMi:
Right, that is tricky. So the start and stop job would need the same matrix, and then there needs to be a way to get the instance name from the respective matrix start sub job.

Is something like this possible?

```
  start-runner:
    id: start-runner-${{ matrix.docker-image }}
    strategy:
      fail-fast: false
      matrix:
        docker-image:
          - horovod
          - horovod-cpu
          - horovod-ray

  docker:
    needs: start-runner # required to start the main job when the runner is ready
    runs-on: ${{ needs.start-runner-${{ matrix.docker-image }}.outputs.label }} # run the job on the newly created runner
    strategy:
      fail-fast: false
      matrix:
        docker-image:
          - horovod
          - horovod-cpu
          - horovod-ray
```

Probably not ...

tgaddair:
I was wondering about that, too. I am not sure, but may be worth discussing in this thread at some point: https://github.com/machulav/ec2-github-runner/issues/8

PR #2923, opened 2021-05-20
add customized data loader

irasit:
Signed-off-by: Peng Zhang <pengz@uber.com>

## Description
To fix the deadlock issue from uneven data in model check pointing and early stopping callback
 - Make data loader’s num_epoch confinable, default as infinite loop
 - Make train_steps_per_epoch configurable, default as “row_count/batch_size/hvd.size()”
 - Make data loader configurable in estimator, default as new async data loader.
 - Add common data loaders interface, and a mixin class to load and produce batches asynchronously. 
    Mainly used to detach data loader with trainer.  It can be used by Ray data loader in future.
 - Enable model checkpoint call back and early stopping call back test cases.

## Review process to land 


horovod/spark/common/data_loader.py
```diff
@@ -0,0 +1,151 @@
...
+                    break
+                if isinstance(b, Exception):
+                    raise b
+                yield b
+        else:
+            for b in self._iterate():
+                yield b
+
+
+class PetastormReaderWrapper(BaseDataLoader):
```
chongxiaoc, line 73: Maybe we don't need `BaseDataLoader` here, name is very confusing.

horovod/spark/common/data_loader.py
```diff
@@ -0,0 +1,151 @@
...
+    def __len__(self):
+        return self.total_size
+
+    def __iter__(self):
+        if self.reader.last_row_consumed:
+            self.reader.reset()
+
+        total_size = 0
+        for b in self.reader:
+            b = b._asdict()
```
chongxiaoc, line 91: I think `_asdict()` will be called in petastorm dataloder itself.
https://github.com/uber/petastorm/blob/7cda582a50c276c99cbeb0405675f103b6765d33/petastorm/pytorch.py#L195

horovod/spark/common/data_loader.py
chongxiaoc: Also, if on line#102, the reader is pure `petastormReader`, then I think we don't need thisReaderWrapper here.

horovod/spark/lightning/estimator.py
```diff
@@ -167,6 +167,12 @@ class TorchEstimator(HorovodEstimator, TorchEstimatorParamsWritable,
     num_gpus = Param(Params._dummy(), 'num_gpus',
                      'Number of gpus per process, default to 1 when CUDA is available in the backend, otherwise 0.')
 
+    data_loader_class = Param(Params._dummy(), 'data_loader_class',
+                              'Name of the dataloader class.')
+
+    loader_num_epochs = Param(Params._dummy(), 'loader_num_epochs',
+                              'Number of epochs whcih data loader reads in each iteration. If set to None, reader will be in infinite loop mode.')
```
chongxiaoc, line 174: typo `whcih `

horovod/spark/lightning/remote.py
```diff
@@ -117,8 +119,14 @@ def train(serialized_model):
 
             model = deserialize(serialized_model)
 
-            _train_steps_per_epoch = train_steps_per_epoch if train_steps_per_epoch else 1.0
-            _val_steps_per_epoch = val_steps_per_epoch if val_steps_per_epoch else 1.0
+            # _train_steps_per_epoch = train_steps_per_epoch if train_steps_per_epoch else 1.0
+            # _val_steps_per_epoch = val_steps_per_epoch if val_steps_per_epoch else 1.0
```
chongxiaoc, line 123: remove comment?

horovod/spark/common/data_loader.py
```diff
@@ -0,0 +1,151 @@
...
+            total_size += len(b)
+
+            yield b
+        self.total_size = total_size
+
+
+class PetastormDataLoader(BaseDataLoader):
```
chongxiaoc, line 98: Maybe we can give a different name here, in case confusing with original petastorm stuff.

horovod/spark/common/data_loader.py
```diff
@@ -0,0 +1,151 @@
...
+            self.reader,
+            batch_size=self.batch_size,
+            shuffling_queue_capacity=self.shuffling_queue_capacity,
+        )
+
+        for batch in data_loader:
+            yield self._process_batch(batch)
+
+
+class PetastormAsyncDataLoader(BaseAsyncDataLoader):
```
chongxiaoc, line 126: Is this covered by tests?

chongxiaoc (changes requested):
Overall it looks good to me. I leave a few comments open to discuss.

chongxiaoc:
Also, I think we should add in README or somewhere in docs, introducing the basic class of pytorch dataloader. In that sense, users know how to inherit and implement a dataloader on their own.
For example,
```
import horovod.spark.common.basePytorchDataLoader  as BaseDataloader

class MyDataLoader(BaseDataloader)
#  customize below

```

horovod/spark/data_loaders/pytorch_data_loaders.py
```diff
@@ -0,0 +1,58 @@
...
+    def __len__(self):
+        return len(self.reader)
+
+    def _iterate(self):
+        if self.reader.last_row_consumed:
+            print(f"Resetting Petastorm reader for {self.reader.dataset.paths}")
+            self.reader.reset()
+
+        # Re-create the data loader for each iterate. There maybe some left over data
+        # from last epoch which will cause petastorm's BatchedDataLoader fail to reset.
```
chongxiaoc, line 22: The comment means this dataloader is expected to fail in some corner cases?

horovod/spark/data_loaders/pytorch_data_loaders.py
```diff
@@ -0,0 +1,58 @@
...
```
chongxiaoc, line 50: same as discussion above for this comment.

horovod/spark/lightning/remote.py
```diff
@@ -74,10 +76,10 @@ def RemoteTrainer(estimator, metadata, ckpt_bytes, run_id, dataset_idx, train_ro
     if sample_weight_col:
         schema_fields.append(sample_weight_col)
 
-    dataloader_cls = _create_dataloader(feature_columns, input_shapes, metadata)
+    data_loader_cls = _create_dataloader(feature_columns, input_shapes, metadata, data_loader_cls)
     make_petastorm_reader = _make_petastorm_reader_fn(transformation, schema_fields,
                                                       batch_size, calculate_shuffle_buffer_size,
-                                                      dataloader_cls)
+                                                      data_loader_cls, loader_num_epochs)
```
chongxiaoc, line 82: I did a quick search here, the `make_petastorm_reader` is not used anymore?
This function actually set dataloader as attribute of lightning model.
If I understand correctly, we should rename this function.

chongxiaoc (commented):
Leave a few comments for 2nd round.
Thanks for refactoring the code, looks better.

and add `mixin`?

thuningxu:
Like it! Looks good to me but also want @tgaddair to take a look.

horovod/spark/data_loaders/pytorch_data_loaders.py
chongxiaoc: Your change guarantees to do reset() when last row is consumed, leaving a comment here is very confusing. How about just saying "need to reset reader() once last row is consumed"?

horovod/spark/data_loaders/data_loader_base.py
```diff
@@ -0,0 +1,116 @@
...
+class AsyncDataLoaderMixin(object):
+    """
+    Async Mixin on top of the implementation of BaseDataLoader. It contains a seperate thread
+    which reads data from self._iterate() and push them in the queue. The self.__iter__() function
+    will pop the data from the queue.
+    For example:
+        class PytorchAsyncDataLoader(AsyncDataLoaderMixin, PytorchDataLoader):
+    """
+
+    def __init__(self, async_loader_queue_size=64, *args, **kwargs):
```
chongxiaoc, line 42: Questions: I remember we discussed offline but I forgot, why `*args, **kwargs` are needed for constructor?

tgaddair: In some cases you may have multiple-inheritance, in which case one of the other superclasses could need initialization as well, so I think this is good practice. (see https://www.educative.io/edpresso/what-is-mro-in-python).

horovod/spark/data_loaders/data_loader_base.py
```diff
@@ -0,0 +1,116 @@
...
+        initialize the async data loader. Need to add this in the __init__() of the implementation
+        """
+        self.async_loader_queue_size = async_loader_queue_size
+        super().__init__(*args, **kwargs)
+
+        print(f"Apply the AsyncDataLoaderMixin on top of the data loader, async_loader_queue_size={async_loader_queue_size}. ")
+
+        if self.async_loader_queue_size > 0:
+            self.finished_event = Event()
+            self.queue = Queue(self.async_loader_queue_size)
```
tgaddair, line 53: Should the option be provided to use a Process instead of a Thread here? Did you try both in your tests? I suppose if most of the work is on I/O it should be okay to use a thread.

docs/spark.rst
```diff
@@ -96,6 +96,8 @@ logging (for Tensorboard) using the Estimator ``Store`` abstraction.  Stores are
 artifacts including intermediate representations of the training data.  Horovod natively supports stores for HDFS
 and local filesystems.
 
+Petastorm based data loader is used by default, but user can define a custom data loader by override the `base_data_loader` interface.
```
tgaddair, line 99: Nit: link Petastorm to the GitHub page.

docs/pytorch.rst
```diff
@@ -139,3 +139,5 @@ Start the training job and specify the number of workers on the command line as
 You can find an example of use pytorch lightning trainer with horovod backend in `pytorch_lightning_mnist.py script <../examples/pytorch/pytorch_lightning_mnist.py>`__
 
 See the PyTorch Lightning `docs <https://pytorch-lightning.readthedocs.io/en/stable/multi_gpu.html#horovod>`_ for more details.
+
+A pytorch-lightning based spark estimator trainer is also added example is in `pytorch_lightning_spark_mnist.py <../examples/spark/pytorch/pytorch_lightning_spark_mnist.py>`__
```
tgaddair, line 143: Nit: capitalize PyTorch Lightning.

horovod/spark/lightning/estimator.py
```diff
@@ -170,7 +170,13 @@ class TorchEstimator(HorovodEstimator, TorchEstimatorParamsWritable,
     logger = Param(Params._dummy(), 'logger', 'optional logger')
 
     log_every_n_steps = Param(Params._dummy(), 'log_every_n_steps', 'control the frequency of logging',
-                                typeConverter=TypeConverters.toInt)
+                              typeConverter=TypeConverters.toInt)
+
+    data_loader_class = Param(Params._dummy(), 'data_loader_class',
+                              'Name of the dataloader class.')
+
+    loader_num_epochs = Param(Params._dummy(), 'loader_num_epochs',
```
tgaddair, line 178: When would the user want to set this? It seems we would always want to set it to `None` and let the trainer decide when stop reading.

horovod/spark/lightning/estimator.py
```diff
@@ -170,7 +170,13 @@ class TorchEstimator(HorovodEstimator, TorchEstimatorParamsWritable,
     logger = Param(Params._dummy(), 'logger', 'optional logger')
 
     log_every_n_steps = Param(Params._dummy(), 'log_every_n_steps', 'control the frequency of logging',
-                                typeConverter=TypeConverters.toInt)
+                              typeConverter=TypeConverters.toInt)
+
+    data_loader_class = Param(Params._dummy(), 'data_loader_class',
```
tgaddair, line 175: Is this a name (string) or is it a class / function? Seems it would be easier to pass around a class directly.

tgaddair (commented):
Looks good, just a few minor things.

chongxiaoc (approved):

PR #2928, opened 2021-05-22
Spin up multiple EC2 instances to build docker images

EnricoMi:
Spin up multiple EC2 instances to build docker images in parallel:

grafik


tgaddair (approved):
Awesome work getting this working!

PR #2930, opened 2021-05-22
Fixed CPU docker images and added additional sanity checks

tgaddair:
Previously, using `all-framemworks` extras for CPU builds was resulting in both `tensorflow-cpu` and `tensorflow` being installed, which broke `horovod.tensorflow` imports.

Also changes EC2 instance used to build the image to a smaller, cheaper instance, as most of the build is single-threaded and does not need high amounts of CPU, memory, or network bandwidth.


EnricoMi (approved):
LGTM!

PR #2931, opened 2021-05-22
Enforce a single concurrent run of docker workflow per branch

EnricoMi:
Pushing a few commits to a branch kicks of the docker workflow for each of them, causing concurrent runs. Firstly, the earlier runs are redundant, secondly, a slow earlier run may overwrite (on master only) images published by a later run.

This enforces a single concurrent run of the docker workflow for each branch, which kills earlier runs.


tgaddair (approved):
Nice!

PR #2942, opened 2021-05-27
torch estimator: check all ranks have same device type

chongxiaoc:

## Description

Fixes # (issue).

## Review process to land 

### DIFF
--- a/horovod/spark/lightning/remote.py
+++ b/horovod/spark/lightning/remote.py
@@ -121,6 +121,12 @@ def RemoteTrainer(estimator, metadata, ckpt_bytes, run_id, dataset_idx, train_ro
             _val_steps_per_epoch = val_steps_per_epoch if val_steps_per_epoch else 1.0
 
             cuda_available = torch.cuda.is_available()
+            # We need to check all ranks have same device type for traning.
+            # Horovod doesn't support heterogeneous allreduce for gradients.
+            cuda_avail_list = hvd.allgather_object(cuda_available, name='device type')
+            if cuda_avail_list.count(cuda_available) != hvd.size():
+                raise RuntimeError("All ranks don't have same device type!")
+
             if cuda_available:
                 # Horovod: pin GPU to local rank or the assigned GPU from spark.
                 torch.cuda.set_device(_get_assigned_gpu_or_default(default=hvd.local_rank()))
--- a/horovod/spark/torch/remote.py
+++ b/horovod/spark/torch/remote.py
@@ -123,6 +123,12 @@ def RemoteTrainer(estimator, metadata, last_checkpoint_state, run_id, dataset_id
             shuffle_buffer_size = user_shuffle_buffer_size
 
         cuda_available = torch.cuda.is_available()
+        # We need to check all ranks have same device type for traning.
+        # Horovod doesn't support heterogeneous allreduce for gradients.
+        cuda_avail_list = hvd.allgather_object(cuda_available, name='device type')
+        if cuda_avail_list.count(cuda_available) != hvd.size():
+            raise RuntimeError("All ranks don't have same device type!")
+
         if cuda_available:
             # Horovod: pin GPU to local rank or the assigned GPU from spark.
             torch.cuda.set_device(_get_assigned_gpu_or_default(default=hvd.local_rank()))


horovod/spark/lightning/remote.py
```diff
@@ -121,6 +121,12 @@ def train(serialized_model):
             _val_steps_per_epoch = val_steps_per_epoch if val_steps_per_epoch else 1.0
 
             cuda_available = torch.cuda.is_available()
+            # We need to check all ranks have same device type for traning.
+            # Horovod doesn't support heterogeneous allreduce for gradients.
+            cuda_avail_list = hvd.allgather_object(cuda_available, name='device type')
+            if hvd.rank() == 0:
+                assert cuda_avail_list.count(cuda_available) == hvd.size(), "All ranks don't have same device type!"
```
tgaddair, line 128: Wouldn't you want to assert this on all ranks, not just rank 0?

horovod/spark/torch/remote.py
```diff
@@ -123,6 +123,12 @@ def train(serialized_model, optimizer_cls, model_opt_state_serialized,
             shuffle_buffer_size = user_shuffle_buffer_size
 
         cuda_available = torch.cuda.is_available()
+        # We need to check all ranks have same device type for traning.
+        # Horovod doesn't support heterogeneous allreduce for gradients.
+        cuda_avail_list = hvd.allgather_object(cuda_available, name='device type')
+        if hvd.rank() == 0:
+            assert cuda_avail_list.count(cuda_available) == hvd.size(), "All ranks don't have same device type!"
```
tgaddair, line 130: Same as above.

horovod/spark/lightning/remote.py
chongxiaoc: That would be the same result, since you have all nodes status in a `cuda_avail_list`.
We are doing a count op here so it covers all cases:
1. all False/True: assert passes
2. The `cuda_avail_list` has both True/False inside: `cuda_avail_list.count(cuda_available) == hvd.size()` will fail no matter `cuda_available` is True or False.

In other words, allgather returns a global result, which rank checks `cuda_avail_list.count(cuda_available) == hvd.size()`doesn't matter.

tgaddair: Right, but because you're only checking this on rank 0, you're relying on the fact that rank.0 will fail, and then some time in the future the other ranks will attempt to perform a collective op (like `hvd.allreduce`) which will cause the other ranks to fail as well. If you perform the check on all ranks, then they all fail at the same time, which is more reliable and fails faster. Does that make sense?

chongxiaoc: I see.

chongxiaoc: fixed

horovod/spark/torch/remote.py
chongxiaoc: fixed

horovod/spark/lightning/remote.py
```diff
@@ -121,6 +121,11 @@ def train(serialized_model):
             _val_steps_per_epoch = val_steps_per_epoch if val_steps_per_epoch else 1.0
 
             cuda_available = torch.cuda.is_available()
+            # We need to check all ranks have same device type for traning.
+            # Horovod doesn't support heterogeneous allreduce for gradients.
+            cuda_avail_list = hvd.allgather_object(cuda_available, name='device type')
+            assert cuda_avail_list.count(cuda_available) == hvd.size(), "All ranks don't have same device type!"
```
tgaddair, line 127: One thing I forgot to point out: this should be an exception (probably `RuntimeError`) rather than an exception. Two reasons for this:

1. Exceptions can be caught, whereas assertions effectively terminate the program, so there's nothing the caller can do to recover.
2. Assertions can be removed by the interpreter at runtime when optimizers are enabled.

Assertions are primarily for debugging application code, whereas Exceptions are for errors at runtime.

chongxiaoc: okay, I was thinking about raise assertion error instead of RuntimeError before.

chongxiaoc: fixed.

chongxiaoc:
@tgaddair Can you take one more look?

tgaddair (approved):
LGTM!

PR #2945, opened 2021-05-28
Added back horovod.tensorflow.keras.Compression

tgaddair:


EnricoMi (approved):
So it was the sphinx version, that makes more sense.

PR #2959, opened 2021-06-08
Move cancellation hook into job from workflow to ensure instance cleanup

tgaddair:
Some instances in EC2 have been dangling for a few days. I suspect it's due to workflow cancellation signals being sent when a new commit is pushed. Canceling the workflow (I suspect) is preventing the cleanup job from running at the end. This change moves the cancellation handler into the job, so that the following cleanup job can still run.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -9,12 +9,6 @@ on:
   pull_request:
     branches: [ master ]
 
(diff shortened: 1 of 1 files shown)


EnricoMi (approved):
I saw the expected behaviour, that a cancelled workflow stopped the runner. But maybe it is not reliable, so I am happy with this narrower cancel approach.

PR #2963, opened 2021-06-09
Implement more asynchronous dependency handling on GPU 

romerojosh:

## Description
This PR introduces changes to Horovod to enable more asynchronous data dependency handling on GPU, for both input data into Horovod and output data back to the framework.

In Horovod prior to this PR, input data dependencies for the GPU are enforced via CPU synchronization on `ReadyEvent`s (i.e.  `cudaEvent`) before scheduling any GPU collective operations. While this does enforce the dependency, it stalls the Horovod background thread unnecessarily. Instead of moving on to coordinate/schedule additional collective operations, the thread waits for the data to be ready on the first operation scheduled in the sequence. To free up the background thread, the CPU synchronization is replaced with calls to `cudaStreamWaitEvent` which allow the background thread to immediately schedule dependent GPU operations, offloading the data dependency handling to the CUDA driver. 

Horovod also enforces output data dependencies via CPU synchronization. Currently, Horovod spawns a completion thread that waits on a `cudaEvent` (for this discussion, I'll refer to this as the completion event) recorded after Horovod GPU operations are scheduled, which ensures that any completion callbacks to the framework are not executed until the Horovod GPU operations have completed. However, this introduces latency on the framework ops dependent on the Horovod results because this implementation makes it impossible for the framework to schedule any of those ops until Horovod operations have completed, exposing the full cost of the callback/setup/launch of the downstream operation. Similar to the input data dependency, this PR introduces a mechanism for asynchronous dependency handling of the output data via `cudaEvent`s and `cudaStreamWaitEvent`. Instead of having the Horovod completion thread wait on the completion event, we can pass the event through to the framework via the callback, and handle it on the framework side:
- For TF/PyT, the completion event is used with a `cudaStreamWaitEvent` call to place the dependency on the framework GPU stream.
- For MXNet, the dependency is still enforced via a CPU sync with `cudaEventSynchronize` on the completion event. This will be updated when an upstream MXNet feature (https://github.com/apache/incubator-mxnet/pull/20331) that can better take advantage of this mechanism is merged. I will be posting a second WIP PR to illustrate this soon.

The output data dependency handling, due to more complex interactions with the framework, is currently disabled by default and enabled via a new env var `HOROVOD_ENABLE_ASYNC_COMPLETION`. This can be changed based on feedback. Additionally, when the timeline is enabled, the input and output dependencies revert to being handled via CPU synchronization in order to provide correct timings. 

Some other related, but more minor changes in this PR:
- `ReadyEvent`s replaced with list of `ReadyEvent`. This is to enable an input tensor having multiple upstream dependencies (required in upcoming MXNet PR)
- In the non-async-completion path, replaced spin waiting on `cudaEventQuery` in `WaitForEvents` with calls to `cudaEventSynchronize` in non-elastic cases to reduce number of CUDA API calls.


horovod/common/ops/mpi_gpu_operations.cc
```diff
@@ -195,6 +199,9 @@ Status MPI_GPUAlltoall::Execute(std::vector<TensorTableEntry>& entries, const Re
   assert(entries.size() == 1);
 
   gpu_op_context_.InitGPU(entries);
+
+  WaitForData(entries);
```
tgaddair, line 203: Is this `InitGPU` call why `WaitForData` needs to be in the `Execute` function instead of sitting before it? Otherwise it seems we could reduce some duplication by calling it beforehand.

tgaddair (approved):
LGTM! Just a small question about `WaitForData`.

horovod/common/ops/mpi_gpu_operations.cc
romerojosh: Yes, the `InitGPU` call is what lazily creates Horovod's internal CUDA stream which is required for the `WaitForData` variant that uses `cudaStreamWaitEvent`. I do also dislike the duplication but wasn't sure if there were better options. If you have any suggestions, I'm all ears.

tgaddair: Thanks for clarifying, I think that's fine then. Let's land it!

PR #2964, opened 2021-06-09
Fixed docker workflow stability

tgaddair:
A number of jobs have been failing due to issues in the Docker workflow. This PR makes a few changes to address this:

1. Removed parallelism, as it seems to result in some race conditions. Because the Buildkite tests take about 90 minutes to complete, it should not be necessary to parallelize the docker workflow, which will complete ahead of the tests even serially.
2. Changed cancellation behavior so `master` branch docker builds are never cancelled. This is important as otherwise we could miss building images for each SHA, version, etc. The assumption here is that `github.head_ref` will be empty on master, so we use the unique `github.sha` instead for the concurrency group.
3. Added metadata to identify EC2 workers in AWS. This will help us discover which jobs led to which instances being created, to help with debugging.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -57,7 +57,7 @@ jobs:
     # we want an ongoing run of this workflow to be canceled by a later commit
     # so that there is only one concurrent run of this workflow for each branch
     concurrency:
-      group: docker-${{ github.head_ref || github.sha }}
+      group: docker-${{ matrix.docker-image }}-${{ github.head_ref || github.sha }}
       cancel-in-progress: true
 
     strategy:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -11,25 +11,16 @@ on:
 
 jobs:
   start-runner:
-    name: Start self-hosted EC2 runner (${{ matrix.docker-image }})
+    name: Start self-hosted EC2 runner
     if: >
       github.event_name == 'schedule' && github.repository == 'horovod/horovod' ||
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -49,35 +40,34 @@ jobs:
...
     if: needs.start-runner.result != 'skipped'
     needs: start-runner # required to start the main job when the runner is ready
-    runs-on: ${{ matrix.label }} # run the job on the newly created runners
+    runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runners
 
     # we want an ongoing run of this workflow to be canceled by a later commit
     # so that there is only one concurrent run of this workflow for each branch
     concurrency:
-      group: docker-${{ github.head_ref }}
+      group: docker-${{ matrix.docker-image }}-${{ github.head_ref || github.sha }}
```
EnricoMi, line 60: What does the `github.head_ref || github.sha` expression do? `github.head_ref`, or if empty `github.sha`? That is not documented, is it?

EnricoMi: Please put your thoughts on this particular expression in a comment here (branch vs. master)

EnricoMi (approved):
Minor question and comment requested.

.github/workflows/docker.yml
tgaddair: Good point. Updated with a comment. And yes, the idea is that on master we want to ensure that every workflow run has a unique ID so we don't cancel any CI runs on master.

PR #2969, opened 2021-06-10
Added GitHub workflow to publish Horovod releases to PyPI

tgaddair:


.github/workflows/upload-pypi.yml
```diff
@@ -0,0 +1,34 @@
+name: Upload to PyPI
+
+on:
+  # Triggers the workflow when a release is created
+  release:
+    types: [created]
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
EnricoMi, line 8: How does this manual triggered workflow know which release to publish? Is there anything that can be selected like a tag or commit?

EnricoMi (approved):

.github/workflows/upload-pypi.yml
tgaddair: For PyPI, it always looks at the metadata in the `setup.py` file. So the version will come from here: https://github.com/horovod/horovod/blob/master/setup.py#L156. This workflow will always run from the tag you use to create the release.

tgaddair: But yeah, in the case of the manual run, I believe it will be HEAD on master.

PR #2970, opened 2021-06-10
Build release Docker images on github-hosted runners

EnricoMi:
We can build release Docker images with GitHub actions and avoid EC2 instances. The `horovod-ray` image is the largest, as the input image is 4.27 GB, where `horovod`'s input Nvidia image is only 2.17 GB.

Several improvements make this work now:
- running `pip install` with `--no-cache-dir` so that downloaded pip packages are not stored in the Docker image
- running `apt-get` together with `apt-get clean && rm -rf /var/lib/apt/lists/*` removes distribution files
- installing PyTorch first, as this requires 2 GB temporary space (GPU)
- pruning docker images before running docker build frees enough disk space to even build `horovod-ray` frees another 4 GB of pre-allocated disk space

It looks like we have 1.7 GB disk space head room: https://github.com/horovod/horovod/runs/2800641461?check_suite_focus=true#step:9:20

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -67,6 +67,9 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - name: Cleanup Docker
+        run: docker image prune -a --force
+
       - name: Build and push
         uses: docker/build-push-action@v2
         with:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,49 +10,9 @@ on:
     branches: [ master ]
 
 jobs:
-  start-runner:
-    name: Start self-hosted EC2 runner
-    if: >
-      github.event_name == 'schedule' && github.repository == 'horovod/horovod' ||
-      github.event_name == 'push' && github.repository == 'horovod/horovod' ||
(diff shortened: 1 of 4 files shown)


EnricoMi:
Building horovod-ray does not work as the underlying image (rayproject/ray:nightly-gpu) is [4.27 GB](https://hub.docker.com/layers/rayproject/ray/nightly-gpu/images/sha256-55484d5dc418a71a79fd990b2fbaff6a849d04f3a7a56e5894622146690adebe?context=explore), where the the NVidia image for horovod gpu (nvidia/cuda:11.2.2-devel-ubuntu18.04) is [2.17 GB](https://hub.docker.com/layers/nvidia/cuda/11.2.2-devel-ubuntu18.04/images/sha256-5f348e0ad543f3e2061b3407c90d0b8ec8cd77ebfa9e8b98057648b641dce73c?context=explore).

tgaddair:
Nice! This is much better than dealing with EC2. Do the optimizations (like removing apt lists) have any implications for users who want to build on top of these images, installing their own packages? I assume this stuff can be refreshed with calling `apt-get update`, right?

tgaddair (approved):

EnricoMi:
> I assume this stuff can be refreshed with calling `apt-get update`, right?

That is right, this only undos the `apt-get update` that is done in order to apt-get install.

EnricoMi:
We can go back to EC2 whenever this does not work any more.

EnricoMi:
Looks like actually pushing the images requires more disk space: https://github.com/horovod/horovod/runs/2807151712

PR #2987, opened 2021-06-16
FP16 compression for MXNet

Adnios:

## Description

Try to solve the problem mentioned in https://github.com/horovod/horovod/pull/2915
Thank you for your time on reviewing this pr.
cc @romerojosh @tgaddair

### DIFF
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -194,7 +194,8 @@ class DistributedTrainer(mx.gluon.Trainer):
                     allreduce_(tensor_compressed, average=False,
                                name=self._prefix + str(i), priority=-i,
                                prescale_factor=1.0 / self._gradient_predivide_factor)
-                    param.list_grad()[0][:] = self._compression.decompress(tensor_compressed, ctx)
+                    if self._compression != Compression.none:
+                        param.list_grad()[0][:] = self._compression.decompress(tensor_compressed, ctx)
 
 # Wrapper to inject Horovod broadcast after parameter initialization
 def _append_broadcast_init(param, root_rank, name):
--- a/horovod/mxnet/__init__.py
+++ b/horovod/mxnet/__init__.py
@@ -194,6 +194,7 @@ class DistributedTrainer(mx.gluon.Trainer):
                     allreduce_(tensor_compressed, average=False,
                                name=self._prefix + str(i), priority=-i,
                                prescale_factor=1.0 / self._gradient_predivide_factor)
+
                     if self._compression != Compression.none:
                         param.list_grad()[0][:] = self._compression.decompress(tensor_compressed, ctx)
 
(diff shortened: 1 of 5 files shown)


horovod/mxnet/__init__.py
```diff
@@ -169,15 +179,22 @@ def _allreduce_grads(self):
...
             # Meanwhile, since horovod requires Python 3.6, there is no need to sort
             # self._params as enumerating a python dict is always deterministic.
             for i, param in enumerate(self._params):
                 if param.grad_req != 'null':
-                    allreduce_(param.list_grad()[0], average=False,
+                    tensor_compressed, ctx = self._compression.compress(param.list_grad()[0])
+                    allreduce_(tensor_compressed, average=False,
                                name=self._prefix + str(i), priority=-i,
                                prescale_factor=1.0 / self._gradient_predivide_factor)
+                    param.list_grad()[0][:] = self._compression.decompress(tensor_compressed, ctx)
```
romerojosh, line 197: In the case without compression enabled (i.e. the tensor is the same on the left and right hand sides of this assignment), does MXNet do any copies or does it automatically skip the assignment? I think it may just skip the assignment so this shouldn't result in extra copies in the case without compression. With that said, it still might be good to add the if statement here to only do the copy assignment conditionally on `self._compression != Compression.none` for safety and consistency.

romerojosh (commented):
Thanks for updating this PR! Just left a small comment.

I think it is good to leave the `DistributedOptimizer` implementation for symbolic alone as well since it will be mostly deprecated moving forward with MXNet 2.0 anyway.

horovod/mxnet/__init__.py
Adnios: Agree. I will add if statement.

horovod/mxnet/__init__.py
```diff
@@ -169,16 +179,25 @@ def _allreduce_grads(self):
                     grads, names = zip(*entries)
                     grouped_allreduce_(tensors=grads, average=False, name="{}:{}".format(names[0], names[-1]), priority=-i,
                                        prescale_factor=1.0 / self._gradient_predivide_factor)
+
+            if self._compression != Compression.none:
+                for i, param in enumerate(self._params):
+                    if param.grad_req != 'null':
+                        param.list_grad()[0][:] = self._compression.decompress(grads[i], ctxs[i])
```
romerojosh, line 186: Using `grads[i]` and `ctx[i]` here is not quite correct. The only gradients/ctx in the `grads` and `ctx` lists are those with `param.grad_req != 'null'` already. You need to use a different index instead of `i` to access these lists.

Adnios: @romerojosh Thanks for your reminder. There is another problem in my code. When the `groups>1`, `grads` only store the last `grouped_allreduce_`'s `grads`.
I think that it can be solved by using an another list `tensors_compressed` to store all the compressed `param.list_grad`. And decompress `tensors_compressed` after `grouped_allreduce_` finish.

romerojosh (approved):
LGTM! Thanks a lot for fixing up this PR @Adnios.

PR #2997, opened 2021-06-25
Move tests to pyspark 3.1.2, 3.0.3, 2.4.8

EnricoMi:
There are new minor releases for PySpark to test against.


EnricoMi:
Duplicates #2999.

PR #3015, opened 2021-07-05
Add pip freeze to spot package version changes

EnricoMi:
This adds a `pip freeze` to the test image build to easily spot python package version changes across builds, especially useful when nightly builds fail.


tgaddair (approved):
Good idea! Looks like our current issue is with numpy.

PR #3016, opened 2021-07-05
Pin pillow!=8.3.0 for nightly torchvision

tgaddair:
See: https://github.com/pytorch/vision/issues/4146

### DIFF
--- a/Dockerfile.test.gpu
+++ b/Dockerfile.test.gpu
@@ -165,12 +165,10 @@ RUN if [[ ${TENSORFLOW_PACKAGE} == "tf-nightly-gpu" ]]; then \
     fi
 
 # Install PyTorch (nightly).
-# Pin Pillow<7.0 for torchvision < 0.5.0: https://github.com/pytorch/vision/issues/1718
+# Pin Pillow!=8.3.0 for torchvision: https://github.com/pytorch/vision/issues/4146
 RUN if [[ ${PYTORCH_PACKAGE} == "torch-nightly-cu"* ]]; then \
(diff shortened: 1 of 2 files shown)


EnricoMi (approved):

Dockerfile.test.gpu
```diff
@@ -165,12 +165,10 @@ RUN if [[ ${TENSORFLOW_PACKAGE} == "tf-nightly-gpu" ]]; then \
     fi
 
 # Install PyTorch (nightly).
-# Pin Pillow<7.0 for torchvision < 0.5.0: https://github.com/pytorch/vision/issues/1718
+# Pin Pillow!=8.3.0 for torchvision: https://github.com/pytorch/vision/issues/4146
 RUN if [[ ${PYTORCH_PACKAGE} == "torch-nightly-cu"* ]]; then \
         pip install --no-cache-dir --pre torch ${TORCHVISION_PACKAGE} -f https://download.pytorch.org/whl/nightly/${PYTORCH_PACKAGE/#torch-nightly-/}/torch_nightly.html; \
-        if [[ "${TORCHVISION_PACKAGE/%+*/}" == torchvision==0.[1234].* ]]; then \
```
EnricoMi, line 171: image `test-gpu-gloo-py3_7-tf1_15_5-keras2_2_4-torch1_3_1-mxnet1_5_1_p0-pyspark3_1_2` uses `torchvision==0.4.2+cu100`, which might still require the `Pillow<7.0` fix, unless `8.2.0` fixes that old bug.

EnricoMi:
Looks like pytorch 1.9.0 also pulls in Pillow 8.3.0, so I had to exclude that version for torchvision releases as well.

Dockerfile.test.gpu
tgaddair: I think since this is only for nightly torch, it should be okay.

PR #3018, opened 2021-07-07
add SyncBatchNorm support for training examples

Sharpiless:

## Description

Fixes # (issue).

## Review process to land 

Replace BatchNorm in pytorch examples with  SyncBatchNorm.

Mutil-GPU training with BatchNorm with lead to different BN mean and var between different GPU, and the final accuracy will be lower.


tgaddair (commented):
Thanks for the fix @Sharpiless! Can you make this optional through a command line arg like `--sync-bn`, as it may have an effect on performance?

PR #3026, opened 2021-07-07
Update Rossmann spark example to work with the latest TensorFlow version

rongou:
A few small changes to make the examples work with the latest TensorFlow release.


tgaddair:
Thanks for the PR! Looks like we'll need to modify our CI to skip testing this for tf 1.15, and test with tf 2.5 instead.

rongou:
Do you still want to maintain compatibility with tf 1.15? Probably can be done with some version checks in the code, but it'll make the examples harder to read.

tgaddair:
Hey @rongou, yeah we have a lot of users who are still on TF 1.15. Nvidia maintains a separate fork for 1.15 that contains up-to-date security and stability patches. We don't have to update the example, but we do need to fix the CI system to skip this for older TF versions.

PR #3029, opened 2021-07-09
Merge CPU and GPU CI jobs into one again

EnricoMi:
The "Build and Test" job has been split in #3025 because it looked like a good idea (separation of concerns). But the job dependency graph is more complex now. And each job has the full `max-parallel`, where before that split the cpu images determined the `max-parallel` value as CPU build and test takes much longer than GPU build.

Merging both jobs together will de-prioritize the GPU jobs again and leave more concurrent jobs for other workflow runs.


tgaddair (approved):
LGTM, just need to regenerate the ci.yaml it seems.

PR #3031, opened 2021-07-09
Exclude cffi==1.14.6

EnricoMi:
Looks like cffi==1.14.6 breaks some tests.


tgaddair (approved):

EnricoMi:
The problem wasn't cffi but mxnet nightlies.

PR #3041, opened 2021-07-13
Push docker test images from Github

EnricoMi:
Pushing test images from master allows to use them as caches for later test builds. This mimics the behaviour that we have seen on Buildkite.


EnricoMi:
@tgaddair I presume those images shoulnd't go to `${{ steps.ecr.outputs.registry }}/buildkite:` but something like `${{ steps.ecr.outputs.registry }}/github:. What are your thoughts?

tgaddair (approved):

PR #3044, opened 2021-07-13
Fix jq query for CI (Fork) workflow

EnricoMi:
The CI (Fork) workflow is broken, it cannot process the workflow result JSON because the jq query is broken.


EnricoMi:
@maxhgerlach sorry for breaking the CI workflow, again: https://github.com/horovod/horovod/runs/3056767615?check_suite_focus=true

This should fix it and you are good to see GPU tests.

maxhgerlach:
(off topic) It seems the whole story of figuring out and testing CI pipelines hasn't really been solved by any provider yet. I feel your pain, @EnricoMi. At least there's very little Jenkins left!

tgaddair (approved):

Issue #2590, opened 2021-01-12
Horovod Ops not XLA compatible


jtchilders:
**Environment:**
1. Framework: TensorFlow
2. Framework version: 2.4.0
3. Horovod version: 0.21.1
4. MPI version: openmpi-4.0.5
5. CUDA version: 11.0
6. NCCL version: nccl_2.8.3-1+cuda11.0_x86_64
7. Python version: 3.8.5
8. Spark / PySpark version: NA
9. OS and version: 5.3.0-62-generic #56~18.04.1-Ubuntu
10. GCC version: 7.5.0
11. CMake version: 3.18.2

I've run into errors when trying to XLA compile my Tensorflow train/test steps. In my custom model, if I use
```python
@tf.function(jit_compile=True)
def train_step(...):
```
to force compilation of the training operations I can run successfully without Horovod with 1 process. Then when I try to run with Horovod, I receive errors like:
```bash
The op is created at:
File "main.py", line 368, in <module>
  main()
File "main.py", line 188, in main
  epoch_loop.one_train_epoch(config,trainds,net,
File "/gpfs/mira-home/parton/git/atlas_dgcnn/epoch_loop.py", line 9, in one_train_epoch
  return one_epoch(config,dataset,net,train_step,loss_func,opt,epoch_num,tbwriter,batches_per_epoch,True)
File "/gpfs/mira-home/parton/git/atlas_dgcnn/epoch_loop.py", line 74, in one_epoch
  loss_value,logits = step_func(net,loss_func,inputs,labels,weights,opt,first_batch,hvd)
File "/gpfs/mira-home/parton/git/atlas_dgcnn/epoch_loop.py", line 244, in train_step
  if hvd and first_batch:
File "/gpfs/mira-home/parton/git/atlas_dgcnn/epoch_loop.py", line 246, in train_step
  hvd.broadcast_variables(opt.variables(), root_rank=root_rank)
File "/home/parton/.local/lib/python3.8/site-packages/horovod/tensorflow/functions.py", line 56, in broadcast_variables
  return broadcast_group(variables, root_rank)
File "/tmp/tmp4h_gfbt8.py", line 53, in broadcast_group
[...]

PR #3053, opened 2021-07-20
XLA HorovodAllreduce for tf.function(jit_compile=True)

trentlo:

- [X] Did you update the docs?
- [X] Did you write any tests to validate this change?  

## Description

For fixing https://github.com/horovod/horovod/issues/2590, although only HorovodAllreduce is implemented in this PR but the reported issue might need other Horovod ops. More ops can be added later based on the same code structure and facilities built.

Mainly, this PR implements the XLA HorovodAllreduce op, which provides a way to lower the op into XLA. The main idea is to lower the op into XLA (HLO) `custom-calls` that enqueue requests to the Horovod runtime and process the response.

This new XLA op is disabled by default now mainly because of its dependency on `ASYNC_COMPLETION`, which is essential to mitigating host latencies of the XLA runtime. An `HOROVOD_ENABLE_XLA_OPS` environment variable is added to control the on/off of the XLA op. We will make the op(s) on by default after ASYNC_COMPLETION is on by default or after we implement a way to enable ASYNC_COMPLETION on demand.

See `xla.rst` for an example of `tf.function(jit_compile=True)` with lowering HorovodAllreduce into XLA.

Bump the c++ version to c++14 as it is needed to compile XLA.

## Review process to land 


trentlo:
@romerojosh FYI.

trentlo:
The build failed because the XLA header files are missing in the container?

```
  /tmp/pip-req-build-tuml17oe/horovod/tensorflow/xla_mpi_ops.cc:22:10: fatal error: tensorflow/compiler/tf2xla/xla_op_kernel.h: No such file or directory
   #include "tensorflow/compiler/tf2xla/xla_op_kernel.h"
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  compilation terminated.
```

Could someone give me some pointers about how I could fix it? Thanks!

maxhgerlach:
> The build failed because the XLA header files are missing in the container?
From the build log it looks like only `python3.7/site-packages/tensorflow_core/include` is added to the include paths, but not `site-packages/tensorflow/include`.

I guess what headers are put into tensorflow_core would be controlled by https://github.com/tensorflow/tensorflow/blob/r2.1/tensorflow/core/BUILD, but I also get the impression that the tensorflow_core package disappeared with version 2.2. When I build locally with TF 2.4, `site-packages/tensorflow/include` is certainly added to the include path and I can also locate that XLA header under that hierarchy.

So maybe it would suffice to limit building the XLA support to TF 2.2+?

trentlo:
> > The build failed because the XLA header files are missing in the container?
> 
> From the build log it looks like only `python3.7/site-packages/tensorflow_core/include` is added to the include paths, but not `site-packages/tensorflow/include`.
> 
> I guess what headers are put into tensorflow_core would be controlled by https://github.com/tensorflow/tensorflow/blob/r2.1/tensorflow/core/BUILD, but I also get the impression that the tensorflow_core package disappeared with version 2.2. When I build locally with TF 2.4, `site-packages/tensorflow/include` is certainly added to the include path and I can also locate that XLA header under that hierarchy.
> 
> So maybe it would suffice to limit building the XLA support to TF 2.2+?

Thanks for taking a look at it. Right, I built locally with TF2.5 and it include only `.../tensorflow/include` (and no `.../tensorflow_core/include`). So, looks like this build issue will go away with newer TF. That is good.

BTW, I committed [some changes](https://github.com/tensorflow/tensorflow/pull/48806) into XLA to make its schedule work better with the Horovod changes in this PR for performance. So, this PR depends on TF2.5+ anyway.

Let's me see if I can find a way to skip building the new codes if TF version is too old.

horovod/common/common.h
```diff
@@ -153,7 +154,7 @@ namespace common {
 #define JOIN_TENSOR_NAME "join.noname"
 
 // List of supported frameworks.
-enum Framework { TENSORFLOW, PYTORCH, MXNET };
+enum Framework { TENSORFLOW, PYTORCH, MXNET, XLA };
```
kuizhiqing, line 157: It seems that XLA is implemented only for tensorflow, why XLA declare here as a framework ? So does the design of cpp namespace level. Thx for your explain in advance.

trentlo: Thanks for the question. My rationale is explained as below:
1. XLA is independent of Tensorflow, although they are in the same repo now. There are talks about separating XLA from Tensorflow although I don't know when that will happen. In addition, XLA (as a DL compiler) is used by frontends other than Tensorflow, such as JAX and Pytorch (not in its main repo, I guess), etc.
2. I certainly can understand that it is inaccurate to say XLA is a framework. However, as XLA is independent of Tensorflow, it does not use Tensorflow constructs and _cannot_ use [TFOpContext](https://github.com/horovod/horovod/blob/0a421945667ab0d914db588bcd48134f48dd0f05/horovod/tensorflow/mpi_ops.cc#L171). It then needs to create a new [XLAOpContext ](https://github.com/horovod/horovod/blob/2337d8180a0c52076ac4224c7a35031fa2af726e/horovod/tensorflow/xla_mpi_ops.cc#L441) with a new "framework" name. I don't see alternatives.
3. If the name is really confusing to people, I may suggest rename "framework" to "backend".

trentlo (commented):
Responded. Let me know if this addresses your comments. Thanks!

horovod/common/common.h
kuizhiqing: Thanks for your reply, you are right that XLA is a layer independent of tf or even torch. XLA is a layer tf/torch can leverage on in my opinion, that's why I propose the question. I'm not sure am I right with following consideration,
1. XLA can not be used/run without a framework(tf/torch/mxnet)
2. XLA will be run in the way tf+XLA (the version you are working on), or torch+XLA, or more.

If so, may be we should keep framework declaration and consider to implement XLA without changing it even by add *backend* ?

Overall, your work is wonderful, it's ok to keep this version if we cannot figure out a better one.

trentlo: You are right that XLA itself is not complete as a framework.

How about let's rename `XLA` to `TF_XLA`?
That is, `enum Framework { TENSORFLOW, PYTORCH, MXNET, TF_XLA };`

This is conceptually clean and simple. The only theoretical drawback I can think of for this approach is that you might later have PYTORCH_XLA, etc. due to combination (of framework and XLA) but I doubt that if this will really become a problem in practice (mainly due to few combinations that really exist in the world). Even if this becomes a problem in practice in the future, we will know better how to deal with it at that time given more data points.

I will make the change accordingly if this makes sense to you. Please let me know.

trentlo: You are right that XLA itself is not complete as a framework.

How about let's rename `XLA` to `TF_XLA`?
That is, `enum Framework { TENSORFLOW, PYTORCH, MXNET, TF_XLA };`

This is conceptually clean and simple. The only theoretical drawback I can think of for this approach is that you might later have PYTORCH_XLA, etc. due to combination (of framework and XLA) but I doubt that if this will really become a problem in practice (mainly due to few combinations that really exist in the world). Even if this becomes a problem in practice in the future, we will know better how to deal with it at that time given more data points.

I will make the change accordingly if this makes sense to you. Please let me know.

trentlo (commented):
Responded. I will act according to your response.

horovod/common/common.h
kuizhiqing: Great, we have clarity on the issues but not achieve a solution yet, maybe we should discuss it with the community.

/cc @maxhgerlach  @EnricoMi

romerojosh (approved):
Thanks for the contribution @trentlo! These XLA updates look quite nice, and it will be good to see the implementation extended to other operations in the future.

Given that the feature is opt-in via `HOROVOD_ENABLE_XLA_OPS`, seems safe to merge this initial implementation in to me.

trentlo:
> Thanks for the contribution @trentlo! These XLA updates look quite nice, and it will be good to see the implementation extended to other operations in the future.
> 
> Given that the feature is opt-in via `HOROVOD_ENABLE_XLA_OPS`, seems safe to merge this initial implementation in to me.

Thanks for taking a look at it, @romerojosh. I do have a plan to add some more Horovod ops for XLA. :-)

trentlo:
There was 1 build failure because CustomCallSchedule actually requires TF2.6+. @romerojosh, could you hit the CI again? Thanks!

Some details:
CustomCallSchedule exists in TF2.6, e.g., [here](https://github.com/tensorflow/tensorflow/blob/r2.6/tensorflow/compiler/xla/service/hlo.proto#L248).

It did not make into TF2.5, i.e., [no such construct](https://github.com/tensorflow/tensorflow/blob/r2.5/tensorflow/compiler/xla/service/hlo.proto#L248).

I was confused by NV's TF source code because I probably back ported the feature into NV's TF2.5.

trentlo:
@romerojosh, looks like I need your help to hit the CI once again.

There was another linking failure due to missing `-l:_pywrap_tensorflow_internal.so` in an old TF version.

A guard is introduced in cmake to fix it as in https://github.com/horovod/horovod/pull/3053/commits/1a031b302a35d69ff1aaf69e8a4ba7e190b23580.

romerojosh:
Failing test is unrelated. This PR looks good to me. Thanks for the contribution @trentlo!

trentlo:
> Great work! @trentlo is there a minimum TF version required?

TF2.6+ is required.

Issue #2727, opened 2021-03-19
Installation via `pip install -e .` doesn't work on OSX


amogkam:
**Environment:**
1. Framework: (TensorFlow, Keras, PyTorch, MXNet): Torch
2. Framework version: torch v1.8.0
3. Horovod version: Master
4. MPI version: N/A
5. CUDA version: N/A
6. NCCL version: N/A
7. Python version: 3.8.6
8. Spark / PySpark version: N/A
9. OS and version: macOS Catalina 10.15.7
10. GCC version: 11.0.3
11. CMake version: 3.17.3

**Bug report:**
I've cloned the repo and am trying to install horovod master on my Mac like so `HOROVOD_WITH_GLOO=1 HOROVOD_WITHOUT_MPI=1 HOROVOD_WITHOUT_MXNET=1 HOROVOD_WITHOUT_TENSORFLOW=1 pip install --no-cache-dir -U -e .`

But it is failing with the following error
```
Obtaining file:///Users/amog/dev/horovod
Requirement already satisfied: cloudpickle in /Users/amog/dev/ray_lightning_accelerators/lib/python3.8/site-packages (from horovod==0.21.3) (1.6.0)
Requirement already satisfied: psutil in /Users/amog/dev/ray_lightning_accelerators/lib/python3.8/site-packages (from horovod==0.21.3) (5.8.0)
Requirement already satisfied: pyyaml in /Users/amog/dev/ray_lightning_accelerators/lib/python3.8/site-packages (from horovod==0.21.3) (5.3.1)
Requirement already satisfied: cffi>=1.4.0 in /Users/amog/dev/ray_lightning_accelerators/lib/python3.8/site-packages (from horovod==0.21.3) (1.14.5)
Requirement already satisfied: pycparser in /Users/amog/dev/ray_lightning_accelerators/lib/python3.8/site-packages (from cffi>=1.4.0->horovod==0.21.3) (2.20)
Installing collected packages: horovod
  Attempting uninstall: horovod
    Found existing installation: horovod 0.21.3
    Uninstalling horovod-0.21.3:
      Successfully uninstalled horovod-0.21.3
  Running setup.py develop for horovod
    ERROR: Command errored out with exit status 1:
[...]

PR #3059, opened 2021-07-28
Support incremental rebuilds.

### LINKED ISSUE #1520: Incremental build support
**Environment:**
1. Framework: (TensorFlow, Keras, PyTorch, MXNet) TensorFlow
2. Framework version: TF 1.14.0
3. Horovod version: tip of master
4. MPI version:
5. CUDA version: 10.0
6. NCCL version: tip of master
7. Python version: 3.6.8
8. OS and version: Ubuntu 18.04
9. GCC version: 7.4.0

**Checklist:**
2. If your question is about hang, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/running.rst)? N/A
3. If your question is about docker, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/docker.rst)? N/A
4. Did you check if you question is answered in the [troubleshooting guide](https://github.com/horovod/horovod/blob/master/docs/troubleshooting.rst)? Yes.

**Your question:**

Is there any way to do an incremental build of horovod?  I can't figure out a way to build horovod from a local copy of the source code, except through `pip install .`, but that seems to build from scratch every time, regardless of changes to the source code.

rb-determined-ai:

- [ ] Did you write any tests to validate this change? *I tested incremental builds manually*

## Description

Support a `HOROVOD_BUILD_IN_REPO` option that will cause the `cmake` commands to point to the repository for the source directory, and to a subdirectory of the repository for the build directory.  As a result, sequential builds can use pre-existing build artifacts to speed up build time.

On my machine, I see around 10x speedups when altering a C++ file:

|task |time  |
--- | ---
|clean build|130s|
|noop rebuild|6s|
|edit to common/operations.cc|12s|

Fixes #1520.

## Review process to land 

### DIFF
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -55,10 +55,16 @@ From *inside* the Horovod root directory, remove any previous build artifacts an
 .. code-block:: bash
 
     $ rm -rf build/ dist/
-    $ HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 python setup.py install
+    $ HOROVOD_WITH_PYTORCH=1 \
+      HOROVOD_WITH_TENSORFLOW=1 \
+      HOROVOD_BUILD_IN_REPO="$PWD" \
+      python setup.py install
 
 Set ``HOROVOD_WITHOUT_[FRAMEWORK]=1`` to disable building Horovod plugins for that framework.
 This is useful when you’re testing a feature of one framework in particular and wish to save time.
+
+The ``HOROVOD_BUILD_IN_REPO`` enables incremental rebuilds by storing the intermediate build artifacts
+as subdirectories of the source repository, rather than in the temporary pip install directory.
 
 For a debug build with checked assertions etc. replace the invocation of setup.py by:
 
--- a/setup.py
+++ b/setup.py
@@ -86,12 +86,21 @@ class custom_build_ext(build_ext):
             cmake_build_args += ['--'] + make_args
 
         cmake_build_dir = os.path.join(self.build_temp, config)
+        cmake_src_dir = self.extensions[0].cmake_lists_dir
+        repo_dir = os.getenv('HOROVOD_BUILD_IN_REPO')
+        if repo_dir:
+            new_build_dir = os.path.join(repo_dir, os.path.basename(cmake_build_dir))
+            print("replacing", cmake_build_dir, "with", new_build_dir)
+            cmake_build_dir = new_build_dir
+            print("replacing", cmake_src_dir, "with", repo_dir)
+            cmake_src_dir = repo_dir
+
         if not os.path.exists(cmake_build_dir):
             os.makedirs(cmake_build_dir)
 
         # Config and build the extension
         try:
-            subprocess.check_call([cmake_bin, self.extensions[0].cmake_lists_dir] + cmake_args,
+            subprocess.check_call([cmake_bin, cmake_src_dir] + cmake_args,
                                   cwd=cmake_build_dir)
             subprocess.check_call([cmake_bin, '--build', '.'] + cmake_build_args,
                                   cwd=cmake_build_dir)


docs/contributors.rst
```diff
@@ -55,11 +55,17 @@ From *inside* the Horovod root directory, remove any previous build artifacts an
...
-    $ HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 python setup.py install
+    $ HOROVOD_WITH_PYTORCH=1 \
+      HOROVOD_WITH_TENSORFLOW=1 \
+      HOROVOD_BUILD_IN_REPO="$PWD" \
+      python setup.py install
 
 Set ``HOROVOD_WITHOUT_[FRAMEWORK]=1`` to disable building Horovod plugins for that framework.
 This is useful when you’re testing a feature of one framework in particular and wish to save time.
 
+The ``HOROVOD_BUILD_IN_REPO`` enables incremental rebuilds by storing the intermediate build artifacts
```
EnricoMi, line 66: What does "in repo" refer to here? The "source repository", which is the git clone directory?

EnricoMi: I am not forced to provide a path to the sources here, right? Can be any directory, so the name is unnecessarily restrictive.

nvcastet:
@rb-determined-ai For incremental builds, I have used `pip install -e` (see https://newbedev.com/when-would-the-e-editable-option-be-useful-with-pip-install). PIP will use your repo directory for the build and then create a symlink of the repo folder in site-packages. In `-e`, because of the symlink, if you also change the python code in the repo, it gets
Note: I had issues with `-e` if not building all the horovod extensions (framework) but i think we can fix that in setup.py with some conditionals.
So, we could fix the cases where `pip install -e` is not working and document the use of `-e` for development. What do you think?

docs/contributors.rst
rb-determined-ai: Sure.  I'm down for any name you see fit.  Maybe `HOROVOD_PERSISTENT_BUILD_DIR`?

rb-determined-ai:
> @rb-determined-ai For incremental builds, I have used `pip install -e` (see https://newbedev.com/when-would-the-e-editable-option-be-useful-with-pip-install). PIP will use your repo directory for the build and then create a symlink of the repo folder in site-packages. In `-e`, because of the symlink, if you also change the python code in the repo, it takes effect right away.
> Note: I had issues with `-e` if not building all the horovod extensions (framework) but i think we can fix that in setup.py with some conditionals.
> So, we could fix the cases where `pip install -e` is not working and document the use of `-e` for development. What do you think?

I have never successfully gotten `pip install -e` to build, even when the exact same variables passed to `pip install` do build.

I'd be fine if that worked (all I care about is incremental builds) but @tgaddair has said recently that `pip install -e` is specifically [_not_ supported](https://github.com/horovod/horovod/issues/2727#issuecomment-803615893).

"Don't use `pip install -e`" was also the result of [my discussion topic](https://github.com/horovod/horovod/discussions/2841) when I couldn't get horovod to compile.

docs/contributors.rst
rb-determined-ai: Wait, actually, unless you also specify the `--use-feature=in-tree-build` feature, you actually do need the source to the repository in order for this to work, because if you point cmake at the copied source code directory that pip creates, I think you won't get the right timestamps and you'll always be rebuilding everything anyway.

nvcastet:
@rb-determined-ai Let me get back to you on `pip install -e` after i retried it.
@tgaddair Did you think there were hard-blockers to make `pip install -e` work?

docs/contributors.rst
EnricoMi: maybe something like `HOROVOD_SOURCE_DIR`

nvcastet:
@rb-determined-ai  Could you look at https://github.com/horovod/horovod/pull/3074 and see if that works for you?

rb-determined-ai:
@nvcastet works for me!

PR #3060, opened 2021-07-28
launch_gloo() kills all workers if any exit non-zero.

rb-determined-ai:

- [ ] Did you update the docs? **no**
- [ ] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users? **no**

## Description

Previously, the following `train.py` script:

```python
import time

if __name__ == "__main__":
    hvd.init()
    print('hi!', hvd.rank())
    time.sleep(1)
    if hvd.rank() == 2:
        raise RuntimeError("worker absolutely refuses to be rank 2")
    time.sleep(120)
    print('exiting normally after 2 minutes')
```

when invoked like this:
```
horovodrun -np 4 -H localhost:4 python test.py
```

Would result in an output like this:

```
[0]<stdout>:hi! 0
[3]<stdout>:hi! 3
[1]<stdout>:hi! 1
[2]<stdout>:hi! 2
[2]<stderr>:Traceback (most recent call last):
[2]<stderr>:  File "/home/rb/code/horovod/meh/test.py", line 11, in <module>
[2]<stderr>:    raise RuntimeError("worker absolutely refuses to be rank 2")
[2]<stderr>:RuntimeError: worker absolutely refuses to be rank 2
Process 2 exit with status code 1.

< two minutes of silence >

[3]<stdout>:exiting normally after 2 minutes
[0]<stdout>:exiting normally after 2 minutes
[1]<stdout>:exiting normally after 2 minutes
  File "/home/rb/miniconda3/envs/hvd/bin/horovodrun", line 8, in <module>
    sys.exit(run_commandline())
  File "/home/rb/miniconda3/envs/hvd/lib/python3.9/site-packages/horovod/runner/launch.py", line 770, in run_commandline
    _run(args)
  File "/home/rb/miniconda3/envs/hvd/lib/python3.9/site-packages/horovod/runner/launch.py", line 760, in _run
    return _run_static(args)
  File "/home/rb/miniconda3/envs/hvd/lib/python3.9/site-packages/horovod/runner/launch.py", line 617, in _run_static
    _launch_job(args, settings, nics, command)
  File "/home/rb/miniconda3/envs/hvd/lib/python3.9/site-packages/horovod/runner/launch.py", line 730, in _launch_job
    run_controller(args.use_gloo, gloo_run_fn,
  File "/home/rb/miniconda3/envs/hvd/lib/python3.9/site-packages/horovod/runner/launch.py", line 706, in run_controller
    gloo_run()
  File "/home/rb/miniconda3/envs/hvd/lib/python3.9/site-packages/horovod/runner/launch.py", line 722, in gloo_run_fn
    gloo_run(settings, nics, env, driver_ip, command)
  File "/home/rb/miniconda3/envs/hvd/lib/python3.9/site-packages/horovod/runner/gloo_run.py", line 284, in gloo_run
    launch_gloo(command, exec_command, settings, nics, env, server_ip)
  File "/home/rb/miniconda3/envs/hvd/lib/python3.9/site-packages/horovod/runner/gloo_run.py", line 268, in launch_gloo
    raise RuntimeError('Horovod detected that one or more processes exited with non-zero '
RuntimeError: Horovod detected that one or more processes exited with non-zero status, thus causing the job to be terminated. The first process to do so was:
Process name: 2
Exit code: 1
```

With this fix, the same script exits in a matter of seconds.

## Review process to land 

### DIFF
--- a/horovod/runner/gloo_run.py
+++ b/horovod/runner/gloo_run.py
@@ -130,6 +130,10 @@ def _exec_command_fn(settings):
     :return:
     :rtype:
     """
+
+    # Non-elastic gloo runs should terminate all workers when any fail.
+    terminate_all_event = None if settings.elastic else threading.Event()
+
     def _exec_command(command, slot_info, events):
         index = slot_info.rank
         host_name = slot_info.hostname
@@ -162,12 +166,18 @@ def _exec_command_fn(settings):
             stdout = MultiFile([sys.stdout, stdout_file])
             stderr = MultiFile([sys.stderr, stderr_file])
 
+        all_events = []
+        if events:
+            all_events += events
+        if terminate_all_event:
+            all_events += [terminate_all_event]
+
         try:
             exit_code = safe_shell_exec.execute(command,
                                                 index=index,
                                                 stdout=stdout,
                                                 stderr=stderr,
-                                                events=events,
+                                                events=all_events,
                                                 prefix_output_with_timestamp=settings.prefix_output_with_timestamp)
             if exit_code != 0:
                 print('Process {idx} exit with status code {ec}.'.format(idx=index, ec=exit_code))
@@ -180,6 +190,10 @@ def _exec_command_fn(settings):
                 stdout_file.close()
             if stderr_file:
                 stderr_file.close()
+        if exit_code != 0 and terminate_all_event:
+            if not any(ev.is_set() for ev in all_events):
+                print('Terminating remaining workers after failure of Process {idx}.'.format(idx=index))
+            terminate_all_event.set()
         return exit_code, time.time()
 
     return _exec_command
--- a/test/integration/test_interactiverun.py
+++ b/test/integration/test_interactiverun.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 
+import time
 import sys
 import unittest
 import warnings
@@ -85,8 +86,12 @@ class InteractiveRunTests(unittest.TestCase):
             rank = hvd.rank()
             if rank == 1:
                 raise RuntimeError()
+            # The other worker waits a while before exiting.
+            time.sleep(120)
 
         assert gloo_built() or mpi_built()
+
+        start = time.time()
 
         if gloo_built():
             with pytest.raises(RuntimeError, match='Horovod detected that one or more processes exited'):
@@ -95,3 +100,6 @@ class InteractiveRunTests(unittest.TestCase):
         if mpi_built():
             with pytest.raises(RuntimeError, match='mpirun failed'):
                 run(fn, np=2, use_mpi=True)
+
+        # The controller should be terminating workers way before the 2-minute delay.
+        assert time.time() - start < 60


chongxiaoc:
@rb-determined-ai  You need to make the same changes for `gloo_runner` in Spark Estimator side.
https://github.com/horovod/horovod/blob/master/horovod/spark/gloo_run.py#L26

rb-determined-ai:
Thanks for the tip @chongxiaoc, I saw the errors but was pretty lost at what was going on.

I have never used spark, so I think I made it so the tests pass but I don't know enough of the what the code is doing to know if I should be trying to use a real `crash_event` or not.

rb-determined-ai:
@chongxiaoc now that I realize the signature of `_exec_command()` is sort of an interface, I have moved the scope of `crash_event` to not alter that interface.

horovod/runner/gloo_run.py
```diff
@@ -280,7 +289,8 @@ def gloo_run(settings, nics, env, server_ip, command):
     # Each thread will use ssh command to launch the job on each remote host. If an
     # error occurs in one thread, entire process will be terminated. Otherwise,
     # threads will keep running and ssh session.
-    exec_command = _exec_command_fn(settings)
+    crash_event = threading.Event()
```
EnricoMi, line 292: Why are we passing this event into the `_exec_command_fn` method when we are not using it from the outside? We can easily create this event inside the method, which would keep the signature of this method unchanged and clean-up the changes inside the method a bit.

Enabling this behaviour could be controlled via settings. Should that even be the default / only useful behaviour?

@tgaddair I guess for Elastic Horovod, this needs to be turned off.

horovod/runner/gloo_run.py
```diff
@@ -180,6 +186,9 @@ def _exec_command(command, slot_info, events):
                 stdout_file.close()
             if stderr_file:
                 stderr_file.close()
+        if exit_code != 0 and crash_event and not any(ev.is_set() for ev in all_events):
```
EnricoMi, line 189: Does it hurt to always set the `crash_event`, even if other events are set already`. We are not having such a logic for any of the other events.

horovod/runner/gloo_run.py
```diff
@@ -180,6 +186,9 @@ def _exec_command(command, slot_info, events):
...
+            print('Terminating remaining workers after failure of Process {idx}.'.format(idx=index))
```
EnricoMi, line 190: Can we call the event `abnormal_termination_event` or `non_zero_exit_event`? Crash is short but maybe not expressing what it encodes.

EnricoMi (changes requested):
Good idea to kill all jobs when one fails and not wait for the other jobs to find our when their communication fails. But only, if there is no way for it to recover, so I guess for elastic horovod, this should be disabled.

horovod/runner/gloo_run.py
rb-determined-ai: Sure, pick a name and I'll change it.  Maybe  `worker_failed_event`?

horovod/runner/gloo_run.py
rb-determined-ai: It doesn't hurt to set it, but it is very annoying to see this terminating message N times, since after one fails non-zero, they will all fail nonzero.

It would also be annoying if this message was printed after hitting ctrl+c, or something like that, because it would hide the real reason why things were terminating.  That is the purpose of checking all the other events.

From the user's perspective, it's very nice to know _why_ things are shutting down, and only emitting this message when we know it's true is very helpful.

horovod/runner/gloo_run.py
rb-determined-ai: > which would keep the signature of this method unchanged

Is this `_exec_command_fn` called more than the two times in this one file?  It seems to me like it is not.

> Enabling this behaviour could be controlled via settings

When I brought this up in slack, everybody thought this _was_ the only behavior.  The other behavior is just asking for hung workloads, I don't know why you would want that.

> I guess for Elastic Horovod, this needs to be turned off.

Isn't that already the case?  The codepath for `gloo_run_elastic`/`launch_gloo_elastic` should be unchanged.

> Why are we passing this event into the _exec_command_fn method when we are not using it from the outside?

I targeted the smallest change possible.  The only thing that seems broken is non-elastic gloo.  Is changing the `settings` something that other backends would have to accomodate?  I wasn't sure, so I avoided it.  But you're right, it could just as easily be a `def _execute_command_fn(settings, exit_after_worker_failure=False):`.

EnricoMi: > Isn't that already the case? The codepath for `gloo_run_elastic`/`launch_gloo_elastic` should be unchanged.

You are right, Elastic Horovod uses its own `_exec_command_fn`.

Then this is the only useful behaviour, great!

Can you move that event into the `_exec_command_fn`? This will clean up things there as well (e.g. the `if crash_event:`).

horovod/runner/gloo_run.py
EnricoMi: What about this logic then:
```
if exit_code != 0:
  if not crash_event.is_set():
    print('Terminating remaining workers after failure of Process {idx}.'.format(idx=index))
  crash_event.set()
```

rb-determined-ai: You would still get incorrect messages when any of the other events are already set, such as hitting `ctrl+c` (currently the only other way to have one of those events set).

In that case, it would be very confusing to tell the user that you are terminating workers as a result of a failed worker, because really the workers were already terminating due to some other event.

horovod/runner/gloo_run.py
EnricoMi: Let's call it `terminate_all_event`, so it describes its effect, not its cause.

horovod/runner/gloo_run.py
```diff
@@ -280,7 +292,7 @@ def gloo_run(settings, nics, env, server_ip, command):
     # Each thread will use ssh command to launch the job on each remote host. If an
     # error occurs in one thread, entire process will be terminated. Otherwise,
     # threads will keep running and ssh session.
-    exec_command = _exec_command_fn(settings)
+    exec_command = _exec_command_fn(settings, terminate_on_worker_failure=True)
```
EnricoMi, line 295: So didn't we say this is the only useful behaviour? Then we don't need to make this configurable from outside `_exec_command_fn`.

rb-determined-ai: Sorry, this is the only useful behavior _for non-elastic gloo_.  Elastic gloo uses the same `_exec_command_fn` and while I don't have experience running elastic gloo, I'm assuming that it's best not to change that codepath.

EnricoMi: Alright, I checked, the `_exec_command_fn` is used by elastic horovod as well.

You can use `if not settings.elastic:` in `_exec_command_fn` to determine if `terminate_all_event` should be used.

Then, on non-zero exit code we can do:
```
if exit_code != 0 and terminate_all_event:
  if not any(ev.is_set() for ev in all_events):
    print('Terminating remaining workers after failure of Process {idx}.'.format(idx=index))
  terminate_all_event.set()
```

rb-determined-ai: Sure, I can use `settings.elastic`.

horovod/runner/gloo_run.py
```diff
@@ -130,6 +130,11 @@ def _exec_command_fn(settings):
     :return:
     :rtype:
     """
+
+    # Non-elastic gloo runs should terminate all workers when any fail.
+    if not settings.elastic:
```
EnricoMi, line 135: I think the `settings.elastic` should be checked only once and `terminate_all_event` set accordingly (`None` or `threading.Event()`). Then further down only check `if terminate_all_event:` and not `settings.elastic`, because the true-clause accesses `terminate_all_event` and not `settings.elastic`.

horovod/runner/gloo_run.py
```diff
@@ -162,12 +167,18 @@ def _exec_command(command, slot_info, events):
             stdout = MultiFile([sys.stdout, stdout_file])
             stderr = MultiFile([sys.stderr, stderr_file])
 
+        all_events = []
+        if events:
+            all_events += events
+        if not settings.elastic:
```
EnricoMi, line 173:
```suggestion
        if terminate_all_event:
```

horovod/runner/gloo_run.py
```diff
@@ -180,6 +191,10 @@ def _exec_command(command, slot_info, events):
                 stdout_file.close()
             if stderr_file:
                 stderr_file.close()
+        if exit_code != 0 and not settings.elastic:
```
EnricoMi, line 194:
```suggestion
        if exit_code != 0 and terminate_all_event:
```

EnricoMi (changes requested):

horovod/runner/gloo_run.py
rb-determined-ai: these seem 100% equivalent to me, but sure.

EnricoMi: it is logically equivalent but cleaner code

EnricoMi (approved):

chongxiaoc (approved):

chongxiaoc:
@tgaddair Can you take a look?

tgaddair (approved):
LGTM!

PR #3064, opened 2021-07-29
Retry macOS build and test step

EnricoMi:
Sometimes macOS test jobs fail without test failures: https://github.com/horovod/horovod/runs/3190715016?check_suite_focus=true#step:3:2406

Rerun those as we do for Linux CPU and GPU tests.


EnricoMi:
Ignore the incomplete test results above, results can be found here: https://github.com/horovod/horovod/runs/3197136312

tgaddair (approved):

PR #3066, opened 2021-07-30
Make test_mxnet.py fit for MXNet 2.0

barry-jin:
## Checklist before submitting

- [x] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)?
- [ ] Did you update the docs?
- [ ] Did you write any tests to validate this change?  
- [ ] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users?

## Description

Fixes #3061.

## Review process to land 

1. All tests and other checks must succeed.
2. At least one member of the [technical steering committee](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md) must review and approve.
3. If any member of the technical steering committee requests changes, they must be addressed.

PR #3071, opened 2021-08-03
Reworking MXTest to support 1.x and 2.x tests

EnricoMi:
This reworks the MXTests class in `test/parallel/test_mxnet.py` to cleanly distinguish MXNet 1.x and 2.x tests, as well as tests that work for both. This contributes to / includes #3066.


EnricoMi:
Moved branch into horovod repo to get better test comparison: #3072.

PR #3074, opened 2021-08-03
Fix develop/editable install mode

### LINKED ISSUE #1520: Incremental build support

nvcastet:

## Description
Fixes develop/editable mode to support direct changes in the source code and incremental C++/CUDA builds.

Note: On the first rebuild,  I have noticed the CUDA code being rebuilt when not modified. It is potentially an issue with the add_cuda_library cmake code that generates the dependency file. On the second rebuild, the issue disappears.
This issue may disappear with using cudatoolkit (newer version of cmake) instead of cuda cmake module. Something we could try.
@romerojosh FYI ^^
Fixes #1520.

## Review process to land 

### DIFF
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -69,6 +69,11 @@ For a debug build with debug symbols, checked assertions etc., use the extra fla
 
     $ HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 pip install --global-option build_ext --global-option --debug -v -e .
 
+To uninstall Horovod installed in dev mode, run the below command from the Horovod root directory:
+
+.. code-block:: bash
+
+    $ python setup.py develop -u
 
 Testing
 -------
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -45,19 +45,18 @@ From *inside* the Horovod root directory, install Horovod in develop/editable mo
 
 Set ``HOROVOD_WITHOUT_[FRAMEWORK]=1`` to disable building Horovod plugins for that framework.
 This is useful when you’re testing a feature of one framework in particular and wish to save time.
+
 Set ``HOROVOD_WITH_[FRAMEWORK]=1`` to generate an error if the Horovod plugin for that framework failed to build.
+
 Set ``HOROVOD_GPU_OPERATIONS=NCCL`` to run on GPUs with NCCL.
+
+Set ``HOROVOD_DEBUG=1`` for a debug build with checked assertions, disabled compiler optimizations etc.
+
 You can install optional dependencies defined in `setup.py <https://github.com/horovod/horovod/blob/master/setup.py>`__ by adding brackets
 at the end of the command line e.g. ``[test]`` for test dependencies.
 
 In develop mode, you can edit the Horovod source directly in the repo folder. For Python code, the changes will take effect
 immediately. For **C++/CUDA code**, the ``... pip install -v -e .`` command needs to be invoked again to perform an incremental build.
-
-For a debug build with debug symbols, checked assertions etc., use the extra flags below:
-
-.. code-block:: bash
-
-    $ HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 pip install --global-option build_ext --global-option --debug -v -e .
 
 Testing
 -------
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -20,7 +20,7 @@ Develop within a virtual environment to avoid dependency issues:
 
 .. code-block:: bash
 
-    $ virtualenv env
+    $ python3 -m venv env
     $ . env/bin/activate
 
 We recommend installing package versions that match with those under test in
@@ -28,11 +28,6 @@ We recommend installing package versions that match with those under test in
 The following versions are recommended (see default versions defined through :code:`ARG` in
 `Dockerfile.test.cpu <https://github.com/horovod/horovod/blob/master/Dockerfile.test.cpu>`__ and
 `Dockerfile.test.gpu <https://github.com/horovod/horovod/blob/master/Dockerfile.test.gpu>`__ file.
-The versions with CPU support can be installed through the provided :code:`setup.py` file:
-
-.. code-block:: bash
-
-    pip install .[dev,test]
 
 You can find all other non-Python packages that need to be installed on your system for Horovod to build
 in the `Dockerfile.test.cpu <https://github.com/horovod/horovod/blob/master/Dockerfile.test.cpu>`__ and
@@ -42,25 +37,20 @@ Specifically, see all :code:`RUN apt-get install` lines.
 Build and Install
 -----------------
 
-First, uninstall any existing version of Horovod.  Be sure to do this *outside* the Horovod root directory:
+From *inside* the Horovod root directory, install Horovod in develop/editable mode:
 
 .. code-block:: bash
 
-    $ pip uninstall -y horovod
-
-From *inside* the Horovod root directory, remove any previous build artifacts and then install Horovod in dev mode:
-
-.. code-block:: bash
-
-    $ rm -rf build/ dist/
     $ HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 pip install -v -e .
 
 Set ``HOROVOD_WITHOUT_[FRAMEWORK]=1`` to disable building Horovod plugins for that framework.
 This is useful when you’re testing a feature of one framework in particular and wish to save time.
 Set ``HOROVOD_WITH_[FRAMEWORK]=1`` to generate an error if the Horovod plugin for that framework failed to build.
 Set ``HOROVOD_GPU_OPERATIONS=NCCL`` to run on GPUs with NCCL.
+You can install optional dependencies defined in `setup.py <https://github.com/horovod/horovod/blob/master/setup.py>`__ by adding brackets
+at the end of the command line e.g. ``[test]`` for test dependencies.
 
-In dev mode, you can edit the Horovod source directly in the repo folder. For Python code, the changes will take effect
+In develop mode, you can edit the Horovod source directly in the repo folder. For Python code, the changes will take effect
 immediately. For **C++/CUDA code**, the ``... pip install -v -e .`` command needs to be invoked again to perform an incremental build.
 
 For a debug build with debug symbols, checked assertions etc., use the extra flags below:
@@ -68,12 +58,6 @@ For a debug build with debug symbols, checked assertions etc., use the extra fla
 .. code-block:: bash
 
     $ HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 pip install --global-option build_ext --global-option --debug -v -e .
-
-To uninstall Horovod installed in dev mode, run the below command from the Horovod root directory:
-
-.. code-block:: bash
-
-    $ python setup.py develop -u
 
 Testing
 -------
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -48,12 +48,13 @@ This is useful when you’re testing a feature of one framework in particular an
 
 Set ``HOROVOD_WITH_[FRAMEWORK]=1`` to generate an error if the Horovod plugin for that framework failed to build.
 
-Set ``HOROVOD_GPU_OPERATIONS=NCCL`` to run on GPUs with NCCL.
-
 Set ``HOROVOD_DEBUG=1`` for a debug build with checked assertions, disabled compiler optimizations etc.
+
+Other environmental variables can be found in the `install documentation <https://github.com/horovod/horovod/blob/master/docs/install.rst#environment-variables>`__.
 
 You can install optional dependencies defined in `setup.py <https://github.com/horovod/horovod/blob/master/setup.py>`__ by adding brackets
 at the end of the command line e.g. ``[test]`` for test dependencies.
+If you have not installed specific DL frameworks yet, add ``[dev]`` to install the CPU version of all supported DL frameworks.
 
 In develop mode, you can edit the Horovod source directly in the repo folder. For Python code, the changes will take effect
 immediately. For **C++/CUDA code**, the ``... pip install -v -e .`` command needs to be invoked again to perform an incremental build.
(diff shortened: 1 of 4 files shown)


docs/contributors.rst
```diff
@@ -69,6 +69,11 @@ For a debug build with debug symbols, checked assertions etc., use the extra fla
 
     $ HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 pip install --global-option build_ext --global-option --debug -v -e .
 
+To uninstall Horovod installed in dev mode, run the below command from the Horovod root directory:
+
+.. code-block:: bash
+
+    $ python setup.py develop -u
```
EnricoMi, line 76: How does this relate to `pip install .[dev,test]` above? Should we mention either or both?

nvcastet: @EnricoMi this uninstall method is when you install the package in dev mode `-e` since the standard `pip uninstall` does not work in that case.

The part about running `pip install .[dev,test]` is a bit redundant. Should we just remove it and modify line 56 with:
```
HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 pip install -v -e .[dev,test]
```
So when installing in dev mode, it checks for dev and test dependencies?

EnricoMi: I misread your comment in the changes, `python setup.py develop -u` is to uninstall an earlier dev installation. What is wrong about `pip uninstall horovod`?

EnricoMi: The `[dev]` adds a specific version for all ML frameworks to the dependencies and `[test]` adds test-specific dependencies as well. So either users install the frameworks in the desired version or simply use `[dev]` to get recommended versions.

nvcastet: `pip uninstall horovod` did not work with the editable mode. But i just tried it using a virtual environment (instead of just the system environment), and it works. I will update the doc since we recommend a venv.
I can mention the requirements and remove the first call to `pip install`.

docs/contributors.rst
```diff
@@ -42,30 +37,27 @@ Specifically, see all :code:`RUN apt-get install` lines.
...
+Set ``HOROVOD_GPU_OPERATIONS=NCCL`` to run on GPUs with NCCL.
+You can install optional dependencies defined in `setup.py <https://github.com/horovod/horovod/blob/master/setup.py>`__ by adding brackets
+at the end of the command line e.g. ``[test]`` for test dependencies.
 
-For a debug build with checked assertions etc. replace the invocation of setup.py by:
+In develop mode, you can edit the Horovod source directly in the repo folder. For Python code, the changes will take effect
+immediately. For **C++/CUDA code**, the ``... pip install -v -e .`` command needs to be invoked again to perform an incremental build.
 
-.. code-block:: bash
+For a debug build with debug symbols, checked assertions etc., use the extra flags below:
```
maxhgerlach, line 56: > debug symbols

I believe the standard build includes debug symbols as well: The CMake config is set to `RelWithDebInfo` and that comes with the compile flag `-g`.

docs/contributors.rst
```diff
@@ -42,30 +37,27 @@ Specifically, see all :code:`RUN apt-get install` lines.
...
 
-    $ HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 python setup.py build_ext --debug install
+.. code-block:: bash
 
+    $ HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 pip install --global-option build_ext --global-option --debug -v -e .
```
maxhgerlach, line 60: I just tried to do an editable debug build with a variation of that command line:

```
HOROVOD_WITH_MPI=1 HOROVOD_WITH_PYTORCH=1 HOROVOD_WITHOUT_MXNET=1 HOROVOD_WITHOUT_TENSORFLOW=1 HOROVOD_GPU_OPERATIONS=NCCL pip install --global-option build_ext --global-option --debug -v -e .
```

This seems to have lead to two invocations of cmake:
1. `Running setup.py develop for horovod`, proceeding to do a debug build
2. `running build_ext`, proceeding to compile everything in _RelWithDebInfo_ mode

Then there was a weird error message:
```
    [100%] Built target horovod_cuda_kernels    
    make[1]: Leaving directory '/mnt/data/max/horovod/build/temp.linux-x86_64-3.7/RelWithDebInfo'
    /learndata4/maxDev/horovod-dev-venv/lib/python3.7/site-packages/cmake/data/bin/cmake -E cmake_progress_start /mnt/data/max/horovod/build/temp.linux-x86_64-3.7/RelWithDebInfo/CMakeFiles 0
    error: can't copy 'build/lib.linux-x86_64-3.7/horovod/tensorflow/mpi_lib.cpython-37m-x86_64-linux-gnu.so': doesn't exist or not a regular file
ERROR: Command errored out with exit status 1: /learndata4/maxDev/horovod-dev-venv/bin/python3 -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/mnt/data/max/horovod/setup.py'"'"'; __file__='"'"'/mnt/data/max/horovod/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' build_ext --debug develop --no-deps Check the logs for full command output.
Exception information:
  File "/learndata4/maxDev/horovod-dev-venv/lib/python3.7/site-packages/pip/_internal/cli/base_command.py", line 180, in _main
    status = self.run(options, args)
  File "/learndata4/maxDev/horovod-dev-venv/lib/python3.7/site-packages/pip/_internal/cli/req_command.py", line 205, in wrapper
    return func(self, options, args)
  File "/learndata4/maxDev/horovod-dev-venv/lib/python3.7/site-packages/pip/_internal/commands/install.py", line 402, in run
    pycompile=options.compile,
  File "/learndata4/maxDev/horovod-dev-venv/lib/python3.7/site-packages/pip/_internal/req/__init__.py", line 85, in install_given_reqs
    pycompile=pycompile,
  File "/learndata4/maxDev/horovod-dev-venv/lib/python3.7/site-packages/pip/_internal/req/req_install.py", line 773, in install
    unpacked_source_directory=self.unpacked_source_directory,
  File "/learndata4/maxDev/horovod-dev-venv/lib/python3.7/site-packages/pip/_internal/operations/install/editable_legacy.py", line 46, in install_editable
    cwd=unpacked_source_directory,
  File "/learndata4/maxDev/horovod-dev-venv/lib/python3.7/site-packages/pip/_internal/utils/subprocess.py", line 244, in call_subprocess
    raise InstallationSubprocessError(proc.returncode, command_desc)
pip._internal.exceptions.InstallationSubprocessError: Command errored out with exit status 1: /learndata4/maxDev/horovod-dev-venv/bin/python3 -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/mnt/data/max/horovod/setup.py'"'"'; __file__='"'"'/mnt/data/max/horovod/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' build_ext --debug develop --no-deps Check the logs for full command output.
```

Apparently it tried to copy the library for TensorFlow although I set `HOROVOD_WITHOUT_TENSORFLOW=1`.

I've attached a log file after running that a second time so all the gcc invocations are skipped over.
[build_log.txt](https://github.com/horovod/horovod/files/6938575/build_log.txt)

I don't really understand how options are supposed to be passed around between pip and setup.py... If we can't find a "canonical" way to configure a debug build for `pip -e`, we might just want to control that via a bespoke environment variable like `HOROVOD_DEBUG`.

maxhgerlach (commented):
I love this. This is going to remove a lot of time from the edit-build-debug loop during development.

Default (release) builds went fine for me, but I faced some trouble trying to get an editable debug install.

docs/contributors.rst
nvcastet: @maxhgerlach Yes good point, I forgot about it. :) Let me update. It would also disable compiler optimizations to facilitate debugging.

docs/contributors.rst
nvcastet: Thanks For trying it @maxhgerlach. I did a quick try in debug mode but did not let the build finish. Let me look into it.

nvcastet: @maxhgerlach Try again with `HOROVOD_DEBUG=1`.

maxhgerlach: Thanks @nvcastet, now it works like a charm for me.

maxhgerlach (approved):
LGTM!

docs/contributors.rst
```diff
@@ -42,30 +37,26 @@ Specifically, see all :code:`RUN apt-get install` lines.
...
+Set ``HOROVOD_WITH_[FRAMEWORK]=1`` to generate an error if the Horovod plugin for that framework failed to build.
 
-.. code-block:: bash
+Set ``HOROVOD_GPU_OPERATIONS=NCCL`` to run on GPUs with NCCL.
+
+Set ``HOROVOD_DEBUG=1`` for a debug build with checked assertions, disabled compiler optimizations etc.
 
-    $ HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 python setup.py build_ext --debug install
+You can install optional dependencies defined in `setup.py <https://github.com/horovod/horovod/blob/master/setup.py>`__ by adding brackets
+at the end of the command line e.g. ``[test]`` for test dependencies.
```
EnricoMi, line 57: I would like to explicitly mention `[dev]` here to install all frameworks at a supported version:

```suggestion
If you haven't installed specific ML framework versions yet, install horovod with ``[dev]`` to install it along with all supported ML frameworks.
```

docs/contributors.rst
```diff
@@ -42,30 +37,26 @@ Specifically, see all :code:`RUN apt-get install` lines.
...
+    $ HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 pip install -v -e .
 
 Set ``HOROVOD_WITHOUT_[FRAMEWORK]=1`` to disable building Horovod plugins for that framework.
 This is useful when you’re testing a feature of one framework in particular and wish to save time.
 
-For a debug build with checked assertions etc. replace the invocation of setup.py by:
+Set ``HOROVOD_WITH_[FRAMEWORK]=1`` to generate an error if the Horovod plugin for that framework failed to build.
 
-.. code-block:: bash
+Set ``HOROVOD_GPU_OPERATIONS=NCCL`` to run on GPUs with NCCL.
```
EnricoMi, line 51: I wouldn't reference any env var other than `HOROVOD_WITH[OUT]_[FRAMEWORK]` and `HOROVOD_DEBUG` here, because there are so many. Rather than mentioning the `HOROVOD_GPU_OPERATIONS` env var here, what about adding a link to https://github.com/horovod/horovod/blob/master/docs/install.rst#environment-variables?

EnricoMi (changes requested):

docs/contributors.rst
nvcastet: Yes I can do that. They are CPU only versions, correct?

docs/contributors.rst
nvcastet: I can do that.

docs/contributors.rst
EnricoMi: yes, that's right

docs/install.rst
```diff
@@ -226,6 +226,7 @@ Optional environment variables that can be set to configure the installation pro
 
 Possible values are given in curly brackets: {}.
 
+* ``HOROVOD_DEBUG`` - {1}. Install a debug build of Horovod with checked assertions, disabled compiler optimizations etc.
```
EnricoMi, line 229: is the `{1}.` a copy-paste artefact?

docs/contributors.rst
```diff
@@ -42,30 +37,27 @@ Specifically, see all :code:`RUN apt-get install` lines.
...
+
+Other environmental variables can be found in the `install documentation <https://github.com/horovod/horovod/blob/master/docs/install.rst#environment-variables>`__.
 
-    $ HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_TENSORFLOW=1 python setup.py build_ext --debug install
+You can install optional dependencies defined in `setup.py <https://github.com/horovod/horovod/blob/master/setup.py>`__ by adding brackets
+at the end of the command line e.g. ``[test]`` for test dependencies.
+If you have not installed specific DL frameworks yet, add ``[dev]`` to install the CPU version of all supported DL frameworks.
 
+In develop mode, you can edit the Horovod source directly in the repo folder. For Python code, the changes will take effect
+immediately. For **C++/CUDA code**, the ``... pip install -v -e .`` command needs to be invoked again to perform an incremental build.
```
EnricoMi, line 60: Won't pip skip the installation saying it is already installed?

docs/install.rst
nvcastet: @EnricoMi No, see the line above and the other environmental variables. This is the possible value for the variable.

docs/contributors.rst
nvcastet: not for dev mode.

EnricoMi (approved):
LGTM!

PR #3083, opened 2021-08-05
Add process_set support to Gluon DistributedTrainer.

romerojosh:

## Description

Adds `process_set` support to MXNet `DistributedTrainer`. Also fixes the use of `size()` rather than `process_set.size()` when setting the rescale grad argument for averaging in `DistributedOptimizer`.


maxhgerlach (approved):
Thanks for fixing the gradient scaling there, @romerojosh. 

I don't have experience with Gluon Trainers, but the changes certainly look right. Unit test failures look unrelated.

While we're at it, we might also want to change the `if size() == 1: return` early returns in the two `_do_allreduce()` methods to check `self._process_set.size()`.

PR #3085, opened 2021-08-06
Test Fork CI workflow from a fork

EnricoMi:
Just testing current Fork CI workflow in Horovod master.


maxhgerlach:
The GPU CI on Buildkite was triggered fine for my last PR via the CI (Fork) workflow! It's great to have tests on GPUs. However, it's still a bit opaque:

- The results don't appear as checks on the Github "conversation" page for the pull request.
- On the page of the action run there is no association to the triggering PR, instead some unrelated master commit is mentioned. For example [this one](https://github.com/horovod/horovod/actions/runs/1121614034) is titled "Reworking MXTest to support 1.x and 2.x tests" and it links c92e0cb, but it really built and tested the code from #3097.

Not sure if these are easy to tackle, but I thought it would be good to mention these observations.

EnricoMi:
@maxhgerlach Thanks for the input.

> - The results don't appear as checks on the Github "conversation" page for the pull request.

With "results" you mean the Buildkite check result that links to the Buildkite build, or the unit test results? Can you provide the link to the PR please?

> - On the page of the action run there is no association to the triggering PR, instead some unrelated master commit is mentioned.

Do you mean this:
grafik

This is because the entire workflow is running as a `workflow_run` event, which is triggered when the pull request workflow on the commit succeeds, but itself does not run on the pull request commit but master head.

But in the ["Trigger Buildkite Pipeline" step](https://github.com/horovod/horovod/runs/3305438148?check_suite_focus=true#step:4:12) you can get to the Buildkite build, and there you can see the right commit and branch:

grafik

We have to use `workflow_run` to trigger the Buildkite tests and publish test results. There is no other safe way provided by Github.

maxhgerlach:
Thanks for the detailed explanations, @EnricoMi! 

> With "results" you mean the Buildkite check result that links to the Buildkite build, or the unit test results? Can you provide the link to the PR please?

What I meant here is the display on the PR "conversation" page (in this case for #3097), which has been my access gate to the CI results so far. Ideally the results from the `CI (Fork)` would be included here as well, especially for the benefit of new contributors who might not know about the separate workflow. Then again, I don't know anything about Github Actions, and this may just not be feasible with reasonable effort. :)
image

> This is because the entire workflow is running as a workflow_run event, which is triggered when the pull request workflow on the commit succeeds, but itself does not run on the pull request commit but master head.
> 
> But in the "Trigger Buildkite Pipeline" step you can get to the Buildkite build, and there you can see the right commit and branch:
>

Found that eventually myself. I guess it may just be something to get used to.

> 
> We have to use workflow_run to trigger the Buildkite tests and publish test results. There is no other safe way provided by Github.

Security trumps UX polish, of course!

EnricoMi:
The test results are further down, they should include the GPU tests from Buildkite:
grafik

maxhgerlach:
Ah, that's perfect then! Thanks for the clarification and sorry for the bother..

I had scrolled through this artifacts popup, but only saw unit test results for CPU test configurations, then mistakenly assumed the others would not be included. But of course, the number of skipped tests would be much higher if no GPUs were available. 

image

EnricoMi:
You are right, that is really confusing and annoying. The GPU tests are uploaded inside the respective `CI (Fork)` workflow: https://github.com/horovod/horovod/actions/runs/1121614034, probably nothing easy to fix.

PR #3105, opened 2021-08-13
Bump Horovod default build to support C++14.

romerojosh:
## Description

TF nightly builds are currently failing due to requiring C++14 support. I recall in the past that we had some conversation about bumping the base Horovod build to an updated C++ standard, so this PR updates it to C++14.

We can also limit the C++14 support to just TensorFlow, similar to what is already done with PyTorch.


tgaddair (approved):
Thanks for making this fix!

PR #3108, opened 2021-08-16
Add timeout to macOS tests

EnricoMi:
Sometimes, macOS tests (run very long](https://github.com/horovod/horovod/runs/3338155166?check_suite_focus=true). This adds a timeout as we have it for Linux tests.


maxhgerlach:
The last CI results for this branch look broken here (label string formatting, skipped stages), or is this a glitch?

image

EnricoMi:
> The last CI results for this branch look broken

That is Github's way to render matrix jobs when they are skipped. They do not apply the matrix and skip the individual jobs but skip the single job with the templated title. Awesome, right?

maxhgerlach:
> That is Github's way to render matrix jobs when they are skipped. They do not apply the matrix and skip the individual jobs but skip the single job with the templated title. Awesome, right?

Ahh, thank you again for clearing up my confusion! 

The jobs were skipped legitimately, of course ("No code changes, no need to build and test"), and earlier runs of the workflow completed more of the stages as necessary.

maxhgerlach (approved):

PR #3110, opened 2021-08-16
fix torch op handles lazy release which may cause oom in elastic scenario

### LINKED ISSUE #3109: 【Elastic Horovod】torch op handles is not recollect right away as new elastic loop begins
Environment:

Framework: (TensorFlow, Keras, PyTorch, MXNet): Pytorch
Framework version: 1.6.0
Horovod version: 0.21.3
MPI version: 4.0.3
CUDA version: 10.2
NCCL version: 2.7.6
Python version: 3.6
Checklist:

Did you search issues to find if somebody asked this question before? Yes.
If your question is about hang, did you read this doc?
If your question is about docker, did you read this doc?
Did you check if you question is answered in the [troubleshooting guide] (https://github.com/horovod/horovod/blob/master/docs/troubleshooting.rst)? Yes

**Bug report:**
horovod torch currently use a dict to hold references to horovod operation results and release such reference in synchronize function.

```python
def synchronize(handle):
    if handle not in _handle_map:
        return
    try:
        output = _handle_map.pop(handle)[-1]
        return output
    except RuntimeError as e:
        raise HorovodInternalError(e)
```
this code, however, does not releases the handle when exception occurs.as a result, in elastic scenario, handle references may still be retained in the dict for some time in new training loop, and will not be released until the handle counter reaches the same point again. this may be a problem when memory/gpumem is heavily used, prone to oom.

I am going to submit a pr to fix this.

woodlgz:
Signed-off-by: guoze.lin <guozelin@tencent.com>

## Description
This is a PR from JIZHI Team & TaiJi AI platform in Tencent.
Fixes #3109.

## Review process to land 

### DIFF
--- a/horovod/torch/mpi_ops.py
+++ b/horovod/torch/mpi_ops.py
@@ -40,8 +40,8 @@ else:
 _NULL = ""
 
 _basics = _HorovodBasics(__file__, 'mpi_lib_v2')
+
 # import basic methods
-#init = _basics.init
 is_initialized = _basics.is_initialized
 start_timeline = _basics.start_timeline
 stop_timeline = _basics.stop_timeline
@@ -61,6 +61,7 @@ ddl_built = _basics.ddl_built
 ccl_built = _basics.ccl_built
 cuda_built = _basics.cuda_built
 rocm_built = _basics.rocm_built
+
 def shutdown(*args, **kwargs):
     mpi_lib.horovod_torch_reset()
     return _basics.shutdown(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


horovod/torch/mpi_ops.py
```diff
@@ -41,7 +41,7 @@
 
 _basics = _HorovodBasics(__file__, 'mpi_lib_v2')
 # import basic methods
-init = _basics.init
+#init = _basics.init
```
chongxiaoc, line 44: remove this comment line?

chongxiaoc (changes requested):
one liner change.
Overall it looks good to me.
Thanks for fixing it.

tgaddair (approved):
LGTM! Thanks for the fix.

chongxiaoc (approved):

PR #3112, opened 2021-08-16
a better way to handle nccl error under elastic scenario

### LINKED ISSUE #3111: nccl error handling in elastic scenario is buggy
**Environment:**
1. Framework: (TensorFlow, Keras, PyTorch, MXNet)
2. Framework version:
3. Horovod version:
4. MPI version:
5. CUDA version:
6. NCCL version:
7. Python version:
8. Spark / PySpark version:
9. Ray version:
10. OS and version:
11. GCC version:
12. CMake version:

**Checklist:**

**Bug report:**
when nccl communication is enabled in horovod, in elastic scenario, evicting a worker instance may cause nccl communicator abort from other workers.

specifically, there existed 3 problems in this scenario:
1. gpu operations event error check (from a thread-pool thread) just throws away exceptions that will never be handled in which case the program aborts. this behaviour simply goes against elastic horovod's will.
2. in the shadow of problem 1, background loop shutting-down and cleaning nccl resources by calling ncclCommDestroy (when controller detects a shutdown condition, for instance, peers exit) may get into race condition with thread pool error check in which ncclCommAbort will be called, potentially causing a double-free corruption.
3. in elastic eviction scenario, a program of a particular rank somehow failed to detect its nccl communicator broken, is prone to hang reporting 100% gpu utilization.

woodlgz:
Signed-off-by: guoze.lin <guozelin@tencent.com>

## Description
This is a PR from JIZHI Team & TaiJi AI platform in Tencent.
Fixes #3111.

## Review process to land 


woodlgz:
@tgaddair do we need a discussion on this?

tgaddair:
Hey @woodlgz, echoing what Max said, apologies for letting this slip through the cracks while we fixed the errors in CI so we could properly test this on GPU. This looks like a good change that we'll definitely prioritize merging in as soon as possible. Once you rebase, the GPU tests should hopefully run as expected.

PR #3127, opened 2021-08-25
separate prediction from transform

irasit:
## Description
Lightning transformer, separate the model prediction function out of _transform function. So subclass can override the behaivor of how model is called.


chongxiaoc:
@irasit you need to sign off commit

chongxiaoc (approved):

PR #3137, opened 2021-08-28
Added support for extraction of storage options from url.

### LINKED ISSUE #3057: Unable to extract storage_options from URL
Integration of fsspec in Horovod was a great step.

Fsspec support extracting storage_options from input url. In Horovod, the url format is restricted i.e the option of extracting the storage options from url is blocked.

It would be of great help if we could have mechanism to extract the storage_options from the input url similar to fsspec.

For above to work, we need to make changes to _get_fs_and_protocol(self) method in store file. The modified def will look like:
```
def _get_fs_and_protocol(self):
        protocol, path = split_protocol(self.prefix_path)
        cls = fsspec.get_filesystem_class(protocol)
        options = cls._get_kwargs_from_urls(self._dataset_url)
        update_storage_options(options, storage_options)
        fs = cls(**options)
        return fs, protocol
```

manjuransari:
Enable extraction of storage_options from url.
Fixes #3057

### DIFF
--- a/docs/mocks.py
+++ b/docs/mocks.py
@@ -39,7 +39,7 @@ MOCK_MODULES = [
 
     'fsspec',
     'fsspec.core',
-    'fsspec.utils'
+    'fsspec.utils',
 
     'pyarrow',
     'pyarrow.parquet',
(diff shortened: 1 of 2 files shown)


chongxiaoc (approved):

tgaddair (approved):
Nice, thanks for making this change!

PR #3139, opened 2021-08-29
Add barrier call to torch module to support easy synchronization for process sets

### LINKED ISSUE #3121: horovod.common.exceptions.HorovodInternalError: Broadcast is not supported with Join at this time.
**Environment:**
1. Framework: Pytorch (Lightning)
2. Framework version: 1.9.0+cu102 (and 1.4.2 for pytorch lightning)
3. Horovod version: 0.22.1
4. MPI version: 2.1.1
5. CUDA version: 10.2
6. NCCL version: 2708
7. Python version: 3.9.6
8. Spark / PySpark version: -
9. Ray version: -
10. OS and version: Ubuntu 18.04
11. GCC version:  gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
12. CMake version: cmake version 3.10.2

**Bug report:**
When running my training loop using horovod as an accelerator in pytorch lightning I encounter the following error after the 2nd epoch in all of my 8 workers:
```
[0]<stderr>:  File "/usr/local/lib/python3.9/dist-packages/pytorch_lightning/callbacks/model_checkpoint.py", line 816, in file_exists
[0]<stderr>:    return trainer.training_type_plugin.broadcast(exists)
[0]<stderr>:           │       │                              └ False
[0]<stderr>:           │       └ <property object at 0x7fa7182721d0>
[0]<stderr>:           └ <pytorch_lightning.trainer.trainer.Trainer object at 0x7fa708526fa0>
[0]<stderr>:
[0]<stderr>:  File "/usr/local/lib/python3.9/dist-packages/pytorch_lightning/plugins/training_type/horovod.py", line 129, in broadcast
[0]<stderr>:    obj = hvd.broadcast_object(obj, src)
[0]<stderr>:          │   │                │    └ 0
[0]<stderr>:          │   │                └ False
[0]<stderr>:          │   └ <function broadcast_object at 0x7fa72eb0c4c0>
[0]<stderr>:          └ <module 'horovod.torch' from '/usr/local/lib/python3.9/dist-packages/horovod/torch/__init__.py'>
[0]<stderr>:
[0]<stderr>:  File "/usr/local/lib/python3.9/dist-packages/horovod/torch/functions.py", line 218, in broadcast_object
[0]<stderr>:    broadcast_(sz, root_rank, name + '.sz')
[0]<stderr>:    │          │   │          └ 'bool'
[0]<stderr>:    │          │   └ 0
[0]<stderr>:    │          └ tensor([4], dtype=torch.int32)
[0]<stderr>:    └ <function broadcast_ at 0x7fa72eb0a040>
[0]<stderr>:
[0]<stderr>:  File "/usr/local/lib/python3.9/dist-packages/horovod/torch/mpi_ops.py", line 741, in broadcast_
[0]<stderr>:    return synchronize(handle)
[0]<stderr>:           │           └ 69713
[0]<stderr>:           └ <function synchronize at 0x7fa72eb0a4c0>
[0]<stderr>:
[0]<stderr>:  File "/usr/local/lib/python3.9/dist-packages/horovod/torch/mpi_ops.py", line 882, in synchronize
[0]<stderr>:    raise HorovodInternalError(e)
[0]<stderr>:          └ <class 'horovod.common.exceptions.HorovodInternalError'>

[0]<stderr>:horovod.common.exceptions.HorovodInternalError: Broadcast is not supported with Join at this time.
```

I'm not quite sure how to avoid the broadcast in the Join since this is all code internal to pytorch lightning. Do you have any idea what might be causing this? If you think this is an issue with pytorch lightning I'm happy to open an issue there. 
Thanks a lot!

Tixxx:

## Description
Add barrier call to torch module to support easy synchronization for process sets. Existing methods that use other apis are either not well known or not designed to be a synchronization point. This also achieves parity with other distributed training frameworks. 

This also fixes a minor issue with is_initialized(). The horovod_is_initialized returns an atomic bool which will be interpreted as a random INT by ctypes on Mac. Changed it to return INT instead.

Fixes #3121  (issue).
https://github.com/horovod/horovod/issues/3121
## Review process to land 

### DIFF
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -1016,8 +1016,8 @@ void horovod_shutdown() {
   }
 }
 
-bool horovod_is_initialized() {
-  return horovod_global.initialization_done;
+int horovod_is_initialized() {
+  return int(horovod_global.initialization_done.load());
 }
 
 int horovod_start_timeline(const char* file_name, bool mark_cycles) {
@@ -1753,6 +1753,22 @@ Status EnqueueJoin(std::shared_ptr<OpContext> context,
     LOG(TRACE, horovod_global.global_controller->GetRank()) << "Enqueued " << name;
   }
   return status;
+}
+
+// Contexts and controller must be initialized and the background thread
+// must be running before this function is called.
+Status CallBarrier(int32_t process_set_id) {
+  auto& process_set = horovod_global.process_set_table.Get(process_set_id);
+  if (!process_set.IsCurrentProcessIncluded()) {
+    return Status::InvalidArgument(
+        "Barrier: Rank " +
+        std::to_string(horovod_global.global_controller->GetRank()) +
+        " is not a member of the provided process set.");
+  }
+
+  process_set.controller->Barrier(Communicator::GLOBAL);
+  LOG(TRACE, horovod_global.global_controller->GetRank()) << "Released from barrier.";
+  return Status::OK();
 }
 
 } // namespace common
--- a/horovod/common/operations.cc
+++ b/horovod/common/operations.cc
@@ -245,11 +245,12 @@ OperationManager* CreateOperationManager(HorovodGlobalState& state) {
 #endif
 
   std::shared_ptr<JoinOp> join_op(new JoinOp(&state));
+  std::shared_ptr<BarrierOp> barrier_op(new BarrierOp(&state));
   std::shared_ptr<ErrorOp> error_op(new ErrorOp(&state));
 
   return new OperationManager(&state.parameter_manager, allreduce_ops,
                               allgather_ops, broadcast_ops, alltoall_ops,
-                              join_op, adasum_ops, error_op);
+                              join_op, adasum_ops, barrier_op, error_op);
 }
 
 // Process a Response by doing a reduction, a gather, a broadcast, or
@@ -259,7 +260,9 @@ void PerformOperation(Response response, ProcessSet& process_set) {
   auto& timeline = horovod_global.timeline;
   process_set.tensor_queue.GetTensorEntriesFromResponse(response, entries,
                                                         process_set.joined);
-  if (response.response_type() != Response::JOIN) {
+
+  if (response.response_type() != Response::JOIN &&
+      response.response_type() != Response::BARRIER) {
     for (auto& e : entries) {
       timeline.Start(e.tensor_name, response.response_type(), e.tensor->size());
     }
@@ -1757,8 +1760,9 @@ Status EnqueueJoin(std::shared_ptr<OpContext> context,
 
 // Contexts and controller must be initialized and the background thread
 // must be running before this function is called.
-Status CallBarrier(int32_t process_set_id) {
+Status EnqueueBarrier(StatusCallback callback, int32_t process_set_id) {
   auto& process_set = horovod_global.process_set_table.Get(process_set_id);
+
   if (!process_set.IsCurrentProcessIncluded()) {
     return Status::InvalidArgument(
         "Barrier: Rank " +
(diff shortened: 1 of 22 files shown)


chongxiaoc (approved):
LGTM

So Lightning has to use this new op to serve as barrier instead of using `join()`?

Tixxx:
> LGTM
> 
> So Lightning has to use this new op to serve as barrier instead of using `join()`?

@chongxiaoc  Yes, otherwise we will keep bumping into the error reported in the issue. I can take a look on the lightning side as well. As a side note, we need to implement join for other collective operations(allgather, bcast, etc) too to fully support join during training.

chongxiaoc:
@tgaddair @romerojosh Can you take a look?

horovod/common/operations.cc
```diff
@@ -1755,5 +1755,21 @@ Status EnqueueJoin(std::shared_ptr<OpContext> context,
   return status;
 }
 
+// Contexts and controller must be initialized and the background thread
+// must be running before this function is called.
+Status CallBarrier(int32_t process_set_id) {
```
maxhgerlach, line 1760: Is it safe to call this function directly from the PyTorch thread? I fear that this will be problematic at least when single-threaded MPI is used as the controller. Horovod's overall design is laid out such that a single persistent thread handles all the collective communication between Horovod processes.

Then I would expect that `process_set.controller->Barrier(Communicator::GLOBAL);` should be called in the Horovod background thread. Then there would need to be an `EnqueueBarrier()` function etc. as we have for Join and the more complex ops.

This comment in operations.cc explains this a bit:
https://github.com/horovod/horovod/blob/ddd36d5be1ac8071dab74daa2883dca679b65bc8/horovod/common/operations.cc#L358

maxhgerlach (commented):
Hi @Tixxx,

this looks like a very useful contribution and would indeed provide a missing feature which is regularly requested.

I left a comment wondering about possible multithreading issues with the current implementation of this PR.

horovod/common/operations.cc
tgaddair: This is my concern as well. @Tixxx @chongxiaoc can you elaborate on why this isn't handled in the Horovod background thread?

Tixxx: Thanks for pointing this out. I have been mainly using openmpi which has been shipped with thread_multiple by default for a while. looks like we still need to support single threaded cases. It makes sense to move this call to the main thread that calls mpi_init then.

Tixxx: @maxhgerlach @tgaddair  I added a new barrier request to do barrier in the background thread.

Tixxx:
> Hi @Tixxx,
> 
> this looks like a very useful contribution and would indeed provide a missing feature which is regularly requested.
> 
> I left a comment wondering about possible multithreading issues with the current implementation of this PR.

@maxhgerlach  @tgaddair  I added a new barrier request to process the op in the background thread. please take a look. Thanks.

tgaddair (approved):
LGTM! @romerojosh can you also take a quick look?

romerojosh (commented):
Thanks for making updates @Tixxx.  While this looks like it gets the job done, the implementation here looks a bit involved just to implement a barrier operation in the background thread. In torch.distributed, a barrier is implemented by scheduling a 1 element allreduce. If we do something similar, this would greatly simplify the implementation since you can just implement the barrier op directly in Python using existing allreduce operations.

What do you think?

Tixxx:
> Thanks for making updates @Tixxx. While this looks like it gets the job done, the implementation here looks a bit involved just to implement a barrier operation in the background thread. In torch.distributed, a barrier is implemented by scheduling a 1 element allreduce. If we do something similar, this would greatly simplify the implementation since you can just implement the barrier op directly in Python using existing allreduce operations.
> 
> What do you think?

@romerojosh You raised a good point. I had thought about it at the beginning, but after going through c10d's code base, it looks like they only go with allreduce approach for the nccl backend which I think it's due to the lack of barrier api in NCCL. For other backends that expose a barrier api, they call the api directly.
This is where the separation of control plane and computation plane in Horovod comes in handy, we don't need to reply on  backend computation primitives for simple control ops like barrier. Also, the complexity of barrier is simpler compared to allreduce. For example, in open mpi, the complexity of mpi_barrier is almost constant when called on mpi_world communicator. Launching an allreduce call for barrier seems a bit of an overkill to me. This might look involved in the implementation's perspective, but IMO it's actually simpler during runtime.
let me know if this makes sense. Thanks!

romerojosh:
@Tixxx Thanks for the comment. Really, the bulk of the wall time overhead from the barrier would be from the control plane processing, not the actual barrier call. The benefits of a faster barrier op will mostly be hidden by that overhead.
However, as you've already implemented the barrier code in the backend, I'm fine with it.  Please do take a look at the generated flatbuffer files though to ensure things are consistent.

Tixxx:
> @Tixxx Thanks for the comment. Really, the bulk of the wall time overhead from the barrier would be from the control plane processing, not the actual barrier call. The benefits of a faster barrier op will mostly be hidden by that overhead. However, as you've already implemented the barrier code in the backend, I'm fine with it. Please do take a look at the generated flatbuffer files though to ensure things are consistent.

@romerojosh thanks a lot for the reminder. I did miss one part in the fbs file. updated to reflect the correct schema.

PR #3146, opened 2021-09-02
Fix the mapping btw pyspark and numpy

voe09:

## Description

Fixes # (issue).
Horovod maps pyspark string type to numpy uint8 type. It blocks users using string type tensor input. I think originally the reason why we map string type to numpy uint8 is because tensorflow does not support string label. But with the mapping, horovod KerasModel would fail in transform stage anyway. And keras itself does not support string label (https://stackoverflow.com/questions/55970851/tensorflow-model-fit-does-not-work-with-strings-as-labels). We don't need to use the incorrect mapping to fail the transform

## Review process to land 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Fix Horovod develop/editable install mode and incremental builds. ([#3074](https://github.com/horovod/horovod/pull/3074))
 - Estimator/Lightning: use lightning datamodule ([#3084](https://github.com/horovod/horovod/pull/3084))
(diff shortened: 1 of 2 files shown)


chongxiaoc:
@voe09 
Surprisingly this function is not used in horovod? Which example did you use to catch this?

```
(hvd-py3.7) chongxiaoc@chongxiaoc-C02DL7WPMD6R horovod % ag data_type_to_numpy
horovod/spark/common/util.py
127:def data_type_to_numpy(dtype):
```

@tgaddair Do you remember a reason to use `uint8` for string type?

chongxiaoc:
@voe09  Just saw your description now I get it.

tgaddair (approved):

chongxiaoc:
@voe09 plz sign off so we can merge this.

PR #2374, opened 2020-10-14
[WIP] Add Intel MPI support in horovodrun

shirosankaku:
Signed-off-by: Yana Shchyokotova <yana.shchyokotova@intel.com>

## Checklist before submitting

- [+ ] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)?
- [ -] Did you update the docs?
- [ +] Did you write any tests to validate this change?  
- [ -] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users?

WIP. Should not be merged.

PR #3148, opened 2021-09-04
fix MPICH implementation

### LINKED ISSUE #1637: Support MPICH when using horovodrun
I'm using MPICH instead of OpenMPI. 
I'm able to train models using mpiexec and mpirun on my cluster but when I try to use horovodrun, it fails with the following error

```python
[ec2-user@master ~]$python3 ./horovodrun -np 8 python3 <train command>
[mpiexec@master.amazon.com] match_arg (utils/args/args.c:159): unrecognized argument allow-run-as-root
[mpiexec@master.amazon.com] HYDU_parse_array (utils/args/args.c:174): argument matching returned error
[mpiexec@master.amazon.com] parse_args (ui/mpich/utils.c:1596): error parsing input array
[mpiexec@master.amazon.com] HYD_uii_mpx_get_parameters (ui/mpich/utils.c:1648): unable to parse user arguments
[mpiexec@master.amazon.com] main (ui/mpich/mpiexec.c:153): error parsing parameters
```

Horovod version: 0.18.2
MPICH version: 3.2

From looking at it, it looks like certain args are not available in MPICH.
Is horovodrun compatible with MPICH?

njzjz:

## Description

Fixes #1637.

I notice horovod's MPICH implementation is broken (#1637) but Intel MPI implementation is good (#2374). Per [Intel MPI's official page](https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/mpi-library.html), it implements the MPICH specification (in fact it's MPICH derivative), so we can safely use Intel MPI's implementation for MPICH.

## Review process to land 

### DIFF
--- a/horovod/runner/mpi_run.py
+++ b/horovod/runner/mpi_run.py
@@ -153,7 +153,7 @@ def mpi_run(settings, nics, env, command, stdout=None, stderr=None):
     if mpi_impl_flags is None:
         raise Exception(_MPI_NOT_FOUND_ERROR_MSG)
 
-    impi = (_IMPI_IMPL == mpi or _MPICH_IMPL == mpi)
+    impi_or_mpich = mpi in (_IMPI_IMPL, _MPICH_IMPL)
 
     ssh_args = []
     if settings.ssh_port:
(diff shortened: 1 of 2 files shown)


chazeon (approved):

horovod/runner/mpi_run.py
```diff
@@ -153,7 +153,7 @@ def mpi_run(settings, nics, env, command, stdout=None, stderr=None):
     if mpi_impl_flags is None:
         raise Exception(_MPI_NOT_FOUND_ERROR_MSG)
 
-    impi = _IMPI_IMPL == mpi
+    impi = (_IMPI_IMPL == mpi or _MPICH_IMPL == mpi)
```
EnricoMi, line 156: Can this logic be rephrased to this, which I think is more readable?
```suggestion
    impi = mpi in (_IMPI_IMPL, _MPICH_IMPL)
```

EnricoMi: I think, `impi` should be renamed to `impi_or_mpich` as this is what it now encodes.

EnricoMi (changes requested):

EnricoMi:
Good catch, thanks for fixing this.

horovod/runner/mpi_run.py
njzjz: Fixed.

EnricoMi (approved):
LGTM!

PR #3155, opened 2021-09-07
Split gpu tests in head and non-head versions

EnricoMi:
Buildkite tests do not complete if some tests fail for head versions. This splits the Buildkite jobs into HEAD and non-HEAD tests so that non-HEAD tests can succeed even if some HEAD tests fail.

This does not fix any broken nightly versions but makes our CI resilient against those as we still have full test coverage over non-nightly versions.


chongxiaoc (approved):

tgaddair (approved):
LGTM!

EnricoMi:
Thanks!

PR #3176, opened 2021-09-23
test_gradient_aggregation with real gradient instead of a constant

### LINKED ISSUE #2664: Fix test_gradient_aggregation for TF 2.4.0
Following #2647, local gradient aggregation will not work on a tf.constant.  We should change this test to use an actual gradient.

ashahab:
This fixes issue #2664 by performing gradient aggregation with a real gradient instead of a constant.

- [ X] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)?
- [ X] Did you update the docs?
- [X ] Did you write any tests to validate this change?  
- [ X] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users?

## Description
PR #2647 shifts the gradient allreduce when the gradient is computed (both through the DistributedOptimizer or through the DistributedGradientTape). Which means that this unittest, by design in TF2.4, doesn't call allreduce in _aggregate_gradients().

Since this unittest provide a gradient as constant (without effectively computing it), the gradient will never be allreduced.
The current change ensure that instead of a constant a real gradient is computed from a loss-function.

Note: The current loss-function intentionally evaluates to zero. A future PR should convert it to a real loss function(e.g. MeanSquaredError) and compute gradients from that to test gradient aggregation.

Fixes #2664.

## Review process to land 

### DIFF
--- a/test/parallel/test_tensorflow2_keras.py
+++ b/test/parallel/test_tensorflow2_keras.py
@@ -31,13 +31,15 @@ from tensorflow.python.keras.optimizer_v2 import optimizer_v2
 import horovod.tensorflow.keras as hvd
 
 
-_PRE_TF_2_4_0 = LooseVersion(tf.__version__) < LooseVersion("2.4.0")
 _PRE_TF_2_2_0 = LooseVersion(tf.__version__) < LooseVersion("2.2.0")
 
-# Set environment variable to enable adding/removing process sets after initializing Horovod.
+# Set environment variable to enable adding/removing process sets after
+# initializing Horovod.
 os.environ["HOROVOD_DYNAMIC_PROCESS_SETS"] = "1"
 
-@pytest.mark.skipif(LooseVersion(tf.__version__) < LooseVersion('2.0.0'), reason='TensorFlow v2 tests')
+
+@pytest.mark.skipif(LooseVersion(tf.__version__) <
+                    LooseVersion('2.0.0'), reason='TensorFlow v2 tests')
 class Tf2KerasTests(tf.test.TestCase):
     """
     Tests for ops in horovod.tensorflow.keras.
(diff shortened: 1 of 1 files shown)


ashahab:
@tgaddair @DEKHTIARJonathan I have a torch test breaking above, but I don't think it's related to this PR. Would love to get your thoughts on this.

test/parallel/test_tensorflow2_keras.py
```diff
@@ -52,13 +54,14 @@ def __init__(self, *args, **kwargs):
...
-            tf.config.experimental.set_visible_devices(gpus[hvd.local_rank()], 'GPU')
+            tf.config.experimental.set_visible_devices(
+                gpus[hvd.local_rank()], 'GPU')
 
     def test_train_model_lr_schedule(self):
         initial_lr = 0.1 * hvd.size()
         opt = tf.keras.optimizers.Adam()
         opt = hvd.DistributedOptimizer(opt)
-        linear_multiplier = lambda epoch: epoch
+        def linear_multiplier(epoch): return epoch
```
tgaddair, line 64: Can we put this as two separate lines?

test/parallel/test_tensorflow2_keras.py
```diff
@@ -234,44 +248,42 @@ def compute_expected_value(batch_id):
...
             # Apply gradient updates in tf.function to reproduce how it is
             # done inside `model.fit()`.
-            hvd_optimizer.apply_gradients(zip(gradient_updates, model_variables), **kwargs)
+            hvd_optimizer.apply_gradients(grads_and_vars, **kwargs)
 
-        gradients = [tf.constant([float(hvd.rank())])]
-        variables = [tf.Variable([0.0])]
+        var = tf.Variable([0.0])
+        variables = [var]
+        def loss(): return (var - var)
```
tgaddair, line 270: Same here, minor style nit, just split this fn into two lines.

tgaddair (commented):
Looks good ,just a couple style changes.

test/parallel/test_tensorflow2_keras.py
ashahab: updated.

test/parallel/test_tensorflow2_keras.py
ashahab: split into two lines

tgaddair (approved):
LGTM! Most likely the torch test is the existing torch-head failure we're investigating.

tgaddair:
@ashahab can you try rebasing master to fix tests?

PR #3196, opened 2021-10-05
Change CMake syntax to be compatible with old versions of CMake

maxhgerlach:

## Description

Ubuntu 16.04 comes with CMake 3.5.1 by default. I tested that with this change Horovod builds successfully with TensorFlow in such a container.

We currently ask for version 2.8.12 or higher in the root `CMakeLists.txt`.


chongxiaoc (approved):
I don't  know old CMake doesn't have `>=`. Thanks for fixing it.

maxhgerlach:
> I don't know old CMake doesn't have >=.

It's pretty surprising sometimes. At least it's improving...

PR #3199, opened 2021-10-05
use trained model if no custom checkpoint call back is specified. 

irasit:

## Description
In lightning estimator, we do not need to load from check point if only want to return the last weights. The model after trianer.fit already have them.

### DIFF
--- a/horovod/spark/lightning/remote.py
+++ b/horovod/spark/lightning/remote.py
@@ -223,8 +223,9 @@ def RemoteTrainer(estimator, metadata, ckpt_bytes, run_id, dataset_idx, train_ro
                     # One more file sync to push profiler result.
                     remote_store.sync(logs_path)
 
-                # rank 0 overwrites model with best checkpoint and returns.
-                best_model = model.load_from_checkpoint(_checkpoint_callback.best_model_path)
+                # Rank 0 overwrites model with best checkpoint and returns.
+                # Need to set strict=False for dynamically added layers.
+                best_model = model.load_from_checkpoint(_checkpoint_callback.best_model_path, strict=False)
                 serialized_checkpoint = io.BytesIO()
                 module = best_model if not is_legacy else best_model._model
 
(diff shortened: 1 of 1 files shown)


horovod/spark/lightning/remote.py
```diff
@@ -128,11 +128,16 @@ def train(serialized_model):
             # Lightning requires to add checkpoint callbacks for all ranks.
             # Otherwise we are seeing hanging in training.
             _checkpoint_callback = checkpoint_callback
```
chongxiaoc, line 130: For simplicity, just relying on this `_checkpoint_callback` variable != None check is enough.
If `_checkpoint_callback` is not None: modify `dir_path` and `filename`, and `callbacks.append(_checkpoint_callback)`

horovod/spark/lightning/remote.py
```diff
@@ -223,8 +228,9 @@ def on_epoch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -
                     # One more file sync to push profiler result.
                     remote_store.sync(logs_path)
 
-                # rank 0 overwrites model with best checkpoint and returns.
-                best_model = model.load_from_checkpoint(_checkpoint_callback.best_model_path)
+                # Rank 0 overwrites model with best checkpoint and returns.
+                best_model = model if use_last_model_weights else model.load_from_checkpoint(_checkpoint_callback.best_model_path, strict=False)
```
chongxiaoc, line 232: best_model = model if not _checkpoint_callback else ...

chongxiaoc (changes requested):
add two nit comments.

horovod/spark/lightning/remote.py
irasit: if _checkpoint_callback is None, it will set to default ModelCheckpoint in line 141.
If user do not pass in a checkpoint callback, we still want to add the default `ModelCheckpoint` in to callback list, right?

horovod/spark/lightning/remote.py
irasit: _checkpoint_callback is always not None because of the line 141

PR #3207, opened 2021-10-08
Test buildkite with tag

EnricoMi:
CI failed when triggering the Buildkite tests for tag v0.23.0. Testing what went wrong.


EnricoMi:
Problem solved. The Buildkite API said that tags are not configured to be built, so I changed the settings in Buildkites to allow building tags. Interestingly, that option was hidden in an disabled part of the settings. Anyway, fixed.

PR #3225, opened 2021-10-15
Job buildkite-heads accesses ci-workflow outputs, add it to the needs

EnricoMi:
The `buildkite-heads` job in `ci-results.yaml` accesses outputs of job `ci-workflow` but does not list it in its `needs`, so the string is empty, which makes the job fail on parsing the empty json string:
https://github.com/horovod/horovod/runs/3897640061?check_suite_focus=true

The `buildkite` job works perfectly well, which has access to `ci-workflow`:
https://github.com/horovod/horovod/runs/3896534243?check_suite_focus=true

### DIFF
--- a/.github/workflows/ci-results.yaml
+++ b/.github/workflows/ci-results.yaml
@@ -112,7 +112,7 @@ jobs:
 
   buildkite-heads:
     name: "Build and Test GPU heads (on Builtkite)"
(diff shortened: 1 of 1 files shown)


EnricoMi:
@maxhgerlach this worked: https://github.com/horovod/horovod/actions/runs/1346323323

PR #3230, opened 2021-10-19
RayExecutor: Dynamic executor for elastic and static jobs

### LINKED ISSUE #3190: Refactor ElasticRayExecutor as part of RayExecutor
Currently, `ElasticRayExecutor` is implemented as a separate class from the normal `RayExecutor`, but this doesn't need to be the case. If we instead add the `ElasticRayExecutor` params to the `RayExecutor` constructor, we can dynamically create an autoscaling worker topology when the user provides `min_workers` or `max_workers`, or any other elastic-only params.

ashahab:
RayExecutor V2: Dynamic executor for elastic and static jobs

This resolves #3190 by adding elastic params to the RayExecutor API for horovod:
This API now supports both static(non-elastic) and elastic horovod jobs.

Example of static job(Identical to current RayExecutor):
```python
from horovod.ray import RayExecutor
ray.init()
hjob = RayExecutor(setting, num_workers=num_workers,
        use_gpu=True
    ))

executor.start()

def simple_fn():
    hvd.init()
    print("hvd rank", hvd.rank())
    return hvd.rank()

result = executor.run(simple_fn)
assert len(set(result)) == hosts * num_slots

executor.shutdown()
```
Example of an elastic job:
```python
from horovod.ray import RayExecutor
import horovod.torch as hvd

def training_fn():
    hvd.init()
    model = Model()
    torch.cuda.set_device(hvd.local_rank())

    @hvd.elastic.run
    def train(state):
        for state.epoch in range(state.epoch, epochs):
            ...
            state.commit()

    state = hvd.elastic.TorchState(model, optimizer, batch=0, epoch=0)
    state.register_reset_callbacks([on_state_reset])
    train(state)
    return

executor = RayExecutor(settings, min_workers=1, use_gpu=True, cpus_per_worker=2)
executor.start()
executor.run(training_fn)
```

### DIFF
--- a/horovod/ray/runner.py
+++ b/horovod/ray/runner.py
@@ -1,19 +1,21 @@
 import ray
 from ray.util.placement_group import get_current_placement_group
 
-import warnings
 from collections import defaultdict
-from dataclasses import dataclass
+from dataclasses import dataclass, asdict
 import os
 from typing import Dict, Callable, Any, Optional, List
 import logging
+import ray.exceptions
+from horovod.ray.adapter import Adapter, BaseParams
 
 from horovod.runner.common.util import secret, timeout, hosts
 from horovod.runner.http.http_server import RendezvousServer
 from horovod.ray.utils import detect_nics, nics_to_env_var, map_blocking
 from horovod.ray.strategy import ColocatedStrategy, PGStrategy
-logger = logging.getLogger(__name__)
+from horovod.ray.elastic_v2 import ElasticParams
 
+logger = logging.getLogger(__name__)
 
 @dataclass
 class MiniSettings:
@@ -28,6 +30,7 @@ class MiniSettings:
     ssh_identity_file: str = None
     timeout_s: int = 300
     placement_group_timeout_s: int = 100
+    elastic: bool = False
 
     @property
     def start_timeout(self):
@@ -126,6 +129,42 @@ class Coordinator:
         }
 
 
+@dataclass
+class StaticParams(BaseParams):
+    """Parameters for non-elastic jobs.
+
+    Args:
+        num_workers (int): Number of workers to use for training.
+        cpus_per_worker (int): Number of CPU resources to allocate to
+            each worker.
+        use_gpu (bool): Whether to use GPU for allocation. TODO: this
+            can be removed.
+        gpus_per_worker (int): Number of GPU resources to allocate to
+            each worker.
+        num_hosts (int): Alternative API to ``num_workers``. Number of
+            machines to execute the job on. Used to enforce equal number of
+            workers on each machine.
+        num_workers_per_host (int): Alternative API to
+            ``num_workers``. Number of workers to be placed on each machine.
+            Used to enforce equal number of workers on each machine. Only
+            used in conjunction with `num_hosts`.
+        use_current_placement_group (bool): Whether to use the current
+            placement group instead of creating a new one. Defaults to True.
+
+    """
+    num_workers: Optional[int] = None
+    num_hosts: Optional[int] = None
+    num_workers_per_host: int = 1
+    use_current_placement_group: bool = True
+
+    @property
+    def elastic(self):
+        return False
+
+    @property
+    def adapter(self):
+        return StaticAdapter
+
 class RayExecutor:
     """Job class for Horovod + Ray integration.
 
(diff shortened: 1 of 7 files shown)


richardliaw (changes requested):
@ashahab Can you please explicitly state what is the delta between the standard RayExecutor and the new RayExecutorV2 in terms of API?

Is it possible to maintain compatibility across the two versions (while being OK with breaking ElasticRayExecutor)?

ashahab:
> @ashahab Can you please explicitly state what is the delta between the standard RayExecutor and the new RayExecutorV2 in terms of API?
> 
> Is it possible to maintain compatibility across the two versions (while being OK with breaking ElasticRayExecutor)?

@richardliaw I see two options here:
Option 1: Keep RayExecutor and RayExecutorV2 API identical. Elastic and non Elastic modes would have different behavior.
I can update the current PR to make RayExecutorV2 identical to RayExecutor, with the difference being when users invoke the elastic API, `start` and `run` methods would change behavior.
The `ElasticRayExecutor` does all of the worker orchestration and execution in the `run` function, whereas the `RayExecutor` performs worker orchestration in `start` and execution in one or more `run` functions.
When users will use the elastic params in `RayExecutorV2`, e.g. `min_np` and `max_np`, `run` would change its behavior.

Option 2: Get rid of the `start` function. This makes elastic and non-elastic modes identical in terms of orchestration and execution.
I had some discussion with @amogkam  on getting rid of the `start` function from RayExecutorV2 completely, which would make the API very different from `RayExecutor`, but the elastic and non-elastic versions both would have orchestration and execution done in the `run` function, and would probably share much more code paths.

Which option is preferred? @tgaddair what's your thought on keeping the current API vs. making the new API identical for elastic and non-elastic

richardliaw:
I think (1) is a safer move given that there are multiple organizations depending on RayExecutor, and introducing instability into the supply chain is highly undesirable.

I'm actually not that worried about the behaviors of "start" being different from "run". That being said, I would be open to user feedback in the future about the APIs being unintuitive. 

I would recommend documenting it clearly though.

tgaddair:
Hey @ashahab, this looks awesome. In general, I think this does a great job of cleanly unifying the elastic and non-elastic use cases. A few general observations based on the above discussion:

1. I agree with @richardliaw that we should try to keep the user-facing `RayExecutor` API unchanged for the non-elastic scenario, since there are organizations using it in production.
2. I also agree that breaking / removing `ElasticRayExecutor` is acceptable (though we should probably start with a deprecation warning), with the plan being to remove it in v1.0.
3. To achieve goals (1) and (2), my suggestion would be to not have the separate `ElasticParams` and `NonElasticParams`. Instead, we use the same set of kwargs for both, and then use the `ElasticAdapter` if `min_np` or `max_np` is set explicitly. Otherwise, we use the existing non-elastic behavior. Alternatively, we can have an `elastic=False` kwarg, I'm fine either way (we just need to do more validation logic with this additional param).
4. Having different behavior for `start()` and `run()` is okay, as Richard said, we should just be clear in the docs what the differences are. For elastic, I think it's fine if `start()` simply creates the driver and begins the discovery process and other background processes, as it does now. In the future, it would also be useful to create the desired set of workers, which could then be updated dynamically as part of the add / removal process. This would help make the elastic API align better with the non-elastic API (so the current set of `workers` can be exposed through the API), but it's not essential at the moment, I think.

There is a useful benefit to `start()`, which is the stateful execution. This is something we make use of in Ludwig, for example:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/backend/ray.py#L304

So to sum up, I think preserving RayExecutor's existing API is important, and it's okay if the elastic mode behaves slightly differently for now, as long as it's documented and there is plan for getting closer alignment in the future (where it makes sense).

@ashahab @richardliaw does this make sense to you, or were there other considerations I missed?

horovod/ray/runner.py
```diff
@@ -360,9 +474,31 @@ def _maybe_call_ray(self, driver_func, *args, **kwargs):
             return driver_func(**kwargs)
 
 
-class _ExecutorDriver:
-    """Base driver for executing Ray calls."""
+class NonElasticAdapter:
```
richardliaw, line 477: non-elastic => Static

horovod/ray/runner.py
```diff
@@ -125,6 +127,98 @@ def establish_rendezvous(self) -> Dict[str, str]:
...
+                             "use_gpu must be True if gpus_per_worker is "
+                             "set. ")
+        if self.use_gpu and isinstance(self.gpus_per_worker,
+                                  int) and self.gpus_per_worker < 1:
+            raise ValueError(
+                f"gpus_per_worker must be >= 1: Got {self.gpus_per_worker}.")
+        self.gpus_per_worker = self.gpus_per_worker or int(self.use_gpu)
+
+@dataclass
+class ElasticParams(BaseParams):
```
richardliaw, line 147: can you move this to the elastic file?

richardliaw (approved):
This diff looks good to me. Can you try merging master again?

PR #3253, opened 2021-11-01
Debugging for lightning data loader and fix for simple profiler.

irasit:
## Description
- Adding debugging logs to lightning data loader.
- skip adding path if profiler is 'simple'
- Make async queue size customizable.

### DIFF
--- a/horovod/data/data_loader_base.py
+++ b/horovod/data/data_loader_base.py
@@ -89,15 +89,18 @@ class AsyncDataLoaderMixin(object):
 
                     c += 1
 
-                    # Force break out if hanging. We assume hanging if already pop enough items from queue.
+                    # Force break out if hanging. We assume hanging if already pop items more than twice of the size of the queue.
                     if c > max(2 * self.async_loader_queue_size, 200):
                         print(f"close_async_loader: ERROR!!! Force break out after {c} times get_nowait.")
                         break
                 except Empty:
                     break
-            print(f"close_async_loader[{self.async_loader_queue_size}], joining...")
+            if self.debug_data_loader:
+                print(f"close_async_loader[{self.async_loader_queue_size}], joining...")
+
             self.thread.join()
-            print(f"close_async_loader[{self.async_loader_queue_size}], joined.")
+
+        print(f"close_async_loader[{self.async_loader_queue_size}] is closed.")
 
     def _async_worker(self):
         """
@@ -126,6 +129,7 @@ class AsyncDataLoaderMixin(object):
             self.queue.put(None)
         finally:
             self.queue.put(None)
+
         print(f"_async_worker[{self.async_loader_queue_size}], stoped")
 
     def __iter__(self):
--- a/horovod/data/data_loader_base.py
+++ b/horovod/data/data_loader_base.py
@@ -75,16 +75,25 @@ class AsyncDataLoaderMixin(object):
         """
         Close the async data loader.
         """
-        print("Closing the AsyncDataLoaderMixin.")
+        print("close_async_loader[{self}:{self.async_loader_queue_size}], Closing the AsyncDataLoaderMixin.")
         if self.async_loader_queue_size > 0 and self.started:
             self.finished_event.set()
+            c = 0
             while True:
                 try:
                     # Drain buffer
-                    self.queue.get_nowait()
+                    batch = self.queue.get_nowait()
+                    msg = f'{batch}'
+                    msg = msg if len(msg) <= 20 else msg[:20]
+                    print(f"close_async_loader[{self}:{self.async_loader_queue_size}], discarding: {msg}")
+                    c += 1
+                    if c == 2000:
+                        break
                 except Empty:
                     break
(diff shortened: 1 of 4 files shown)


horovod/data/data_loader_base.py
```diff
@@ -75,34 +76,57 @@ def close_async_loader(self):
...
                 try:
                     # Drain buffer
                     self.queue.get_nowait()
+                    if self.debug_data_loader:
+                        print(f"close_async_loader[{self.async_loader_queue_size}], discarded batch #{c} from Queue.")
+
+                    c += 1
+
+                    # Force break out if hanging. We assume hanging if already pop enough items from queue.
+                    if c > max(2 * self.async_loader_queue_size, 200):
```
Tixxx, line 93: why multiplying by 2 here?

horovod/spark/lightning/datamodule.py
```diff
@@ -99,7 +101,7 @@ def train_dataloader(self):
             # To avoid loading too much data in memory, need to calculate the queue size
             # dynamicaly, and limit the data loaded in queue.
             # Add 1 in size for storing the None in the end of each epoch.
-            kwargs['async_loader_queue_size'] = max(1, min(100000 // kwargs['batch_size'], kwargs['limit_step_per_epoch'] // 4)) + 1
+            kwargs['async_loader_queue_size'] = max(1, min(2000000 // kwargs['batch_size'], kwargs['limit_step_per_epoch'] // 4)) + 1
```
Tixxx, line 104: How is this 100000 or 2000000 calculated based on? This looks like a significant increase. If this is not dynamically determined, maybe we can define some constants.

irasit: It estimated based on deepETA's data. Only an estimation because the real data size in MB is unknown here.

horovod/data/data_loader_base.py
irasit: If already popped 2 times the queue size,  there must be an issue for stoping the asyn data loading thread.
We should break out instead looping forever.

horovod/spark/lightning/datamodule.py
Tixxx: hmm ok, seems arbitrary. if possible, would be good to pass this in as an argument from our side so that we dont need to modify the horovod part again if anything changes.

Tixxx (approved):

horovod/data/data_loader_base.py
```diff
@@ -75,34 +76,57 @@ def close_async_loader(self):
...
+                        print(f"close_async_loader: ERROR!!! Force break out after {c} times get_nowait.")
+                        break
                 except Empty:
                     break
+            print(f"close_async_loader[{self.async_loader_queue_size}], joining...")
```
chongxiaoc, line 98: make these prints all under `debug_data_loader=True`

horovod/data/data_loader_base.py
```diff
@@ -75,34 +76,57 @@ def close_async_loader(self):
...
+                    print(f"_async_worker[{self.async_loader_queue_size}], finish reading at #{c}, reset debugging counter, append None to queue.")
+                    c = 0
+
                 self.queue.put(None)
         except Exception as ex:
             self.queue.put(ex)
             self.queue.put(None)
         finally:
             self.queue.put(None)
+        print(f"_async_worker[{self.async_loader_queue_size}], stoped")
```
chongxiaoc, line 129: put this print under `debug_data_loader`

horovod/spark/lightning/remote.py
```diff
@@ -207,7 +210,7 @@ def on_epoch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -
             print("Creating trainer with: \n ", kwargs)
             trainer = Trainer(**kwargs)
 
-            if trainer.profiler:
+            if profiler != 'simple' and trainer.profiler:
```
chongxiaoc, line 213: why `simple_profiler` doesn't need to set `dirpath` here?

chongxiaoc (changes requested):
overall it looks good to me. just a few nits to put prints under debug=True and a question of simple profiler.

horovod/spark/lightning/datamodule.py
irasit: Modify the code to let user pass the queue size.

horovod/spark/lightning/remote.py
irasit: it will be easier to just print the simple profiler log to stdout. It is not big, no need to put in file.

horovod/data/data_loader_base.py
irasit: yes, will add this.

horovod/data/data_loader_base.py
irasit: Like to keep this outside, it is only 1 time printing in the end. We saw some data loader hang issue during close, it  helps to detect that.

chongxiaoc (approved):

PR #3258, opened 2021-11-03
Call process_set._setup in init() to point to the correct native lib path

Tixxx:

## Description
In remote trainers of the horovod spark module, if multiple frameworks are imported in the same process, there's a chance that the _basics in horovod.common.process_set is pointing to the wrong shared lib path resulting in horovod not initialized error even after calling hvd.init(). This calls _setup again before fitting the model to make sure the shared lib is correct.

Fixes # (issue).

## Review process to land 

### DIFF
--- a/horovod/spark/keras/remote.py
+++ b/horovod/spark/keras/remote.py
@@ -109,6 +109,7 @@ def RemoteTrainer(estimator, metadata, keras_utils, run_id, dataset_idx):
 
         hvd = get_horovod()
         hvd.init()
+
         pin_gpu(hvd, tf, k)
 
         if not user_shuffle_buffer_size:
@@ -128,6 +129,9 @@ def RemoteTrainer(estimator, metadata, keras_utils, run_id, dataset_idx):
 
         # Verbose mode 1 will print a progress bar
         verbose = user_verbose if hvd.rank() == 0 else 0
+
+        if verbose:
+            print(f"Shared lib path is pointing to: {_horovod.common.process_sets._basics.MPI_LIB_CTYPES}")
 
         transform_spec = None
         if transformation:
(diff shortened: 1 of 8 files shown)


chongxiaoc (approved):
I think it is non-relevant, but `(test-cpu-gloo-py3_8-tf2_5_1-keras_none-torch1_8_1-mxnet1_7_0_p2-pyspark3_1_2)` also fails on lightning pyspark mnist example.

Tixxx:
> I think it is non-relevant, but `(test-cpu-gloo-py3_8-tf2_5_1-keras_none-torch1_8_1-mxnet1_7_0_p2-pyspark3_1_2)` also fails on lightning pyspark mnist example.

Yea, it's because Ray was upgraded to 1.7 which needed 3.19.1 protobuf, but tf 2.5.1 was compiled against an older version. I think some of the tests are complaining about protobuf mismatch. Let me try to pin the protobuf version to make tf2.5.1 happy, let's see how Ray reacts..

horovod/spark/keras/remote.py
```diff
@@ -227,6 +227,12 @@ def train(serialized_model, train_rows, val_rows, avg_row_size):
                 reader_factory = make_batch_reader
                 is_batch_reader = True
 
+            # Call _setup again in process set module to point shared lib to tensorflow's module
+            # since the lib path might be overwritten in remote trainer.
+            _horovod.common.process_sets._setup(_horovod.tensorflow.mpi_ops._basics)
```
tgaddair, line 232: This is an interesting one. Is this something we can fix a higher level of abstraction, like by calling `hvd.init()` again? I'm wondering if other non-estimator users might run into the same problem (like for `RayExecutor`).

maxhgerlach:
Woah, good catch!

horovod/spark/keras/remote.py
Tixxx: Yea, that's a good suggestion. I think we can hide it in init(), let me test it out.

Tixxx: updated to move this logic to init()

tgaddair (approved):
LGTM! Can land once we fix the Keras tests on master.

PR #3259, opened 2021-11-04
Fixing up current CI test failures. 

romerojosh:
PR to fix up some of the recently failing tests.

PR #3263, opened 2021-11-08
Reactivate tests removed in #3259

### LINKED ISSUE #3262: Fix flaky spark integration tests
Several Spark tests have been disabled in https://github.com/horovod/horovod/pull/3259, we should address the underlying flakiness and re-enable.

EnricoMi:
Reactivates Spark tests disabled in #3259. Fixes #3262.

### DIFF
--- a/test/integration/test_spark_lightning.py
+++ b/test/integration/test_spark_lightning.py
@@ -32,6 +32,17 @@ from torch.nn import functional as F
 from pyspark.ml.linalg import VectorUDT
 from pyspark.sql.types import FloatType, IntegerType
 
+# Spark PyTorch Lightning tests conflict with Tensorflow 2.5.x: https://github.com/horovod/horovod/pull/3263
+skip_lightning_tests = False
+try:
+    # tensorflow has to be imported BEFORE pytorch_lightning, otherwise we see the segfault right away
+    import tensorflow as tf
+    from distutils.version import LooseVersion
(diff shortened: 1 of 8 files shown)


test/integration/test_spark_lightning.py
```diff
@@ -119,6 +130,10 @@ def __init__(self, *args, **kwargs):
         warnings.simplefilter('module')
 
     def test_fit_model(self):
+        if skip_lightning_tests:
+            self.skipTest('Spark PyTorch Lightning tests conflict with Tensorflow 2.5.x: '
+                          'https://github.com/horovod/horovod/pull/3263')
+
```
maxhgerlach, line 136: It looks as if such a block might be missing for `test_terminate_on_nan_flag()` right now. Could the skip condition be integrated into the 
```
@pytest.mark.skipif(LooseVersion(pl.__version__) < LooseVersion(MIN_PL_VERSION), reason='Pytorch lightning version is not supported.')
class SparkLightningTests(unittest.TestCase):
```
decorator otherwise?

EnricoMi: > It looks as if such a block might be missing for `test_terminate_on_nan_flag()` right now.

`test_terminate_on_nan_flag()` does not fail

> Could the skip condition be integrated into the ...

Sure, but then all tests would be skipped, half of the tests do not fail.

maxhgerlach: I see, thanks!

maxhgerlach (approved):

PR #3272, opened 2021-11-11
Exit check_build via sys.exit to flush streams

EnricoMi:
Running `horovodrun --check-build` without a terminal (e.g. piping the output to another process) does not retrieve the actual check output:

```
horovodrun --check-build | tee
```

The reason might be that the respective `print` does not flush and the subsequent `os._exit(0)` [does not flush stdout](https://docs.python.org/3/library/os.html#os._exit) either. There are two possible fixes to this:

1. flush the print: `print(textwrap.dedent(output), flush=True)`
2. use `sys.exit(0)` rather than `os._exit(0)`

Since `horovodrun` is not run in a forked process, `sys.exit` should be fine.

Thought about replacing all other places as well, but decided against it as impact on open file handlers shared across processes is unknown to me. There are more `sys.exit` places in Horovod than `os._exit`.

@abditag2 any particular reason to use `os._exit` here?
@tgaddair if this seems to much of a risk we can use fix no. 1 instead.

### DIFF
--- a/horovod/runner/driver/driver_service.py
+++ b/horovod/runner/driver/driver_service.py
@@ -80,9 +80,11 @@ def _launch_task_servers(all_host_names, local_host_names, driver_addresses,
                                                 stdout=host_output,
                                                 stderr=host_output)
             if exit_code != 0:
-                print('Launching horovod task function was not '
-                      'successful:\n{host_output}'.format(host_output=host_output.getvalue()))
-                sys.exit(exit_code)
+                print(
+                    'Launching horovod task function was not '
+                    'successful:\n{host_output}'
+                    .format(host_output=host_output.getvalue()))
+                os._exit(exit_code)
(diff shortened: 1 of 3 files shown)


EnricoMi:
Running horovod with `PYTHONUNBUFFERED=1` has the same effect as the suggested two approches.

akihironitta:
Encountered this when debugging installation in CI. Any updates?

> Running horovod with `PYTHONUNBUFFERED=1` has the same effect as the suggested two approaches.

@EnricoMi Thanks for the workaround! I'll utilise it for now :)

EnricoMi:
Lets stay on the safe side here and only change this for `horovodrun -cb`.

maxhgerlach (approved):
LGTM. 

I don't know what could have motivated using `os._exit()` here.

Issue #3314, opened 2021-12-13
`TorchTests::test_delta_optimizer` causes deadlocks in later tests


maxhgerlach:
**Environment:**
1. Framework: PyTorch
2. Framework version: 1.10
3. Horovod version: master
4. MPI version: OpenMPI

**Bug report:**
With PyTorch 1.10 we occasionally observed deadlocks in `TorchTests::test_dynamic_requires_grad`. These went away by skipping `test_delta_optimizer` (the test before the previous test in alphabetical order), see https://github.com/horovod/horovod/pull/3291#issuecomment-988021354. Since that test only applies with MPI and when run on GPUs, only a few CI configurations can encounter the problem.

The skip is a workaround for some more fundamental problem that is not understood. It should be fixed and `test_delta_optimizer` should be re-eneabled then.

Tixxx:
This is an interesting issue. I think this might be related to how the in-place tensor computations are done in the function hooks in delta optimizer. I can take a look if no on has started.

PR #3319, opened 2021-12-13
Support resurrecting blacklisted hosts

### LINKED ISSUE #1926: Implement more forgiving host blacklist policy in elastic mode
Currently, process failure results in permanent blacklist of the offending host.  This is problematic, as often a host failure will be transient, or a host will be temporarily removed and later restored (in the case of graceful removal).

As such, we should add two policies:

1. Cooldown on failing hosts that allows them to be assigned again once the cooldown period has elapsed.
2. Do not blacklist hosts that are removed from the discovered set and are removed gracefully.

This will be a follow-up to #1849.

ashahab:
This adds support for resurrecting blacklisted hosts in elastic mode.
Currently hosts that get blacklisted remain in the blacklist for the lifetime of the job.
This cannot handle transient host failure or a scale-up after as scale-down.
This is especially the case for the Kubeflow mpi-operator on Kubernetes, as it always
gives pods known hostnames from its hostfile.

This patch will allow blacklisted hosts to become whitelisted after a configured cooldown period.
Users can configure the cooldown range by providing the `--blacklist-cooldown-range` parameter like this:

```bash
$ horovodrun -np 8 --blacklist-cooldown-range 10 100 --min-np 4 --max-np 12 --host-discovery-script discover_hosts.py python train.py
```

The above example configures the minimum cooldown period to 10 seconds and the maximum cooldown period to 100 seconds.
The intial cooldown period would be 10 seconds. For repeat failures the cooldown period would grow with an exponential
backoff delay (with a constant exponent of 2): 10s, 20s, 40s, and so on. However, the maximum cooldown period would be
capped at 100 seconds, regardless of failure count. A random backoff fraction of the cooldown lower limit is added
to the cooldown delay.
The default behavior is to have no cooldown period, and blacklisted hosts would remain in blacklist.

## Description

Fixes #1926 (issue).

## Review process to land 

### DIFF
--- a/docs/elastic.rst
+++ b/docs/elastic.rst
@@ -332,8 +332,23 @@ The maximum np can be used to cap the number of processes (to prevent over-utili
 as a reference point for learning rate scales and data partitions (in cases where these need to be held constant
 regardless of the current number of workers).  If unspecified, maximum np also defaults to ``-np``.
 
-Instances that fail will be added to a blacklist, as they may have faulty hardware.  Ranks that fail repeatedly
-will result in job failure, as it may be the case that the training process cannot make progress.
+Instances that fail will be added to a blacklist, as they may have faulty hardware. Hosts will remain in blacklist for a configured cooldown period.
+After the cooldown period ends, the hosts will be whitelisted back. This is to account for transient failures, and cases where the same host
+is added back to a job.
+Cooldown periods can be configured with the ``--blacklist-cooldown-range`` parameter like this:
+
+.. code-block:: bash
+
+    $ horovodrun -np 8 --blacklist-cooldown-range 10 100 --min-np 4 --max-np 12 --host-discovery-script discover_hosts.py python train.py
+
+The above example configures the minimum cooldown period to 10 seconds and the maximum cooldown period to 100 seconds.
+The intial cooldown period would be 10 seconds. For repeat failures the cooldown period would grow with an exponential
+backoff delay (with a constant exponent of 2): 10s, 20s, 40s, and so on. However, the maximum cooldown period would be
+capped at 100 seconds, regardless of failure count. A random backoff fraction of the cooldown lower limit is added
+to the cooldown delay.
+The default behavior is to have no cooldown period, and blacklisted hosts would remain in blacklist.
+
+Ranks that fail repeatedly will result in job failure, as it may be the case that the training process cannot make progress.
 
 
 Running on Ray
--- a/horovod/ray/elastic_v2.py
+++ b/horovod/ray/elastic_v2.py
@@ -1,4 +1,4 @@
-from typing import Callable, List, Any, Dict, Optional
+from typing import Callable, List, Any, Dict, Optional, Tuple
 import logging
 import ray.exceptions
 import socket
@@ -162,6 +162,11 @@ class ElasticParams(BaseParams):
         reset_limit (int): Maximum number of times that the training
             job can scale up or down the number of workers after
             which the job is terminated.
+        cooldown_range(Tuple[int, int]): Range(in seconds) a failing
+            host will remain in blacklist.
+            Example: cooldown_range=(10, 100)
+            This sets the minimum cooldown period to 10 seconds,
+            and the maximum cooldown period to 100 seconds.
         elastic_timeout (int): Timeout for elastic initialisation after
             re-scaling the cluster. The default value is 600 seconds.
             Alternatively, the environment variable
@@ -177,6 +182,7 @@ class ElasticParams(BaseParams):
     min_workers: int = 1
     max_workers: int = None
     reset_limit: int = None
+    cooldown_range: Optional[Tuple[int, int]] = None
     elastic_timeout: int = 600
     override_discovery: bool = True
 
@@ -205,6 +211,11 @@ class ElasticAdapter(Adapter):
         reset_limit (int): Maximum number of times that the training
             job can scale up or down the number of workers after
             which the job is terminated.
+        cooldown_range (Tuple[int, int]): Range(in seconds) a failing
+            host will remain in blacklist.
+            Example: cooldown_range=(10, 100)
+            This sets the minimum cooldown period to 10 seconds,
+            and the maximum cooldown period to 100 seconds.
         elastic_timeout (int): Timeout for elastic initialisation after
             re-scaling the cluster. The default value is 600 seconds.
             Alternatively, the environment variable
@@ -228,6 +239,7 @@ class ElasticAdapter(Adapter):
                 gpus_per_worker: Optional[int] = None,
                 override_discovery: bool=True,
                 reset_limit: int = None,
+                cooldown_range: Optional[Tuple[int, int]] = None,
                 elastic_timeout: int = 600):
         self.settings = settings
         if override_discovery:
@@ -243,6 +255,7 @@ class ElasticAdapter(Adapter):
         self.max_workers = max_workers
         self.num_workers = min_workers
         self.reset_limit = reset_limit
+        self.cooldown_range = cooldown_range
         self.elastic_timeout = elastic_timeout
         self.driver = None
         self.rendezvous = None
@@ -275,6 +288,7 @@ class ElasticAdapter(Adapter):
             max_np=self.max_workers,
             timeout=self.elastic_timeout,
             reset_limit=self.reset_limit,
+            cooldown_range=self.cooldown_range,
             verbose=self.settings.verbose)
         handler = create_rendezvous_handler(self.driver)
         logger.debug("[ray] starting rendezvous")
--- a/horovod/runner/elastic/discovery.py
+++ b/horovod/runner/elastic/discovery.py
@@ -15,20 +15,48 @@
 
 import io
 import logging
+import random
 import threading
-
+import time
 from collections import defaultdict
 
 from horovod.runner.common.util import safe_shell_exec
 from horovod.runner.elastic.worker import HostUpdateResult
 
+# The default lower bound for cooldown period. If a range is provided,
+# the provided lower limit must be at or above this lower bound
+DEFAULT_COOLDOWN_LOWER_LIMIT_SECONDS = 1
+# The default upper bound for cooldown period. If a range is provided,
+# the provided upper limit must be at or below this upper bound
+DEFAULT_COOLDOWN_UPPER_LIMIT_SECONDS = 1 * 60 * 60
 
 class HostState(object):
-    def __init__(self):
+
+    def __init__(self, cooldown_range=None):
         self._event = threading.Event()
 
-        # TODO(travis): blacklisted hosts should have a timeout period that increases with each failure
         self._blacklisted = False
+        self._blacklist_count = 0
+        if cooldown_range:
+            HostState._validate_cooldown_range(cooldown_range)
+            self._cooldown_lower_limit, self._cooldown_upper_limit = cooldown_range
+        else:
+            self._cooldown_lower_limit = -1
+            self._cooldown_upper_limit = -1
+        self._cooldown_period_end_ts = 0
+
+    @staticmethod
+    def _validate_cooldown_range(cooldown_range):
+        cooldown_lower_limit, cooldown_upper_limit = cooldown_range
+
+        if (cooldown_lower_limit < DEFAULT_COOLDOWN_LOWER_LIMIT_SECONDS):
+            raise ValueError(f"Provided cooldown lower limit: {cooldown_lower_limit} \
+                             cannot be lower than default cooldown lower limit: {DEFAULT_COOLDOWN_LOWER_LIMIT_SECONDS}")
+
+
+        if (cooldown_upper_limit > DEFAULT_COOLDOWN_UPPER_LIMIT_SECONDS):
+            raise ValueError(f"Provided cooldown upper limit: {cooldown_upper_limit} \
+                             cannot be higher than default cooldown upper limit: {DEFAULT_COOLDOWN_UPPER_LIMIT_SECONDS}")
 
     def get_event(self):
         if self._event.is_set():
@@ -39,12 +67,47 @@ class HostState(object):
     def set_event(self):
         self._event.set()
 
+    def _in_cooldown_period(self, current_time):
+        return self._cooldown_period_end_ts > current_time
+
+
+    def _set_cooldown_period(self, current_time):
+        if self._cooldown_lower_limit == -1 or self._cooldown_upper_limit == -1:
+            return
+        self._blacklist_count += 1
+
+        cooldown_delay = self._cooldown_lower_limit * (1 << self._blacklist_count) + (random.uniform(0,1) * self._cooldown_lower_limit)
+        logging.debug(f"{self._blacklist_count}:{self._cooldown_period_end_ts} cooldown_delay: {cooldown_delay}")
+        # We need to ensure that the cooldown upper limit is the upper bound of the delay
+        cooldown_delta_seconds = max(self._cooldown_lower_limit, min(self._cooldown_upper_limit, cooldown_delay))
+
+        self._cooldown_period_end_ts = current_time + cooldown_delta_seconds
+        logging.debug(f"cooldown delta seconds: {cooldown_delta_seconds}")
+
     def blacklist(self):
+        """Moves this host to a blacklist, and starts the cooldown period."""
         self._blacklisted = True
+        now = time.time()
+        if self._in_cooldown_period(now):
+            return
+        self._set_cooldown_period(now)
         self.set_event()
 
+    def whitelist(self):
+        """Ends the cooldown period and moves this host out of blacklist."""
+        self._cooldown_period_end_ts = 0
+        self._blacklisted = False
+
     def is_blacklisted(self):
+        """Checks if the host is in the blacklist."""
         return self._blacklisted
+
+    def is_resurrected(self):
+        """Checks if host is in an expired cooldown period."""
+        if self._cooldown_period_end_ts > 0:
+            return not self._in_cooldown_period(time.time())
+        return False
+
 
 
 class DiscoveredHosts(object):
@@ -76,15 +139,17 @@ class DiscoveredHosts(object):
                                        if not hosts_state[host].is_blacklisted()]
         return self
 
+    def __str__(self):
+        return f"slots: {self._host_slots} order: {self._host_assignment_order}"
+
 
 class HostManager(object):
-    def __init__(self, discovery):
+    def __init__(self, discovery, cooldown_range=None):
         self._current_hosts = DiscoveredHosts(host_slots={}, host_assignment_order=[])
-        self._hosts_state = defaultdict(HostState)
+        self._hosts_state = defaultdict(lambda: HostState(cooldown_range))
         self._discovery = discovery
 
     def update_available_hosts(self):
-        # TODO(travis): also check for hosts removed from the blacklist in the future
         def check_update(cur_host_slots, prev_host_slots):
             res = HostUpdateResult.no_update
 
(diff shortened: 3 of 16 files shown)


ashahab:
@tgaddair @zw0610 This implements https://github.com/horovod/horovod/issues/1926. Please take a look.

horovod/runner/elastic/discovery.py
```diff
@@ -15,20 +15,27 @@
...
 from horovod.runner.common.util import safe_shell_exec
 from horovod.runner.elastic.worker import HostUpdateResult
-
-
+# Maximum time hosts are in cooldown period
+COOLDOWN_UPPER_LIMIT_SECONDS = 5 * 60
+# Minimum time hosts are in cooldown
+COOLDOWN_LOWER_LIMIT_SECONDS = 10
+# Unit of increasing cooldown for repeat failure
+COOLDOWN_PERIOD_DELAY_SECONDS = 10
```
zw0610, line 30: https://github.com/horovod/horovod/issues/1926#issuecomment-992722712

Should it be `Delta` or `Delay`?

ashahab: I removed this in light of our conversation on the `--cooldown-range`

ashahab:
Some of the tests are failing with exit code 124/sigkill. I'll look at the CI scripts to see if they time-bound the tests(some of my tests take long because of waiting for cooldown.

zw0610:
Generally LGTM. Just to make sure the tests go well.

ashahab:
@EnricoMi I'm getting a flaky error in one of the masOS tests. Do you know if this has something to do with the time limits on the tests?
```
[1,0]<stdout>:test_torch.py::TorchTests::test_horovod_join_allreduce [1,1]<stdout>:
[1,1]<stdout>:test_torch.py::TorchTests::test_horovod_join_allreduce 
Error: Process completed with exit code 124.
```

EnricoMi:
@maxhgerlach is above macOS flakiness related to your fixes in #3300 / #3291?

maxhgerlach:
Hmm, we saw a similar problem with join and _allgather_ (https://github.com/horovod/horovod/pull/3291#issuecomment-992220741), which apparently went away with @Tixxx's fix in https://github.com/horovod/horovod/pull/3313. If this one is with allreduce though, the cause may be different.

We observed another hang with torch 1.10 that went away by skipping a specific test: https://github.com/horovod/horovod/issues/3314 

Something's fishy here.

ashahab:
@maxhgerlach I see this GPU buildkite failures in other PRs too, what can be done about this?

EnricoMi:
Which other PRs have the same failures? You can ignore the "Unit Test Results (with flaky tests)" failures, but there are no "Unit Test Results" failure in master (https://github.com/horovod/horovod/runs/4696939351) and haven't been for the last four commits (https://github.com/horovod/horovod/actions/workflows/ci.yaml?query=branch%3Amaster+event%3Apush).

It looks like your branch is already rebased to master HEAD. Maybe this is triggering some hidden issue or introducing it.

ashahab:
Thanks for the tip @EnricoMi . I figured out the issue, we need to look at why ray tests become flaky on GPU heads. I created the following issue: https://github.com/horovod/horovod/issues/3347

@tgaddair @zw0610 can you review this PR?

zw0610:
LGTM. An additional question: if the hostname of a worker is resurrected after cool down period but the worker is still down, as the worker is Not reporting itself, I suppose this worker will not be added to the group until it starts and reports itself. Am I right?

ashahab:
@zw0610 No, the nodes would still be added even if they are down. The discovery script is supposed to figure out which nodes are up and return only healthy nodes. If we want to separate the discovery of healthy nodes from the discovery script, we can design in a health check hook, but that is not included in this PR.

zw0610:
> @zw0610 No, the nodes would still be added even if they are down. The discovery script is supposed to figure out which nodes are up and return only healthy nodes. If we want to separate the discovery of healthy nodes from the discovery script, we can design in a health check hook, but that is not included in this PR.

@ashahab  Sorry for not explaining clearly. What I meant is such worker is not reported by `discover_host.sh` script but it's already past its cool down time. I suppose the `HostManager` checks if a worker should be still blacklisted or whitelisted only after this worker being reported by `discover_host.sh` script again.

ashahab:
@zw0610 that is an interesting test case, I'll add that in.
However, the answer to that is no, hosts not in discovery script would never get added:
```python
        def whitelist_all_hosts():
            for host in host_slots.keys():
                if self._hosts_state[host].is_resurrected():
                    self._hosts_state[host].whitelist()

        def has_resurrected_hosts():
            resurrected_hosts = [host for host in host_slots.keys() if self._hosts_state[host].is_resurrected()]
            return len(resurrected_hosts) > 0
```

zw0610:
> hosts not in discovery script would never get added

@ashahab That's great! It's exactly the behavior I expect from horovod elastic.

chongxiaoc (approved):
LGTM. Need @tgaddair to take one more look.

horovod/ray/elastic_v2.py
```diff
@@ -162,6 +162,8 @@ class ElasticParams(BaseParams):
         reset_limit (int): Maximum number of times that the training
             job can scale up or down the number of workers after
             which the job is terminated.
+        cooldown_range(int range): Range of seconds(min, max) a failing
+            host will remain in blacklist.
```
EnricoMi, line 166: Can you given an example, I think `int range` is not very common to know that two integers are expected here.

horovod/ray/elastic_v2.py
```diff
@@ -205,6 +208,8 @@ class ElasticAdapter(Adapter):
         reset_limit (int): Maximum number of times that the training
             job can scale up or down the number of workers after
             which the job is terminated.
+        cooldown_range (int range): Range of seconds(min, max) a failing
+            host will remain in blacklist.
```
EnricoMi, line 212: same here

horovod/runner/elastic/discovery.py
```diff
@@ -15,20 +15,33 @@
...
 from horovod.runner.common.util import safe_shell_exec
 from horovod.runner.elastic.worker import HostUpdateResult
-
-
+# Maximum time hosts are in cooldown period
+COOLDOWN_UPPER_LIMIT_SECONDS = 5 * 60
+# Minimum time hosts are in cooldown
+COOLDOWN_LOWER_LIMIT_SECONDS = 10
+# Unit of increasing cooldown for repeat failure
+COOLDOWN_RANGE_SECONDS = [COOLDOWN_LOWER_LIMIT_SECONDS, COOLDOWN_UPPER_LIMIT_SECONDS]
```
EnricoMi, line 30: Constant `COOLDOWN_RANGE_SECONDS` is not used anywhere. What is it good for?

horovod/runner/elastic/discovery.py
```diff
@@ -15,20 +15,33 @@
...
 class HostState(object):
```
EnricoMi, line 31: I think here should go an empty line

horovod/runner/elastic/discovery.py
```diff
@@ -15,20 +15,33 @@
...
+import random
 import threading
-
+import time
 from collections import defaultdict
 
 from horovod.runner.common.util import safe_shell_exec
 from horovod.runner.elastic.worker import HostUpdateResult
-
-
```
EnricoMi, line 25: I think we should keep at least one empty line here

horovod/runner/elastic/discovery.py
EnricoMi, line 30: Are these defaults? Then the constants should indicate that in their name.

horovod/runner/elastic/discovery.py
```diff
@@ -15,20 +15,33 @@
...
-    def __init__(self):
+
+    def __init__(self, cooldown_range=None):
         self._event = threading.Event()
 
-        # TODO(travis): blacklisted hosts should have a timeout period that increases with each failure
         self._blacklisted = False
+        self._blacklist_count = 0
+        if cooldown_range:
+            self._cooldown_lower_limit = cooldown_range[0]
```
EnricoMi, line 39: Does this require some sanity check on the values in `cooldown_range`?

docs/elastic.rst
```diff
@@ -332,8 +332,20 @@ The maximum np can be used to cap the number of processes (to prevent over-utili
...
+After the cooldown period ends, the hosts will be whitelisted back. This is to account for transient failures, and cases where the same host
+is added back to a job.
+Cooldown periods can be configured with the ``--blacklist-cooldown-range`` parameter like this:
+
+.. code-block:: bash
+    $ horovodrun -np 8 --blacklist-cooldown-range 10 100 --min-np 4 --max-np 12 --host-discovery-script discover_hosts.py python train.py
+
+The above example configures the minimum cooldown period to 10 seconds and the maximum cooldown period to 100 seconds.
+The intial cooldown period would be 10 seconds. For repeat failures the cooldown period would grow with an exponential
+backoff delay: 10s, 20s, 30s, and so on. However, the maximum cooldown period would be capped at 100 seconds, regardless
```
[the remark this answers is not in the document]
EnricoMi: There is a random lower backoff added to the exponential series (https://github.com/horovod/horovod/pull/3319/files#diff-3d599d62df6fc82afb3202d7abaacc169b292f800bae36b5355867ae83769aa4R64). This should be stated here.

horovod/runner/elastic/discovery.py
```diff
@@ -39,13 +52,48 @@ def get_event(self):
...
 
+    def _in_cooldown_period(self, current_time):
+        return self._cooldown_period_end_ts > current_time
+
+
+    def _set_cooldown_period(self, current_time):
+        if self._cooldown_lower_limit == -1 or self._cooldown_upper_limit == -1:
+            return
+        self._blacklist_count += 1
+        def _exponential_backoff_time():
```
EnricoMi, line 63: please add an empty line before and after this def

horovod/runner/elastic/discovery.py
```diff
@@ -39,13 +52,48 @@ def get_event(self):
...
+            cooldown_delay = self._cooldown_lower_limit * (1 << self._blacklist_count)
+            + (random.uniform(0,1) * self._cooldown_lower_limit)
```
EnricoMi, line 65: Can you move the `+` to the previous line and intend the this line with the `self` above?
Also the brackets are redundant.

Does this code even work without a `\` at the end of the previous line?

horovod/runner/elastic/discovery.py
```diff
@@ -39,13 +52,48 @@ def get_event(self):
...
+            logging.debug(f"{self._blacklist_count}:{self._cooldown_period_end_ts} cooldown_delay: {cooldown_delay}")
+            return max(self._cooldown_lower_limit, min(self._cooldown_upper_limit, cooldown_delay))
+        cooldown_delta_seconds = _exponential_backoff_time()
```
EnricoMi, line 68: Why is this computation in an inner `def` anyway?

horovod/runner/elastic/discovery.py
```diff
@@ -39,13 +52,48 @@ def get_event(self):
...
+
+    def _set_cooldown_period(self, current_time):
+        if self._cooldown_lower_limit == -1 or self._cooldown_upper_limit == -1:
+            return
+        self._blacklist_count += 1
+        def _exponential_backoff_time():
+            cooldown_delay = self._cooldown_lower_limit * (1 << self._blacklist_count)
+            + (random.uniform(0,1) * self._cooldown_lower_limit)
+            logging.debug(f"{self._blacklist_count}:{self._cooldown_period_end_ts} cooldown_delay: {cooldown_delay}")
+            return max(self._cooldown_lower_limit, min(self._cooldown_upper_limit, cooldown_delay))
```
EnricoMi, line 67: What is the edge case that requires the `max`here?

horovod/runner/elastic/discovery.py
```diff
@@ -123,7 +188,7 @@ def current_hosts(self):
 
     def blacklist(self, host):
         if not self._hosts_state[host].is_blacklisted():
-            logging.warning('blacklist failing host: {}'.format(host))
+            logging.debug('blacklist failing host: {}'.format(host))
```
EnricoMi, line 191: I think this is useful in  normal operation, so what about moving this to `info` level?

horovod/ray/elastic_v2.py
```diff
@@ -177,6 +179,7 @@ class ElasticParams(BaseParams):
     min_workers: int = 1
     max_workers: int = None
     reset_limit: int = None
+    cooldown_range: List[int] = None
```
EnricoMi, line 182: This implies that any size is allowed, where we only allow for exactly two ints. Can this be changed to `Tuple[int, int]`, or even `Optional[Tuple[int, int]]`?

EnricoMi (changes requested):

horovod/runner/elastic/discovery.py
EnricoMi: Are these default values? Then the name should indicate this.

docs/elastic.rst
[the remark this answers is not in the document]
ashahab: @EnricoMi thanks for the review. I have updated my description. I would like to keep the exponent to be constant for now as users already can configure the cooldown range([min,max]).

horovod/ray/elastic_v2.py
ashahab: Done

horovod/ray/elastic_v2.py
ashahab: Done

horovod/ray/elastic_v2.py
ashahab: Done

horovod/runner/elastic/discovery.py
ashahab: Done

horovod/runner/elastic/discovery.py
ashahab: Removed

ashahab: Removed

horovod/runner/elastic/discovery.py
ashahab: Added

horovod/runner/elastic/discovery.py
ashahab: Added

horovod/runner/elastic/discovery.py
ashahab: Added

horovod/runner/elastic/discovery.py
ashahab: Done

horovod/runner/elastic/discovery.py
ashahab: I added it for symmetry with the `min` computation, but I don't see the need unless we support fractional cooldown limits.

horovod/runner/elastic/discovery.py
ashahab: I didn't feel I should expose the _exponential_backoff_time calculation to any other methods even in this class :). Are you suggesting I make it a regular private method?

horovod/runner/elastic/discovery.py
ashahab: Done

tgaddair (approved):

chongxiaoc:
Hi @ashahab , have you addressed @EnricoMi 's change request? We can land this PR soon.

ashahab:
@chongxiaoc  I just pushed the changes, lets wait for the tests. Thank you!

docs/elastic.rst
```diff
@@ -332,8 +332,22 @@ The maximum np can be used to cap the number of processes (to prevent over-utili
...
+After the cooldown period ends, the hosts will be whitelisted back. This is to account for transient failures, and cases where the same host
+is added back to a job.
+Cooldown periods can be configured with the ``--blacklist-cooldown-range`` parameter like this:
+
+.. code-block:: bash
+    $ horovodrun -np 8 --blacklist-cooldown-range 10 100 --min-np 4 --max-np 12 --host-discovery-script discover_hosts.py python train.py
+
+The above example configures the minimum cooldown period to 10 seconds and the maximum cooldown period to 100 seconds.
+The intial cooldown period would be 10 seconds. For repeat failures the cooldown period would grow with an exponential
+backoff delay(with a constant exponent of 2): 10s, 20s, 30s, and so on. However, the maximum cooldown period would be
```
EnricoMi, line 345: please add a whitespace between `delay` and `(`

EnricoMi: This example is still not exponential, should be `10s, 20s, 40s`.

docs/elastic.rst
```diff
@@ -332,8 +332,22 @@ The maximum np can be used to cap the number of processes (to prevent over-utili
...
+capped at 100 seconds, regardless of failure count. Note: a random backoff fraction of the cooldown lower limit is added
```
EnricoMi, line 346: Maybe without `Note:`, otherwise this should be moved to a new line / paragraph.

horovod/ray/elastic_v2.py
```diff
@@ -162,6 +162,11 @@ class ElasticParams(BaseParams):
         reset_limit (int): Maximum number of times that the training
             job can scale up or down the number of workers after
             which the job is terminated.
+        cooldown_range(Tuple[int, int]): Range of seconds(min, max) a failing
+            host will remain in blacklist.
+            e.g: cooldown_ranage: [10, 100]. This sets the minimum cooldown
```
EnricoMi, line 167: Well, not that this is a tuple, the example should read `(10, 100)`. Also the notation `cooldown_ranage:` is not really useful as this is not how you would call this method with the example value.

```suggestion
            Example: cooldown_ranage=(10, 100)
            This sets the minimum cooldown
```

horovod/ray/elastic_v2.py
EnricoMi: The same example as above would be as useful here, I think.

horovod/runner/elastic/discovery.py
EnricoMi: Sorry, I meant before `class HostState`.

horovod/runner/elastic/discovery.py
```diff
@@ -15,20 +15,50 @@
...
+            self._cooldown_upper_limit = cooldown_range[1]
+        else:
+            self._cooldown_lower_limit = -1
+            self._cooldown_upper_limit = -1
+        self._cooldown_period_end_ts = 0
+
+    @staticmethod
+    def _validate_cooldown_range(cooldown_range):
+        cooldown_lower_limit = cooldown_range[0]
+        cooldown_upper_limit = cooldown_range[1]
```
EnricoMi, line 52: Now that `cooldown_range` is a tuple, why not:

```
        cooldown_lower_limit, cooldown_upper_limit = cooldown_range
```

Here and in a few similar other places.

horovod/runner/elastic/discovery.py
EnricoMi: no, just inline the code into this `_set_cooldown_period` method

horovod/runner/elastic/discovery.py
EnricoMi: well, since you never assert that `cooldown_lower_limit` is smaller than `cooldown_upper_limit`, the `max` is needed here

EnricoMi (changes requested):

EnricoMi:
@ashahab Updated my review. Can you please update the PR description, specially this part:

```
For repeat failures the cooldown period grows with an exponential backoff delay: 10s, 20s, 30s. Cooldown period is capped
at 5 minutes.
```

I think you are referring to the default here, which is different I think. And the example is not exponential.

horovod/ray/elastic_v2.py
```diff
@@ -162,6 +162,11 @@ class ElasticParams(BaseParams):
         reset_limit (int): Maximum number of times that the training
             job can scale up or down the number of workers after
             which the job is terminated.
+        cooldown_range(Tuple[int, int]): Range of seconds(min, max) a failing
```
EnricoMi, line 165: What do you want to express with the notation `seconds(min, max)` here? Or is there just a whitespace missing before the `(`?

The same further down.

docs/elastic.rst
ashahab: Done, thanks!

docs/elastic.rst
ashahab: Done

horovod/ray/elastic_v2.py
ashahab: Done

horovod/runner/elastic/discovery.py
ashahab: Done

horovod/ray/elastic_v2.py
ashahab: I wanted to express that the first element if the range implies the min and the second element implies the max.  Thinking over it, that's the definition of a range, so I'll remove that notation.

horovod/runner/elastic/discovery.py
```diff
@@ -39,13 +67,50 @@ def get_event(self):
...
+            return
+        self._blacklist_count += 1
+
+        def _exponential_backoff_time():
+            cooldown_delay = self._cooldown_lower_limit * (1 << self._blacklist_count) + (random.uniform(0,1) * self._cooldown_lower_limit)
+            logging.debug(f"{self._blacklist_count}:{self._cooldown_period_end_ts} cooldown_delay: {cooldown_delay}")
+            # We need to ensure that the cooldown upper limit is the upper bound of the delay
+            return max(self._cooldown_lower_limit, min(self._cooldown_upper_limit, cooldown_delay))
+
+        cooldown_delta_seconds = _exponential_backoff_time()
```
EnricoMi, line 85: The inner `_exponential_backoff_time` is not really needed:

```suggestion
    def _set_cooldown_period(self, current_time):
        if self._cooldown_lower_limit == -1 or self._cooldown_upper_limit == -1:
            return
        self._blacklist_count += 1

        cooldown_delay = self._cooldown_lower_limit * (1 << self._blacklist_count) + (random.uniform(0,1) * self._cooldown_lower_limit)
        logging.debug(f"{self._blacklist_count}:{self._cooldown_period_end_ts} cooldown_delay: {cooldown_delay}")
        # We need to ensure that the cooldown upper limit is the upper bound of the delay
        cooldown_delta_seconds = max(self._cooldown_lower_limit, min(self._cooldown_upper_limit, cooldown_delay))
```

EnricoMi (changes requested):

horovod/runner/elastic/discovery.py
ashahab: @EnricoMi I was looking for a clean way to separate it from the rest of the cooldown calculation for readability. However, I can inline it as you have suggested.

EnricoMi: Understood. I think two empty lines around that block should make it readable as well.

EnricoMi (approved):
LGTM!

EnricoMi:
@ashahab the docs are failing: https://readthedocs.org/projects/horovod/builds/15848706/

EnricoMi:
> @ashahab the docs are failing: https://readthedocs.org/projects/horovod/builds/15848706/

Never mind, looks like this exists in master. Will be fixed in #3377.

PR #3323, opened 2021-12-15
Fix build with torch-nightly, remove unused code

maxhgerlach:
Following changes from https://github.com/pytorch/pytorch/pull/69039


maxhgerlach:
Spark PyTests apparently failed three times in a row for test-cpu-gloo-py3_8-tf2_5_1-keras2_4_3-torch1_9_1-mxnet1_7_0_p2-pyspark3_2_0 because of `test_spark_task_service_abort_command`: https://github.com/horovod/horovod/runs/4532327101?check_suite_focus=true 

The test is flaky for the same reason in other builds, too, so I've added a commit to monkey patch `safe_shell_exec.GRACEFUL_TERMINATION_TIME_S` to a slightly larger value. I hope this does not hide a fundamental problem.

maxhgerlach:
Build and test went fine for dc0b657, which included an unrelated extra commit that I've removed again. (I thought I could fix the flaky test that failed the first build, but that did not work)

EnricoMi (approved):
LGTM! Wait for the tests to complete.

Tixxx:
Great fix! Thanks for taking a look.

Tixxx (approved):

PR #3326, opened 2021-12-16
Fixes empty hostname returned from HostDiscoveryScript

plliao:

## Description

Fixes empty hostname returned from HostDiscoveryScript and make sure hostname is not empty.

Context:
When discovery script return nothing, HostDiscoveryScript will treat the empty result i.e. "" string as the available host. 
It will fail training job as launcher cannot ssh on the host with empty address. Discovery script can return nothing when no workers are available or new workers are not ready yet.

## Review process to land 

### DIFF
--- a/horovod/runner/elastic/discovery.py
+++ b/horovod/runner/elastic/discovery.py
@@ -169,7 +169,9 @@ class HostDiscoveryScript(HostDiscovery):
             if ':' in line:
                 host, slots = line.split(':')
                 host_slots[host] = int(slots)
-            else:
+            # Make sure the host is not empty. The discovery script might
+            # return empty string when all workers are not ready or available.
+            elif host:
                 host_slots[host] = self._default_slots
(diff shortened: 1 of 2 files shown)


horovod/runner/elastic/discovery.py
```diff
@@ -169,7 +169,9 @@ def find_available_hosts_and_slots(self):
             if ':' in line:
                 host, slots = line.split(':')
                 host_slots[host] = int(slots)
-            else:
+            # Make sure the host is not empty. The discovery script might
+            # return empty string when all workers are not ready or available.
+            elif host:
```
ashahab, line 174: Can you add some tests? Also this would result in an empty map being returned to the caller (update_available_hosts) , which would result in the HostManager treating this as an update in the set of hosts, which is not intended.

ashahab (changes requested):

PR #3338, opened 2021-12-28
Make docker fail on the first command that fails in RUN

EnricoMi:
Currently, docker only fails if the last command in `RUN` fails, otherwise it keeps on building the image. The resulting image fails during tests (in the best case) and a thorough investigation of the build process is the consequence. There have been a lot of situations in the past where errors in the build process have been hidden by this, e.g. https://github.com/horovod/horovod/runs/4648818225?check_suite_focus=true#step:10:1588.

This makes the `RUN` command fail on the first bash command that fails, not only on the last one.

Related, macOS tests run even if the build step fails. This is also fixed here.

### DIFF
--- a/.github/gen-workflow-ci.py
+++ b/.github/gen-workflow-ci.py
@@ -515,7 +515,7 @@ def main():
                 '\n'.join([f'      - name: Test [attempt {attempt} of {attempts}]\n'
                            f'        id: test-{attempt}\n'
                            f'        continue-on-error: {"true" if attempt < attempts else "false"}\n'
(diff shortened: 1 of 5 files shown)


maxhgerlach (approved):

maxhgerlach:
This seems to have unveiled another hidden error:

> No matching distribution found for keras==2.6.2

Keras 2.6.0 is available on PyPI.

EnricoMi:
Yep, fixed. This simple thing hid a lot of build errors in the past.

PR #3358, opened 2022-01-12
Replaced use of tempfile.mktemp() with tempfile.mkstemp()

ashahab:
This addresses use of tempfile.mktemp() which [is not secure](https://docs.python.org/3/library/tempfile.html#deprecated-functions-and-variables). Another process may create a file with the same name in the time between the call to mktemp() and subsequent attempt to create the file by the first process.

- [ x] Did you update the docs?

## Description

Fixes # (issue).

## Review process to land 

### DIFF
--- a/horovod/runner/js_run.py
+++ b/horovod/runner/js_run.py
@@ -126,7 +126,12 @@ def generate_jsrun_rankfile(settings, path=None):
             slots=settings.num_proc))
 
     # Generate rankfile
-    path = tempfile.mktemp() if path is None else path
+    # using mkstemp here instead of insecure mktemp.
+    # note that the caller is responsible for cleaning up this file
+    if path is None:
+        fd, path = tempfile.mkstemp()
+        fd.close()
+
     with open(path, 'w') as tmp:
         tmp.write('overlapping_rs: allow\n')
         tmp.write('cpu_index_using: logical\n')
--- a/test/utils/common.py
+++ b/test/utils/common.py
@@ -112,15 +112,12 @@ def tempdir():
 
 @contextlib.contextmanager
 def temppath():
-    path = tempfile.mktemp()
+    dir_path = tempfile.TemporaryDirectory()
+    path = os.path.join(dir_path.name,'temp_test_file')
     try:
         yield path
     finally:
-        if os.path.exists(path):
-            if os.path.isfile(path):
-                os.remove(path)
-            else:
-                shutil.rmtree(path)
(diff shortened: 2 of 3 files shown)


Tixxx:
Thanks for fixing this. It seems like test/utils/common.py has another usage of mktemp(), could you change that too?

ashahab:
> 

Done, thanks!

test/utils/common.py
```diff
@@ -112,7 +112,8 @@ def tempdir():
 
 @contextlib.contextmanager
 def temppath():
-    path = tempfile.mktemp()
+    _, path = tempfile.mkstemp()
```
EnricoMi, line 115: `mkstemp` has the argument `suffix`, that should make `path += '.tmp''` obsolete.

EnricoMi: Further, `mkstemp` **creates** a file, and then we come up with a different name and create and delete that one, but not the original one. Not good.

EnricoMi (changes requested):
left a few comments, I think this needs improvement

test/utils/common.py
EnricoMi: maybe we should use `tempfile.mkdtemp` here since we use the path as a directory

horovod/runner/js_run.py
```diff
@@ -126,7 +126,10 @@ def generate_jsrun_rankfile(settings, path=None):
             slots=settings.num_proc))
 
     # Generate rankfile
-    path = tempfile.mktemp() if path is None else path
+    # using mkstemp here instead of insecure mktemp.
+    # note that the caller is responsible for cleaning up this file
+    if path is None:
+        _, path = tempfile.mkstemp()
```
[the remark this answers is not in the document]
nvcastet: +1 yes cleaner to close the fd with os.close()

test/utils/common.py
ashahab: @EnricoMi I have changed this to mkdtemp as that makes the most sense for these tests. Thanks for the feedback!

horovod/runner/js_run.py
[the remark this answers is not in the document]
ashahab: I thought of using the higher level tempfile.TemporaryFile, but that will disappear at the call of close()
I have closed the fd.

ashahab:
> left a few comments, I think this needs improvement

Resolved the comments

EnricoMi (approved):
LGTM!

PR #3362, opened 2022-01-13
Fixes the checkpoint directory structure for pytorch and pytorch lightning

### LINKED ISSUE #3268: Unable to load most recent checkpoint for Pytorch and Pytorch lightning Estimator
**Environment:**
1. Framework: PyTorch
2. Framework version: 1.8.1
3. Horovod version: 0.23.0
4. MPI version:
5. CUDA version:
6. NCCL version:
7. Python version: 3.8
8. Spark / PySpark version: 3.1.2
9. Ray version:
10. OS and version:
11. GCC version:
12. CMake version:

**Bug report:**
In case of pytorch lightning estimator, the _read_checkpoint() API does not return the latest checkpoint stored in the run path. 
Reason: Pytorch lightning estimator calls store.get_checkpoints() which looks for a folder named 'checkpoint' in run path while there is no folder named checkpoint, instead there is a temp folder generated via tempfile.TemporaryDirectory()

In case of pytorch estimator, the checkpoint stored in run path is not overwritten if multiple iterations are done using the same run path, which leads to _load_checkpoint() API returning the stale checkpoint.

kamalsharma2:

## Description
Fixes #3268. Creating a ```copy()``` method to replace ```put()``` while syncing from local temporary directory to run path. ```put()``` is inconsistent with cp and currently has no mechanism to paste the contents from source directory to target without pasting the source directory itself if the target directory is already present, this case is unwanted for horovod and replacing this creates consistency for checkpoints path for both pytorch and pytorch lightning.

### DIFF
--- a/horovod/spark/common/store.py
+++ b/horovod/spark/common/store.py
@@ -267,9 +267,6 @@ class AbstractFilesystemStore(Store):
     def get_logs_subdir(self):
         return 'logs'
 
-    def get_wildcard(self):
-        return '*'
-
     def _get_full_path_or_default(self, path, default_key):
         if path is not None:
             return self.get_full_path(path)
@@ -315,13 +312,17 @@ class FilesystemStore(AbstractFilesystemStore):
 
         def fn(local_run_path):
             print(f"Syncing dir {local_run_path} to dir {run_path}")
-            if self.fs.exists(run_path):
-                local_run_path = os.path.join(local_run_path, self.get_wildcard())
             self.copy(local_run_path, run_path, recursive=True, overwrite=True)
 
         return fn
 
     def copy(self, lpath, rpath, recursive=False, callback=_DEFAULT_CALLBACK,**kwargs):
+        """ 
+        This method copies the contents of the local source directory to the target directory.
+        This is different from the fsspec's put() because it does not copy the source folder 
+        to the target directory in the case when target directory already exists.
+        """
+
         from fsspec.implementations.local import LocalFileSystem, make_path_posix
         from fsspec.utils import other_paths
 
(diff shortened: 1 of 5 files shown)


EnricoMi (commented):
minor formatting, can't say much about the logic

kamalsharma2:
@EnricoMi resolved the comments, please take a look

EnricoMi:
@tgaddair @irasit what do you think?

horovod/spark/common/store.py
```diff
@@ -311,10 +315,35 @@ def sync_fn(self, run_id):
 
         def fn(local_run_path):
             print(f"Syncing dir {local_run_path} to dir {run_path}")
-            self.fs.put(local_run_path, run_path, recursive=True, overwrite=True)
+            if self.fs.exists(run_path):
+                local_run_path = os.path.join(local_run_path, self.get_wildcard())
+            self.copy(local_run_path, run_path, recursive=True, overwrite=True)
```
chongxiaoc, line 320: Why this `copy()` function is not required in HDFS store?

horovod/spark/common/store.py
```diff
@@ -248,7 +249,7 @@ def get_run_path(self, run_id):
         return os.path.join(self.get_runs_path(), run_id)
 
     def get_checkpoint_path(self, run_id):
-        return os.path.join(self.get_run_path(run_id), self.get_checkpoint_filename()) \
+        return self.get_run_path(run_id) \
```
chongxiaoc, line 252: This is actually conflicting with my on-going work for some Keras/TF1 checkpoint stuff at Uber, since I'm calling this interface directly to get file name.
But I'm fine to add additional path join function to compose file name.

chongxiaoc (commented):
left a question.

horovod/spark/common/store.py
kamalsharma2: HDFS store's ```sync_fn()``` does not make use of ```put()```. While ```put()``` copies the source folder along with it's contents to the target folder when the target folder exists, this issue is eliminated in the HDFS store's ```sync_fn()``` because it copies from the source folder to the target folder only.

horovod/spark/common/store.py
kamalsharma2: ```get_checkpoint_path()``` is used for all the estimators to check if the checkpoints exists, plus to fetch the checkpoint( except for lightning), if we yield the whole path here then all these functionalities can't be fulfilled hence returning the run_path

chongxiaoc (approved):
LGTM

tgaddair (approved):

zyluo:
fsspec version requirement  introducee from this commit

https://github.com/horovod/horovod/issues/3450

PR #3371, opened 2022-01-17
Build Horovod with temporarily installed CMake if necessary

maxhgerlach:

## Description

This is a follow-up to PR #3261, which would require users of many distros to install a recent CMake from an external source (like `pip`).

If such a situation is identified in `setup.py`, we can automatically install a sufficiently recent CMake to a temporary directory and use that to build Horovod. That directory is cleaned up afterwards and the user or system environment is not affected. If `HOROVOD_CMAKE` is set, only that CMake binary is used without any change of behavior.

## Review process to land 

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -100,9 +100,6 @@ jobs:
         sed -i -e "s%^# setup ssh service$%${command//$'\n'/\\n}\n\n# setup ssh service%" Dockerfile.test.?pu
 
         command=$(cat <<EOF
-        # Install recent CMake
-        RUN pip install --no-cache-dir -U cmake~=3.13.0
-
         # Setup CodeQL tracing
         RUN mkdir -p /home/runner/work/horovod
         RUN ln -s /horovod /home/runner/work/horovod/horovod
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -100,6 +100,9 @@ jobs:
         sed -i -e "s%^# setup ssh service$%${command//$'\n'/\\n}\n\n# setup ssh service%" Dockerfile.test.?pu
 
         command=$(cat <<EOF
+        # Install recent CMake
+        RUN pip install --no-cache-dir -U cmake~=3.13.0
+
         # Setup CodeQL tracing
         RUN mkdir -p /home/runner/work/horovod
         RUN ln -s /horovod /home/runner/work/horovod/horovod
--- a/Dockerfile.test.cpu
+++ b/Dockerfile.test.cpu
@@ -36,6 +36,7 @@ RUN add-apt-repository ppa:ubuntu-toolchain-r/test
 RUN apt-get update -qq && apt-get install -y --no-install-recommends \
         wget \
         ca-certificates \
+        cmake \
         openssh-client \
         openssh-server \
         git \
@@ -56,9 +57,6 @@ RUN wget --progress=dot:mega https://bootstrap.pypa.io/get-pip.py && python get-
 # pinning pip to 21.0.0 as 22.0.0 cannot fetch pytorch packages from html linl
 # https://github.com/pytorch/pytorch/issues/72045
 RUN pip install --no-cache-dir -U --force pip~=21.0.0 setuptools requests pytest mock pytest-forked parameterized
-
-# Install recent CMake.
-RUN pip install --no-cache-dir -U cmake~=3.13.0
 
 # Add launch helper scripts
 RUN echo "env SPARK_HOME=/spark SPARK_DRIVER_MEM=512m PYSPARK_PYTHON=/usr/bin/python${PYTHON_VERSION} PYSPARK_DRIVER_PYTHON=/usr/bin/python${PYTHON_VERSION} \"\$@\"" > /spark_env.sh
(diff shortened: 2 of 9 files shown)


Dockerfile.test.cpu
```diff
@@ -33,6 +33,7 @@ RUN add-apt-repository ppa:ubuntu-toolchain-r/test
 RUN apt-get update -qq && apt-get install -y --no-install-recommends \
         wget \
         ca-certificates \
+        cmake \
```
EnricoMi, line 36: we could rely here on `setup.py` installing the latest cmake so we always test this really works

.github/workflows/codeql-analysis.yml
```diff
@@ -100,6 +100,9 @@ jobs:
         sed -i -e "s%^# setup ssh service$%${command//$'\n'/\\n}\n\n# setup ssh service%" Dockerfile.test.?pu
 
         command=$(cat <<EOF
+        # Install recent CMake
```
EnricoMi, line 103: Why is installing `cmake` through `apt-get install`in the `Dockerfile`s not sufficient here?

Dockerfile.test.cpu
maxhgerlach: I believe that this apt-get line currently installs cmake 3.10 on the Ubuntu 18.04 we use for testing (too old to build Horovod). So yes, as it stands, we rely on setup.py to install a temporary CMake 3.13 to build Horovod and the mechanism from this PR is tested. This would change once we update to Ubuntu 20.04 or later. 

I left cmake in here for any non-Horovod builds in this Docker container, e.g., for oneCCL.

.github/workflows/codeql-analysis.yml
maxhgerlach: We need CMake 3.13 to compile the C++ parts of Horovod. On Ubuntu 18.04 and without extra package sources like a PPA from Kitware `apt-get install cmake` installs CMake 3.10 IIRC.

The mechanism of this PR does not work with the build for CodeQL because we don't use setup.py there.

Dockerfile.test.cpu
EnricoMi: Ah, our ancient Ubuntu 18.04 means this `apt-get intall cmake` is not sufficient. Then it makes sense we need to add that extra code to the CodeQL yaml. Once we update to Ubuntu 20.04 we will lose testing that auto-install. We should remember when we upgrade to keep Ubuntu 18.04 for one of the test combinations, e.g. TF 1.15 to see that auto-install feature still being tested (as long as we want to support Ubuntu 18.04 in our test setup).

EnricoMi (commented):
I link this, leave it to @tgaddair and @nvcastet to approve.

Dockerfile.test.cpu
maxhgerlach: > We should remember when we upgrade to keep Ubuntu 18.04 for one of the test combinations, e.g. TF 1.15 to see that auto-install feature still being tested (as long as we want to support Ubuntu 18.04 in our test setup).

That's a good plan!

> https://github.com/horovod/horovod/runs/4852098212?check_suite_focus=true#step:10:4652

Yep, here we can see that it used a CMake binary from `/tmp`:

>   11.63   Could not find a recent CMake to build Horovod. **Attempting to install CMake 3.13 to a temporary location via pip.**
>   ...
>   13.92   Running CMake in build/temp.linux-x86_64-3.7/RelWithDebInfo:
>   13.92   **/tmp/horovod-cmake-tmpmj4n5l40/bin/run_cmake** /tmp/pip-req-build-fzb3ontk -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_LIBRARY_OUTPUT_DIRECTORY_RELWITHDEBINFO=/tmp/pip-req-build-fzb3ontk/build/lib.linux-x86_64-3.7 -DPYTHON_EXECUTABLE:FILEPATH=/usr/bin/python
>

.github/workflows/codeql-analysis.yml
EnricoMi: This can probably be removed again as we have moved to Ubuntu 20.04 for our test images in the meantime.

Dockerfile.test.cpu
EnricoMi: As a side note: In the meantime we have moved our test images to Ubuntu 20.04, but a few test images still use Ubuntu 18.04, so we now test both.

EnricoMi (changes requested):

.github/workflows/codeql-analysis.yml
maxhgerlach: I removed this and also went back to installing CMake via apt in the Dockerfiles based on Ubuntu 20.04 (`docker/*/Dockerfile`).

EnricoMi (approved):
LGTM!

nvcastet (approved):

EnricoMi:
@maxhgerlach I have rebased this with master, lets see if we now get the tests terminate

EnricoMi:
@maxhgerlach looks like the last minute change to the `horovod` and `horovod-cpu` images broke Docker image generation. The GitHub action steps take forever (what I observed in this branch before merge, and which is the case for branches rebased to master since):

https://github.com/horovod/horovod/runs/5385819691?check_suite_focus=true
https://github.com/horovod/horovod/runs/5385981381?check_suite_focus=true

maxhgerlach:
Great find, thanks @EnricoMi!

Somehow Github Actions doesn't show me any logs for the hanging steps. Is there any way to get them?

EnricoMi:
Note to self: add timeout to docker build jobs ;-)

maxhgerlach:
> While a job is running, only new lines are shown.
And the only way to get to see those appears to be to keep a tab open and to not let your computer go to sleep...

PR #3375, opened 2022-01-20
fix resume logger

irasit:
Signed-off-by: Peng Zhang <pengz@uber.com>

## Description
The comet logger will fail in local if change save_dir without resume experience. 
Need to always resume the experience with the new save_dir path in arguments.

### DIFF
--- a/horovod/spark/lightning/remote.py
+++ b/horovod/spark/lightning/remote.py
@@ -123,15 +123,26 @@ def RemoteTrainer(estimator, metadata, ckpt_bytes, run_id, dataset_idx, train_ro
                 train_logger = TensorBoardLogger(logs_path)
                 print(f"Setup logger: Using TensorBoardLogger: {train_logger}")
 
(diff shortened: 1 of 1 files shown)


horovod/spark/lightning/remote.py
```diff
@@ -123,15 +123,15 @@ def train(serialized_model):
                 train_logger = TensorBoardLogger(logs_path)
                 print(f"Setup logger: Using TensorBoardLogger: {train_logger}")
 
-            elif isinstance(logger, CometLogger) and logger._experiment_key is None:
-                # Resume logger experiment key if passed correctly from CPU.
+            elif isinstance(logger, CometLogger) and logger_experiment_key is None:
```
Tixxx, line 126: Did you mean "is not None"? this looks like it's creating a new CometLogger with an empty experiment.

irasit: Good catch, yes, should be is not none.

chongxiaoc (approved):

Tixxx (approved):

PR #3377, opened 2022-01-21
Docs: Clarify that Windows is not supported; fix Sphinx build

maxhgerlach:
Some people make futile attempts to build Horovod on their Windows systems and seek help about it, e.g., #3373.


EnricoMi (approved):
Good idea!

PR #3353, opened 2022-01-07
PyTorch: Release Python GIL when yielding in foreground thread

maxhgerlach:
## Checklist before submitting

- [x] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)?
- [ ] Did you update the docs?
- [ ] Did you write any tests to validate this change?  
- [ ] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users?

## Description

Fixes #3352. I've run the test suite from `tests/parallel/test_torch.py` 15 times in the environment described in the issue and did not encounter any hang.

I hope releasing the GIL here is not unsafe in any way.

PR #3379, opened 2022-01-22
re-enable delta optimizer test and fix a bug in adasum communicator init logic

Tixxx:
re-enable delta optimizer test and fix a bug in adasum communicator init logic
Signed-off-by: TJ <tix@uber.com>

## Description
This addresses [issue 3314](https://github.com/horovod/horovod/issues/3314). Torch's delta_optimizer test was disabled due to a deadlock issue with torch 1.10, this is to enable the test with a barrier at the end to properly synchronize. While the main root cause was due to the python GIL which was fixed by this [pr](https://github.com/horovod/horovod/pull/3353), another bug with adasum reduction comm init logic was discovered during debugging, adasum was always initialized relative to mpi_world, the correct relative communicator should be mpi_context->mpi_comm.

Fixes # (issue).
https://github.com/horovod/horovod/issues/3314
## Review process to land 

### DIFF
--- a/horovod/common/ops/adasum/adasum_mpi.cc
+++ b/horovod/common/ops/adasum/adasum_mpi.cc
@@ -28,8 +28,8 @@ AdasumMPI::~AdasumMPI() {
 
 void AdasumMPI::InitializeVHDDReductionComms() {
   int rank, size;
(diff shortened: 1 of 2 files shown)


maxhgerlach (approved):
Great, good to hear that this hang was also fixed by releasing the GIL.

PR #3391, opened 2022-01-30
Pinning pip to 21.0.0 as 22.0.0 cannot install pytorch from html link

EnricoMi:
See https://github.com/pytorch/pytorch/issues/72045.

### DIFF
--- a/Dockerfile.test.cpu
+++ b/Dockerfile.test.cpu
@@ -49,7 +49,10 @@ RUN apt-get update -qq && apt-get install -y python${PYTHON_VERSION} python${PYT
 RUN ln -s -f /usr/bin/python${PYTHON_VERSION} /usr/bin/python
 RUN ln -s -f /usr/bin/python${PYTHON_VERSION} /usr/bin/python${PYTHON_VERSION/%.*/}
 RUN wget --progress=dot:mega https://bootstrap.pypa.io/get-pip.py && python get-pip.py && rm get-pip.py
-RUN pip install --no-cache-dir -U --force pip setuptools requests pytest mock pytest-forked parameterized
+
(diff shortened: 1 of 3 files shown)


Dockerfile.test.cpu
```diff
@@ -49,7 +49,10 @@ RUN apt-get update -qq && apt-get install -y python${PYTHON_VERSION} python${PYT
 RUN ln -s -f /usr/bin/python${PYTHON_VERSION} /usr/bin/python
 RUN ln -s -f /usr/bin/python${PYTHON_VERSION} /usr/bin/python${PYTHON_VERSION/%.*/}
 RUN wget --progress=dot:mega https://bootstrap.pypa.io/get-pip.py && python get-pip.py && rm get-pip.py
-RUN pip install --no-cache-dir -U --force pip setuptools requests pytest mock pytest-forked parameterized
+
+# pinning pip to 21.0.0 as 22.0.0 cannot fetch pytorch packages from html linl
+# https://github.com/pytorch/pytorch/issues/72045
+RUN pip install --no-cache-dir -U --force pip~=21.0.0 setuptools requests pytest mock pytest-forked parameterized
```
pradyunsg, line 55:
```suggestion
RUN pip install --no-cache-dir -U --force pip~=21.0 setuptools requests pytest mock pytest-forked parameterized
```

Otherwise, you'll get 21.0.*, which is probably not what you want.

pradyunsg (changes requested):

Dockerfile.test.cpu
EnricoMi: Even a fixed version that just works would be fine for the time being until pytorch fixes their HTML file.

maxhgerlach (approved):

tgaddair (approved):

PR #3395, opened 2022-02-01
Do apt upgrade in released docker images

EnricoMi:
Let's see if this removes known issues.


EnricoMi:
Doesn't, the base image is almost up-to-date.

PR #3409, opened 2022-02-14
Deprecate use of np, min_np, max_np, replace with num_proc

EnricoMi:
Horovod uses `np` and `num_proc` synonymously across APIs, but not consistently one or the other.

This deprecates the use of `np` in favour of the more meaningful `num_proc`. The `np` variants are kept in public APIs to avoid breaking user code, while a deprecation warning is given. The deprecated variants should be removed in `v1.0.0` or `v1.1.0`.

### DIFF
--- a/docs/elastic.rst
+++ b/docs/elastic.rst
@@ -294,7 +294,7 @@ to allow Horovod to discover available hosts is to provide a ``--host-discovery-
 
 .. code-block:: bash
 
-    $ horovodrun -np 8 --host-discovery-script discover_hosts.sh python train.py
+    $ horovodrun --num-proc 8 --host-discovery-script discover_hosts.sh python train.py
 
 The host discovery script must have user executable permissions, and return one host with its available slots per line
 of the form: ``<hostname>:<slots>``.  For example:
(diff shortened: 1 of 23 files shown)


docs/elastic.rst
```diff
@@ -294,7 +294,7 @@ to allow Horovod to discover available hosts is to provide a ``--host-discovery-
 
 .. code-block:: bash
 
-    $ horovodrun -np 8 --host-discovery-script discover_hosts.sh python train.py
+    $ horovodrun --num-proc 8 --host-discovery-script discover_hosts.sh python train.py
```
tgaddair, line 297: For the docs, I personally prefer keeping `-np` here, which will be very intuitive for mpi users and reinforce the fact that `horovodrun` is analogous to `mpirun`. Curious to get others' take on it.

EnricoMi: I kept the examples everywhere else because there was only the `-np` and not the `--min-np` / `--max-np`.

Here, in elastic, `--num-proc` comes together with `--min-num-proc` and `--max-num-proc`, and `mpirun` does not have those, so I went for this.

But I am also fine with reverting the `elastic.rst`.

maxhgerlach (approved):

nvcastet (approved):

PR #3426, opened 2022-02-27
Add TF 2.8.0 to test space, remove TF 2.5.x, upgrade patch versions

EnricoMi:
Upgrades to TF 2.7.1, Torch 1.10.2, Spark 3.1.3 and 3.2.1


EnricoMi:
Using torchvision in CPU test images without `+cpu` breaks `Spark PyTests` and `Spark Lightning MNIST` tests for some tensorflow versions: https://github.com/horovod/horovod/runs/5363000181?check_suite_focus=true#step:49:63

```
[libprotobuf FATAL google/protobuf/stubs/common.cc:83] This program was compiled against version 3.9.2 of the Protocol Buffer runtime library, which is not compatible with the installed version (3.19.4).  Contact the program author for an update.  If you compiled the program yourself, make sure that your headers are from the same version of Protocol Buffers as your link-time library.  (Version verification failed in "bazel-out/k8-opt/bin/tensorflow/core/framework/tensor_shape.pb.cc".)
```

This is fixed by e50bdf3c.

maxhgerlach (approved):
LGTM, provided that tests pass

PR #3430, opened 2022-02-28
Fix flaky ray tests

EnricoMi:
Ray tests assert that available ray resources after the test are identical to before the test. This turns out to be flaky.

This removes the `check_resources` assertion as it has race conditions. Further, this restores the `CUDA_VISIBLE_DEVICES` environment variable after tests finish that modify it.

And this improves assertion context to further debug the issue of 8 GPUs though 4 workers have been started.


test/single/test_ray.py
```diff
@@ -38,19 +38,27 @@ def ray_start_4_cpus():
...
-    yield address_info
-    # The code after the yield will run as teardown code.
-    ray.shutdown()
-    del os.environ["CUDA_VISIBLE_DEVICES"]
+    try:
+        yield address_info
+        # The code after the yield will run as teardown code.
+        ray.shutdown()
+    finally:
+        if orig_devices:
```
tgaddair, line 58: `None` and `""` have different meanings for `CUDA_VISIBLE_DEVICES`. Probably safer to be explicit:

```
if orig_devices is not None:
    ...
```

tgaddair (approved):

test/single/test_ray.py
EnricoMi: Alright, will better do that then.

PR #3434, opened 2022-03-01
Use pytest-xdist instead of pytest-forked

EnricoMi:
This runs all tests that previously ran with `--forked` in a similar separate process but this time multiple in parallel.


EnricoMi:
Looks like the elastic spark tests cannot run multiple times on one host. They spin up a Spark standalone cluster, where there cannot be multiple on a host (there could).

Also some other non-Spark elastic tests fail.

PR #3446, opened 2022-03-03
Bump version to 0.24.1

EnricoMi:


EnricoMi:
@tgaddair any particular reason to have all those empty lines between the bullets? I think without MD renders nicer: https://github.com/horovod/horovod/blob/u0241/CHANGELOG.md

tgaddair (approved):

tgaddair:
@EnricoMi I believe the keep a changelog folks used to recommend it to help with avoiding merge conflicts, but I don't see anything about that now, so maybe fine to remove.

EnricoMi:
Merge conflicts, that is a good point. We could have the empty lines for the `[UNRELEASED]` section and remove them on release.

PR #3452, opened 2022-03-05
Update min framework versions, test that Horovod builds with them

### LINKED ISSUE #3453: Horovod does not compile with old framework versions that setup.py accept
Building Horovod, the log says:

But actually building Horovod against PyTorch 1.2.0, it fails with:

```
    /tmp/pip-req-build-13fs79hx/horovod/torch/mpi_ops_v2.cc:61:12: error: ‘class at::Tensor’ has no member named ‘floor_divide_’
         tensor.floor_divide_(divisor);
                ^~~~~~~~~~~~~
```

Looks like the minimum PyTorch that Horovod compiles with is 1.5.0.

EnricoMi:
Horovod detects versions of ML frameworks while enforcing a minimum version. These minimum versions may be lower than our lowest test versions (which are tested that Horovod fully works with) but may still compile and partially work for users.

This PR tests that those minimal versions actually compile, but does not test that Horovod **fully** supports them.

Fixes #3453.


maxhgerlach (approved):
Looks great to me!

Issue #3458, opened 2022-03-07
Maybe a bug in group allreduce


Infi-zc:
**Environment:**
1. Framework: (TensorFlow, Keras, PyTorch, MXNet) Tensorflow
2. Framework version: v1.15.4
3. Horovod version: v0.21.1
4. MPI version: 4.1.0
5. CUDA version: - 
6. NCCL version: - 
7. Python version: 3.7.8
8. Spark / PySpark version: - 
9. Ray version: - 
10. OS and version: ubuntu 18.04
11. GCC version: 7.5
12. CMake version:  - 

**Checklist:**
1. Did you search issues to find if somebody asked this question before? yes
2. If your question is about hang, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/running.rst)? yes
3. If your question is about docker, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/docker.rst)? yes
4. Did you check if you question is answered in the [troubleshooting guide](https://github.com/horovod/horovod/blob/master/docs/troubleshooting.rst)? yes

**Bug report:**
I'm trying to run a google bert model written by tensorflow v1.15 and horovod, butI find that when num_groups is set to larger than 1,  Tensorflow will report an error of shape mismatch, like:
```bash
hvd optimizer in model scripts : 
optimizer = hvd.DistributedOptimizer(optimizer, num_groups = 6)

the trackback is : 
[1,11]<stderr>: File "run_squad.py", line 674, in model_fn
[1,11]<stderr>: total_loss, learning_rate, num_train_steps, num_warmup_steps, use_tpu, FLAGS.use_horovod)
[1,11]<stderr>: File "/xxxx/google_bert/optimization.py", line 82, in create_optimizer
[1,11]<stderr>: grads_and_vars = optimizer.compute_gradients(loss, tvars)
[1,11]<stderr>: File "/usr/local/python3/lib/python3.7/site-packages/horovod/tensorflow/__init__.py", line 462, in compute_gradients
[1,11]<stderr>: avg_grads = self._allreduce_grads(grads)
[1,11]<stderr>: File "/usr/local/python3/lib/python3.7/site-packages/horovod/tensorflow/__init__.py", line 383, in allreduce_grads
[1,11]<stderr>: postscale_factor=postscale_factor)
[1,11]<stderr>: File "/usr/local/python3/lib/python3.7/site-packages/horovod/tensorflow/__init__.py", line 268, in _grouped_allreduce_cond
[...]

PR #3459, opened 2022-03-07
Fix gradient and var name mismatch when num_groups > 0

Infi-zc:

## Description

Fixes [Mismatch of grads and vars in group allreduce](https://github.com/horovod/horovod/issues/3458)

## Review process to land 

### DIFF
--- a/horovod/tensorflow/__init__.py
+++ b/horovod/tensorflow/__init__.py
@@ -464,6 +464,7 @@ if _LegacyOptimizer is not None:
             super(_DistributedOptimizer, self).__init__(name=name, use_locking=use_locking)
 
             self._optimizer = optimizer
+            self._groups = groups
             self._allreduce_grads = _make_allreduce_grads_fn(
                 name, device_dense, device_sparse, compression, sparse_as_dense, op,
                 gradient_predivide_factor, groups, process_set=process_set)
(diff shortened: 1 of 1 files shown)


horovod/tensorflow/__init__.py
```diff
@@ -495,6 +496,8 @@ def compute_gradients(self, *args, **kwargs):
             """
             gradients = self._optimizer.compute_gradients(*args, **kwargs)
             grads, vars = zip(*gradients)
+            if self._group is not None and not isinstance(self._groups, list) and self._group > 0:
```
maxhgerlach, line 499: Could you motivate the logic in this condition a bit?

`self._group` is probably a typo for `self._groups`?

Is this fix not necessary for the case where `self._groups` is a list?

maxhgerlach (changes requested):
Thanks for giving this a go! 

It would be good to add a unit test to `tests/parallel/test_tensorflow.py` that fails without the fix.

Note that you would need to sign the [Developer Certificate of Origin](https://probot.github.io/apps/dco/) by [signing off your commits](https://github.com/horovod/horovod/pull/3436/checks?check_run_id=5376922918).

horovod/tensorflow/__init__.py
Infi-zc: Sorry for the typo here, I have corrected it. 
Yes, I think this fix is unnecessary for the case where self._groups is a list. In this case (self._groups: list) the code handles both var name and grad, which will not lead to the error discribed in the issue. 
Next I will try to add a unit test to test_tensorflow.py to reproduce the problem.

EnricoMi:
@maxhgerlach @Infi-zc what is the status of this PR?

maxhgerlach:
I haven't quite understood why the problem does not appear in case the `groups` are specified as an explicit list (there is some related discussion in the PR that introduced the feature: https://github.com/horovod/horovod/pull/2523#issuecomment-771391796). A test that showcases the problem could be helpful here.

PR #3487, opened 2022-03-22
skip environment variables with dot ('.')

yundai424:

- [✓] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)?
- [ NA ] Did you update the docs?
- [✓] Did you write any tests to validate this change?  
`horovodrun -np 1 -H localhost:1 FOO.BAR=1 echo hello` would fail without this change with `FOO.BAR=1: command not found`, and would succeed after this change
- [NA] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users?

## Description

Bash doesn't support environment variables with dots. We can make them as non-exportable so the job will not fail with invalid environment variables.

## Review process to land 

### DIFF
--- a/horovod/runner/common/util/env.py
+++ b/horovod/runner/common/util/env.py
@@ -21,7 +21,7 @@ from horovod.runner.common.util import secret
 LOG_LEVEL_STR = ['FATAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG', 'TRACE']
 
 # List of regular expressions to ignore environment variables by.
-IGNORE_REGEXES = {'BASH_FUNC_.*', 'OLDPWD', secret.HOROVOD_SECRET_KEY}
+IGNORE_REGEXES = {'BASH_FUNC_.*', 'OLDPWD', secret.HOROVOD_SECRET_KEY, '.*\..*'}
 
 KUBEFLOW_MPI_EXEC = '/etc/mpi/kubexec.sh'
 


EnricoMi:
You are right, the command:

    $ horovodrun -np 1 -H localhost:1 FOO.BAR=1 echo hello
    [0]<stderr>:/bin/sh: 1: FOO.BAR=1: not found

fails because `FOO.BAR=1 echo hello` is not a proper command.

Prepend the command with `env`, which is capable of defining an environment variable. So

    $ horovodrun -np 1 -H localhost:1 env FOO.BAR=1 env
    [0]<stdout>:FOO.BAR=1

works fine with Bash 5, even the `.` is allowed.

So I wouldn't say Horovod needs adjustment. Only use environment variables that your cluster can handle, or upgrade your installed bash. You could even filter the list of environment variables before constructing your `horovodrun` command.

EnricoMi:
Your example is misleading, as it can be fixed to work (see above).

The following example does not workwith Gloo:

    env FOO.BAR=1 horovodrun -np 1 -H localhost:1 --gloo --verbose env

It works perfectly fine with MPI:

    env FOO.BAR=1 horovodrun -np 1 -H localhost:1 --mpi --verbose env

So if the environment variable is around (for whatever reason), `horovodrun` cannot start the remote command, because the environment variable is being forwarded to the target machine, and the way this is done breaks the command to run.

EnricoMi:
If gloo uses `env` as the remote shell, then `horovodrun` works with `FOO.BAR` present. Even on the remote host the variable then exists in bash. See #3489.

yundai424:
Hi @EnricoMi thanks for the insight! Agree that #3489 is a better approach to solving the issue. I'll close this PR.

PR #3489, opened 2022-03-22
Use env to set environment vars in remote shell

EnricoMi:
Some environment variables on the host that runs `horovodrun` might not be supported by the remote shell (#3487). Looks like `env` does support those.


yundai424 (approved):

Tixxx (approved):

EnricoMi:
@tgaddair do you see problems with gloo / ssh using `env` as the shell to run the remote command? It is more capable of setting the environment variables than `bash`.

tgaddair (approved):

PR #3490, opened 2022-03-23
Fixes empty hostname returned from HostDiscoveryScript

plliao:

## Description

Fixes empty hostname returned from HostDiscoveryScript and make sure hostname is not empty.

Context:
When discovery script return nothing, HostDiscoveryScript will treat the empty result i.e. "" string as the available host.
It will fail training job as launcher cannot ssh on the host with empty address. Discovery script can return nothing when no workers are available or new workers are not ready yet.

## Review process to land 

### DIFF
--- a/test/single/test_elastic_discovery.py
+++ b/test/single/test_elastic_discovery.py
@@ -37,6 +37,9 @@ class HostDiscoveryScriptTests(unittest.TestCase):
         ["host-1:2", {"host-1": 2}],
         ["host-1\nhost-5", {"host-1": DEFAULT_SLOTS, "host-5": DEFAULT_SLOTS}],
         ["host-1:2\nhost-4:3", {"host-1": 2, "host-4": 3}],
+        ["\nhost-1:2\nhost-4:3", {"host-1": 2, "host-4": 3}],
+        ["host-1:2\n\nhost-4:3", {"host-1": 2, "host-4": 3}],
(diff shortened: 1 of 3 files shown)


test/single/test_elastic_discovery.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super(HostDiscoveryScriptTests, self).__init__(*args, **kwargs)
+
+    @parameterized.expand([
+        ["", {}],  # No host found.
+        ["host-1", {"host-1": DEFAULT_SLOTS}],
+        ["host-1:2", {"host-1": 2}],
+        ["host-1\nhost-5", {"host-1": DEFAULT_SLOTS, "host-5": DEFAULT_SLOTS}],
+        ["host-1:2\nhost-4:3", {"host-1": 2, "host-4": 3}],
```
EnricoMi, line 39: Can we also add tests where multiple hosts exist, but also an empty line, because the new code should handle that case gracefully as well:

    ["\nhost-1:2\nhost-4:3", ...],
    ["host-1:2\n\nhost-4:3", ...],
    ["host-1:2\nhost-4:3\n", ...],

EnricoMi (changes requested):

test/single/test_elastic_discovery.py
plliao: Sure! I added those cases. Please review again. Thanks!

EnricoMi (approved):
LGTM!

Issue #2846, opened 2021-04-19
Make Horovod build with OneCCL in GitHub CI


EnricoMi:
Moving the CI to GitHub Workflow, Horovod does not build with OnceCCL anymore. This has to be investigated and resolved here.

ashiqimranintel:
Is there any update in this issue?

EnricoMi:
@shirosankaku @mshiryaev do you want to look into adding oneccl test coverage to our CI?

i-kosarev:
Hi @EnricoMi! Yes, we would like to add oneCCL coverage back to CI

EnricoMi:
@i-kosarev glad to hear that, I managed to get the OneCCL test image build on GitHub, so we can put it back into the CI. But some tests fail (5) for OneCCL. If you could look into these and see if you can fix them: https://github.com/horovod/horovod/runs/5728199621

EnricoMi:
As @francares [pointed out](https://github.com/horovod/horovod/pull/3492#issuecomment-1082020956), OneCCL should use a fixed version (e.g. https://github.com/oneapi-src/oneCCL/tree/2021.5.2), not master.

i-kosarev:
> @i-kosarev glad to hear that, I managed to get the OneCCL test image build on GitHub, so we can put it back into the CI. But some tests fail (5) for OneCCL. If you could look into these and see if you can fix them: https://github.com/horovod/horovod/runs/5728199621
> 
> You can use this branch to get tests fixed: [branch-ci-add-oneccl-image](https://github.com/horovod/horovod/tree/branch-ci-add-oneccl-image)

Hi @EnricoMi! Sorry for the very long wait. I have linked the PR which is now merged. From my side I see that this fix make all the test for oneCCL pass. I think this unblocks proceeding oneCCL integration. I am not sure what are the next steps, but I can try to assist if that can be helpful

PR #3492, opened 2022-03-24
Fix broken Dockerfile due to change in oneCCL deps paths

francares:

## Description

When trying to build a Docker image using oneCCL as dependency, it fails due to a path change in latest master oneCCL repository (https://github.com/oneapi-src/oneCCL).

## Review process to land 

### DIFF
--- a/.github/gen-workflow-ci.py
+++ b/.github/gen-workflow-ci.py
@@ -296,10 +296,7 @@ def main():
                            f''.join([f'            {test}: true\n'
                                      for test in sorted(list(tests_per_image.get(image, [])))]) +
                            f'            build_timeout: {30 if "-cpu-" in image else 40}\n'
-                           for image in sorted(images)
-                           # oneccl does not compile on GitHub Workflows:
-                           # https://github.com/horovod/horovod/issues/2846
-                           if '-oneccl-' not in image]) +
+                           for image in sorted(images)]) +
                 f'\n'
                 f'    steps:\n'
                 f'      - name: Clean up disk space\n'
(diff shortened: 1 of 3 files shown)


EnricoMi:
Thanks for fixing this. Why is the oneccl installation doing all this copying anyway? Their README states, installation is as simple as

    mkdir build
    cd build
    cmake ..
    make -j install

https://github.com/oneapi-src/oneCCL#installation

EnricoMi:
It even has CMake integration: https://github.com/oneapi-src/oneCCL#using-oneccl-package-from-cmake

francares:
> Thanks for fixing this. Why is the oneccl installation doing all this copying anyway? Their README states, installation is as simple as
> 
> ```
> mkdir build
> cd build
> cmake ..
> make -j install
> ```
> 
> https://github.com/oneapi-src/oneCCL#installation

@EnricoMi, these steps are already there (https://github.com/horovod/horovod/blob/4b8cc49826e21abc84112e43e90b46258a7af528/Dockerfile.test.cpu#L104) but you are right, the steps related to copying dependencies are not needed. I'm going to remove those steps from the Docker file.
I'm also going to add a new step to remove `oneccl.tar.gz` and the tmp build folder `/tmp/oneCCL-${CCL_PACKAGE}/` after install.
OTH, I'm seeing the oneCCL [build](<https://github.com/horovod/horovod/runs/5688852371?check_suite_focus=true>) is stuck. I can't see the details on why. Can you take a look to it?

EnricoMi:
That is the strange bit: GitHub CI cannot build that image, it gets stuck. Looks like the VM itself dies: https://github.com/horovod/horovod/issues/2846

francares:
@EnricoMi Good catch on the cmake flag. Thanks for fixing it. I'm seeing now that some tests from `MPI_Cluster_PyTests_ONECCL_MPI` and `MPI_Cluster_PyTests_ONECCL_OFI` are failing with an unexpected error code 255 (expected error code is 1). It might be due to `HOROVOD_CPU_OPERATIONS=CCL` is not set. I'm working on it.

EnricoMi:
Good point, could be part of #2846.

PR #3495, opened 2022-03-25
Replace '+' with '.' to get more consistent package version with the …

arde171:
…environment HOROVOD_LOCAL_VERSION set

## Description

Replace '+' with '.' to get more consistent package version when the environment HOROVOD_LOCAL_VERSION is set.

Issue: Presently,  `get_package_version()` appended the HOROVOD_LOCAL_VERSION using '+'.  For example, with `HOROVOD_LOCAL_VERSION` is set to `1` and the horovod version is `0.24.2`, the package version becomes `0.24.2+1`. 
By replacing `+` with `.`,  we get more consistent package name  (e.g., `0.24.2.1`) for our build process.

## Review process to land 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - MXNet: Updated allreduce functions to newer `op` API. ([#3299](https://github.com/horovod/horovod/pull/3299))
+- Replace '+' with '.' to get more consistent package version when the environment HOROVOD_LOCAL_VERSION is set. ([#3495](https://github.com/horovod/horovod/pull/3495))
 
 ### Deprecated
 
--- a/setup.py
+++ b/setup.py
@@ -203,7 +203,7 @@ if not os.environ.get('HOROVOD_WITHOUT_PYTORCH'):
 
 
 def get_package_version():
-    return __version__ + "+" + os.environ['HOROVOD_LOCAL_VERSION'] if 'HOROVOD_LOCAL_VERSION' in os.environ else __version__
+    return __version__ + "." + os.environ['HOROVOD_LOCAL_VERSION'] if 'HOROVOD_LOCAL_VERSION' in os.environ else __version__
 
 
 setup(name='horovod',


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - MXNet: Updated allreduce functions to newer `op` API. ([#3299](https://github.com/horovod/horovod/pull/3299))
+- Replace '+' with '.' to get more consistent package version when the environment HOROVOD_LOCAL_VERSION is set. ([#3495](https://github.com/horovod/horovod/pull/3495))
```
EnricoMi, line 16: This `+` refers to build metadata, not version increment; https://semver.org/#spec-item-10

In which situations is this `HOROVOD_LOCAL_VERSION` being used?

setup.py
```diff
@@ -203,7 +203,7 @@ def build_extensions(self):
 
 
 def get_package_version():
-    return __version__ + "+" + os.environ['HOROVOD_LOCAL_VERSION'] if 'HOROVOD_LOCAL_VERSION' in os.environ else __version__
+    return __version__ + "." + os.environ['HOROVOD_LOCAL_VERSION'] if 'HOROVOD_LOCAL_VERSION' in os.environ else __version__
```
EnricoMi, line 206: While touching this line, I would suggest to hardening it against `HOROVOD_LOCAL_VERSION=`
```suggestion
    return __version__ + "." + os.environ['HOROVOD_LOCAL_VERSION'] if os.environ.get('HOROVOD_LOCAL_VERSION') else __version__
```

EnricoMi (changes requested):

EnricoMi:
This `+` refers to build metadata, not version increments: https://semver.org/#spec-item-10

This `HOROVOD_LOCAL_VERSION` is used to build a package with version differing from a release version (a local version). A "local" version `0.24.2.1` would collide with a prospect future official release with the same version, hence the `+`.

You could use the git SHA hash of HEAD, e.g. `0.24.2+6964eb4`.

EnricoMi:
What is your use case for having your own "consistent" Horovod build release versions? What qualifies as "consistent" for you?

EnricoMi:
Please re-open if you want to follow up on this.

PR #3496, opened 2022-03-25
Fixes gradient update timing in TF AggregationHelperEager

plliao:

## Description

Fixes gradient update timing in TF LocalGradientAggregationHelperEager.

Why?
The gradient updating timing in LocalGradientAggregationHelperEager was set one step before the locally aggregated gradients are allreduced. Therefore, the applied gradient are the locally aggregated gradient instead of the allreduced gradient. This PR fixes the issue.

The PR also improves the gradient aggregation test case to cover tf.IndexedSlices gradient and validates the updated values. Before the change, the grad and values are always zero which does not test anything. 

## Review process to land 

### DIFF
--- a/horovod/tensorflow/gradient_aggregation_eager.py
+++ b/horovod/tensorflow/gradient_aggregation_eager.py
@@ -141,11 +141,6 @@ class LocalGradientAggregationHelperEager:
             return increment_optimizer_iteration()
 
         def is_aggregation_step():
-            if _POST_TF_2_4_0:
-                # In TF 2.4+ we evaluate whether it's time to aggregated gradients before
-                # calling `_aggregate_gradients()`.
-                return tf.equal(tf.add(self.counter, 1), self.backward_passes_per_step)
-
             return tf.equal(self.counter, 0)
 
         return tf.cond(
--- a/test/parallel/test_tensorflow2_keras.py
+++ b/test/parallel/test_tensorflow2_keras.py
@@ -22,12 +22,13 @@ import os
 import warnings
 
 from distutils.version import LooseVersion
+from parameterized import parameterized
 
 import pytest
 
 from tensorflow import keras
 
-from horovod.common.util  import is_version_greater_equal_than
+from horovod.common.util import is_version_greater_equal_than
 
 if is_version_greater_equal_than(tf.__version__, "2.6.0"):
     if LooseVersion(keras.__version__) < LooseVersion("2.9.0"):
@@ -70,6 +71,7 @@ class Tf2KerasTests(tf.test.TestCase):
         initial_lr = 0.1 * hvd.size()
         opt = tf.keras.optimizers.Adam()
         opt = hvd.DistributedOptimizer(opt)
+
         def linear_multiplier(epoch):
             return epoch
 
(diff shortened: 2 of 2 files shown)


Tixxx:
I'm not entirely sure if we should remove it. From the comments in "non_aggregation_step":

>                 # In TF 2.4+ `_aggregate_gradients()` is called from inside of `apply_gradients()`.
                # We account for this by calling `_aggregate_gradients()` for steps where we do
                # not call `apply_gradients()`.

My understanding is that they return 'true' 1 step early to let apply_gradients() do the final step which will then call _allreduce again which returns reduced tensors.
@tgaddair  @aaron276h

plliao:
> I'm not entirely sure if we should remove it. From the comments in "non_aggregation_step":
> 
> > ```
> >             # In TF 2.4+ `_aggregate_gradients()` is called from inside of `apply_gradients()`.
> > ```
> 
> ```
>                # We account for this by calling `_aggregate_gradients()` for steps where we do
>                # not call `apply_gradients()`.
> ```
> 
> My understanding is that they return 'true' 1 step early to let apply_gradients() do the final step which will then call _allreduce again which returns reduced tensors. @tgaddair @aaron276h

I think behavior was changed in this PR. After TF 2.4, the gradients are allreduced in `_compute_gradients` instead of `_aggregate_gradients`. Hence, the applied gradients is not reduced and averaged.
https://github.com/horovod/horovod/pull/2647

Tixxx:
> > I'm not entirely sure if we should remove it. From the comments in "non_aggregation_step":
> > > ```
> > >             # In TF 2.4+ `_aggregate_gradients()` is called from inside of `apply_gradients()`.
> > > ```
> > 
> > 
> > ```
> >                # We account for this by calling `_aggregate_gradients()` for steps where we do
> >                # not call `apply_gradients()`.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > My understanding is that they return 'true' 1 step early to let apply_gradients() do the final step which will then call _allreduce again which returns reduced tensors. @tgaddair @aaron276h
> 
> I think behavior was changed in this PR. After TF 2.4, the gradients are allreduced in `_compute_gradients` instead of `_aggregate_gradients`. Hence, the applied gradients is not reduced and averaged. #2647

`_compute_gradients` then calls `_allreduce` which calls agg helper's `compute_gradients` -> increment counter by 1 -> counter == backward_steps -> `_all_reduce_and_clear_aggregated_variables`, finally returns with reduced gradients. Did I miss anything in the flow?

Tixxx:
> > I'm not entirely sure if we should remove it. From the comments in "non_aggregation_step":
> > > ```
> > >             # In TF 2.4+ `_aggregate_gradients()` is called from inside of `apply_gradients()`.
> > > ```
> > 
> > 
> > ```
> >                # We account for this by calling `_aggregate_gradients()` for steps where we do
> >                # not call `apply_gradients()`.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > My understanding is that they return 'true' 1 step early to let apply_gradients() do the final step which will then call _allreduce again which returns reduced tensors. @tgaddair @aaron276h
> 
> I think behavior was changed in this PR. After TF 2.4, the gradients are allreduced in `_compute_gradients` instead of `_aggregate_gradients`. Hence, the applied gradients is not reduced and averaged. #2647

I see, thanks for the explanation.

shaowei-su:
Quick follow up on this issue: @Tixxx could you help push this issue fixes (with the other PR: https://github.com/horovod/horovod/pull/3499) into production releases? the latest release `0.24.3` does not seem to come with this fix. Thanks!

Tixxx:
> Quick follow up on this issue: @Tixxx could you help push this issue fixes (with the other PR: #3499) into production releases? the latest release `0.24.3` does not seem to come with this fix. Thanks!

Done, added the two PRs to 0.25.0 release milestone. You can check them out [here](https://github.com/horovod/horovod/milestone/4?closed=1)

shaowei-su:
Thank you @Tixxx for your quick responses and help!

PR #3503, opened 2022-04-04
Add elastic run api

EnricoMi:
Currently, the elastic training mode can only be used through `horovodrun` and not the existing `horovod.run` API.

This allows to run `horovod.run` with `min_num_proc` or `host_discovery_script` set to run a `func` in elastic  mode.


tgaddair (approved):

EnricoMi:
I managed to move the `KVStoreServer` code from `launch.py` into `gloo_run.py`, which is where some remote hosts are known and the common interface can be determined. This removes the ugly need to provide all driver ips to `run_task.py` with very low timeout.

PR #3505, opened 2022-04-06
clear locally accumulated gradient by assigning with zeros_like to avoid infinite gradient not correctly cleared up

### LINKED ISSUE #3504: hvd.DistributedOptimizer gradient accumulation doesn't clean up infinite gradient correctly
**Environment:**
1. Framework: (TensorFlow, Keras, PyTorch, MXNet) Keras
2. Framework version: 2.4
3. Horovod version: 2.3
4. MPI version: 
5. CUDA version: 
6. NCCL version:
7. Python version:
8. Spark / PySpark version:
9. Ray version:
10. OS and version:
11. GCC version:
12. CMake version:

**Checklist:**

**Bug report:**

We were training in TensorFlow [FP16 mixed precision](https://www.tensorflow.org/guide/mixed_precision) with keras `model.fit()` and with gradient accumulation/aggregation (`backward_pass_per_step` in `hvd.DistributedOptimizer`) and noticed that the [GradientAggregationHelperEager](https://github.com/horovod/horovod/blob/master/horovod/tensorflow/gradient_aggregation_eager.py#L8) doesn't work correctly with FP16 when the loss goes infinite. Details:

It is kind of expected that at the very first 2-15 steps of the training, the gradient out of TF [LossScaleOptimizer](https://github.com/keras-team/keras/blob/v2.8.0/keras/mixed_precision/loss_scale_optimizer.py#L258-L844) is infinite (because the default initial loss scale factor is as large as `2**15`). Dynamic LSO can handle this gracefully, it just skips applying gradient of that step and divides the scale factor by half. However horovod GradientAggregationHelper will anyway add the infinite gradient up locally, and the infinite gradient will never be correctly cleaned up in [this way](https://github.com/horovod/horovod/blob/133ef0725253db83cfb82a4ed4003df76d189829/horovod/tensorflow/gradient_aggregation_eager.py#L119-L123):
```
    def _clear_vars(self):
        self.counter.assign(0)
        for idx in self.locally_aggregated_grads.keys():
            self.locally_aggregated_grads[idx].assign_add(
                -1 * self.locally_aggregated_grads[idx])
```

as the result of adding inf value by its negative val will be NaN.

yundai424:

## Description

Fixes #3504

When clearing the locally aggregated/added gradient, we should rather assign with `tf.zeros_like`, because substracting by the local aggregated value doesn't work for infinite gradients.

Added a unit test for this. Without the change it'll fail.

## Review process to land 

### DIFF
--- a/horovod/tensorflow/gradient_aggregation_eager.py
+++ b/horovod/tensorflow/gradient_aggregation_eager.py
@@ -119,8 +119,8 @@ class LocalGradientAggregationHelperEager:
     def _clear_vars(self):
         self.counter.assign(0)
         for idx in self.locally_aggregated_grads.keys():
-            self.locally_aggregated_grads[idx].assign_add(
-                -1 * self.locally_aggregated_grads[idx])
+            self.locally_aggregated_grads[idx].assign(
+                tf.zeros_like(self.locally_aggregated_grads[idx]))
 
     def apply_gradients(self, apply_grads_closure, optimizer, *args, **kwargs):
         def increment_optimizer_iteration():
--- a/test/parallel/test_tensorflow2_keras.py
+++ b/test/parallel/test_tensorflow2_keras.py
@@ -182,6 +182,33 @@ class Tf2KerasTests(tf.test.TestCase):
         aggregation_counter = opt._agg_helper.counter.numpy()
         assert aggregation_counter == training_steps % backward_passes_per_step
 
+    def test_grad_aggregation_with_inf_grad(self):
+        backward_passes_per_step = 2
+        step_count = tf.Variable(0, trainable=False, dtype=tf.int32)
+        opt = tf.optimizers.SGD()
+        opt = hvd.DistributedOptimizer(
+            opt, 
+            backward_passes_per_step=backward_passes_per_step,
+            sparse_as_dense=True
+        )
+        x = tf.Variable(0.)
+        var = [x]
+
+        def loss():
+            step_count.assign_add(1)
+            return tf.cond(
+                pred=tf.greater(step_count, 1),
+                true_fn=lambda: x,
+                false_fn=lambda: x * float('inf')
+            )
+        for _ in range(2 * backward_passes_per_step):
+            # in the first aggregation cycle the gradient is infinite,
+            # and it should be cleaned up to zero after apply_gradients
+            # and doesn't affect the 2nd aggregation cycle
+            grads_and_vars = opt._compute_gradients(loss=loss, var_list=var)
+            opt.apply_gradients(grads_and_vars)
+        assert tf.math.is_finite(grads_and_vars[0][0])
+
     def test_from_config(self):
         opt = keras.optimizers.Adam()
         hopt = hvd.DistributedOptimizer(opt)


chongxiaoc (approved):
LGTM

Tixxx (approved):

chongxiaoc:
@EnricoMi Mind having a 2nd pair of eye on this minor change to catch inf?

EnricoMi:
I have no experience with that part of the code, I'd rather like to hear @maxhgerlach's opinion.

horovod/tensorflow/gradient_aggregation_eager.py
```diff
@@ -119,8 +119,8 @@ def _allreduce_helper(self, grads, vars):
     def _clear_vars(self):
         self.counter.assign(0)
         for idx in self.locally_aggregated_grads.keys():
-            self.locally_aggregated_grads[idx].assign_add(
-                -1 * self.locally_aggregated_grads[idx])
```
EnricoMi, line 123: Is this a breaking change for existing user code?

maxhgerlach: inf - inf is NaN, so semantically it would definitely be an improvement to assign zero here.

However, I suspect that `tf.assign_add()` has been used here originally to avoid an intermediate extra memory allocation for the result of `tf.zeros_like()`. In the past I've seen a similar effect actually cause perceivable memory waste.

I wonder if this is still the case or if recent releases of TensorFlow can optimize `x.assign(tf.zeros_like(x))` appropriately..

chongxiaoc: @tgaddair Is there a reason of introducing `tf.assign_add()` originally?

Tixxx: @maxhgerlach i had the same doubt initially, but according to [this](https://www.tensorflow.org/guide/tensor), it seems like even the old assign_add created a new buffer. I guess ultimately the proper way is to use the [in-place](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/inplace-update) c++ api which is a bit involved than the scope of the pr.

romerojosh: I would assume:
```
a.assign_add(-1 * a)
```
consumes the same amount of memory as:
```
a.assign(tf.zeros_like(a))
```
since both likely create temporaries, as mentioned by @Tixxx.

Is there a TF API to just set all values of a tensor to a scalar value?

yundai424: We should go into [this](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/assign_op.h#L90-L104) route when assigning zero to local gradient. What happens afterwards will be the same for [Assign](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/dense_update_ops.cc#L39-L40) and [AssignAdd](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/dense_update_ops.cc#L80-L82), the only difference is for assign the update is `params.device(d) = update;` and for assignAdd it's `params.device(d) += update;`. In this sense it looks like pretty much in place. My understanding could be wrong :)

tgaddair (commented):
cc @rb-determined-ai, can you also review? I believe Determined is using this feature, right?

rb-determined-ai (approved):
This all seems very reasonable to me.  I dug up the [original PR](https://github.com/determined-ai/horovod/pull/22/files#diff-51347bf71dea1501689daad891b071e0a53dc77df8c67f146f59e6d6daa9879aR88) where we landed this line of code, back before we upstreamed the feature.  I read through the comments there and didn't see any indication of why it was originally written this way.  I even pinged @aaron276h who reviewed it, but he hasn't responded yet and I'm about 10 minutes from going on vacation.

PR #3506, opened 2022-04-06
Rework network interfaces args in horovod.run and horovodrun

EnricoMi:
Reworks network interfaces arguments in two places: the `horovod.run` API and the `horovodrun` CLI:

The `horovod.run` API deprecates `network_interface` argument in favour of `network_interfaces`, which becomes a list of strings, rather than a comma separated string that is to be split.

The `horovodrun` CLI replaces `--network-interface` with semantically more accurate `--network-interfaces`, which still takes a comma separated string of network interfaces, but also provides the new `--network-interface` argument, that semantically correctly expects a single network interface. That argument can be used multiple times to set multiple NICs.

Arguably, `--network-interfaces` could be removed completely, as `--network-interface` allows to specify multiple NICs. Unless, there is any use of comma separated strings.

### DIFF
--- a/horovod/runner/launch.py
+++ b/horovod/runner/launch.py
@@ -240,6 +240,49 @@ def make_deprecated_bool_action(override_args, bool_value, replacement_option, r
     return StoreOverrideBoolAction
 
 
+def make_nic_action(deprecated):
+    # This is an append Action that splits the values on ','
+    # where that splitting can be marked deprecated, depending on the actual command line argument used
+    class NicAction(argparse.Action):
+        def __init__(self,
+                     option_strings,
+                     dest,
+                     default=None,
+                     type=None,
+                     choices=None,
+                     required=False,
+                     help=None):
+            super(NicAction, self).__init__(
+                option_strings=option_strings,
+                dest=dest,
+                nargs=1,
+                default=default,
+                type=type,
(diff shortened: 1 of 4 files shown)


horovod/runner/launch.py
```diff
@@ -273,10 +316,18 @@ def parse_args():
                              'HOROVOD_START_TIMEOUT can also be used to '
                              'specify the initialization timeout.')
 
-    parser.add_argument('--network-interface', action='store', dest='nics',
+    parser.add_argument('--network-interfaces', action=make_nic_action(deprecated=False), dest='nics',
                         help='Network interfaces that can be used for communication separated by '
                              'comma. If not specified, Horovod will find the common NICs among all '
```
maxhgerlach, line 321: Should it be singular "common NIC" here or should it be "use them" a bit later in the sentence?

maxhgerlach (approved):
LGTM; the new interface is clearer!

Left a minor comment regarding the argument help strings.

horovod/runner/launch.py
EnricoMi: good point, I have simplified that bit

maxhgerlach: great!

PR #3509, opened 2022-04-15
Avoid redundant ptx generation for maximum specififed compute capability

trevor-m:
Use 
```
-gencode=arch=compute_xy,code=\"sm_xy,compute_xy\"
```
instead of
```
-gencode=arch=compute_xy,code=compute_xy
-gencode=arch=compute_xy,code=sm_xy
```
for the highest arch, which avoids generating ptx a second time. This slightly improves nvcc compilation time.

### DIFF
--- a/cmake/build_utils.py
+++ b/cmake/build_utils.py
@@ -109,9 +109,10 @@ def get_nvcc_flags():
 
     # Build native kernels for specified compute capabilities
     cc_list = full_cc_list if cc_list_env is None else [int(x) for x in cc_list_env.split(',')]
-    for cc in cc_list:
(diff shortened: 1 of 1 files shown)


maxhgerlach (approved):
LGTM, thanks for the contribution @trevor-m!

Test failures for framework head versions are unrelated and already fixed on master.

nvcc docs show a slightly different syntax `code=[sm_xy,compute_xy]`, but `code=\"sm_xy,compute_xy\"` works just the same. https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#extended-notation

PR #3510, opened 2022-04-18
Fix bugs in DBFSLocalStore

WeichenXu123:
Signed-off-by: Weichen Xu <weichen.xu@databricks.com>

## Description

Fix bugs in DBFSLocalStore.
1. Make DBFSLocalStore support "file:/dbfs/..." style path
2. Make DBFSLocalStore implement `get_localized_path` correctly.

## Review process to land 

3. At least one member of the [technical steering committee](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md) must review and approve.
4. If any member of the technical steering committee requests changes, they must be addressed.

### DIFF
--- a/horovod/spark/common/store.py
+++ b/horovod/spark/common/store.py
@@ -559,21 +559,23 @@ class DBFSLocalStore(FilesystemStore):
 
     @classmethod
     def matches_dbfs(cls, path):
-        return path.startswith("dbfs:/") or path.startswith("/dbfs/") or path.startswith("file:///dbfs/") \
-                or (path.startswith("file:/dbfs/") and not path.startswith("file://dbfs/"))
+        return (path.startswith("dbfs:/") and not path.startswith("dbfs://")) or \
+               path.startswith("/dbfs/") or \
+               path.startswith("file:///dbfs/") or \
+               path.startswith("file:/dbfs/")
 
     @staticmethod
     def normalize_path(path):
         """
         Normalize the path to the form `/dbfs/...`
         """
-        if path.startswith("dbfs:/"):
+        if path.startswith("dbfs:/") and not path.startswith("dbfs://"):
             return "/dbfs" + path[5:]
         elif path.startswith("/dbfs/"):
             return path
         elif path.startswith("file:///dbfs/"):
             return path[7:]
-        elif path.startswith("file:/dbfs/") and not path.startswith("file://dbfs/"):
+        elif path.startswith("file:/dbfs/"):
             return path[5:]
         else:
             raise ValueError(DBFSLocalStore.DBFS_PATH_FORMAT_ERROR.format(path))
--- a/horovod/spark/common/store.py
+++ b/horovod/spark/common/store.py
@@ -181,8 +181,7 @@ class AbstractFilesystemStore(Store):
         super().__init__()
 
     def exists(self, path):
-        localized_path = self.get_localized_path(path)
-        return self.fs.exists(localized_path) or self.fs.isdir(localized_path)
+        return self.fs.exists(self.get_localized_path(path)) or self.fs.isdir(path)
 
     def read(self, path):
         with self.fs.open(self.get_localized_path(path), 'rb') as f:
--- a/horovod/spark/common/store.py
+++ b/horovod/spark/common/store.py
@@ -533,16 +533,17 @@ class HDFSStore(AbstractFilesystemStore):
         return path.startswith(cls.FS_PREFIX)
 
 
-# If `_DBFS_PATH_MAPPING_TO_PATH` is not None, map `/dbfs/...` path to `{_DBFS_PATH_MAPPING_TO_PATH}/...`
-# This is used in testing.
-_DBFS_PATH_MAPPING_TO_PATH = None
+# If `_DBFS_PREFIX_MAPPING` is not None, map `/dbfs/...` path to `{_DBFS_PREFIX_MAPPING}/...`
+# This is used in testing, and this mapping only applies to `DBFSLocalStore.get_localized_path`
+_DBFS_PREFIX_MAPPING = None
 
 
 class DBFSLocalStore(FilesystemStore):
     """Uses Databricks File System (DBFS) local file APIs as a store of intermediate data and
     training artifacts.
 
-    Initialized from a `prefix_path` starts with `/dbfs/...`, `file:///dbfs/...` or `dbfs:/...`, see
+    Initialized from a `prefix_path` starts with `/dbfs/...`, `file:///dbfs/...`, `file:/dbfs/...`
+    or `dbfs:/...`, see
     https://docs.databricks.com/data/databricks-file-system.html#local-file-apis.
     """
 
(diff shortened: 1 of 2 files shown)


WeichenXu123:
@tgaddair Could you help review this PR ? thanks!

and is it possible to make a patch release after this PR merged ? Thanks!

horovod/spark/common/store.py
```diff
@@ -181,7 +181,8 @@ def __init__(self, prefix_path, train_path=None, val_path=None, test_path=None,
         super().__init__()
 
     def exists(self, path):
-        return self.fs.exists(self.get_localized_path(path)) or self.fs.isdir(path)
+        localized_path = self.get_localized_path(path)
+        return self.fs.exists(localized_path) or self.fs.isdir(localized_path)
```
EnricoMi, line 185: The semantics of `fs.isdir` looks like it includes `fs.exists`, so the ` or self.fs.isdir(localized_path)` looks redundant. Can you please specify in a comment why this bit is needed? Otherwise, someone might remove it in the future.

horovod/spark/common/store.py
```diff
@@ -532,23 +533,37 @@ def matches(cls, path):
...
-        if not prefix_path.startswith("/dbfs/"):
-            warnings.warn("The provided prefix_path might be ephemeral: {} Please provide a "
-                          "`prefix_path` starting with `/dbfs/...`".format(prefix_path))
+        if not DBFSLocalStore.matches_dbfs(prefix_path):
+            raise ValueError(DBFSLocalStore.DBFS_PATH_FORMAT_ERROR.format(prefix_path))
         super(DBFSLocalStore, self).__init__(prefix_path, *args, **kwargs)
 
+    @staticmethod
+    def _get_dbfs_path_format_error(path):
+        return
```
EnricoMi, line 561: this always returns `None`

horovod/spark/common/store.py
```diff
@@ -532,23 +533,37 @@ def matches(cls, path):
...
+
     @classmethod
     def matches_dbfs(cls, path):
-        return path.startswith("dbfs:/") or path.startswith("/dbfs/") or path.startswith("file:///dbfs/")
+        return path.startswith("dbfs:/") or path.startswith("/dbfs/") or path.startswith("file:///dbfs/") \
+                or path.startswith("file:/dbfs/")
 
     @staticmethod
     def normalize_path(path):
```
EnricoMi, line 569: I think this static method is easy to test. Please add a unit test for the use cases of this method.

horovod/spark/common/store.py
```diff
@@ -532,23 +533,37 @@ def matches(cls, path):
...
+# If `_DBFS_PATH_MAPPING_TO_PATH` is not None, map `/dbfs/...` path to `{_DBFS_PATH_MAPPING_TO_PATH}/...`
+# This is used in testing.
+_DBFS_PATH_MAPPING_TO_PATH = None
+
+
 class DBFSLocalStore(FilesystemStore):
     """Uses Databricks File System (DBFS) local file APIs as a store of intermediate data and
     training artifacts.
 
     Initialized from a `prefix_path` starts with `/dbfs/...`, `file:///dbfs/...` or `dbfs:/...`, see
```
EnricoMi, line 545: I think this also needs to mention `file:/dbfs/`

horovod/spark/common/store.py
```diff
@@ -532,23 +533,37 @@ def matches(cls, path):
         return path.startswith(cls.FS_PREFIX)
 
 
+# If `_DBFS_PATH_MAPPING_TO_PATH` is not None, map `/dbfs/...` path to `{_DBFS_PATH_MAPPING_TO_PATH}/...`
+# This is used in testing.
+_DBFS_PATH_MAPPING_TO_PATH = None
```
EnricoMi, line 538: Maybe the naming could be improved here, e.g. `_DBFS_PATH_PREFIX` or `_DBFS_PREFIX_MAPPING`.

The comment should also indicate that this mapping only applies to `get_localized_path`.

EnricoMi (changes requested):

EnricoMi:
> and is it possible to make a patch release after this PR merged ?

I could cut a 0.24.3 release.

horovod/spark/common/store.py
WeichenXu123: I am not sure why it includes `or self.fs.isdir(localized_path)` here, it is in `AbstractFilesystemStore`, maybe other filesystem (such as HDFS) require this ? 
For DBFSLocal and local fs, `fs.isdir` is redundant.

WeichenXu123: I would like to keep calling `fs.isdir` in `AbstractFilesystemStore.exists` and let author of `AbstractFilesystemStore` to update it.

WeichenXu123:
@EnricoMi PR updated and addressed your comments. Could you take another look ? Thanks!

horovod/spark/common/store.py
EnricoMi: You are right, I did not see that this existed before. It has been introduced here: https://github.com/horovod/horovod/commit/7a697111eef7d88899551c176e31cde5ab61545c#diff-68621af6dea1527be13ddb5823a2d54a48a3f7b8fa98dd17cfa6ec1e3d677fccR180

However, you are changing the argument of calling into `self.isdir` from `path` to `localized_path`.

@tgaddair do you remember why you introduced the seemingly redundand `isdir(path)` here?

@WeichenXu123 why do you think `AbstractFilesystemStore` is the right place for `exists` to use the localized path rather than the given path? This affects **all** FilesystemStore implementations

horovod/spark/common/store.py
```diff
@@ -557,9 +569,25 @@ def normalize_path(path):
         """
         if path.startswith("dbfs:/"):
```
EnricoMi, line 570:
```suggestion
        if path.startswith("dbfs:/") and not path.startswith("dbfs://"):
```

horovod/spark/common/store.py
```diff
@@ -557,9 +569,25 @@ def normalize_path(path):
...
             return "/dbfs" + path[5:]
-        if path.startswith("file:///dbfs/"):
+        elif path.startswith("/dbfs/"):
+            return path
+        elif path.startswith("file:///dbfs/"):
             return path[7:]
-        return path
+        elif path.startswith("file:/dbfs/") and not path.startswith("file://dbfs/"):
```
EnricoMi, line 576: `"file://dbfs/"` does not start with `"file:/dbfs/"` already

EnricoMi (changes requested):

horovod/spark/common/store.py
WeichenXu123: OK I can make `DBFSLocalStore` override "exists" method.

EnricoMi (approved):
LGTM

WeichenXu123:
@EnricoMi Thank you very much ! When will the 0.24.3 release be cut ? :)

PR #3514, opened 2022-04-20
Add framework RC versions to test space

EnricoMi:
Besides head versions (nightlies) of tested frameworks, this also tests latest release candidates of those frameworks.


EnricoMi:
Tensorflow `2.9.0rc1` can only be tested on Linux with Python 3.7: https://github.com/tensorflow/tensorflow/issues/55725

PR #3520, opened 2022-04-21
Move pytorch_lightning requirement to pytorch-spark extras

tgaddair:
The PyTorch Lightning dependency is only needed if using the PyTorchEstimator, which requires the `pytorch-spark` extras. For users only using the `horovod.torch` submodule, this imposes an extra dependency they do not necessarily need.


chongxiaoc (approved):

EnricoMi (approved):
LGTM!

EnricoMi:
Failed CI jobs unrelated and no blockers.

Issue #3524, opened 2022-04-23
Expose an option to choose whether to use GPU or not for spark estimators


Tixxx:
Sometimes when the estimators are created on GPU hosts, the model params are already initialized on GPU, so pinning to GPU again will yield strange behaviors for keras with errors complaining "variable resource not found".
Expose an option in the interface to disable pinning to GPU if requested.

PR #3526, opened 2022-04-25
Add an option to pin to gpu for all estimators

Tixxx:
Signed-off-by: TJ <tix@uber.com>

## Description
Sometimes the estimator can be created on a GPU host which initializes model weights on one device. Pinning to GPU again in remote trainer can have strange behaviors such as "Variable resource not found" error for Tensorflow since tf thinks the model weights were initialized on a different device.

Fixes # (issue).
https://github.com/horovod/horovod/issues/3524
## Review process to land 

### DIFF
--- a/horovod/spark/common/params.py
+++ b/horovod/spark/common/params.py
@@ -96,6 +96,14 @@ class EstimatorParams(Params):
 
     label_shapes = Param(Params._dummy(), 'label_shapes', 'specifies the shape (or shapes) of the label column (or columns)')
 
+    inmemory_cache_all = Param(Params._dummy(), 'inmemory_cache_all',
+                               'Cache the data in memory for training and validation.',
+                               typeConverter=TypeConverters.toBoolean)
+
+    pin_gpu = Param(Params._dummy(), 'pin_gpu',
+                               'Whether to pin the traininig process to the GPU. Defaults to True.',
+                               typeConverter=TypeConverters.toBoolean)
+
     def __init__(self):
         super(EstimatorParams, self).__init__()
 
@@ -129,7 +137,9 @@ class EstimatorParams(Params):
             train_reader_num_workers=2,
             val_reader_num_workers=2,
             reader_pool_type='process',
-            label_shapes=None)
+            label_shapes=None,
+            inmemory_cache_all=False,
+            pin_gpu=True)
 
     def _check_params(self, metadata):
         model = self.getModel()
@@ -334,6 +344,17 @@ class EstimatorParams(Params):
     def getLabelShapes(self):
         return self.getOrDefault(self.label_shapes)
 
+    def setInMemoryCacheAll(self, value):
+        return self._set(inmemory_cache_all=value)
+
+    def getInMemoryCacheAll(self):
+        return self.getOrDefault(self.inmemory_cache_all)
+
+    def setPinGpu(self, value):
+        self._set(pin_gpu=value)
+
+    def getPinGpu(self):
+        return self.getOrDefault(self.pin_gpu)
 
 class ModelParams(HasOutputCols):
     history = Param(Params._dummy(), 'history', 'history')
--- a/horovod/spark/common/params.py
+++ b/horovod/spark/common/params.py
@@ -100,9 +100,9 @@ class EstimatorParams(Params):
                                'Cache the data in memory for training and validation.',
                                typeConverter=TypeConverters.toBoolean)
 
-    pin_gpu = Param(Params._dummy(), 'pin_gpu',
-                    'Whether to pin the traininig process to the GPU. '
-                    'Setting this to False will skipping binding to GPU even when GPU is available.'
+    use_gpu = Param(Params._dummy(), 'use_gpu',
+                    'Whether to use the GPU for training. '
+                    'Setting this to False will skipping binding to GPU even when GPU is available. '
                     'Defaults to True.',
                     typeConverter=TypeConverters.toBoolean)
 
@@ -141,7 +141,7 @@ class EstimatorParams(Params):
             reader_pool_type='process',
             label_shapes=None,
             inmemory_cache_all=False,
-            pin_gpu=True)
+            use_gpu=True)
 
     def _check_params(self, metadata):
         model = self.getModel()
@@ -352,11 +352,11 @@ class EstimatorParams(Params):
     def getInMemoryCacheAll(self):
         return self.getOrDefault(self.inmemory_cache_all)
 
-    def setPinGpu(self, value):
-        self._set(pin_gpu=value)
+    def setUseGpu(self, value):
+        self._set(use_gpu=value)
 
-    def getPinGpu(self):
-        return self.getOrDefault(self.pin_gpu)
+    def getUseGpu(self):
+        return self.getOrDefault(self.use_gpu)
 
 class ModelParams(HasOutputCols):
     history = Param(Params._dummy(), 'history', 'history')
--- a/horovod/spark/common/params.py
+++ b/horovod/spark/common/params.py
@@ -101,8 +101,10 @@ class EstimatorParams(Params):
                                typeConverter=TypeConverters.toBoolean)
 
     pin_gpu = Param(Params._dummy(), 'pin_gpu',
-                               'Whether to pin the traininig process to the GPU. Defaults to True.',
-                               typeConverter=TypeConverters.toBoolean)
+                    'Whether to pin the traininig process to the GPU. '
+                    'Setting this to False will skipping binding to GPU even when GPU is available.'
+                    'Defaults to True.',
+                    typeConverter=TypeConverters.toBoolean)
 
     def __init__(self):
         super(EstimatorParams, self).__init__()
--- a/horovod/spark/keras/remote.py
+++ b/horovod/spark/keras/remote.py
@@ -52,6 +52,7 @@ def RemoteTrainer(estimator, metadata, keras_utils, run_id, dataset_idx):
     user_verbose = estimator.getVerbose()
     checkpoint_callback = estimator.getCheckpointCallback()
     inmemory_cache_all = estimator.getInMemoryCacheAll()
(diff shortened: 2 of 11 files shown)


chongxiaoc (changes requested):
Generic param of Spark Estimator can be put in `EstimatorParams `, and also add Getter and Setter there.
https://github.com/horovod/horovod/blob/master/horovod/spark/common/params.py#L56

chongxiaoc (approved):
LGTM

chongxiaoc:
@EnricoMi Can you help do a sanity check as well?

chongxiaoc:
@EnricoMi Hold this PR unmerge for now.
I just synced with @Tixxx and we need to wait for Ray Tune to test this PR for integration test.

EnricoMi:
> Hold this PR unmerge for now.

I have set this to draft, press "Ready for review" when you are happy to see this merged.

horovod/spark/common/params.py
```diff
@@ -96,6 +96,14 @@ class EstimatorParams(Params):
 
     label_shapes = Param(Params._dummy(), 'label_shapes', 'specifies the shape (or shapes) of the label column (or columns)')
 
+    inmemory_cache_all = Param(Params._dummy(), 'inmemory_cache_all',
+                               'Cache the data in memory for training and validation.',
+                               typeConverter=TypeConverters.toBoolean)
+
+    pin_gpu = Param(Params._dummy(), 'pin_gpu',
+                               'Whether to pin the traininig process to the GPU. Defaults to True.',
```
EnricoMi, line 104: indentation is a bit off here

Tixxx:
The [fix](https://github.com/ray-project/ray/pull/22708) for the horovod incompatibility on ray tune's side was checked in just now. I'm running the integration test now and will merge it once that's done.

horovod/spark/keras/remote.py
```diff
@@ -111,7 +112,16 @@ def train(serialized_model, train_rows, val_rows, avg_row_size):
         hvd = get_horovod()
         hvd.init()
 
-        pin_gpu(hvd, tf, k)
+        # Verbose mode 1 will print a progress bar
+        verbose = user_verbose if hvd.rank() == 0 else 0
+
+        if should_pin_gpu:
+            if verbose:
+                print(f"Pinning current process to the GPU.")
```
EnricoMi, line 120: no `f` needed

horovod/spark/keras/remote.py
```diff
@@ -111,7 +112,16 @@ def train(serialized_model, train_rows, val_rows, avg_row_size):
...
+            pin_gpu(hvd, tf, k)
+        else:
+            if verbose:
+                print(f"Skip pinning current process to the GPU.")
```
EnricoMi, line 124: same here

horovod/spark/lightning/remote.py
```diff
@@ -194,7 +195,10 @@ def on_epoch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -
                       f"Val rows: {val_rows}, Val batch size: {val_batch_size}, Val_steps_per_epoch: {_val_steps_per_epoch}\n"
                       f"Checkpoint file: {remote_store.checkpoint_path}, Logs dir: {remote_store.logs_path}\n")
 
-            cuda_available = torch.cuda.is_available()
+            if not should_pin_gpu and verbose:
+                print("Skip pinning current process to the GPU.")
+
+            cuda_available = torch.cuda.is_available() and should_pin_gpu
```
EnricoMi, line 201: It should be mentioned in the API that `pin_gpu==False` disables CUDA.

Here should be a **warning** that though CUDA is available is is not used as pin_gpu is false. Would otherwise be hard to find out why CUDA is not used.

horovod/spark/lightning/remote.py
```diff
@@ -194,7 +195,10 @@ def on_epoch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -
                       f"Val rows: {val_rows}, Val batch size: {val_batch_size}, Val_steps_per_epoch: {_val_steps_per_epoch}\n"
                       f"Checkpoint file: {remote_store.checkpoint_path}, Logs dir: {remote_store.logs_path}\n")
 
-            cuda_available = torch.cuda.is_available()
+            if not should_pin_gpu and verbose:
+                print("Skip pinning current process to the GPU.")
```
EnricoMi, line 199: Why not using the logger? Why is there `verbose` when there is a logger?

EnricoMi (commented):
Some minor improvements possible.

horovod/spark/lightning/remote.py
Tixxx: makes sense, added a check to emit message when cuda is available but pin_gpu=false

horovod/spark/keras/remote.py
Tixxx: good catch

horovod/spark/common/params.py
Tixxx: updated.

horovod/spark/lightning/remote.py
Tixxx: The logger doesn't write to stdout and stderr properly in this function since it's run in ray executor. The train_logger is for passing some specialized loggers(not writing to stdout and stderr directly) to pytorch lightning. I have tried using some generic logger modules here, they either failed to serialize or no output.

horovod/spark/torch/remote.py
```diff
@@ -139,6 +139,12 @@ def train(serialized_model, optimizer_cls, model_opt_state_serialized,
             print("Skip pinning current process to the GPU.")
 
         cuda_available = torch.cuda.is_available() and should_pin_gpu
```
EnricoMi, line 141: The last bit of the condition has to go:
```suggestion
        cuda_available = torch.cuda.is_available()
```

horovod/spark/torch/remote.py
```diff
@@ -139,6 +139,12 @@ def train(serialized_model, optimizer_cls, model_opt_state_serialized,
...
+
+        if cuda_available and not should_pin_gpu:
+            print("GPU is available but pin_gpu is set to False."
+                    "Training will proceed without GPU support.")
```
EnricoMi, line 145: identation

horovod/spark/common/params.py
```diff
@@ -101,8 +101,10 @@ class EstimatorParams(Params):
                                typeConverter=TypeConverters.toBoolean)
 
     pin_gpu = Param(Params._dummy(), 'pin_gpu',
-                               'Whether to pin the traininig process to the GPU. Defaults to True.',
-                               typeConverter=TypeConverters.toBoolean)
+                    'Whether to pin the traininig process to the GPU. '
+                    'Setting this to False will skipping binding to GPU even when GPU is available.'
+                    'Defaults to True.',
```
EnricoMi, line 106: If not pinning to GPU means not using a GPU at all, then this should be called `use_gpu`. I thought "pin" has a subtle different meaning than "use".

Then, `use_gpu=False` becomes clear what it means even if GPU is available.

I would still keep the warning prints below.

Tixxx: i see, then i think we can rename it to use_gpu

Tixxx: renamed

Tixxx:
some builds are failing with 
```
#6 4.879 W: GPG error: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64  InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A4B469963BF863CC
  #6 4.879 E: The repository 'https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64  InRelease' is not signed.
```

@EnricoMi  Do you know if this is because nv's key expired or we need to refresh on our side?

Tixxx:
Integration test passed on my side. Changing this back to ready state.

EnricoMi:
> @EnricoMi Do you know if this is because nv's key expired or we need to refresh on our side?

@Tixxx This is something NVidia has to fix on their side.

@romerojosh @nvcastet NVidia's CUDA docker image cannot `apt update` because one of NVidia's repos' key is invalid:

```
docker run --rm -it nvidia/cuda:10.0-devel-ubuntu18.04 apt-get update
...
Err:9 https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64  InRelease                        
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A4B469963BF863CC
```

Something you can escalate?

EnricoMi (approved):
LGTM! Please wait with merge for fixed CI.

Tixxx:
Looks like nvidia rolled out a[ rotating key mechanism](https://forums.developer.nvidia.com/t/notice-cuda-linux-repository-key-rotation/212771). I will try to fix it by downloading the corresponding keys.

EnricoMi:
@Tixxx this is how @tgaddair fixed it for Ludwig: https://github.com/ludwig-ai/ludwig/commit/1a4f67914bc3894c7ada67a5b985c9c1ca2a3cc0

EnricoMi:
@Tixxx looks like your fix works!

Tixxx:
> 

yea, I tried that first, but we need to support 18.04 too which doesn't have wget installed by default. It needs to call `apt-get update` which will fail with invalid key. My fix is the alternative approach from nvidia which hopefully works for both 18.04 and 20.04.

Tixxx:
> @Tixxx this is how @tgaddair fixed it for Ludwig: [ludwig-ai/ludwig@1a4f679](https://github.com/ludwig-ai/ludwig/commit/1a4f67914bc3894c7ada67a5b985c9c1ca2a3cc0)
yea, I tried that first, but we need to support 18.04 too which doesn't have wget installed by default. It needs to call apt-get update which will fail with invalid key. My fix is the alternative approach from nvidia which hopefully works for both 18.04 and 20.04.

tgaddair (approved):

PR #3555, opened 2022-05-30
Upgrade tf and mxnet, fix libcudnn and libnccl on ubuntu18.04

EnricoMi:
Upgrades to latest Tensorflow and mxnet versions.


maxhgerlach (approved):
LGTM! I think we can deal with the failures on TF head in upcoming PRs.

PR #3563, opened 2022-06-02
Split Buildkite job for better re-run

EnricoMi:
This splits the job that triggers Buildkite CI and downloads results into two jobs:

grafik

is split into trigger and download jobs:

grafik

In case of a download timeout or any other error while waiting for the results, rerunning the failed job does not rerun the CI but only downloads the results.

Example run (with reduced tests): https://github.com/horovod/horovod/actions/runs/2429186983/attempts/1
Re-run does not trigger a new CI: https://github.com/horovod/horovod/actions/runs/2429186983/attempts/2


maxhgerlach (approved):
Very nice!

PR #3558, opened 2022-05-31
Bump build partially to C++17 for TensorFlow 2.10+

othakkar:
TF nightly builds are currently failing due to requiring C++17 support. Hence, this PR updates it to C++17.

Fixes #3566.

PR #3513, opened 2022-04-20
Fix for TF 2.9 API change: tensorflow_accelerator_device_info

maxhgerlach:
## Checklist before submitting

- [x] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)?
- [ ] Did you update the docs?
- [ ] Did you write any tests to validate this change?  
- [ ] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users?

## Description

The function has been renamed for TF 2.9-rc0: https://github.com/tensorflow/tensorflow/commit/51950258ea4cd5a6e50b926db7efee30358c1ce8

## Review process to land 

1. All tests and other checks must succeed.
2. At least one member of the [technical steering committee](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md) must review and approve.
3. If any member of the technical steering committee requests changes, they must be addressed.

PR #3571, opened 2022-06-08
Update docs and changelog for PR #3558

maxhgerlach:

## Description

Follow up for #3558 with new compiler requirements; also updating the changelog for #3513.

> 
> > Do we need to amend the documentation (where it mentions CMake dependency) and CHANGELOG.md?
> > 
> > We can do that in a follow up PR to avoid full CI cycles for these changes.
> 
> Good point! And I agree that it would be wise to do this in a separate PR.
> 
> CMake supports C++17 from version 3.8, so hopefully we should be good there with 3.13+. https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html
> 
> `gcc` seems to support all features of C++17 starting with version 8. https://gcc.gnu.org/projects/cxx-status.html#cxx17
> 
> We could add a phrase then like: "To install Horovod with TensorFlow 2.10 or later you will need a compiler that supports C++17 like `g++8` or newer."
> 

## Review process to land 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Spark/Lightning: add missing tranform_spec for Petastorm datamodule. ([#3543](https://github.com/horovod/horovod/pull/3543))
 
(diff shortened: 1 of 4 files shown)


EnricoMi:
Please wait with merge until backlog in CI cleared.

nvcastet (approved):

PR #3572, opened 2022-06-08
Test heads and mins with MPI

EnricoMi:
Test head versions not only with Gloo but also with MPI.

Once #3557 is (fixed and) merged, this can be merged as well.


EnricoMi:
@maxhgerlach this adds MPI to head testing.

maxhgerlach (approved):
LGTM!

Issue #3570, opened 2022-06-08
CI for torchhead hangs after `TorchTests::test_horovod_join_allgather`


maxhgerlach:
**Environment:**
1. Framework: PyTorch
2. Framework version: Nightly
3. Horovod version: master

**Bug report:**
As observed in https://github.com/horovod/horovod/pull/3558#issuecomment-1148685657, having fixed the build for TF 2.10 nightly, tests for `torchhead` now seem to hang just after `TorchTests::test_horovod_join_allgather`.

maxhgerlach:
Testing locally (Horovod master built with torch nighty), it appears that `test_horovod_join_allreduce` fails for one worker, while the other worker hangs. That's the test right after `..._join_allgather`:
```
test/parallel/test_torch.py::TorchTests::test_horovod_join_allgather PASSED [ 73%]PASSED [ 73%]
test/parallel/test_torch.py::TorchTests::test_horovod_join_allreduce
test/parallel/test_torch.py::TorchTests::test_horovod_join_allreduce FAILED [ 74%]
test/parallel/test_torch.py::TorchTests::test_horovod_join_broadcast
# hang
```
Running `test_horovod_join_allreduce` individually, I get a clear failure message:
```
>               assert torch.allclose(averaged_b, div(tensor_b * (size - 1), size), threshold), \
                    'hvd.join with hvd.allreduce produces incorrect results'
E               AssertionError: hvd.join with hvd.allreduce produces incorrect results
E               assert False
E                +  where False = <built-in method allclose of type object at 0x7fb4076c8840>(tensor([ 32, -34, -37,  37,  14,  21, -29,  27,  41,  31,  21,  48,  24, -29,\n         41,  -1, -34], dtype=torch.int32), tensor([ 32, -34, -38,  37,  14,  21, -30,  27,  41,  31,  21,  48,  24, -29,\n         41,  -1, -34], dtype=torch.int32), 0)
E                +    where <built-in method allclose of type object at 0x7fb4076c8840> = torch.allclose
[...]

PR #3573, opened 2022-06-09
Build Horovod in debug mode in test images

### LINKED ISSUE #3307: CI: Build Horovod in test images with `HOROVOD_DEBUG=1`
**Is your feature request related to a problem? Please describe.**
Some times unit tests may trigger bugs in Horovod's C++ backend. These bugs may cause segmentation faults at times or, if we are unlucky, may go unnoticed although some internal state is corrupted. We have some `assert()` macros all over the code base, however these assertions are not checked in release mode. They would be useful though to identify bugs before they trigger segmentation faults. Assertion failure messages are also more specific and easier to understand than segfaults.

One example would be part 2 (related to `hvd.barrier`) of PR https://github.com/horovod/horovod/pull/3300. In local debug builds an assertion failure would be raised before the segmentation fault was triggered.

**Describe the solution you'd like**
I propose to set the environment variable `HOROVOD_DEBUG=1` when building Horovod in CI test containers. Then debug symbols will be included and assertions will be checked at runtime.

**Describe alternatives you've considered**
Counter arguments I can think of:

- Tests might take a bit longer to run with debug code: I don't believe that there would be significant slowdowns, but of course I could be proven wrong.
- Certain bugs might only be observable in release builds, not in debug builds: This would be bad, but personally I would expect that we miss more problems because assertions are not checked. Some test cases could still be built in release mode to partially cover this situation.

maxhgerlach:

## Description

This enables assertion checking in the C++ code base, which should make it much easier to spot certain bugs early.

Fixes #3307. See that issue for some extended motivation.

For now I have just hard coded `HOROVOD_DEBUG=1` in `Dockerfile.test.{cpu,gpu}`. Alternatively we could configure this via `docker-compose.test.yml` and `HOROVOD_BUILD_FLAGS` only for a subset of test configuartions.

### DIFF
--- a/Dockerfile.test.cpu
+++ b/Dockerfile.test.cpu
@@ -231,7 +231,7 @@ RUN if [[ ${MPI_KIND} == "ONECCL" ]]; then \
     fi; \
     cd /horovod && \
     python setup.py sdist && \
-    bash -c "${HOROVOD_BUILD_FLAGS} HOROVOD_WITH_TENSORFLOW=1 HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_MXNET=1 pip install --no-cache-dir -v $(ls /horovod/dist/horovod-*.tar.gz)[spark,ray]"
+    bash -c "${HOROVOD_BUILD_FLAGS} HOROVOD_DEBUG=1 HOROVOD_WITH_TENSORFLOW=1 HOROVOD_WITH_PYTORCH=1 HOROVOD_WITH_MXNET=1 pip install --no-cache-dir -v $(ls /horovod/dist/horovod-*.tar.gz)[spark,ray]"
 
 # Show the effective python package version to easily spot version differences
 RUN pip freeze | sort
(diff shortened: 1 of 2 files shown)


EnricoMi:
Will this slow down tests?

maxhgerlach:
> Will this slow down tests?
Not sure, let's compare run times with master when the CI is done.

I mainly thought of this idea because there's a chance that we could see an error message in torchhead tests (which hang at the moment, #3570).

maxhgerlach:
> > Will this slow down tests?
> 
> Not sure, let's compare run times with master when the CI is done.

The variation in run times between different instances of the same job on GitHub actions is quite large. You can't see the impact of setting `HOROVOD_DEBUG=1` or not that easily unfortunately. While the data is inconclusive, it shows that there is no reason to fear massive slowdowns from turning on debug mode.

In the table I compared some data points for this PR (https://github.com/horovod/horovod/runs/6812448999?check_suite_focus=true) with master (https://github.com/horovod/horovod/runs/6796474820?check_suite_focus=true). I took the aggregate time for the entire "Build and Test" pipelines here, but zooming in on individual tests you see similar variations.

Screenshot 2022-06-14 at 17 15 38
[debug-build_vs_master.csv](https://github.com/horovod/horovod/files/8901108/debug-build_vs_master.csv)

PR #3541, opened 2022-05-09
clang compile failed

zhouronghua:
## Checklist before submitting

- [x] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)?
- [ ] Did you update the docs?
- [ ] Did you write any tests to validate this change?  
- [ ] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users?

## Description

Fixes # (issue).

## Review process to land 

1. All tests and other checks must succeed.
2. At least one member of the [technical steering committee](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md) must review and approve.
3. If any member of the technical steering committee requests changes, they must be addressed.

PR #3574, opened 2022-06-09
Add NVTX op tracing for Reducescatter, make base class destructors virtual

maxhgerlach:

## Description

In #3299 I forgot to register the `"HorovodReducescatter"` string with NVTX to use it in named op ranges. This means that `Reducescatter` ops would appear as "[Unknown]" in Nsight Systems profiles.

Besides that change I've defined virtual destructors for two abstract polymorphic base classes. This should fix a warning that has been mentioned in https://github.com/horovod/horovod/pull/3541. Generally I think having them virtual is a safer practice, but currently there probably is no actual problem because smart pointers (`std::shared_ptr` or `std::unique_ptr`) are used rather than raw base class pointers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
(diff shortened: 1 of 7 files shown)


EnricoMi:
LGTM! @tgaddair @romerojosh what do you think?

romerojosh (approved):
LGTM!

PR #3580, opened 2022-06-15
Expose an parameter through all spark estimators to set start method for multiprocessing

Tixxx:
Expose an parameter through all spark estimators to set start method for multiprocessing.

## Description
This is to avoid resource leaks in situations where CUDA calls are made in multiprocessing environment. According to CUDA [guidelines](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#um-fork-managed-memory), calling CUDA kernels on pinned memory region in forked processes will lead to undefined behavior such as handles not being cleared. Some leaks persist even when the process terminates.
Fixes # (issue).

## Review process to land 


chongxiaoc (approved):
LGTM

EnricoMi (approved):

PR #3584, opened 2022-06-22
Remove std::regex use for better cxx11 abi compatibility

trevor-m:

## Description

Since TensorFlow has moved to `_GLIBCXX_USE_CXX11_ABI=1`, we have ran into some issues with Horovod related to the use of `std::regex` when we build TF/Horovod on manylinux2014. `HorovodBroadcastInplaceOp` was crashing when `std::regex` was used.

To fix the problem, I replaced the regex with something simpler which is probably also faster and more lightweight.

## Review process to land 

### DIFF
--- a/horovod/tensorflow/mpi_ops.cc
+++ b/horovod/tensorflow/mpi_ops.cc
@@ -16,9 +16,9 @@
 // limitations under the License.
 // =============================================================================
 
+#include <algorithm>
 #include <memory>
 #include <queue>
-#include <regex>
 #include <thread>
 #include <unordered_map>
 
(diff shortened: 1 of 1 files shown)


maxhgerlach:
> I definitely understand and agree with your concern about more hidden corruption. Perhaps we can look into it a little more and do a follow-up PR?

I'm definitely in favor of merging this current PR (once the CI passes). But it would be great to better understand the root cause.

>  I can check if changing the horovod _GLIBCXX_USE_CXX11_ABI makes a difference with the error I was seeing.

If you find the time, that would be fabulous!

maxhgerlach (approved):

maxhgerlach:
> Would it help if the rest of Horovod (sources in horovod/common) was also compiled with _GLIBCXX_USE_CXX11_ABI=1 then?

Actually, I was mistaken there: It is already the case that all Horovod sources are compiled with `-D_GLIBCXX_USE_CXX11_ABI` set to match the TensorFlow build when the object files are linked into the `horovod/tensorflow/mpi_lib.cpython-*.so`. So I don't think your crash is related to how that is set when compiling Horovod.

Alternatively, could the crash be related to some other library that your are linking to / importing in Python, where `D_GLIBCXX_USE_CXX11_ABI` is set differently?

In any case, making this conflict less likely to turn up by dropping `std::regex` makes sense to me.

trevor-m:
Thanks @maxhgerlach!

PR #3585, opened 2022-06-22
fix local_rank typo in TF 2 example

NanoNabla:

## Description

In `tensorflow_mnist.py` loss value was printed on every node (`local_rank == 0`) not only on global master process. I do not see any benefit from doing that. Moreover, it may confuse people why the printed loss values differ across nodes (I know loss values are not synchronized). 

I think it was simply a typo to use `local_rank == 0` instead of `rank() == 0`. If not, feel free to reject and close this PR.

### DIFF
--- a/examples/elastic/tensorflow2/tensorflow2_mnist_elastic.py
+++ b/examples/elastic/tensorflow2/tensorflow2_mnist_elastic.py
@@ -84,7 +84,7 @@ def main():
             state.batch = start_batch + batch_idx
             loss_value = training_step(images, labels)
 
-            if state.batch % 10 == 0 and hvd.local_rank() == 0:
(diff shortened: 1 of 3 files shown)


EnricoMi:
Nice catch. There are a few other examples that use the same approach:
- examples/elastic/tensorflow2/tensorflow2_mnist_elastic.py
- examples/tensorflow/tensorflow_mnist_eager.py
- examples/tensorflow2/tensorflow2_mnist.py

Would you mind fixing all occurrences of this issue?

NanoNabla:
I only took a look into `examples/tensorflow2` but of course I'll take a look into the other files as well and update this PR

EnricoMi (approved):
LGTM!

PR #3593, opened 2022-07-06
Fix Typo in FindPytorch.cmake

Magnus1cheng:
## Description

Fixes typo in FindPytorch.cmake. REGEX could not match for ROCm error message.

### DIFF
--- a/cmake/Modules/FindPytorch.cmake
+++ b/cmake/Modules/FindPytorch.cmake
@@ -24,18 +24,18 @@ endif()
 
 execute_process(COMMAND ${PY_EXE} -c "import torch; from torch.utils.cpp_extension import CUDA_HOME; print(True if ((torch.version.cuda is not None) and (CUDA_HOME is not None)) else False)"
                 OUTPUT_VARIABLE Pytorch_CUDA OUTPUT_STRIP_TRAILING_WHITESPACE)
-string(REGEX REPLACE "No CUDA runtime[^\n]*\n?" "" Pytorch_CUDA "${Pytorch_CUDA}")
+string(REGEX REPLACE "No (CUDA|ROCm) runtime[^\n]*\n?" "" Pytorch_CUDA "${Pytorch_CUDA}")
 string(TOUPPER "${Pytorch_CUDA}" Pytorch_CUDA)
 
 execute_process(COMMAND ${PY_EXE} -c "import torch; from torch.utils.cpp_extension import ROCM_HOME; print(True if ((torch.version.hip is not None) and (ROCM_HOME is not None)) else False)"
                 OUTPUT_VARIABLE Pytorch_ROCM OUTPUT_STRIP_TRAILING_WHITESPACE)
-string(REGEX REPLACE "No CUDA runtime[^\n]*\n?" "" Pytorch_ROCM "${Pytorch_ROCM}")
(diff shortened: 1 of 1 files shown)


Magnus1cheng:
or we can use this way to match both strings.
`string(REGEX REPLACE "(No CUDA runtime[^\n]*\n)|(No ROCm runtime[^\n]*\n)" "" Pytorch_CUDA "${Pytorch_CUDA}")`
`string(REGEX REPLACE "(No CUDA runtime[^\n]*\n)|(No ROCm runtime[^\n]*\n)" "" Pytorch_ROCM "${Pytorch_ROCM}")`

maxhgerlach:
Thank you for the contribution, @Magnus1cheng! Please note that you would need to sign off your commits (DCO), see https://github.com/horovod/horovod/pull/3593/checks?check_run_id=7214166372 for some pointers.

> or we can use this way to match both strings.
> ```
> string(REGEX REPLACE "(No CUDA runtime[^\n]*\n)|(No ROCm runtime[^\n]*\n)" "" Pytorch_CUDA "${Pytorch_CUDA}")a
> string(REGEX REPLACE "(No CUDA runtime[^\n]*\n)|(No ROCm runtime[^\n]*\n)" "" Pytorch_ROCM "${Pytorch_ROCM}")
> ```

I'm not sure under which conditions these error messages appear. If either could show up here, it might make sense to filter both in any case.

There's also this second spot: https://github.com/horovod/horovod/blob/9d56c5a6d55e9619e100d19480a11952d116f4b3/cmake/Modules/FindPytorch.cmake#L38

Should that filter also be extended?

Magnus1cheng:
Hi, @maxhgerlach, I updated my commit. This cause error on my environment before so I match both Regex by using my commit, and the error gone. Somehow I cannot reproduce it now.
And for line 38, yes, we may need to change it as well, but I didn't encounter error in line 38, but I guess that changing would not harm anything because it includes the original regex match logic, but extend it. So you decide if we need that line changing.

EnricoMi:
Instead of `"(No CUDA runtime[^\n]*\n)|(No ROCm runtime[^\n]*\n)"`,
why not `"No (CUDA|ROCm) runtime[^\n]*\n"`?

Magnus1cheng:
Yes, that's better. Modified.

cmake/Modules/FindPytorch.cmake
```diff
@@ -24,18 +24,18 @@ endif()
 
 execute_process(COMMAND ${PY_EXE} -c "import torch; from torch.utils.cpp_extension import CUDA_HOME; print(True if ((torch.version.cuda is not None) and (CUDA_HOME is not None)) else False)"
                 OUTPUT_VARIABLE Pytorch_CUDA OUTPUT_STRIP_TRAILING_WHITESPACE)
-string(REGEX REPLACE "No CUDA runtime[^\n]*\n?" "" Pytorch_CUDA "${Pytorch_CUDA}")
+string(REGEX REPLACE "(No (CUDA|ROCm) runtime[^\n]*\n)" "" Pytorch_CUDA "${Pytorch_CUDA}")
```
EnricoMi, line 27:
```suggestion
string(REGEX REPLACE "No (CUDA|ROCm) runtime[^\n]*\n" "" Pytorch_CUDA "${Pytorch_CUDA}")
```
These brackets are not needed, I think.

Magnus1cheng: Updated

maxhgerlach (approved):
Thanks for the updates! The changes look good to me.

PR #3594, opened 2022-07-06
Add hvd.grouped_allgather and hvd.grouped_reducescatter

maxhgerlach:

## Description

This adds two new multi-tensor functions to Horovod: `hvd.grouped_allgather()` and `hvd.grouped_reducescatter()`. They are very similar to the existing `hvd.grouped_allreduce()`, introduced with PR #2453, giving users the equivalent extra control over tensor fusion for their `Allgather` and `Reducescatter` ops.

To implement this functionality I've added a new attribute `output_index` to `TensorTableEntry`. This allows an `AllgatherOp` or `ReducescatterOp` that has been enqueued as part of a group to allocate its output tensor properly (in contrast to `Allreduce` this cannot be done beforehand because the resulting output size is only known after cross-process coordination). I've also added warning log messages in case `AllocateOutput` ever fails because TensorFlow would crash with just a default "Unknown error." exception message in such a case.

Supported frameworks:
- [x] TensorFlow
- [x] PyTorch
- [x] MXNet

## Review process to land 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Added support for batched memory copies in GPUAllgather. ([#3590](https://github.com/horovod/horovod/pull/3590))
-- Added support for batched memory copies in GPUReducescatter.
+- Added support for batched memory copies in GPUReducescatter. ([#3621](https://github.com/horovod/horovod/pull/3621))
+- Added `hvd.grouped_allgather()` and `hvd.grouped_reducescatter()` operations. ([#3594](https://github.com/horovod/horovod/pull/3594))
+- TensorFlow: Added doc string for `hvd.grouped_allreduce()`. ([#3594](https://github.com/horovod/horovod/pull/3594))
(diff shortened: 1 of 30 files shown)


maxhgerlach:
Rebased to master to pick up test fixes and #3590.

maxhgerlach:
I believe this is ready to be reviewed now. For once all the test suites have passed.

For PyTorch and MXNet I have extended the respective `OpContext` classes so they can refer to all output tensors of a group. But ultimately they only need to allocate memory for one specific output index (to do an Allgather or a Reducescatter). In contrast, the `TFOpContext` already knew about all outputs of the group op. Maybe the design could be cleaned up a bit here to reduce confusion with how the two outputs of Alltoall are handled (tensor and splits), but that potential issue didn't seem pressing to me so far.

romerojosh (approved):
Excellent work @maxhgerlach! Very well implemented and I have no comments. Also, thanks for catching adding and cleaning up the documentation.

maxhgerlach:
Thanks for taking the time to review, @romerojosh! Fortunately, your careful earlier work was easy to transfer, so this was quite straightforward.

PR #3603, opened 2022-07-13
Add custom data loading (e.g. NVTabular) in KerasEstimator

### LINKED ISSUE #3602: Support custom data loaders in KerasEstimator
Currently, the Spark integration for PyTorchLightning [TorchEstimator supports customizable data_modules](https://github.com/horovod/horovod/blob/9d56c5a6d55e9619e100d19480a11952d116f4b3/horovod/spark/lightning/estimator.py#L109-L110), but the [RemoteTrainer for the KerasEstimator is hard-coded to use petastorm](https://github.com/horovod/horovod/blob/9d56c5a6d55e9619e100d19480a11952d116f4b3/horovod/spark/keras/remote.py#L112).

Ideally, the KerasEstimator should support customizable data modules/loaders as well.

leewyang:

## Description

Fixes #3602.

This introduces a `DataModule` base class for Spark integrations like `KerasEstimator`, along with a default `PetastormDataModule` and a new `NVTabularDataModule` that leverages [NVTabular](https://github.com/NVIDIA-Merlin/NVTabular) for GPU-accelerated data loading of tabular datasets.  It is loosely based on the existing `DataModule` for PyTorch Lightning's `TorchEstimator`, which is unfortunately tied directly to the `pl.LightningDataModule`.  Ideally, this base class could also be used for the plain PyTorch `TorchEstimator` as well.

## Review process to land 


leewyang:
@tgaddair Here is the NVTabular PR with a higher-level API to abstract away the Readers and a reduced image size (18.1GB vs. 37GB) for `Dockerfile.test.gpu`.

Dockerfile.test.gpu
```diff
@@ -159,16 +167,21 @@ RUN if [[ ${MXNET_PACKAGE} != "mxnet-nightly-cu"* ]]; then \
     fi
 
 # Prefetch Spark MNIST dataset.
-RUN mkdir -p /work
-RUN mkdir -p /data && wget --progress=dot:mega https://horovod-datasets.s3.amazonaws.com/mnist.bz2 -O /data/mnist.bz2
+RUN mkdir -p /work && mkdir -p /data && wget --progress=dot:mega https://horovod-datasets.s3.amazonaws.com/mnist.bz2 -O /data/mnist.bz2
```
EnricoMi, line 170:
```suggestion
RUN mkdir -p /work /data && wget --progress=dot:mega https://horovod-datasets.s3.amazonaws.com/mnist.bz2 -O /data/mnist.bz2
```

Dockerfile.test.gpu
```diff
@@ -123,7 +131,7 @@ RUN if [[ ${MPI_KIND} != "None" ]]; then \
 RUN if [[ ${TENSORFLOW_PACKAGE} != "tf-nightly-gpu" ]]; then \
         PROTOBUF_PACKAGE=""; \
         if [[ ${TENSORFLOW_PACKAGE} == tensorflow-gpu==1.15.* ]] || \
-           [[ ${TENSORFLOW_PACKAGE} == tensorflow-gpu==2.[012345].* ]]; then \
+           [[ ${TENSORFLOW_PACKAGE} == tensorflow-gpu==2.[0123456789].* ]]; then \
```
EnricoMi, line 134: Are you sure this is required for Tensorflow >= 2.6?

Dockerfile.test.gpu
```diff
@@ -159,16 +167,21 @@ RUN if [[ ${MXNET_PACKAGE} != "mxnet-nightly-cu"* ]]; then \
...
+RUN mkdir -p /work && mkdir -p /data && wget --progress=dot:mega https://horovod-datasets.s3.amazonaws.com/mnist.bz2 -O /data/mnist.bz2
 
 # Prefetch Spark Rossmann dataset.
-RUN mkdir -p /work
 RUN mkdir -p /data && wget --progress=dot:mega https://horovod-datasets.s3.amazonaws.com/rossmann.tgz -O - | tar -xzC /data
 
 # Prefetch PyTorch datasets.
 RUN wget --progress=dot:mega https://horovod-datasets.s3.amazonaws.com/pytorch_datasets.tgz -O - | tar -xzC /data
 
+# Patches (leveraging cached layers above)
```
EnricoMi, line 178: What do you mean with "leveraging cached layers"? In which sense is this comment specifically important here?

Dockerfile.test.gpu
```diff
@@ -52,28 +54,33 @@ RUN CUDNN_MAJOR=$(cut -d '.' -f 1 <<< "${CUDNN_VERSION}"); \
...
 # setup ssh service
 RUN ssh-keygen -f /root/.ssh/id_rsa -q -N ''
 RUN cp -v /root/.ssh/id_rsa.pub /root/.ssh/authorized_keys
 
-# Install Python.
-RUN apt-get update -qq && apt-get install -y python${PYTHON_VERSION} python${PYTHON_VERSION}-dev python${PYTHON_VERSION}-distutils
-RUN ln -s -f /usr/bin/python${PYTHON_VERSION} /usr/bin/python
-RUN ln -s -f /usr/bin/python${PYTHON_VERSION} /usr/bin/python${PYTHON_VERSION/%.*/}
-RUN wget --progress=dot:mega https://bootstrap.pypa.io/get-pip.py && python get-pip.py && rm get-pip.py
+RUN wget -q https://repo.continuum.io/miniconda/Miniconda3-py38_4.12.0-Linux-x86_64.sh -O miniconda.sh && \
```
EnricoMi, line 64: Do we really want to base our test images on miniconda? Why is this needed?

EnricoMi (commented):
What do you mean with `reduced image size (18.1GB vs. 37GB) for Dockerfile.test.gpu.`? Does adding NVtabular inflate the image to 37GB? It is currently 15.3GB: https://github.com/horovod/horovod/runs/7841398530?check_suite_focus=true#step:231:16

I'd prefer to split this into to PRs, one for reducing the image size and one for the custom data loading.

leewyang:
@EnricoMi Thanks for the review!

> What do you mean with `reduced image size (18.1GB vs. 37GB) for Dockerfile.test.gpu.`?

This was just a reference to an earlier version of the dockerfile, which was 37GB due to the use of a different base image.  By going back to the `nvidia/cuda` base image, this image was reduced down to 18.1GB.  So, this was not a separate action to try to reduce the existing image, but just to produce a reasonable-size image for this PR.

Dockerfile.test.gpu
leewyang: I think this may have been leftover from prior experiments.  Just tried another image without this, which seemed to be OK, so will revert this line.

Dockerfile.test.gpu
leewyang: Not particularly important, so will remove.

Dockerfile.test.gpu
leewyang: Unfortunately, this is required for [installing nvtabular](https://pypi.org/project/nvtabular/).  There is an option to install it w/ only CPU support (via pip), so maybe that would be a reasonable option for CI/CD purposes?  But for end-user runtimes, they'd probably want the full GPU support.

horovod/spark/keras/remote.py
```diff
@@ -242,55 +233,39 @@ def train(serialized_model, train_rows, val_rows, avg_row_size):
...
+                'inmemory_cache_all': inmemory_cache_all,
+                'cur_shard': hvd.rank(),
+                'shard_count': hvd.size(),
+                'schema_fields': schema_fields,
+                'storage_options': storage_options,
+                'steps_per_epoch_train': steps_per_epoch,
+                'steps_per_epoch_val': validation_steps,
+                'verbose': verbose,
+            }
+            print("data_module: {}".format(data_module))
```
EnricoMi, line 264: This print out should be done only `if verbose`, and please add a bit more context to the message like other `print`s in the file.

horovod/spark/common/datamodule.py
```diff
@@ -0,0 +1,60 @@
...
+    _data_modules[module_name] = data_module
+
+def datamodule_from_name(module_name):
+    """Returns a DataModule implementation associated with a string key."""
+    if module_name in _data_modules:
+        return _data_modules[module_name]
+    else:
+        # try dynamic import
+        module = importlib.import_module(module_name)
+        module.register()
```
EnricoMi, line 59: When `module` does not implement `register`, or `register` does not call into `register_datamodule`, this will produce errors with very little meaning.

horovod/spark/common/datamodule.py
```diff
@@ -0,0 +1,60 @@
+import importlib
+
+
+_data_modules = {}
+class DataModule:
+    """Context manager base class for data module/loader implementation.s"""
```
EnricoMi, line 6:
```suggestion
    """Context manager base class for data module/loader implementations."""
```

horovod/spark/common/datamodule.py
```diff
@@ -0,0 +1,60 @@
+import importlib
+
+
+_data_modules = {}
+class DataModule:
```
EnricoMi, line 5: I think we can use two empty lines before `class DataModule`.

horovod/spark/common/datamodule.py
EnricoMi, line 59: Looks like any `DataModule` is expected to implement `register`, maybe you should add abstract method `def register(...)` to `DataModule`. That is then a good place to document this contract.

horovod/spark/common/datamodule.py
```diff
@@ -0,0 +1,60 @@
...
+        self.train_async_data_loader_queue_size = train_async_data_loader_queue_size
+        self.val_async_data_loader_queue_size = val_async_data_loader_queue_size
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, type, value, traceback):
+        pass
+
+    def train_data(self, reader=None):
```
EnricoMi, line 41: Looks like `train_data` and `val_data` could be abstract methods.

Dockerfile.test.gpu
EnricoMi, line 64: The `NVTabularDataModule` does not contain any GPU-specific code, it simply calls into `nvtabular.loader.tensorflow.KerasSequenceLoader`, which I presume is tested with GPUs thoroughly. I do not see the benefits to test `NVTabularDataModule` with GPU support, especially given this has an impact on the entire Python environment.

Dockerfile.test.gpu
```diff
@@ -32,12 +34,12 @@ SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 # Extract ubuntu distribution version and download the corresponding key.
 # This is to fix CI failures caused by the new rotating key mechanism rolled out by Nvidia.
 # Refer to https://forums.developer.nvidia.com/t/notice-cuda-linux-repository-key-rotation/212771 for more details.
-#RUN DIST=$(echo ${CUDA_DOCKER_VERSION#*ubuntu} | sed 's/\.//'); \
-#    apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${DIST}/x86_64/3bf863cc.pub && \
-#    apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu${DIST}/x86_64/7fa2af80.pub
+RUN DIST=$(echo ${CUDA_DOCKER_VERSION#*ubuntu} | sed 's/\.//'); \
```
EnricoMi, line 37: This is not needed and can stay un-commented, I think.

Dockerfile.test.gpu
```diff
@@ -32,12 +34,12 @@ SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
...
+    apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${DIST}/x86_64/3bf863cc.pub && \
+    apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu${DIST}/x86_64/7fa2af80.pub
 
 # Prepare to install specific g++ versions
-RUN apt-get update -qq && apt-get install -y --no-install-recommends software-properties-common
+RUN apt-get update -qq && apt-get install -y --no-install-recommends software-properties-common && rm -rf /var/lib/apt/lists/*
```
EnricoMi, line 42: All these improvements on the Docker image size should also go into `Dockerfile.test.cpu`.

Dockerfile.test.gpu
```diff
@@ -52,28 +54,33 @@ RUN CUDNN_MAJOR=$(cut -d '.' -f 1 <<< "${CUDNN_VERSION}"); \
...
+RUN wget -q https://repo.continuum.io/miniconda/Miniconda3-py38_4.12.0-Linux-x86_64.sh -O miniconda.sh && \
+    bash miniconda.sh -b && \
+    rm miniconda.sh && \
+    # Source conda.sh for all login and interactive shells.
+    ln -s /root/miniconda3/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
+    echo ". /etc/profile.d/conda.sh" >> ~/.bashrc && \
+    # Set always_yes for non-interactive shells.
+    conda config --system --set always_yes True && \
+    # Install NVTabular
+    conda install -y -c nvidia -c rapidsai -c numba -c conda-forge nvtabular python=${PYTHON_VERSION} cudatoolkit=${CUDA_DOCKER_VERSION:0:4} && \
```
EnricoMi, line 73: nvtabular should also be installed in `Dockerfile.test.cpu` via pip.

horovod/spark/common/datamodule.py
leewyang: Good point, looking into this further.

leewyang: Now moved into `DataModule`.

docs/spark.rst
```diff
@@ -96,7 +96,27 @@ logging (for Tensorboard) using the Estimator ``Store`` abstraction.  Stores are
 artifacts including intermediate representations of the training data.  Horovod natively supports stores for HDFS
 and local filesystems.
 
-`Petastorm <https://github.com/uber/petastorm/blob/master/petastorm/pytorch.py#L258>` based data loader is used by default, but user can define a custom data loader by override the `BaseDataLoader` interface. A async data loader mixin can also added on top of the data loader.
+`Petastorm <https://github.com/uber/petastorm/blob/master/petastorm/pytorch.py#L258>` based data loader is used by default,
+but user can define a custom data loader by overriding the `BaseDataLoader` interface. A async data loader mixin can also
```
EnricoMi, line 100:
```suggestion
but user can define a custom data loader by overriding the `BaseDataLoader` interface. An async data loader mixin can also
```

horovod/spark/common/datamodule.py
```diff
@@ -0,0 +1,79 @@
...
+    Alternate implemntations of DataModule can be referenced by their fully-scoped name, e.g. `x.y.z.CustomDataModule`
+    """
+    if module_name in _data_modules:
+        # return data module class from registry
+        return _data_modules[module_name]
+    else:
+        # otherwise, try to dynamically import data module
+        try:
+            splits = module_name.split('.')
+            m, c = '.'.join(splits[:-1]), splits[-1]
```
EnricoMi, line 74: using descriptive variable names helps understand what kind of value they hold:
```suggestion
            module_name, class_name = '.'.join(splits[:-1]), splits[-1]
```
For this to work, method argument `module_name` should be renamed to something more precise like `datamodule_name`.

horovod/spark/common/datamodule.py
```diff
@@ -0,0 +1,79 @@
...
+        pass
+
+    @classmethod
+    def register(cls):
+        """Adds this DataModule implementation to a global registry keyed by short name."""
+        _data_modules[cls.short_name] = cls
+
+
+def datamodule_from_name(module_name):
+    """Returns a DataModule implementation associated with a string key.
```
EnricoMi, line 63:
```suggestion
    """Returns a DataModule implementation (the class) associated with a name.
```

horovod/spark/common/datamodule.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Alternate implemntations of DataModule can be referenced by their fully-scoped name, e.g. `x.y.z.CustomDataModule`
```
EnricoMi, line 65:
```suggestion
    Alternatively, implementations of DataModule can be referenced by their fully qualified class names, e.g. `horovod.spark.keras.datamodule.PetastormDataModule`
```

horovod/spark/common/datamodule.py
```diff
@@ -0,0 +1,79 @@
...
+
+    @abstractmethod
+    def val_data(self, reader=None):
+        """Returns the validation data in a form required by the target DL framework."""
+        pass
+
+    @classmethod
+    def register(cls):
+        """Adds this DataModule implementation to a global registry keyed by short name."""
+        _data_modules[cls.short_name] = cls
```
EnricoMi, line 59: What if that class data module name exists already in `_data_modules`? This then silently overwrites that class, which would be hard to discover.

This method should check if `cls.short_name` already exists in `_data_modules`, and log a warning if it does. Further, `cls` should not be registered in that case by its short name. It can still be accesses by the fully qualified class name (which should be logged in the warning as well).

Alternative is to also drop the existing class from `_data_modules`, so neither datamodule is accessible through the collided class name and either has to be accessed through the fully qualified datamodule class name.

horovod/spark/common/datamodule.py
```diff
@@ -0,0 +1,79 @@
...
+def datamodule_from_name(module_name):
+    """Returns a DataModule implementation associated with a string key.
+
+    Alternate implemntations of DataModule can be referenced by their fully-scoped name, e.g. `x.y.z.CustomDataModule`
+    """
+    if module_name in _data_modules:
+        # return data module class from registry
+        return _data_modules[module_name]
+    else:
+        # otherwise, try to dynamically import data module
```
EnricoMi, line 71: Here, `module_name` has two possible meanings:
1. a simple datamodule name (string used to register the class in `_data_modules`
2. a fully qualified class name

In the first case, a message should be logged that the datamodule name is not registered, plus a full list of registered names (to easily spot typos or registration issues).

Only in the second case (`module_name` contains `'.'`) the module should be tried to be loaded. That will avoid the following exception to be raised (which is pretty meaningless to the user) if `module_name` is not registered and does not contain a `.`:
```
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1011, in _gcd_import
  File "<frozen importlib._bootstrap>", line 950, in _sanity_check
ValueError: Empty module name
```

EnricoMi (changes requested):

horovod/spark/common/datamodule.py
leewyang: Ended up removing this registry, since it's likely more trouble than it's worth.  Now, user can just import the datamodule class itself and use the class directly.

horovod/spark/keras/datamodule.py
```diff
@@ -0,0 +1,168 @@
...
+        import tensorflow as tf
+        hvd.init()
+
+        min_int, max_int = tf.int32.limits
+        max_rand = max_int // hvd.size()
+        # Generate a seed fragment on each worker
+        seed_fragment = cupy.random.randint(0, max_rand).get()
+        # Aggregate seed fragments from all Horovod workers
+        seed_tensor = tf.constant(seed_fragment)
+        reduced_seed = hvd.allreduce(seed_tensor, name="shuffle_seed", op=hvd.Sum)
```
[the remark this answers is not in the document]
leewyang: Good point, updated the code to just broadcast the rank zero seed.

Dockerfile.test.gpu
leewyang: Tried to just `pip install nvtabular` into the original `Dockerfile.test.gpu`, but that fails to install other dependencies that require conda, e.g. `cudf`.  So, if I understand correctly, we can just leave the original `Dockerfile.test.gpu` (without `nvtabular`), which would skip the newly added `NVTabularDataModule` tests, but would still test the `PetastormDataModule` and related infrastructure.  For users interested in `NVTabular`, we can always park a separate dockerfile/image somewhere else outside of the horovod build pipeline (maybe something like `horovod/docker/horovod-nvtabular`?).

If you're OK with that, I can revert the edits to both dockerfiles (since they'd be largely unrelated to this PR then, and can always create a separate PR for the image slimming bits if desired).

EnricoMi: Firstly, given Horovod with NVTabular on GPU requires Conda, this should be noted in the docs, maybe in a new section between Spark and Controllers (at the level of Spark).

Secondly, I would prefer the `horovod/docker/horovod-nvtabular` image that simplifies building a GPU NVTabular environment for Horovod. Then, `nvtabular` and `miniconda` can be kept out of `Dockerfile.test.gpu`, but the size improvements (`rm -rf /var/lib/apt/lists/*`) should stay. We can add the image to the CI so we always know this image builds:
https://github.com/horovod/horovod/blob/32b5d3ecbb419bcebedf1a456f713ef0c05b578a/.github/gen-workflow-ci.py#L817

horovod/spark/keras/estimator.py
```diff
@@ -93,6 +95,7 @@ class KerasEstimator(HorovodEstimator, KerasEstimatorParamsReadable,
 
     Args:
         num_proc: Number of Horovod processes.  Defaults to `spark.default.parallelism`.
+        data_module: (Optional) DataModule used for training and validation, if not set, defaults to the PetastormDataModule.
```
EnricoMi, line 98:
```suggestion
        data_module: (Optional) DataModule class used for training and validation, if not set, defaults to the PetastormDataModule.
```

docs/spark.rst
```diff
@@ -96,7 +96,27 @@ logging (for Tensorboard) using the Estimator ``Store`` abstraction.  Stores are
...
-`Petastorm <https://github.com/uber/petastorm/blob/master/petastorm/pytorch.py#L258>` based data loader is used by default, but user can define a custom data loader by override the `BaseDataLoader` interface. A async data loader mixin can also added on top of the data loader.
+`Petastorm <https://github.com/uber/petastorm/blob/master/petastorm/pytorch.py#L258>` based data loader is used by default,
+but user can define a custom data loader by overriding the `BaseDataLoader` interface. An async data loader mixin can also
+be added on top of the data loader.  Additionally, the KerasEstimator now supports a DataModule argument that is similar
+to the Lightning DataModule which abstracts the data loading and allows for alternative implementations:
+
+.. code-block:: python
+
+    keras_estimator = hvd.KerasEstimator(
+        data_module='nvtabular',   # default: 'petastorm'
```
EnricoMi, line 107: this needs to be updated

horovod/spark/keras/datamodule.py
```diff
@@ -0,0 +1,169 @@
...
+
+    @staticmethod
+    def seed_fn():
+        """
+        Generate consistent dataloader shuffle seeds across workers
+        Reseeds each worker's dataloader each epoch to get fresh a shuffle
+        that's consistent across workers.
+        """
+        import horovod.tensorflow.keras as hvd
+        import cupy
```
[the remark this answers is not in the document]
leewyang: Probably can get rid of this particular dependency, since it's just generating a random number.  Will test replacing w/ alternatives.

docs/spark.rst
leewyang: good point, missed that.

chongxiaoc:
It looks this PR needs to be rebased after https://github.com/horovod/horovod/pull/3665 is merged.
`make_dataset_fn()` in keras is changed there.

horovod/spark/keras/datamodule.py
```diff
@@ -0,0 +1,168 @@
...
+
+        super().__init__(**kwargs)
+        self.reader_pool_type = reader_pool_type
+        self.train_reader_worker_count = train_reader_worker_count
+        self.val_reader_worker_count = val_reader_worker_count
+
+        self.make_dataset = make_dataset
+        self.random_seed = random_seed
+
+        hvd.init()
```
chongxiaoc, line 35: I don't think we need `hvd.init()` here.

horovod/spark/keras/datamodule.py
```diff
@@ -0,0 +1,168 @@
...
+            self.train_reader.stop()
+        super().__exit__(type, value, traceback)
+
+    @contextlib.contextmanager
+    def empty_batch_reader(self):
+        yield None
+
+    def train_data(self):
+        return self.make_dataset(self.train_reader, self.train_batch_size, self.shuffle_size,
+                                 self.is_batch_reader, shuffle=True if self.shuffle_size > 0 else False,
```
chongxiaoc, line 93: As mentioned outside, I would like to hold this PR till https://github.com/horovod/horovod/pull/3665 is merged due to Petastorm API changes.

chongxiaoc (changes requested):

Dockerfile.test.gpu
leewyang: OK, updated `docs/spark.rst` with some comments about conda/docker w.r.t. `NVTabularDataModule`, since currently the only mention and it's fairly specific to the `KerasEstimator` at the moment.  Also, moved the NVTabular dockerfile into `horovod/docker/horovod-nvtabular` and removed `nvtabular` and `conda` from the `Dockerfile.test.*` files.

horovod/spark/keras/datamodule.py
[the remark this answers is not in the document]
leewyang: Replaced `cupy` with `numpy` here, since this is just generating a random number.  As such, we probably don't want to add `nvtabular` as an `extras_require`, since it's unlikely to successfully install via `pip install`.  Instead, we can refer folks to the `docker/horovod-nvtabular` image.

horovod/spark/keras/datamodule.py
leewyang: Removed, thanks!

Dockerfile.test.gpu
leewyang: Now removed conda and nvtabular from the Dockerfile.test.* images in favor of the `docker/horovod-nvtabular/Dockerfile`.

horovod/spark/keras/datamodule.py
```diff
@@ -0,0 +1,166 @@
...
+        for col, tensor in X.items():
+            if isinstance(tensor, tuple):
+                retX[col] = tf.RaggedTensor.from_row_lengths(values=tensor[0][:,0], row_lengths=tensor[1][:,0]).to_tensor()
+            else:
+                retX[col] = tensor
+        return retX, labels
+
+    def train_data(self):
+        import horovod.tensorflow.keras as hvd
+        from nvtabular.loader.tensorflow import KerasSequenceLoader
```
chongxiaoc, line 139: Does this line require `nvtabular` to be a required package of horovod?
In that case, we should change `setup.py` part https://github.com/horovod/horovod/blob/master/setup.py#L169 ?

chongxiaoc: As long as `dm.train_data()` with NVTabular data module is not called, we are safe then?

EnricoMi: Yes, I would say using `NVTabularDataModule` requires `nvtabular` to be installed, but I would not make it a dependency. If you are not calling into this code path, `nvtabular` is not required.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased] - YYYY-MM-DD
 
 ### Added
+- Spark Estimator: Add support for custom data loaders in KerasEstimator.
```
EnricoMi, line 11: I would also add another line mentioning the main contribution here:
```suggestion
Spark Estimator: Add NVTabular data loader for KerasEstimator. ([#3603](https://github.com/horovod/horovod/pull/3603))
```
Please add that link to the PR also to above CHANGELOG entry.

docs/spark.rst
```diff
@@ -96,7 +96,32 @@ logging (for Tensorboard) using the Estimator ``Store`` abstraction.  Stores are
...
 and local filesystems.
 
-`Petastorm <https://github.com/uber/petastorm/blob/master/petastorm/pytorch.py#L258>` based data loader is used by default, but user can define a custom data loader by override the `BaseDataLoader` interface. A async data loader mixin can also added on top of the data loader.
+`Petastorm <https://github.com/uber/petastorm/blob/master/petastorm/pytorch.py#L259>`__ based data loader is used by default,
+but user can define a custom data loader by overriding the `BaseDataLoader` interface. An async data loader mixin can also
+be added on top of the data loader.  Additionally, the KerasEstimator supports a DataModule argument, similar
+to the Lightning DataModule, which abstracts the data loading and allows for alternative implementations.  For example,
+the NVTabularDataModule integrates the `KerasSequenceLoader <https://github.com/NVIDIA-Merlin/NVTabular/blob/main/nvtabular/loader/tensorflow.py>`__
+from NVTabular to enable GPU-accelerated data loading.  Note, however, due to the complexity installation, NVTabular
+recommends the use of a `conda` environment or a pre-built docker image.
```
EnricoMi, line 105: You can also mention `docker/horovod-nvtabular/Dockerfile` here as an example to install Horovod with NVTabular.

leewyang: Good point, done.

docs/spark.rst
```diff
@@ -96,7 +96,33 @@ logging (for Tensorboard) using the Estimator ``Store`` abstraction.  Stores are
...
+`Petastorm <https://github.com/uber/petastorm/blob/master/petastorm/pytorch.py#L259>`__ based data loader is used by default,
+but user can define a custom data loader by overriding the `BaseDataLoader` interface. An async data loader mixin can also
+be added on top of the data loader.  Additionally, the KerasEstimator supports a DataModule argument, similar
+to the Lightning DataModule, which abstracts the data loading and allows for alternative implementations.  For example,
+the NVTabularDataModule integrates the `KerasSequenceLoader <https://github.com/NVIDIA-Merlin/NVTabular/blob/main/nvtabular/loader/tensorflow.py>`__
+from NVTabular to enable GPU-accelerated data loading.  Note, however, due to the complexity of installation, NVTabular
+recommends the use of a `conda` environment or a pre-built docker image.  For users who want to build their own docker images,
+there is an `example Dockerfile <../docker/horovod-nvtabular/Dockerfile>`__ for building Horovod with NVTabular support.
```
EnricoMi, line 106: This documentation will be hosted separate from the sources, so this link will be broken:
https://horovod--3603.org.readthedocs.build/en/3603/spark_include.html#horovod-spark-estimators

Please link to https://github.com/horovod/horovod/blob/master/docker/horovod-nvtabular/Dockerfile. That link will become live once this is merged into master.

EnricoMi (approved):
Excellent work!

EnricoMi:
@chongxiaoc are you happy with this?

chongxiaoc:
@leewyang @EnricoMi I will take a look asap. Thanks.

horovod/spark/common/params.py
```diff
@@ -23,6 +24,7 @@
 
 class EstimatorParams(Params):
     num_proc = Param(Params._dummy(), 'num_proc', 'number of processes')
+    data_module = Param(Params._dummy(), 'data_module', 'data module class to use when reading data')
```
chongxiaoc, line 27: I suggest to move `data_module` filed under KerasEstimator since it is specific to KerasEstimator.

Refer the example of lightning estimator at https://github.com/horovod/horovod/blob/master/horovod/spark/lightning/estimator.py#L209

horovod/spark/common/params.py
```diff
@@ -134,6 +140,7 @@ def __init__(self):
 
         self._setDefault(
             num_proc=None,
+            data_module=None,
```
chongxiaoc, line 143: same here, move `data_module` to KerasEstimator.

horovod/spark/keras/datamodule.py
```diff
@@ -0,0 +1,173 @@
...
+import contextlib
+import horovod.tensorflow.keras as hvd
+
+from horovod.spark.common import constants
+from horovod.spark.common.datamodule import DataModule
+
+
+class PetastormDataModule(DataModule):
+    """Default Petastorm-based DataModule for KerasEstimator."""
+    def __init__(self, reader_pool_type: str="process",
```
chongxiaoc, line 24: reader_pool_type: str='thread'.

To be consistent with default value here: https://github.com/horovod/horovod/blob/master/horovod/spark/common/params.py#L167

horovod/spark/keras/datamodule.py
```diff
@@ -0,0 +1,173 @@
...
+            self.reader_factory_kwargs['pyarrow_serialize'] = True
+            self.is_batch_reader = False
+        else:
+            self.reader_factory = make_batch_reader
+            self.is_batch_reader = True
+
+    def __enter__(self):
+        super().__enter__()
+        self.train_reader = self.reader_factory(self.train_dir,
+                                num_epochs=self.num_train_epochs,
```
chongxiaoc, line 58: `num_epochs=1`
To be consistent with https://github.com/horovod/horovod/blob/master/horovod/spark/keras/remote.py#L253

Petastorm will repeat internally at https://github.com/horovod/horovod/blob/master/horovod/spark/keras/util.py#L82

horovod/spark/keras/util.py
```diff
@@ -35,15 +36,15 @@ class TFKerasUtil(object):
...
-                validation_data=val_data,
-                steps_per_epoch=steps_per_epoch,
-                validation_steps=validation_steps,
-                callbacks=callbacks,
-                verbose=verbose,
-                epochs=epochs)
+        def fn(model, data_module, steps_per_epoch, validation_steps, callbacks, verbose):
+            with data_module as dm:
+                return model.fit(dm.train_data(),
+                                 validation_data=dm.val_data() if dm.has_val else None,
```
chongxiaoc, line 42: nice!
```
dm.train_data(),
validation_data=dm.val_data()
```

chongxiaoc (changes requested):
just catch a few inconsistencies.

horovod/spark/common/datamodule.py
```diff
@@ -0,0 +1,64 @@
...
+# ==============================================================================
+from abc import ABC, abstractmethod
+
+
+class DataModule(ABC):
+    """Context manager base class for data module/loader implementations."""
+    short_name = None     # implementations should provide a short name for easy reference, e.g. 'petastorm', 'nvtabular', etc.
+
+    def __init__(self, train_dir: str, val_dir: str, num_train_epochs: int=1, has_val: bool=True,
+                 train_batch_size: int=32, val_batch_size: int=32, shuffle_size: int=1000,
```
chongxiaoc, line 23: `shuffle_size` is not used.
Change to `shuffle : bool=True`?

horovod/spark/common/datamodule.py
```diff
@@ -0,0 +1,64 @@
...
+                 debug_data_loader: bool=False, train_async_data_loader_queue_size: int=None,
+                 val_async_data_loader_queue_size: int=None, **kwargs):
+        super().__init__()
+        self.train_dir = train_dir
+        self.val_dir = val_dir
+        self.num_train_epochs = num_train_epochs
+        self.has_val = has_val
+        self.train_batch_size = train_batch_size
+        self.val_batch_size = val_batch_size
+        self.shuffle_size = shuffle_size
```
chongxiaoc, line 36: self.shuffle = shuffle

horovod/spark/common/params.py
```diff
@@ -199,6 +206,12 @@ def setNumProc(self, value):
     def getNumProc(self):
         return self.getOrDefault(self.num_proc)
 
+    def setDataModule(self, value):
+        return self._set(data_module=value)
+
+    def getDataModule(self):
+        return self.getOrDefault(self.data_module)
+
```
chongxiaoc, line 214: remove `setDataModule` and `getDataModule` here since `datamodule` is specific to KerasEstimator and LightningEstimator.

horovod/spark/keras/datamodule.py
```diff
@@ -0,0 +1,173 @@
...
+    Note: requires `label_cols`, `categorical_cols`, and `continuous_cols` to be explicitly provided."""
+    def __init__(self, label_cols=[], categorical_cols=[], continuous_cols=[], **kwargs):
+        super().__init__(**kwargs)
+        self.label_cols = label_cols
+        self.categorical_cols = categorical_cols
+        self.continuous_cols = continuous_cols
+
+    @staticmethod
+    def seed_fn():
+        """
```
chongxiaoc, line 118: Question to NVTabular specific usage: maybe we should use `random_seed` in constructor?
In that case we don't need `hvd.broadcast` here right?
User can specify `random_seed` for reproducible results.

horovod/spark/keras/datamodule.py
```diff
@@ -0,0 +1,173 @@
...
+    def train_data(self):
+        import horovod.tensorflow.keras as hvd
+        from nvtabular.loader.tensorflow import KerasSequenceLoader
+        return KerasSequenceLoader(self.train_dir,
+                                   batch_size=self.train_batch_size,
+                                   label_names=self.label_cols,
+                                   cat_names=self.categorical_cols,
+                                   cont_names=self.continuous_cols,
+                                   engine="parquet",
+                                   shuffle=True,
```
chongxiaoc, line 153: shuffle = self.shuffle?

I think there are many cases that users don't want training samples to be shuffled.

horovod/spark/keras/estimator.py
```diff
@@ -155,12 +158,14 @@ class KerasEstimator(HorovodEstimator, KerasEstimatorParamsReadable,
     custom_objects = Param(Params._dummy(), 'custom_objects', 'custom objects')
     checkpoint_callback = Param(Params._dummy(), 'checkpoint_callback',
                                 'model checkpointing callback')
+    data_module = Param(Params._dummy(), 'data_module', 'data module class to use when reading data')
```
chongxiaoc, line 161: Add getter and setter of datramodule below, for example, at line 253.

```
    def setDataModule(self, value):
        return self._set(data_module=value)

    def getDataModule(self):
        return self.getOrDefault(self.data_module)
```

horovod/spark/keras/remote.py
```diff
@@ -236,56 +225,42 @@ def train(serialized_model, train_rows, val_rows, avg_row_size):
...
-
-                    history = fit(model, train_data, val_data, steps_per_epoch,
-                                  validation_steps, callbacks, verbose)
+
+            data_module_kwargs = {
+                'label_cols': label_columns,                    # nvtabular
+                'continuous_cols': continuous_columns,          # nvtabular
+                'categorical_cols': categorical_columns,        # nvtabular
+                'make_dataset': make_dataset,                   # petastorm
+                'random_seed': random_seed,                     # petastorm
```
chongxiaoc, line 234: I think `random_seed` can be used for NVT.

horovod/spark/keras/remote.py
```diff
@@ -236,56 +225,42 @@ def train(serialized_model, train_rows, val_rows, avg_row_size):
...
+                'train_dir': remote_store.train_data_path,
+                'val_dir': remote_store.val_data_path,
+                'num_train_epochs': epochs,
+                'has_val': should_validate is not None,
+                'train_batch_size': batch_size,
+                'val_batch_size': val_batch_size,
+                'shuffle': shuffle,
+                'random_seed': random_seed,
```
chongxiaoc, line 242: duplicated `random_seed` here. remove.

horovod/spark/keras/remote.py
```diff
@@ -236,56 +225,42 @@ def train(serialized_model, train_rows, val_rows, avg_row_size):
...
+                'num_reader_epochs': epochs,
```
chongxiaoc, line 243: `num_reader_epochs` is not used in Petastorm for Keras/TF, remove it.
You see that we use` epoch=1` and `loop infinitely` in dataloader to avoid introducing duplicated samples in every epoch.
I think you got it from Lightning datamodule usage. Sorry about confusion.

I agree Petastorm has difference of Keras/TF and Lightning, I will try to clean them a little bit after merging this PR.

horovod/spark/keras/remote.py
```diff
@@ -236,56 +225,42 @@ def train(serialized_model, train_rows, val_rows, avg_row_size):
...
+
+            data_module_kwargs = {
+                'label_cols': label_columns,                    # nvtabular
+                'continuous_cols': continuous_columns,          # nvtabular
+                'categorical_cols': categorical_columns,        # nvtabular
+                'make_dataset': make_dataset,                   # petastorm
+                'random_seed': random_seed,                     # petastorm
+                'train_dir': remote_store.train_data_path,
+                'val_dir': remote_store.val_data_path,
+                'num_train_epochs': epochs,
```
chongxiaoc, line 237: It is okay to pass `num_train_epochs` for generic `DataModule` constructor, though it is not used in either Petastorm or NVTabular. But I think `number of epochs` is a generic attribute of data loader.

chongxiaoc (changes requested):
New round of checks. Just a few nits. For the NVTabular part, feel free to comment it since I'm not sure.

horovod/spark/keras/datamodule.py
leewyang: This `seed_fn` is invoked on each worker to seed and re-seed the shuffle operation (consistently across workers), so probably should leave as is.

horovod/spark/keras/remote.py
leewyang: see comment above.

horovod/spark/keras/remote.py
leewyang: Also removed `debug_data_loader`, `train_async_data_loader_queue_size`, and `val_async_data_loader_queue_size`, since these were similar.

horovod/spark/keras/datamodule.py
chongxiaoc: SG. Leave as it.

horovod/spark/keras/remote.py
chongxiaoc: SG.

horovod/spark/keras/remote.py
chongxiaoc: sure.

chongxiaoc (approved):
Great work. Wait for CI completes and merge.

PR #3621, opened 2022-07-27
Add support for batched memory copies in GPUReducescatter

maxhgerlach:

## Description

In a similar fashion to the recent PR #3590 for Allgather, this adds support for more efficient, batched memory copies in the GPU implementation of Reducescatter.

I've also fixed a related minor bug in `Controller::FuseResponses()`, where extra padding bytes were counted mistakenly for Reducescatter when aggregating tensors up to fusion buffer size.

To improve memory alignment this padding could be introduced for Reducescatter (and Allgather), too, but it would be a bit trickier to get right than for Allreduce.

## Review process to land 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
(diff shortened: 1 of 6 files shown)


romerojosh (approved):
LGTM @maxhgerlach! Glad to see this batched memcopy kernel getting more use.

PR #3627, opened 2022-07-30
[CI] Upgrade PySpark and Torch

EnricoMi:
Upgrades CI test frameworks PySpark to 3.3.0 and 3.2.2 and torch to 1.12.1.

Also upgrades ML frameworks versions for macOS.

GPU heads test did not test torch head (1.13) but torch 1.12. Added a step to `Dockerfile.test.[cg]pu` to check installed python package versions against configured versions to discover those situations.


maxhgerlach (approved):
LGTM, good idea to cross check the framework package versions!

We probably should update PyTorch for the macOS pipeline, too.

EnricoMi:
Thanks for approval, good point re macOS. I have upgraded where possible. Waiting for CI to complete.

PR #3628, opened 2022-08-02
Add register_local_source and use_generic_names funtionality to DistributedGradientTape for TF.

romerojosh:

## Description
This PR adds a couple of additional functionalities to `DistributedGradientTape` to enable better gradient handling for model parallel cases.

1. Added new method `register_local_source` to `DistributedGradientTape`: This enables users to mark a source/variable as worker local so Horovod will skip performing any global averaging on gradients associated with that variable during a call to `gradients` and will instead return the unmodified local gradient. For example:
```
    tape = hvd.DistributedGradientTape(tape)

      if <var is worker local>:
        tape.register_local_source(var)

    # Compute gradients. Any gradient associated with a var passed to register_local_source will not be modified by Horovod.
    gradients = tape.gradient(loss, model.trainable_variables)

```

2. Added a new option `use_generic_names` to the `gradients` method: In model parallel scenarios, it is often the case that common gradients on workers that need to be allreduced might be generated on different workers using different logical paths/graphs. TF naming scheme depends on the code paths that create the variables and this can lead to naming mismatches and deadlocks. This issue was resolved for Horovod operations on training variables in #2328 but not the allreduce on gradients. The solution for training variables was to use custom names and also strip the outer name scope applied by TF using the introduced `ignore_name_scope` option. The `use_generic_names` option applies the same fix to gradients by supplying custom generic names (e.g. `grad_0`, `grad_1`,...) and the `ignore_name_scope` option to the underlying allreduce calls on the gradients.

### DIFF
--- a/horovod/tensorflow/__init__.py
+++ b/horovod/tensorflow/__init__.py
@@ -539,7 +539,7 @@ def _make_cached_allreduce_grads_fn(name, device_dense, device_sparse,
                                                                postscale_factor=postscale_factor,
                                                                process_set=process_set,
                                                                name=f"grad_{i}" if use_generic_names else None,
-                                                               ignore_name_scope=True if use_generic_names else False)
+                                                               ignore_name_scope=use_generic_names)
                     for i in range(len(index_group)):
                         reduce_ops[index_group[i]] = reduce_ops_group[i]
                 return reduce_ops
@@ -553,7 +553,7 @@ def _make_cached_allreduce_grads_fn(name, device_dense, device_sparse,
                                     postscale_factor=postscale_factor,
                                     process_set=process_set,
                                     name=f"grad_{i}" if use_generic_names else None,
-                                    ignore_name_scope=True if use_generic_names else False)
+                                    ignore_name_scope=use_generic_names)
                     if grad is not None else grad
                     for i, grad in enumerate(grads)]
 
--- a/horovod/tensorflow/__init__.py
+++ b/horovod/tensorflow/__init__.py
@@ -44,6 +44,7 @@ from horovod.tensorflow.sync_batch_norm import SyncBatchNormalization
 from horovod.tensorflow.gradient_aggregation import LocalGradientAggregationHelper
 
 import tensorflow as tf
+_IS_TF2 = LooseVersion(tf.__version__) >= LooseVersion('2.0.0')
 
 # @DEKHTIARJonathan: Do not remove, this fixes issues:
 # - https://github.com/tensorflow/tensorflow/issues/38516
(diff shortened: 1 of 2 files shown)


MrAta:
@romerojosh we're using pretty much the same patch for our model parallel training, though we call it `PartialDistributedGradientTape` since it's a different use case than the default data parallel use case. 
What do you think about an `PartialDistributedGradientTape(tape, [local_layers])` API which is a wrapper around the `DistributedGradientTape` with local layers? I'ld be happy to raise a PR after this one is merged.

Also, regarding registering local variables, not sure if there will be use cases that need variable level granularity as opposed to layer level granularity, and hence the `[local_layers]` in  `PartialDistributedGradientTape(tape, [local_layers])`. Any thoughts on that?

romerojosh:
Hey @MrAta, thanks for the comment!

Adding an optional list argument to `DistributedGradientTape` like `DistributedGradientTape(tape, local_vars=[...])`  was an alternative approach instead of the `register_local_variables` I considered, but thought it would be more flexible to allow users to to add variables to this list after creating the tape (perhaps in situations where the variables aren't known at tape creation time). 

I think adding a `PartialDistributedGradientTape(tape, [local_layers])` wrapper in a follow up PR would be great! In terms of using layers vs variables, at least for this PR, I wanted to use the same level of granularity that Horovod uses internally to schedule the communication operations (so, variables). For the higher level wrapper you mention, I think layers could make more sense from a user convenience standpoint.

MrAta:
Thanks for the insight, @romerojosh. Great, I'll raise a PR for `PartialDistributedGradientTape` once this PR is merged then.

horovod/tensorflow/__init__.py
```diff
@@ -520,7 +535,9 @@ def allreduce_grads(grads, vars=None):
                                                                op=op,
                                                                prescale_factor=prescale_factor,
                                                                postscale_factor=postscale_factor,
-                                                               process_set=process_set)
+                                                               process_set=process_set,
+                                                               name=f"grad_{i}" if use_generic_names else None,
+                                                               ignore_name_scope=True if use_generic_names else False)
```
MrAta, line 542: is there any reason we don't do `ignore_name_scope=use_generic_names` given that both are booleans?

romerojosh: No particular reason. Your suggestion is cleaner so I made that change. Thanks!

maxhgerlach (approved):
Changes look good to me and the new functionality seems very useful!

The one suggestion that came to mind would be to add a small unit test for `register_local_source()`. However, as it stands we have quite little test coverage for the various existing options of `DistributedGradientTape` and `DistributedOptimizer`, so that feels optional to me. 🙂

PR #3631, opened 2022-08-03
fix the optimizer iteration count when gradient accumulation is enabled

kvignesh1420:

- [x] Did you update the docs? NA
- [x] Did you write any tests to validate this change?  Current tests suffice.

## Description

When gradient accumulation is enabled (`backward_passes_per_step > 1`), the underlying optimizer iteration count update uses the `optimizer._iterations` private variable which might not be exposed for custom optimizers such as the [LossScaleOptimizer](https://github.com/keras-team/keras/blob/059781f3b02ff01d419637c76de833002869ff92/keras/mixed_precision/loss_scale_optimizer.py#L587) in keras. 

The users who have been using such optimizers would not have incremented the underlying iterations property as we just return a `tf.no_op()`, which in turn affects the update calculation. For instance, when we wrap  `Adam` optimizer with `LossScaleOptimizer`, the adam updates would have been miscalculated. This PR fixes this issue by using the safer `iterations` property.

## Review process to land 

### DIFF
--- a/horovod/tensorflow/gradient_aggregation.py
+++ b/horovod/tensorflow/gradient_aggregation.py
@@ -241,8 +241,14 @@ class LocalGradientAggregationHelper:
         # If optimizer tracks iterations, we increment it on steps where we
         # are not going to call `apply_gradients()`.
         def increment_optimizer_iteration():
-            if hasattr(optimizer, "_iterations") and optimizer._iterations is not None:
-                return optimizer._iterations.assign_add(1).op
+            # (kvignesh1420): Since all `tf.OptimizerV2` instances have the `iterations`
+            # property for modifying the underlying `optimizer._iterations`, it is safe to use
+            # the property instead of the private variable. For instance, the keras
+            # `LossScaleOptimizer` inherits `tf.Optimizer` and exposes the cleaner `iterations`
+            # property instead of the unsafe `_iterations`.
+
+            if hasattr(optimizer, "iterations") and optimizer.iterations is not None:
+                return optimizer.iterations.assign_add(1).op
             return tf.no_op()
 
         with tf.control_dependencies([tf.group(*get_not_none_from_list(flattended_args0))]):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Tensorflow: Fixed the optimizer iteration increments when `backward_passes_per_step > 1`. ([#3631](https://github.com/horovod/horovod/pull/3631))
 - Updated Eigen submodule to fix build on macOS with aarch64. ([#3619](https://github.com/horovod/horovod/pull/3619))
 - Fix FuseResponses() on BATCHED_D2D_PADDING edge cases for Reducescatter and/or ROCm. ([#3621](https://github.com/horovod/horovod/pull/3621))
 - PyTorch: Fixed Reducescatter functions to raise `HorovodInternalError` rather than `RuntimeError`. ([#3594](https://github.com/horovod/horovod/pull/3594))
--- a/horovod/tensorflow/gradient_aggregation_eager.py
+++ b/horovod/tensorflow/gradient_aggregation_eager.py
@@ -124,8 +124,14 @@ class LocalGradientAggregationHelperEager:
 
     def apply_gradients(self, apply_grads_closure, optimizer, *args, **kwargs):
         def increment_optimizer_iteration():
-            if hasattr(optimizer, "_iterations") and optimizer._iterations is not None:
-                return optimizer._iterations.assign_add(1).op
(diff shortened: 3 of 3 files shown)


kvignesh1420:
cc: @chongxiaoc @maxhgerlach @romerojosh

kvignesh1420:
cc: @chongxiaoc @maxhgerlach @romerojosh, would be great to get a review on this 🙂 . Thanks!

horovod/tensorflow/gradient_aggregation.py
```diff
@@ -241,8 +241,17 @@ def apply_gradients(self, apply_grads_closure, optimizer, *args, **kwargs):
...
+            # (kvignesh1420): Since all `tf.OptimizerV2` instances have the `iterations`
+            # property for modifying the underlying `optimizer._iterations`, it is safe to use
+            # the property instead of the private variable. For instance, the keras
+            # `LossScaleOptimizer` inherits `tf.Optimizer` and exposes the cleaner `iterations`
+            # property instead of the unsafe `_iterations`.
+
+            # NOTE: For users who wrap their optimizers with `LossScaleOptimizer`, the increment op
+            # before this change would have incorrectly updated the underlying optimizer iterations count,
+            # i.e only once every `backward_passes_per_step`.
+            if hasattr(optimizer, "iterations") and optimizer.iterations is not None:
```
maxhgerlach, line 253: These comments will be confusing after the change has landed. I think it would be sufficient to have their content in the PR description text on GitHub.

maxhgerlach (commented):
Hi @kvignesh1420,

the change looks sensible to me under the condition that there is no standard TensorFlow Optimizer that has an attribute `_iterations`, but not `iterations`.

As far as I can tell that assumption should hold:

- For `OptimizerV2` the property `iterations()` has been added before TF version 1.12.1 (https://github.com/tensorflow/tensorflow/commit/1a4e13725f04d8b27ed8f6c5feeed5ea7b2e54e9) and we only support TF 1.15 and TF 2.
- TF1-style `Optimizer` does not have a member variable `_iterations`. I can't find that string anywhere in `tensorflow/python/training/optimizer.py` or neighboring files like `adam.py` in a search on TF master: https://cs.opensource.google/search?q=_iterations&sq=&ss=tensorflow%2Ftensorflow:tensorflow%2Fpython%2Ftraining%2F

Did I get that right?

Note that there was a test failure, but it looks like a spurious timeout that should be fixed once https://github.com/horovod/horovod/pull/3632 has landed.

horovod/tensorflow/gradient_aggregation.py
kvignesh1420: removed them in the commit

kvignesh1420:
> Hi @kvignesh1420,
> 
> the change looks sensible to me under the condition that there is no standard TensorFlow Optimizer that has an attribute `_iterations`, but not `iterations`.
> 
> As far as I can tell that assumption should hold:
> 
> * For `OptimizerV2` the property `iterations()` has been added before TF version 1.12.1 ([tensorflow/tensorflow@1a4e137](https://github.com/tensorflow/tensorflow/commit/1a4e13725f04d8b27ed8f6c5feeed5ea7b2e54e9)) and we only support TF 1.15 and TF 2.
> * TF1-style `Optimizer` does not have a member variable `_iterations` at all. I can't find that string anywhere in `tensorflow/python/training/optimizer.py` or neighboring files like `adam.py` in a search on TF master: https://cs.opensource.google/search?q=_iterations&sq=&ss=tensorflow%2Ftensorflow:tensorflow%2Fpython%2Ftraining%2F
> 
> Did I get that right?
> 
> Note that there was a test failure (https://github.com/horovod/horovod/runs/7664566671?check_suite_focus=true), but it looks like a spurious timeout that should be fixed once #3632 has landed.

@maxhgerlach yes you got that right. `OptimizerV1` doesn't have a `_iterations` property but it does have the `iterations` property. For ex: the [SGD V1 optimizer](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/optimizer_v1.py#L178). Thus, for backward compatibility it makes sense that `OptimizerV2` also has the `iterations` property, but is implemented as a getter/setter.  Also, as you mentioned the versions of TF that horovod supports has optimizers which have the `iterations` property so we shouldn't be having any issues AFAIK.

maxhgerlach:
> OptimizerV1 doesn't have a _iterations property but it does have the iterations property. For ex: the [SGD V1 optimizer](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/optimizer_v1.py#L178).

Ah, very good! I missed that third type of optimizers, Keras V1.

The new, fourth type of optimizers `tf.keras.optimizers.experimental.Optimizer` still has the `iterations` property (https://github.com/keras-team/keras/blob/v2.9.0/keras/optimizers/optimizer_experimental/optimizer.py#L207), so we are good there, too. Though I'm not sure if Horovod supports that one already.

maxhgerlach (approved):
LGTM!

PR #3632, opened 2022-08-04
TF: Make most tests use statically initialized process sets

maxhgerlach:

## Description

As observed in #3557 there can be significant slowdowns with TensorFlow 2.9 when `HOROVOD_DYNAMIC_PROCESS_SETS=1` is set and MPI is used rather than Gloo. This PR refactors most process-set-specific tests for TensorFlow so they work with statically initialized process sets. This should avoid timeouts in CI pipelines on GPU-equipped AWS hosts.

## Review process to land 

### DIFF
--- a/test/parallel/test_tensorflow2_keras_process_sets.py
+++ b/test/parallel/test_tensorflow2_keras_process_sets.py
@@ -17,11 +17,6 @@ from horovod.runner.common.util.env import get_env_rank_and_size
 
 _PRE_TF_2_2_0 = LooseVersion(tf.__version__) < LooseVersion("2.2.0")
 
-# Set environment variable to enable adding/removing process sets after
-# initializing Horovod.
-os.environ["HOROVOD_DYNAMIC_PROCESS_SETS"] = "1"
-
-
 @pytest.mark.skipif(LooseVersion(tf.__version__) <
                     LooseVersion('2.0.0'), reason='TensorFlow v2 tests')
 class Tf2KerasProcessSetsTests(tf.test.TestCase):
--- a/test/parallel/test_tensorflow2_keras_process_sets.py
+++ b/test/parallel/test_tensorflow2_keras_process_sets.py
@@ -6,14 +6,10 @@ tests for multiple process sets in this script that initializes Horovod with sta
 """
 
 import tensorflow as tf
-import os
-import sys
 import warnings
 from distutils.version import LooseVersion
 import pytest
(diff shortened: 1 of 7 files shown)


test/parallel/test_tensorflow_process_sets.py
```diff
@@ -23,88 +23,86 @@
...
-    Tests for ops in horovod.tensorflow that add/remove process sets after initialization.
+    Tests for ops in horovod.tensorflow using multiple process sets.
     """
     def __init__(self, *args, **kwargs):
         super(TensorFlowProcessSetsTests, self).__init__(*args, **kwargs)
 
+    @classmethod
+    def setUpClass(cls):
+        """Initializes Horovod with two process sets"""
+        _, mpi_size = mpi_env_rank_and_size()
```
EnricoMi, line 42: You can use `horovod.runner.common.util.env.get_env_rank_and_size`, which is agnostic to the specific MPI kind.

I think general use of `common.mpi_env_rank_and_size` can be replaced with that.

maxhgerlach: Thanks, I've replaced that for the new static process set tests. As it stands, the existing `test_process_sets_static.py` also needs to know whether it's going to run with MPI or Gloo before it initializes Horovod, so I've left it as it is for now.

BTW `get_env_rank_and_size()` could be useful to expose as general user API. Everybody who wants to use static process sets needs that information from some source before they initialize Horovod.

test/parallel/test_tensorflow2_keras_process_sets.py
```diff
@@ -1,21 +1,22 @@
...
+import sys
 import warnings
-
 from distutils.version import LooseVersion
-
 import pytest
-
 from tensorflow import keras
 
+sys.path.append(os.path.join(os.path.dirname(__file__), os.pardir, 'utils'))
```
EnricoMi, line 16: Is this still needed?

maxhgerlach: It isn't. Thanks for pointing that out!

I've cleaned up the imports and `sys.path` manipulation in these test scripts and removed one extra `os.environ["HOROVOD_DYNAMIC_PROCESS_SETS"] = "1"` that fell through the cracks before, also rebased to master.

EnricoMi (approved):

maxhgerlach:
Thanks for the approval!

PR #3637, opened 2022-08-09
add PartialDistributedGradientTape API

MrAta:

## Description
This PR adds a new API called `PartialDistributedGradientTape(tape, [local_layers], ...)` for model parallel use cases where unlike `DistributedGradientTape()` ranks only sync a subset of gradients and skip the gradients of their "local layers". A sample usage is shown in the unit test included with this PR.

This is a somehow parallel work to #3628 and we would ideally rebase and merge when that PR is merged as it gives more fine grained control for local gradients through local variables as opposed to local layers.
I'm raising this PR to get early feedbacks on the API itself from folks and would rebase and merge after #3628  as discussed there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added `hvd.grouped_allgather()` and `hvd.grouped_reducescatter()` operations. ([#3594](https://github.com/horovod/horovod/pull/3594))
 - TensorFlow: Added doc string for `hvd.grouped_allreduce()`. ([#3594](https://github.com/horovod/horovod/pull/3594))
 - Added warning messages if output tensor memory allocations fail. ([#3594](https://github.com/horovod/horovod/pull/3594))
+- Added `register_local_source` and `use_generic_names` funtionality to DistributedGradientTape. ([#3628](https://github.com/horovod/horovod/pull/3628))
 
 ### Changed
 
--- a/horovod/tensorflow/__init__.py
+++ b/horovod/tensorflow/__init__.py
@@ -16,6 +16,7 @@
 # ==============================================================================
 # pylint: disable=g-short-docstring-punctuation
 
+from distutils.version import LooseVersion
 import os
 import warnings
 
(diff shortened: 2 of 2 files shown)


test/parallel/test_tensorflow.py
```diff
@@ -3842,6 +3842,40 @@ def test_horovod_grouped_allgather_gpu(self):
                     tf.equal(tf.cast(rank_tensor, tf.int32), value))) for rank_tensor in rank_tensors),
                     "hvd.grouped_allgather produces incorrect gathered tensor")
 
+    def test_partial_distributed_gradient_tape_process_sets(self):
```
maxhgerlach, line 3845: Hi @MrAta, just commenting on this test for now (thanks for including one with the PR!).

- It should not have `process_sets` in its name as it does not use multiple process sets as far as I can tell.
- Should it be moved to `test_tensorflow2_keras.py`? At least TF1 test configurations will fail executing this test.

There is a failure message in the one CI pipeline that runs without explicit approval, PPC64 with TF 2.6.0 (https://powerci.osuosl.org/job/Horovod_PPC64LE_GPU_PIPELINE/view/change-requests/job/PR-3637/1/console). Not sure if this is may be related to the second point above, if it's specific to that build configuration or if its something regarding eager vs. graph-based execution (all tests in `test_tensorflow.py` are run in both configurations, might make sense to detect that and skip if appropriate).

MrAta: Hi @maxhgerlach thanks for the inputs!

-  Sure, I'll remove the process_sets name from the test name.
- Regarding moving it to `test_tensorflow2_keras.py` I initially thought would do that as we're testing with a keras model. But, since the API is under `horovod.tensorflow` not `horovod.keras` I thought I'd keep it in `test_tensorflow.py`. What do you think?

As for the test failure, yeah I have to investigate it further. It passes with our setup (tf2.4, hvd0.23). I was unable to build a docker image out of dockerfile in the master branch to test it with the latest versions. Will give it a try again and fix the test.

Thanks!

maxhgerlach: > But, since the API is under horovod.tensorflow not horovod.keras I thought I'd keep it in test_tensorflow.py. What do you think?

True, that is a good argument to keep it in `test_tensorflow.py`. Maybe it will need some annotations for certain versions of TF then. I've kicked off the rest of the CI workflows, maybe it passes in some configurations already.

> I was unable to build a docker image out of dockerfile in the master branch to test it with the latest versions.

The easiest way is to pick one of the configurations from [`docker-compose.test.yml`](https://github.com/horovod/horovod/blob/master/docker-compose.test.yml), e.g., `test-cpu-gloo-py3_8-tf2_9_1-keras2_9_0-torch1_11_0-mxnet1_9_1-pyspark3_2_1`. Then run
```
docker-compose -f docker-compose.test.yml build --pull test-cpu-gloo-py3_8-tf2_9_1-keras2_9_0-torch1_11_0-mxnet1_9_1-pyspark3_2_1```

PR #3639, opened 2022-08-09
Fix race condition in PyTorch allocation handling.

romerojosh:

## Description
This PR fixes what appears to be a quite serious bug in the handling of buffer allocations in the PyTorch backend that was discovered and reported by @eordentlich at NVIDIA. The issue is that the current implementation does not enforce any type of synchronization after PyTorch buffers are allocated. This is a problem because the PyTorch allocations can be completed asynchronously and not enforcing a sync or recording a dependency using an event results in a race condition when writing to these buffers. This was handled appropriately in the TensorFlow implementation (see: https://github.com/horovod/horovod/blob/ed15ddc8f269664d4a89671b33960d0fab9f2dd9/horovod/tensorflow/mpi_ops.cc#L404-L411), but similar code was never introduced for the PyTorch backend. I've added equivalent synchronization code to the PyTorch backend in this PR.

The cases that are most impacted by this bug are collectives that need to allocate output buffers, since the size is not known at enqueue time (e.g. `allgather`, `reduce_scatter`, and `alltoall`). In these cases, the output data from Horovod can be written to the buffers prior to them being available. This issue can also impact the allocation of the persistent buffer used for tensor fusion; however, I think that allocation occurs very early during job startup, so maybe is somehow less likely to occur. Since `allreduce` and `broadcast` do not allocate output buffers within the Horovod backend, those collectives should be unimpacted (outside of the possible issue with the fusion buffer).

I've taken this as an opportunity to also extend the performance improvement from #3464 to the PyTorch backend where appropriate.


chongxiaoc (approved):
LGTM

PR #3646, opened 2022-08-11
Enable use of native ncclAvg op for NCCL allreduces.

romerojosh:

## Description
NCCL 2.10+ introduced a native `ncclAvg` operation (https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/types.html#c.ncclAvg). This PR enables the use of this feature for versions of NCCL that support it.

### DIFF
--- a/horovod/common/ops/nccl_operations.cc
+++ b/horovod/common/ops/nccl_operations.cc
@@ -206,7 +206,7 @@ Status NCCLAllreduce::Execute(std::vector<TensorTableEntry>& entries,
   // Copy (and possibly scale) tensors into the fusion buffer.
   if (entries.size() > 1) {
     ScaleMemcpyInFusionBuffer(entries, fused_input_data, buffer_data,
-                              buffer_len, response.prescale_factor());
+                              buffer_len, prescale_factor);
     if (global_state_->timeline.Initialized()) {
       gpu_context_->RecordEvent(gpu_op_context_.event_queue,
                                 MEMCPY_IN_FUSION_BUFFER,
@@ -220,7 +220,7 @@ Status NCCLAllreduce::Execute(std::vector<TensorTableEntry>& entries,
         buffer_len / DataType_Size(first_entry.tensor->dtype());
     if (prescale_factor != 1.0) {
       // Execute prescaling op
-      ScaleBuffer(response.prescale_factor(), entries, fused_input_data,
+      ScaleBuffer(prescale_factor, entries, fused_input_data,
                   buffer_data, num_elements);
       fused_input_data = buffer_data; // for unfused, scale is done out of place
     }
--- a/horovod/common/ops/nccl_operations.cc
+++ b/horovod/common/ops/nccl_operations.cc
@@ -189,6 +189,20 @@ Status NCCLAllreduce::Execute(std::vector<TensorTableEntry>& entries,
   void* buffer_data;
   size_t buffer_len;
 
+  ncclRedOp_t ncclOp = ncclSum;
+  double prescale_factor = response.prescale_factor();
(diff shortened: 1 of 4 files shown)


horovod/mxnet/__init__.py
```diff
@@ -47,10 +47,16 @@ def __init__(self, optimizer, gradient_predivide_factor=1.0, num_groups=0, proce
...
 
         self._optimizer = optimizer
-        # Normalizing rescale_grad by Horovod size, which is equivalent to
-        # performing average in allreduce, has better performance.
-        self._optimizer.rescale_grad *= (gradient_predivide_factor / process_set.size())
         self._gradient_predivide_factor = gradient_predivide_factor
+        self._average_in_framework=False
+        # C++ backend will apply additional 1 / size() factor to postscale_factor for op == Average.
+        self._postscale_factor = self._gradient_predivide_factor
+        if rocm_built() or nccl_built() < 21000:
```
maxhgerlach, line 54: It's a bit surprising to see this case distinction on this level. I would have expected it in C++ `horovod_mxnet_allreduce_async()`, but if that's not possible easily, it will be fine like this.

Overall the code path with averaging to be quite convoluted with MXNet. Why do we have more branches in framework-specific code with MXNet than we do with TensorFlow for instance? Does it make Horovod with MXNet faster than with the other frameworks?

maxhgerlach (approved):
Sorry for keeping you waiting a bit, @romerojosh. This is a great addition.

Can using the `ncclAvg` op lead to performance or numerical stability improvements with some systems?

I've left a comment / question regarding the MXNet code, but nothing that would block a merge.

horovod/common/ops/nccl_operations.cc
```diff
@@ -189,10 +189,24 @@ Status NCCLAllreduce::Execute(std::vector<TensorTableEntry>& entries,
...
   size_t buffer_len;
 
+  ncclRedOp_t ncclOp = ncclSum;
+  double prescale_factor = response.prescale_factor();
+  double postscale_factor = response.postscale_factor();
+#ifdef NCCL_AVG_SUPPORTED
+  auto& process_set =
+      global_state_->process_set_table.Get(entries[0].process_set_id);
+  if (prescale_factor == 1.0 &&
+      postscale_factor == 1.0 / process_set.controller->GetSize()) {
```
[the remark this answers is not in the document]
romerojosh: Yes, this is a great comment. Because we set `postscale_factor` ourselves, I believe this comparison should be safe.

horovod/mxnet/__init__.py
romerojosh: The reason for this is because MXNet, unlike the other frameworks, has the ability to rescale gradients in the optimizer (via `rescale_grad`), usually with the scaling fused into the weight update kernel. In the past, this was a nice performance optimization because Horovod required calling a separate kernel for scaling gradients (or even worse, called a kernel per individual gradient for scaling). By using MXNet's gradient rescaling functionality, we were able to cut down on the number of kernels launched and improve performance. Since I fused scaling into the buffer packing, this has been of less benefit; however, I maintained the older path for platforms (e.g. ROCM) that didn't have the CUDA kernels I wrote to do the scaling/packing and could still benefit from the gradient scaling in the optimizer. Perhaps this is no longer needed since #3486, but it is hard to say without any CI testing of that platform.

maxhgerlach: Ah, thank you, I didn't know the `rescale_grad` feature would be that powerful. That's useful!

EnricoMi:

maxhgerlach:
Tests look good. Are we ready to merge?

Issue #3642, opened 2022-08-10
Tensorflow: support int8 allreduce


kvignesh1420:
**Is your feature request related to a problem? Please describe.**
Currently, `HorovodAllreduce` tensorflow op is registered to support allreduce on `int32, int64, float16, float32, float64` types but not on `int8` even though the backends support it (ex: NCCL, MPI). 

**Describe the solution you'd like**
I plan to add `int8` to the supported types of `HorovodAllreduce`.

**Describe alternatives you've considered**
An alternative approach would be to use `allgather` on `int8` tensors and then do the addition/averaging in the user code, which is not efficient.

**Additional context**
With this in place, I believe we can also add an `Int8Compressor` which is extreme but people have used `1bit` compressors in the literature so `8bit` compression can definitely help.

maxhgerlach:
Is there a specific reason HorovodAllreduce only supports `int32, int64, float16, float32` (other than avoiding integer overflows)? `HorovodAllgather` and `HorovodBroadcast` have supported `int8` and `uint8` since the initial commit.

kvignesh1420:
I am not familiar with the initial design decisions but to address your question of integer overflows, I was thinking that in the compressor, we can:
1.  Threshold the values to be in the range `[-128, 127]`
2. Cast the thresholded tensor to int8
3. Cast it back to fp32 during decompression.

There will be loss of information with thresholding but at-least it would be deterministic.

maxhgerlach:
In combination with a compressor `int8` could make a lot of sense!

I was just wondering if anybody recalled any TF kernel incompatibilities. For instance Horovod's in-place broadcast op for TF does not support `(u)int16` because there were no `DenseUpdate` kernels for those data types.

romerojosh:
I don't recall there being any incompatibilities. Probably just an oversight early on in Horovod development.

kvignesh1420:
I see. So shall I go ahead and enable int8 allreduce and add the Int8Compressor? WDYT @maxhgerlach @romerojosh

romerojosh:
Yes, I think that is a great idea. I'd suggest splitting this into an initial PR that adds (u)int8 support to TF (and the other frameworks if necessary) with extended testing, then a PR for the compressor if that sounds good to you.

kvignesh1420:
sounds good!

maxhgerlach:
#3649 for TensorFlow is merged. Do you also want to look into PyTorch and MXNet?

[...]

PR #3649, opened 2022-08-15
feat: support uint8 and int8 allreduce in tensorflow

kvignesh1420:

## Description

This PR adds support for `int8` and `uint8` allreduce in tensorflow along with the relevant tests. 
Corresponding issue:  https://github.com/horovod/horovod/issues/3642

Signed-off-by: Vignesh Kothapalli <k.vignesh1420@gmail.com>

## Review process to land 

### DIFF
--- a/test/parallel/test_tensorflow.py
+++ b/test/parallel/test_tensorflow.py
@@ -117,19 +117,22 @@ class TensorFlowTests(BaseTensorFlowTests):
         """Test on CPU that the allreduce correctly sums 1D, 2D, 3D tensors."""
         hvd.init()
         size = hvd.size()
-        dtypes = self.filter_supported_types([tf.int32, tf.int64, tf.float16, tf.float32, tf.float64])
+        dtypes = self.filter_supported_types([tf.uint8, tf.int8, tf.int32, tf.int64, tf.float16, tf.float32, tf.float64])
         dims = [1, 2, 3]
         for dtype, dim in itertools.product(dtypes, dims):
             with tf.device("/cpu:0"):
-                tensor = self.random_uniform(
-                    [17] * dim, -100, 100, dtype=dtype)
+                # prevent underflows/overflows in uint8, int8
+                tensor = self.random_uniform([17] * dim, -100, 100)
+                tensor = tf.cast(tensor, dtype=dtype)
                 summed = hvd.allreduce(tensor, average=False)
             multiplied = tensor * size
-            max_difference = tf.reduce_max(tf.abs(summed - multiplied))
+            difference = summed - multiplied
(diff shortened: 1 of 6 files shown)


test/parallel/test_tensorflow.py
```diff
@@ -117,19 +117,22 @@ def test_horovod_allreduce_cpu(self):
...
         size = hvd.size()
-        dtypes = self.filter_supported_types([tf.int32, tf.int64, tf.float16, tf.float32, tf.float64])
+        dtypes = self.filter_supported_types([tf.uint8, tf.int8, tf.int32, tf.int64, tf.float16, tf.float32, tf.float64])
         dims = [1, 2, 3]
         for dtype, dim in itertools.product(dtypes, dims):
             with tf.device("/cpu:0"):
-                tensor = self.random_uniform(
-                    [17] * dim, -100, 100, dtype=dtype)
+                # prevent underflows/overflows in uint8, int8
+                tensor = self.random_uniform([17] * dim, 0, 10)
```
maxhgerlach, line 125: I'd suggest to leave range at [-100,100] in general and add a special case for `uint8` if necessary (like taking the absolute value and casting). 

Removing any negative tensor values from the test coverage does not sound wise and I'd also prefer not to change the range with this PR where it's not necessary.

(comment applies to all tests adapted for `(u)int8`)

maxhgerlach (commented):
Thanks for taking a stab at this, @kvignesh1420!

I've triggered the CI pipeline and have left one comment regarding the test script.

test/parallel/test_tensorflow.py
kvignesh1420: sure, will change it.

kvignesh1420:
@maxhgerlach I reverted the range back to `(-100, 100)` and fixed the tests for `int8, uint8` when `average=True`. Please take a look.

maxhgerlach:
Thank you @kvignesh1420, that looks great!

There is a bunch of additional TensorFlow allreduce tests, scattered over a couple of separate files:
- test_tensorflow_process_sets.py
- test_xla.py
- test_xla_process_sets.py

I think it would be good to extend the `(u)int8` test coverage at least to `test_xla.py` as that one uses a different backend than `test_tensorflow.py`. 

There would be less to gain from the various tests with multiple process sets in the two other files, so I'm not sure if those are worth the busy work.

maxhgerlach (approved):
LGTM, let's just wait for the tests to pass. Thank you for the contribution!

maxhgerlach:
Tests are all good. The failure at "CI / Build docker image horovod-ray" is unrelated.

PR #3643, opened 2022-08-11
add PartialDistributedGradientTape API

MrAta:
Signed-off-by: Ata FatahiBaarzi <afatahibaarzi@linkedin.com>

## Checklist before submitting

- [x] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)?
- [ ] Did you update the docs?
- [x] Did you write any tests to validate this change?  
- [x] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users?

## Description
This PR builds upon #3628 and adds a new API called `PartialDistributedGradientTape(tape, [local_layers], ...)` for model parallel use cases where unlike DistributedGradientTape() ranks only sync a subset of gradients and skip the gradients of their "local layers". A sample usage is shown in the unit test included with this PR.

## Review process to land 

1. All tests and other checks must succeed.
2. At least one member of the [technical steering committee](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md) must review and approve.
3. If any member of the technical steering committee requests changes, they must be addressed.

PR #3663, opened 2022-08-23
TF: Add register_local_var to distributed optimizers and gradient aggregators

MrAta:

## Description
This is to continue work done in #3628 and #3643 and adds the same functionality to the Distributed Optimizers and Local Gradient Aggregators. This is useful for model parallel use cases that use distributed optimizers and want to skip syncing their "local" gradients.

An example usage is shown in the unit tests included PR, but in short it's like the following:
```python
    ...

      if <var is worker local>:
          opt.register_local_var(var)

    # Compute gradients. Any gradient associated with a var passed to register_local_var will not be modified by Horovod.
    gradients = opt.compute_gradients(loss, model.trainable_variables, tape)
```

If this change get merged, similar to #3643 we can possibly add a new API called `PartialDistributedOptimizer`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added `register_local_source` and `use_generic_names` funtionality to DistributedGradientTape. ([#3628](https://github.com/horovod/horovod/pull/3628))
 - Added `transformation_edit_fields` and `transformation_removed_fields` param for EstimatorParams. ([#3651](https://github.com/horovod/horovod/pull/3651))
 - Added `PartialDistributedGradientTape()` API for model parallel use cases. ([#3643](https://github.com/horovod/horovod/pull/3643))
+- Added `register_local_var` functionality to distributed optimizers and local gradient aggregators. ([3663](https://github.com/horovod/horovod/pull/3663))
 
 ### Changed
 
--- a/horovod/_keras/__init__.py
+++ b/horovod/_keras/__init__.py
@@ -79,7 +79,7 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
             elif _IS_TF2:
                 self._local_vars.add(var.ref())
             else:
-                self._local_vars.add(var.ref())
+                self._local_vars.add(var)
 
         def _compute_gradients(self, loss, var_list, grad_loss=None, tape=None):
             """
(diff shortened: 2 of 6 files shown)


MrAta:
cc @romerojosh @maxhgerlach would appreciate your review on this PR.

MrAta:
cc @romerojosh @maxhgerlach would appreciate your review on this PR.

romerojosh (approved):
Thanks for extending this feature @MrAta! Changes LGTM!

romerojosh:
On a related topic, Horovod has a callback to broadcast initial variables for Keras (https://github.com/horovod/horovod/blob/master/horovod/_keras/callbacks.py#L23) that currently doesn't take into account any differences between local and global variables. Is this something that has come up in any of your use cases? I'm trying to come up with a nice way to handle this in the callback.

MrAta:
> On a related topic, Horovod has a callback to broadcast initial variables for Keras (https://github.com/horovod/horovod/blob/master/horovod/_keras/callbacks.py#L23) that currently doesn't take into account any differences between local and global variables. Is this something that has come up in any of your use cases? I'm trying to come up with a nice way to handle this in the callback.

That's a good catch, @romerojosh! I just double-checked and for one of our cases since we have a custom loop, we are using `hvd.broadcast_variables(<non-local variables>)` (for both model and optimizer) instead of that callback. But handling it in the callback would definitely be a great feature to add!

MrAta:
@romerojosh re the callback, I can take a stab at it for your review if you haven't already.

skyw:
The callback is needed when using Keras model.fit, in which data loader is fed as a whole. To the minimu, if nothing else, I think a call back take the same list as partial tape should be provided.

romerojosh:
@MrAta Yes, feel free to take a stab at it if you have the time. Thanks!

maxhgerlach:
@MrAta, I fear we missed a test failure in the GPU pipeline with `test_distributed_optimizer_with_local_vars`. Could you give this another look?

It's a bit convoluted because there were timeouts as well, but you can see it on this log line for instance: https://buildkite.com/horovod/horovod/builds/8207#0182d288-5319-4257-83cb-3f4b83a5c764/6-4579 (that's the last Buildkite pipeline that ran for this PR)

romerojosh:
@MrAta Sorry for the preemptive merge. It seems that for some reason, the full CI pipeline didn't run for this PR so I didn't see the problem. Please do reopen a PR for this feature and I will review again.

PR #3681, opened 2022-09-07
feat: support gradient accumulation in spark torch estimator

thinkall:

## Description

We can use `backward_passes_per_step` in [torch](https://github.com/horovod/horovod/blob/master/examples/pytorch/pytorch_imagenet_resnet50.py#L273-L278), although we need to adjust the [training code](https://github.com/horovod/horovod/blob/master/examples/pytorch/pytorch_imagenet_resnet50.py#L73-L86) to make it work as expected. However, in [spark torch estimator](https://github.com/horovod/horovod/blob/master/horovod/spark/torch/estimator.py), there is no implementation for supporting it.

In this PR, we added support to spark torch estimator, thus we can apply gradient accumlation by simply set up param `backward_passes_per_step` in TorchEstimator. [Example](https://github.com/horovod/horovod/blob/master/examples/spark/pytorch/pytorch_spark_mnist.py) and [test](https://github.com/horovod/horovod/blob/master/test/integration/test_spark_torch.py) are modified accordingly.

## Review process to land 


EnricoMi:

thinkall:
> Please rebase with latest master, which includes a fix for the CI.

Thanks @EnricoMi , done!

chongxiaoc:
OOO this week. I will take a look when I'm back.

horovod/spark/torch/remote.py
```diff
@@ -194,6 +195,7 @@ def train(serialized_model, optimizer_cls, model_opt_state_serialized,
             # Pass the compression arg only if it is specified by the user.
             dist_optimizer_args['compression'] = gradient_compression
         # Horovod: wrap optimizer with DistributedOptimizer.
+        dist_optimizer_args['backward_passes_per_step'] = backward_passes_per_step
```
Tixxx, line 198: I just realized both torch distributed optimizer and this estimator don't check the value for this. Could you add checks in those places to make sure they are always > 0? Since you are doing division with this number, we need to careful.

horovod/spark/torch/remote.py
```diff
@@ -454,13 +458,16 @@ def _validate(epoch):
...
-        optimizer.zero_grad()
+    def train_minibatch(model, optimizer, transform_outputs, loss_fn, inputs, labels, sample_weights, backward_passes_per_step, batch_idx):
+        if batch_idx % backward_passes_per_step == 0:
+            if batch_idx != 0:
+                optimizer.step()
+            optimizer.zero_grad()
         outputs = model(*inputs)
         outputs, labels = transform_outputs(outputs, labels)
         loss = loss_fn(outputs, labels, sample_weights)
+        loss.div_(float(backward_passes_per_step))
```
Tixxx, line 469: Would be great if we do this only for accumulation >1, we dont want to add new computation nodes for default behavior.

Tixxx: sometimes we will convert gradients to fp16 for allreduce, this pre-division might have an impact on the numerical precision. I suggest to do this after accumulation i.e before weight update()

thinkall: > sometimes we will convert gradients to fp16 for allreduce, this pre-division might have an impact on the numerical precision. I suggest to do this after accumulation i.e before weight update()

If I understand it correctly, allreduce is applied to gradients only, so I guess we need to do loss div before accumulation in order to obtain the correct gradients?

Tixxx: ideally (loss / n) translates to (gradients / n), that's why we can linearly scale learning rate directly when adding more workers, but pre-division might cause underflow because of small gradients. This is minor for now as we dont plan to introduce fp16 allreduce for estimators anytime soon.

Tixxx (approved):

chongxiaoc (approved):

horovod/spark/torch/remote.py
thinkall: > ideally (loss / n) translates to (gradients / n), that's why we can linearly scale learning rate directly when adding more workers, but pre-division might cause underflow because of small gradients. This is minor for now as we dont plan to introduce fp16 allreduce for estimators anytime soon.

Thanks for the clarification.

PR #3685, opened 2022-09-08
add support for local variables for BroadcastGlobalVariablesCallback

MrAta:

## Description
This PR is continuation of #3628, #3643, and #3600 and adds support for local variables for BroadcastGlobalVariablesCallback. So in model parallel use cases, the only the initial value of non-local variables are broadcasted.
An example use case is provided in the unit test included with this PR, but, in short, the callback will be used as follows:

```bash
...
local_variables = <model's local variables>
callbacks = [hvd.callbacks.BroadcastGlobalVariablesCallback(root_rank=root_rank, local_variables=local_variables)]
model.fit(..., callbacks=callbacks)
...
```


MrAta:
cc @romerojosh @maxhgerlach  would appreciate your review for this PR.
One thing that I wasn't sure about is that whether TF1 code uses this callback or not? currently the PR doesn't handle the case for TF1.

EnricoMi:

PR #3686, opened 2022-09-08
Revert "TF: Add register_local_var to distributed optimizers and gradient aggregators"

maxhgerlach:
Reverts horovod/horovod#3663 because `test_distributed_optimizer_with_local_vars` fails on GPU


EnricoMi (approved):
Thank you!

PR #3699, opened 2022-09-13
Fix reducescatter() and grouped_reducescatter() to raise clean exceptions for scalar inputs

### LINKED ISSUE #3698: reducescatter() and grouped_reducescatter() crash for scalar inputs
`hvd.reducescatter(3.14)` currently leads to a C++ assertion failure in debug builds or a segmentation fault in release builds. 

There should either be a clear error message or it should behave similarly to `hvd.reducescatter([3.14])`, which returns a one-element tensor on the root rank and an empty tensor on the other ranks.

maxhgerlach:

## Description

This fixes Horovod crashing when scalar inputs are passed to `hvd.reducescatter` or `hvd.grouped_reducescatter`. Attempts to do so now raise exceptions that users can deal with appropriately.

Initially I thought it would be a good idea to have `hvd.reducescatter(3.14)` do exactly the same thing as `hvd.reducescatter([3.14])`, i.e., return a single-element 1D tensor on rank 0 and an empty 1D tensor on the other ranks. However, having given this some consideration, I feel that it would be confusing and potentially error prone if zero-rank inputs produced higher ranked outputs. It seems less surprising to deal with such a possibility explicitly in user code if necessary (just as it is the case with `hvd.allgather`).

Fixes #3698.


romerojosh (approved):
LGTM @maxhgerlach! I introduced a conflict merging in another PR, but once that is resolved, good to merge.

maxhgerlach:
Thanks @romerojosh

PR #3703, opened 2022-09-19
Support registering local variables for BroadcastGlobalVariablesCallback

MrAta:

## Description
This PR is continuation of #3695, #3643, and #3628  and adds support for local variables for BroadcastGlobalVariablesCallback. So in model parallel use cases, the only the initial value of non-local variables are broadcasted.
An example use case is provided in the unit test included with this PR, but, in short, the callback will be used as follows:

```bash
...
local_variables = <local variables to current rank>
callbacks = [hvd.callbacks.BroadcastGlobalVariablesCallback(root_rank=root_rank, local_variables=local_variables)]
model.fit(..., callbacks=callbacks)
...
```


MrAta:
@romerojosh @maxhgerlach reopening this one again. Do we need support for TF1 in this? currently it's assume TF2 only.

romerojosh:
@MrAta Thanks for posting this PR again! 

> Do we need support for TF1 in this?
I think supporting TF2 only is ok for now. If there is a user request to extend support for TF1, we can add that in a follow up PR. I assume this is because TF1 doesn't have a convenient way to access the model variables (like `self.model.variables` in TF2)? If we do want to limit support to TF2 only, we just need to be sure there is an error raised if someone attempts to use this feature with TF1.

romerojosh (approved):
LGTM @MrAta! Could you rebase on the current master branch to pick up updates? Hopefully this should make all tests pass.

EnricoMi:
All green!

PR #3719, opened 2022-09-27
TensorFlow: scale the gradients of local variables 

MrAta:

## Description

This PR is a followup to #3695 and the discussion made [here](https://github.com/horovod/horovod/discussions/3705) to scale down the gradients of local variables in partial distributed gradient tape, distributed optimizer, and the local gradient aggregators.

Fixes # (issue).
#3705

### DIFF
--- a/horovod/_keras/__init__.py
+++ b/horovod/_keras/__init__.py
@@ -31,7 +31,8 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
                                  compression, sparse_as_dense, gradient_predivide_factor,
                                  op, backward_passes_per_step=1,
                                  average_aggregated_gradients=False,
-                                 groups=None, process_set=hvd.global_process_set):
+                                 groups=None, process_set=hvd.global_process_set,
+                                 scale_local_gradients=True):
     class _DistributedOptimizer(keras.optimizers.Optimizer):
         _HAS_AGGREGATE_GRAD = True
 
@@ -54,6 +55,7 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
 
             self._local_vars = set()
             self.process_set = process_set
+            self.scale_local_gradients = scale_local_gradients
             self._agg_helper = None
             if backward_passes_per_step > 1:
                 if hvd._executing_eagerly():
@@ -62,7 +64,8 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
                         allreduce_func=self._allreduce_grads,
                         sparse_as_dense=sparse_as_dense,
                         average_aggregated_gradients=average_aggregated_gradients,
-                        process_set=process_set
+                        process_set=process_set,
+                        scale_local_gradients=scale_local_gradients
                     )
                 else:
                     self._agg_helper = LocalGradientAggregationHelper(
@@ -72,7 +75,8 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
                         average_aggregated_gradients=average_aggregated_gradients,
                         rank=rank(),
                         optimizer_type=LocalGradientAggregationHelper._OPTIMIZER_TYPE_KERAS,
-                        process_set=process_set
+                        process_set=process_set,
+                        scale_local_gradients=scale_local_gradients
                     )
 
         def register_local_var(self, var):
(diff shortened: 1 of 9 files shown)


MrAta:
cc @skyw @plliao; would appreciate your review here.

test/parallel/test_tensorflow.py
```diff
@@ -4153,14 +4153,14 @@ def test_partial_distributed_gradient_tape(self):
...
+                        self.assertAllClose(grad*hvd.size(), var_grad[var.ref()])
                     else:
                         # non-local gradients shouldn't be equal given that the initial weights are set to ranks
                         self.assertNotAllClose(grad, var_grad[var.ref()])
                 else:
                     if var in local_vars:
                         # local gradients should not change.
-                        self.assertAllClose(grad, var_grad[var])
+                        self.assertAllClose(grad*hvd.size(), var_grad[var])
                     else:
```
plliao, line 4164: Can you do the following instead? Also, can you add comment about why we are doing this?

`
avg_local_grad = var_grad[var.ref()] / hvd.size()
self.assertAllClose(grad, avg_local_grad)
`

horovod/tensorflow/__init__.py
```diff
@@ -622,6 +622,7 @@ def __init__(self, optimizer, name=None, use_locking=False, device_dense='',
                     average_aggregated_gradients=average_aggregated_gradients,
                     rank=rank(),
                     optimizer_type=LocalGradientAggregationHelper._OPTIMIZER_TYPE_LEGACY,
+                    process_set=process_set
                 )
 
         def register_local_var(self, var):
```
plliao, line 628: Can you update the doc? (and all the other register_local_var fn)

plliao (changes requested):

skyw:
Should we make it optional? I like it to be the default behavior because that is the use case I care about.

But still, there are model parallel cases in which local gradient doesn't need to be scaled. Do we need to care about those case right now?

MrAta:
> Should we make it optional? I like it to be the default behavior because that is the use case I care about.
> 
> But still, there are model parallel cases in which local gradient doesn't need to be scaled. Do we need to care about those case right now?

@skyw are there any usecases that strictly require not scaling the local gradients? 
I'm asking because making it optional requires adding a new argument/flag to the current `DistributedOptimizer()` which I'm not sure I should do that or not given there's no plan for major version upgrade right now. @romerojosh what do you think?

romerojosh:
I think adding a new flag with a default to `DistributedOptimizer` is a good idea. I have marked this PR to be included in our v0.26.0 release so it is ok to add an argument. The default can be to scale the local gradients by the number of workers.

horovod/_keras/__init__.py
```diff
@@ -30,7 +31,8 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
                                  compression, sparse_as_dense, gradient_predivide_factor,
                                  op, backward_passes_per_step=1,
                                  average_aggregated_gradients=False,
-                                 groups=None, process_set=hvd.global_process_set):
+                                 groups=None, process_set=hvd.global_process_set,
+                                 scale_local_gradients=True):
```
plliao, line 35: Instead of a bool, can we use floating points?

plliao (changes requested):

horovod/_keras/__init__.py
MrAta: can you elaborate? the floating point will be the scaling value?

plliao: yes

plliao (approved):

romerojosh:
@MrAta It seems the test failures are unrelated to this PR (spark and Elastic tests). Also, the docs are failing to build, but I can't quite tell if there is anything wrong with those changes either.

@EnricoMi @maxhgerlach Any ideas?

MrAta:
Hi @romerojosh, yes seems in spark tests, horovod is not initialized :-?
```bash
/usr/local/lib/python3.8/dist-packages/horovod/tensorflow/keras/__init__.py:136: in DistributedOptimizer
    horovod_size = size_op(process_set_id=process_set.process_set_id) if int(os.environ.get("HOROVOD_ELASTIC", 0)) else process_set.size()
/usr/local/lib/python3.8/dist-packages/horovod/common/process_sets.py:56: in size
    return _basics._process_set_size(self.process_set_id)
...
if result == self.HOROVOD_PROCESS_SET_ERROR_INIT:
>           raise ValueError('Horovod has not been initialized; use hvd.init().')
E           ValueError: Horovod has not been initialized; use hvd.init().
```
I'll take a closer look later today.

EnricoMi:
Follow the [this check](https://github.com/horovod/horovod/runs/8652966664) link in the test results above:

See the "Raw Output" of the first failure:
```
        if self.even_set.included():
            self.assertAlmostEqual(computed_value,
                                   sum(range(0, size, 2)) / self.even_set.size())
        else:
>           self.assertAlmostEqual(computed_value, float(hvd.rank()))
E           TypeError: type numpy.ndarray doesn't define __round__ method

test_tensorflow2_keras_process_sets.py:93: TypeError
```

I think this is the best place to start.

MrAta:
Thanks for the inputs @maxhgerlach @EnricoMi! Seems all the tests pass now.

romerojosh (approved):

PR #3720, opened 2022-09-28
Upgrade ML frameworks in Docker images

### LINKED ISSUE #3689: Update ML frameworks of released Docker images for v0.26.0
Before releasing the next version, versions in `docker/*/Dokerfile` should be updated to latest versions.

EnricoMi:
Fixes #3689.


.github/gen-workflow-ci.py
```diff
@@ -635,14 +635,14 @@ def publish_docker_images(needs: List[str], images: List[str]) -> str:
                 f'            type=semver,pattern={{{{major}}}}\n'
                 f'            type=sha\n'
                 f'\n'
-                f'      - name: Set up QEMU\n'
-                f'        uses: docker/setup-qemu-action@v1\n'
```
maxhgerlach, line 639: Just out of curiosity: Was QEMU ever needed before?

maxhgerlach (approved):
LGTM! 

It's great to have these quick checks that trainings actually run in the generated images.

.github/gen-workflow-ci.py
EnricoMi: It does not look like it ever was.

PR #3727, opened 2022-10-05
Pad fused GPU Allgathers for better memory alignment

maxhgerlach:
# Performance of NCCL Allgather depends on memory alignment

I noticed in profiling that grouped Allgather operations can have vastly differing run times, in particular with FP16, although they add up to similar fused tensor sizes. For instance, an operation sending 6,292,992 bytes would take < 2 msec, but an operation sending 6,294,530 bytes would take > 7 msec. The decisive difference between these two examples is that, in the faster case, the buffer size is divisible by `2**9`, while there is only one factor of 2 in the other buffer size. 

image

Sylvain Jeaugey has pointed out in comments that Allgather and Reducescatter in particular benefit from memory alignment to 16 bytes (`2**4`): https://github.com/NVIDIA/nccl/issues/413#issuecomment-718862629, https://github.com/NVIDIA/nccl/issues/256#issuecomment-534192899

I could also measure performance differences depending on input size alignment with both `all_gather_perf` and `reduce_scatter_perf` from [nccl-tests](https://github.com/NVIDIA/nccl-tests). The effect was less pronounced with Reducescatter than with Allgather, and it was less extreme than the profiling suggested with Horovod. I suspect the reason for the latter point is that I wasn't able to set the input element count to a value that's not a multiple of 4 with nccl-tests (so with FP16 the finest granularity would be 8 bytes rather than 2 bytes). 

# Changes proposed with this PR

For GPU implementations of Allgather this PR changes slightly how tensors are copied into the fusion buffer: The fused Allgather input from each rank is padded to a multiple of 16 bytes. This significantly improves performance in the profiling example shown above: Both Allgathers now take < 2 msec and the aggregate effect significantly improves average training step times.

image

While we could also pad the input of individual (rather than fused) Allgather operations (at the cost of an extra mem copy), I think the problem is more likely to surface for fused tensors: Large tensors will typically be set up to be divisible by many powers of 2, but they may for instance be fused with scalar values.

I've implemented the change for both NCCL and GPU-based MPI. Besides that I've made a pass over some class definitions and mostly went with `clang-tidy` suggestions to reduce warning clutter a bit.

# Alternative padding approach

In a first attempt I added extra padding for each individual tensor incorporated in a fused Allgather. An advantage was that this would also improve memory alignment for `BatchedD2DMemcpy*` before and after the Allgather. However, while this approach also evened out transfer times in profiling, it slowed down otherwise well-aligned Allgathers. In my profiling example both fused Allgathers would take ~2.6 msec. (Changes in branch at https://github.com/horovod/horovod/compare/master...maxhgerlach:horovod:pad-allgather)

### DIFF
--- a/horovod/common/controller.cc
+++ b/horovod/common/controller.cc
@@ -956,7 +956,21 @@ void Controller::FuseResponses(std::deque<Response>& responses,
           // sequence causing breakups in fusion. To counter this some look
           // ahead is allowed.
 
-          skipped_size += new_tensor_size;
+          if (new_response.response_type() ==
+              Response::ResponseType::ALLGATHER) {
+            // Allgather: Not counting padding for skipped_size estimate.
+            assert(new_response.tensor_names().size() == 1);
+            const auto& new_entry =
+                tensor_queue_.GetTensorEntry(new_response.tensor_names()[0]);
+            SetTensorByteSizesForAllgatheredTensors(
+                allgather_new_tensor_byte_sizes, new_response.tensor_sizes(),
+                new_entry);
+            skipped_size +=
+                std::accumulate(allgather_new_tensor_byte_sizes.begin(),
+                                allgather_new_tensor_byte_sizes.end(), 0LL);
+          } else {
+            skipped_size += new_tensor_size;
+          }
           if (tensor_size + skipped_size <= TensorFusionThresholdBytes()) {
             // Skip response and look ahead for more to fuse.
             skipped_responses.push_back(std::move(new_response));
@@ -979,6 +993,7 @@ void Controller::FuseResponses(std::deque<Response>& responses,
 
       int rankwise_padding_bytes = 1;
 #if HAVE_CUDA || HAVE_ROCM
+      // 16 byte pad for efficient allgather
       rankwise_padding_bytes = BATCHED_D2D_PADDING;
 #endif
       // response.tensor_sizes(): Size of first dimension for each rank.
@@ -1026,12 +1041,18 @@ void Controller::FuseResponses(std::deque<Response>& responses,
 
           auto total_byte_size_of_output =
               SumPadded(allgather_tensor_byte_sizes, rankwise_padding_bytes);
-          // The skipped_size estimate is valid even if new_response is not of
-          // type ALLGATHER, but, e.g., ALLREDUCE. Not counting padding here.
-          skipped_size +=
-              std::accumulate(allgather_new_tensor_byte_sizes.begin(),
-                              allgather_new_tensor_byte_sizes.end(), 0LL);
-
+          // Not counting padding for skipped_size estimate.
+          if (new_response.response_type() ==
+              Response::ResponseType::ALLGATHER) {
+            skipped_size +=
+                std::accumulate(allgather_new_tensor_byte_sizes.begin(),
+                                allgather_new_tensor_byte_sizes.end(), 0LL);
+          } else {
+            skipped_size += new_response.tensor_sizes().empty()
+                                ? 0
+                                : new_response.tensor_sizes()[0] *
+                                      GetTypeSize(new_response.tensor_type());
+          }
           if (total_byte_size_of_output + skipped_size <=
               TensorFusionThresholdBytes()) {
             // Skip response and look ahead for more to fuse.
--- a/horovod/common/controller.cc
+++ b/horovod/common/controller.cc
@@ -887,6 +887,12 @@ void AddVectorToVector(std::vector<int64_t>& vec1,
 void Controller::FuseResponses(std::deque<Response>& responses,
                                HorovodGlobalState& state,
                                ResponseList& response_list) {
+  // Avoid re-allocating these vectors inside loop
+  std::vector<int64_t> allgather_tensor_byte_sizes;
+  allgather_tensor_byte_sizes.reserve(GetSize());
+  std::vector<int64_t> allgather_new_tensor_byte_sizes;
+  allgather_new_tensor_byte_sizes.reserve(GetSize());
+
   while (!responses.empty()) {
 
     auto response = responses.front();
@@ -976,8 +982,8 @@ void Controller::FuseResponses(std::deque<Response>& responses,
       rankwise_padding_bytes = BATCHED_D2D_PADDING;
 #endif
       // response.tensor_sizes(): Size of first dimension for each rank.
-      auto tensor_byte_sizes =
-          ByteSizeOfEachAllgatheredTensor(response.tensor_sizes(), entry);
+      SetTensorByteSizesForAllgatheredTensors(allgather_tensor_byte_sizes,
+                                              response.tensor_sizes(), entry);
 
       std::deque<Response> skipped_responses;
       int64_t skipped_size = 0;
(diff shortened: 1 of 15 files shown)


maxhgerlach:
Allgather tests for this PR also pass if Horovod is built with `HOROVOD_GPU_OPERATIONS=MPI` (this is not covered by Horovod CI). I verified that in an NGC container that comes with CUDA-aware MPI. So changes related to `MPI_GPUAllgather` appear to be fine.

     $ docker run --gpus all -it --rm nvcr.io/nvidia/tensorflow:22.02-tf2-py3

     root@bb61561db91a:/workspace# ompi_info --parsable --all | grep mpi_built_with_cuda_support:value
     mca:mpi:base:param:mpi_built_with_cuda_support:value:true

     root@bb61561db91a:/workspace# pip list | grep horovod
     nvidia-horovod                0.23.0+nv22.2
     root@bb61561db91a:/workspace# pip uninstall nvidia-horovod
     # ...

     root@bb61561db91a:/workspace# HOROVOD_GPU_OPERATIONS=MPI HOROVOD_WITHOUT_GLOO=1 HOROVOD_WITHOUT_PYTORCH=1 HOROVOD_WITHOUT_MXNET=1 HOROVOD_WITH_TENSORFLOW=1 HOROVOD_LOCAL_VERSION=pr3727_gpu_mpi pip install -vU git+https://github.com/horovod/horovod.git@refs/pull/3727/head
     # ...
     root@bb61561db91a:/workspace# pip list | grep horovod
     horovod                       0.25.0+pr3727.gpu.mpi

     root@bb61561db91a:/workspace# git clone https://github.com/horovod/horovod.git --recursive
     # ...
     root@bb61561db91a:/workspace# cd horovod/test
     root@bb61561db91a:/workspace/horovod/test# pip install 'mock' 'pytest' 'pytest-forked' 'pytest-subtests' 'parameterized'

     root@bb61561db91a:/workspace/horovod/test# mpirun --allow-run-as-root -np 2 python -m pytest -v parallel/test_tensorflow.py -k allgather
     # ...
     ============================================================= 16 passed, 96 deselected, 1 warning in 81.58s (0:01:21) =====================================================
     # success!

maxhgerlach:
Just for completeness: I've also tried out equivalent padding for Reducescatter, but could not see performance improvements in my experiments (changes in https://github.com/horovod/horovod/compare/master...maxhgerlach:horovod:pad-reducescatter-rankwise).

maxhgerlach:
I've added a commit to improve the bookkeeping in `Controller::FuseResponses()`: https://github.com/horovod/horovod/pull/3727/commits/e351437d44da5e3f8eff7f3af2d9c89ed733c21e

If a user sets up their fusion buffer to precisely hold a certain number of (well-aligned) tensors, there should be no more surprises due to unnecessary padding.

horovod/common/controller.cc
```diff
@@ -961,17 +992,18 @@ void Controller::FuseResponses(std::deque<Response>& responses,
         const auto& new_entry =
             tensor_queue_.GetTensorEntry(new_response.tensor_names()[0]);
 
-        int64_t new_total_byte_size_of_output = TotalByteSizeOfAllgatherOutput(
+        auto new_tensor_byte_sizes = ByteSizeOfEachAllgatheredTensor(
```
romerojosh, line 995: This is creating a new vector every iteration of this loop, which may increase latency. Do you think it'd be worth it to try to avoid it?

horovod/common/ops/gpu_operations.cc
```diff
@@ -471,12 +462,11 @@ void GPUAllgather::MemcpyInFusionBuffer(
   buffer_data = const_cast<void*>(buffer->AccessData(first_entry.context));
   auto& process_set =
       global_state_->process_set_table.Get(first_entry.process_set_id);
+  // offset incorporates rank padding via displcmnts
```
romerojosh, line 465: nit: displcmnts -> displacements

horovod/common/ops/mpi_gpu_operations.cc
```diff
@@ -148,20 +157,36 @@ Status MPI_GPUAllgather::Execute(std::vector<TensorTableEntry>& entries, const R
   }
 
   timeline.ActivityStartAll(entries, ALLOCATE_OUTPUT);
-  Status status = AllocateOutput(entries, response, entry_component_sizes, recvcounts);
+  Status status = AllocateOutput(entries, response, entry_component_sizes);
   if (!status.ok()) {
+    for (size_t ec = 0; ec < entries.size(); ++ec) {
```
romerojosh, line 162: Great catch on this memory leak! Unsure why this file is using `new` and `delete` rather than `std::vector` but guess it's always been that way. No need to change it now since the path is very rarely used anyway.

romerojosh (commented):
Excellent PR @maxhgerlach and very interesting performance results! It's nice to see some improvements to the `allgather` implementation.

Left a few comments, mostly just nits.

horovod/common/ops/mpi_gpu_operations.cc
maxhgerlach: I think all Allgather implementations use these manually allocated buffers (and they had leaks originally, #1845). There's really no reason not to use `std::vector` here as the buffers are zero-initialized anyway.

Refactoring these and all helper functions of `AllgatherOp` to use `std::vector` makes a lot of sense, but I'd suggest to do that in an extra PR.

horovod/common/controller.cc
```diff
@@ -857,6 +856,34 @@ void Controller::CoordinateCacheAndState(CacheCoordinator& cache_coordinator) {
...
+  for (size_t i = 0; i < vec1.size(); ++i) {
+    result += padding * ((vec1[i] + vec2[i] + padding - 1) / padding);
+  }
+  return result;
+}
+
+void AddVectorToVector(std::vector<int64_t>& vec1,
+                       const std::vector<int64_t>& vec2) {
+  assert(vec1.size() == vec2.size());
+  for (std::size_t i = 0; i < vec1.size(); ++i) {
```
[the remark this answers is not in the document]
maxhgerlach: No reason, they are equivalent. Changed this one for consistency in the file.

horovod/common/ops/gpu_operations.cc
maxhgerlach: changed

horovod/common/controller.cc
maxhgerlach: Very good point. Avoiding latency is important here. 

I've changed this in https://github.com/horovod/horovod/pull/3727/commits/df6c4385dab5c5c7784f9e938d20528b14f34ab3 so the memory for those two allgather size vectors is allocated just once at the start of `FuseResponses()`. A slight drawback might be that the control flow in this long function is less local now, making it a bit easier to introduce bugs.

maxhgerlach:
@romerojosh, going over `FuseResponses()` again today, I noticed that the `skipped_size` estimate (when looking ahead) didn't take into account that `response.tensor_sizes()` differs between allgather and allreduce etc. responses: For allreduce it's the total number of elements, while it's only the size of the first dimension for allgather, which can be much smaller of course. It should only matter in edge cases when people interleave allgather with other operations, but nevertheless I've expanded the logic in https://github.com/horovod/horovod/pull/3727/commits/984955f4cbd01c9088613fbb08abf5dbb43e5aa5.

I'm not sure if there's a good reason that this quantity is measured in two different "units", though.

horovod/common/ops/mpi_gpu_operations.cc
romerojosh: Yes, no need to fix it in this PR. I'm not sure if it really needs to be fixed, given that it now works fine (after your memory leak fix).

horovod/common/controller.cc
romerojosh: Thanks for updating this! I think avoiding the allocations is well worth the added complexity of the implementation, given that `FuseResponses` is on the critical path and always called, even in cases where the responses are cached.

romerojosh (approved):
Thanks for addressing the review comments @maxhgerlach! LGTM, though it looks like an unrelated spark test is failing at the moment.

horovod/common/ops/mpi_gpu_operations.cc
maxhgerlach: Agree that there's no urgent need to refactor. It would mostly make future changes safer and easier.

maxhgerlach:
Thanks for the thorough review, @romerojosh! Happy to get this one in.

There's an unrelated problem on torchhead with one of the `SparkLightningTests`, seen also on master: https://github.com/horovod/horovod/actions/runs/3336443880/jobs/5523408422

PR #3737, opened 2022-10-11
Support pinning to local rank GPU index in Spark estimators

thinkall:
Signed-off-by: Li Jiang <bnujli@gmail.com>

## Description

When using spark estimators (torch/lightning/keras) in GPU clusters, `get_available_devices` will be called

https://github.com/horovod/horovod/blob/d29bc04488f00426b5a18946d667ee04b4544956/horovod/spark/task/task_info.py#L25-L28

in `_get_assigned_gpu_or_default`.

https://github.com/horovod/horovod/blob/d29bc04488f00426b5a18946d667ee04b4544956/horovod/spark/common/util.py#L766-L774

And the `get_available_devices` will get the devices by calling `_get_resources`.

https://github.com/horovod/horovod/blob/d29bc04488f00426b5a18946d667ee04b4544956/horovod/spark/task/task_service.py#L98-L107

It works well in single-GPU per node clusters. However, when it comes to clusters with multi-GPUs per node. Wrong GPU index may be returned.

For spark clusters, usually [getGpusResources.sh](https://github.com/apache/spark/blob/master/examples/src/main/scripts/getGpusResources.sh) is used for discovering GPU resources. 

For clusters with multi-GPUs per node, if multi-GPUs are assigned to one executor, then the resources will be like:

image

In this case, `_get_resources` will return {"name": "gpu", "addresses":["0","1"]}, thus `_get_assigned_gpu_or_default` will always return 0, which then be pined to different horovod processes. As a result, only one GPU (always index 0) of each node can be used by horovod spark estimators.

If one GPU is assigned to one executor, then the resources will be like:

image

In this case, still only one GPU (always index 0) of each node can be used by horovod spark estimators.

Only when each node has multi-executors and each executor has been assigned different GPUs, will all the GPUs be available to horovod spark estimators. Like below:

image

However, in this case, GPUs must be configured in `EXCLUSIVE_PROCESS` mode. Unfortunately, for spark clusters with GPUs, spark-rapids is usually needed. Which means, if we configure GPUs in `EXCLUSIVE_PROCESS` mode, we can only use spark-rapids or horovod, but not both.

To solve the issue, and let us use all GPUs in all nodes, and leverage both spark-rapids and horovod seems to be meaningful.

In fact, it's quite easy to handle. Just add an env `USE_DEFAULT_GPU_INDEX` for manually pin to default GPU index (local rank) will do the trick.

```python
import os
def get_available_devices():
    if 'gpu' not in _info.resources or os.getenv('USE_DEFAULT_GPU_INDEX') == '1':
        return []
    return _info.resources['gpu'].addresses
```

With this small change, users can maually pin to default GPU index (local rank) just like they do for horovod runners.

## Review process to land 

### DIFF
--- a/docs/spark.rst
+++ b/docs/spark.rst
@@ -354,8 +354,7 @@ for the spark task from which ``get_available_devices()`` is called.
 See `keras_spark3_rossmann.py <../examples/spark/keras/keras_spark3_rossmann.py>`__ for an example of using
 ``get_available_devices()`` with the Run API.
 
-In some cases, ``get_available_devices()`` may not return the correct list of GPUs if the GPUs resource are not configured correctly
-in the spark cluster. In this case, you can set environment variable ``HOROVOD_USE_DEFAULT_GPU_INDEX`` to ``1`` to use the default
-GPU index (local rank) for each task.
+In some cases, you may want to set the GPU index manually. You can set environment variable ``HOROVOD_USE_DEFAULT_GPU_INDEX`` to
+``1`` to use the default GPU index (local rank) for each task.
 
 .. inclusion-marker-end-do-not-remove
--- a/docs/spark.rst
+++ b/docs/spark.rst
@@ -354,4 +354,8 @@ for the spark task from which ``get_available_devices()`` is called.
 See `keras_spark3_rossmann.py <../examples/spark/keras/keras_spark3_rossmann.py>`__ for an example of using
 ``get_available_devices()`` with the Run API.
 
+In some cases, ``get_available_devices()`` may not return the correct list of GPUs if the GPUs resource are not configured correctly
+in the spark cluster. In this case, you can set environment variable ``HOROVOD_USE_DEFAULT_GPU_INDEX`` to ``1`` to use the default
+GPU index (local rank) for each task.
+
 .. inclusion-marker-end-do-not-remove
--- a/docs/spark.rst
+++ b/docs/spark.rst
@@ -354,7 +354,8 @@ for the spark task from which ``get_available_devices()`` is called.
 See `keras_spark3_rossmann.py <../examples/spark/keras/keras_spark3_rossmann.py>`__ for an example of using
 ``get_available_devices()`` with the Run API.
 
-In some cases, you may want to set the GPU index manually. You can set environment variable ``HOROVOD_USE_DEFAULT_GPU_INDEX`` to
-``1`` to use the default GPU index (local rank) for each task.
+In some cases, you may want to ignore GPU devices assigned by Spark and always use the local rank as the GPU index.
+You can set environment variable ``HOROVOD_SPARK_USE_LOCAL_RANK_GPU_INDEX`` to ``1`` to have Horovod use the local rank
+as the GPU index for each task.
 
 .. inclusion-marker-end-do-not-remove
--- a/horovod/spark/common/util.py
+++ b/horovod/spark/common/util.py
@@ -763,15 +763,16 @@ def to_list(var, length):
     return var
 
 
-def _get_assigned_gpu_or_default(default):
+def _get_assigned_gpu_or_local_rank(local_rank):
     from horovod.spark.task import get_available_devices
     available_devices = get_available_devices()
-    if available_devices:
+    if available_devices and os.getenv('HOROVOD_SPARK_USE_LOCAL_RANK_GPU_INDEX', '0') == '0':
         # if GPU-aware scheduling is available, pin to the assigned GPU index
+        # this is always the first GPU index in available_devices as only one GPU is expected to be assigned
         return int(available_devices[0])
     else:
-        # pin to default GPU index (local rank)
-        return default
+        # pin to local rank GPU index
+        return local_rank
 
 
 def is_databricks():
--- a/horovod/spark/common/util.py
+++ b/horovod/spark/common/util.py
@@ -763,17 +763,12 @@ def to_list(var, length):
     return var
 
 
-def _get_assigned_gpu_or_default(default, hvd=None):
+def _get_assigned_gpu_or_default(default):
     from horovod.spark.task import get_available_devices
     available_devices = get_available_devices()
     if available_devices:
-        if hvd and len(available_devices) >= hvd.local_size():
-            # if GPU-aware scheduling is available, pin to one of the assigned GPUs
-            # pin to GPU based on local rank index
-            return int(available_devices[hvd.local_rank()])
-        else:
-            # pin to first assigned GPU
-            return int(available_devices[0])
+        # if GPU-aware scheduling is available, pin to the assigned GPU index
+        return int(available_devices[0])
     else:
         # pin to default GPU index (local rank)
         return default
--- a/horovod/spark/common/util.py
+++ b/horovod/spark/common/util.py
@@ -763,12 +763,17 @@ def to_list(var, length):
     return var
 
 
-def _get_assigned_gpu_or_default(default):
+def _get_assigned_gpu_or_default(default, hvd=None):
     from horovod.spark.task import get_available_devices
     available_devices = get_available_devices()
     if available_devices:
-        # if GPU-aware scheduling is available, pin to the assigned GPU index
-        return int(available_devices[0])
+        if hvd and len(available_devices) >= hvd.local_size():
+            # if GPU-aware scheduling is available, pin to one of the assigned GPUs
+            # pin to GPU based on local rank index
+            return int(available_devices[hvd.local_rank()])
+        else:
+            # pin to first assigned GPU
+            return int(available_devices[0])
     else:
         # pin to default GPU index (local rank)
         return default
--- a/horovod/spark/common/util.py
+++ b/horovod/spark/common/util.py
@@ -763,20 +763,20 @@ def to_list(var, length):
     return var
 
 
-def _get_assigned_gpu_or_default(local_rank, local_size):
+def _get_assigned_gpu_or_default(default, hvd=None):
     from horovod.spark.task import get_available_devices
     available_devices = get_available_devices()
     if available_devices:
-        if len(available_devices) >= local_size:
+        if hvd and len(available_devices) >= hvd.local_size():
             # if GPU-aware scheduling is available, pin to one of the assigned GPUs
             # pin to GPU based on local rank index
-            return int(available_devices[local_rank])
+            return int(available_devices[hvd.local_rank()])
         else:
             # pin to first assigned GPU
             return int(available_devices[0])
     else:
         # pin to default GPU index (local rank)
-        return local_rank
+        return default
 
 
 def is_databricks():
--- a/horovod/spark/common/util.py
+++ b/horovod/spark/common/util.py
@@ -763,20 +763,20 @@ def to_list(var, length):
     return var
 
 
-def _get_assigned_gpu_or_default(default, hvd=None):
+def _get_assigned_gpu_or_default(local_rank, local_size):
     from horovod.spark.task import get_available_devices
     available_devices = get_available_devices()
     if available_devices:
-        if hvd and len(available_devices) >= hvd.local_size():
+        if len(available_devices) >= local_size:
             # if GPU-aware scheduling is available, pin to one of the assigned GPUs
             # pin to GPU based on local rank index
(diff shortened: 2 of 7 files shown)


horovod/spark/task/task_info.py
```diff
@@ -23,7 +24,7 @@ def __init__(self):
 
 
 def get_available_devices():
-    if 'gpu' not in _info.resources:
+    if 'gpu' not in _info.resources or os.getenv('USE_DEFAULT_GPU_INDEX') == '1':
```
Tixxx, line 27: If this is specific to horovod, i suggest to append it with "HOROVOD_" like "HOROVOD_USE_DEFAULT_GPU_INDEX" to keep it consistent with other horovod envs.

Tixxx (approved):

Tixxx:
in general LGTM, just a minor comment

horovod/spark/task/task_info.py
thinkall: Env name updated. Thanks.

horovod/spark/task/task_info.py
```diff
@@ -23,7 +24,7 @@ def __init__(self):
 
 
 def get_available_devices():
-    if 'gpu' not in _info.resources:
+    if 'gpu' not in _info.resources or os.getenv('HOROVOD_USE_DEFAULT_GPU_INDEX') == '1':
```
EnricoMi, line 27: This env var has to be added to documentation.

EnricoMi (changes requested):

EnricoMi:
I'd say the problem is here:

     if available_devices: 
         # if GPU-aware scheduling is available, pin to the assigned GPU index 
         return int(available_devices[0]) 

This should read

     if available_devices:
        # if GPU-aware scheduling is available, pin to one of the assigned GPUs
        if len(available_devices) >= hvd.local_size():
            # pin to GPU based on local rank index
            return int(available_devices[hvd.local_rank()])
        else:
            # pin to first available GPU
            return int(available_devices[0])

EnricoMi:
With

    --conf spark.executor.resource.gpu.amount=2

your executors get two GPUs assigned.

With

    --conf spark.task.resource.gpu.amount=1

each task running on those multi-GPU executors get a single GPU assigned.

Shouldn't `available_devices` look like `['0']` for one task and `['1']` for the other task on the same executor? Shouldn't `available_devices` always be length equal to `spark.task.resource.gpu.amount`, even if `spark.task.resource.gpu.amount` is larger than one?

thinkall:
> 

This works only if GPUs are configured in EXCLUSIVE_PROCESS mode. Unfortunately, in many clusters, GPUs are too expensive to run only one process.

> With
> 
> ```
> --conf spark.executor.resource.gpu.amount=2
> ```
> 
> your executors get two GPUs assigned.
> 
> With
> 
> ```
> --conf spark.task.resource.gpu.amount=1
> ```
> 
> each task running on those multi-GPU executors get a single GPU assigned.
> 
> Shouldn't `available_devices` look like `['0']` for one task and `['1']` for the other task on the same executor? Shouldn't `available_devices` always be length equal to `spark.task.resource.gpu.amount`, even if `spark.task.resource.gpu.amount` is larger than one?

thinkall:
> I'd say the problem is here:
> 
> ```
>  if available_devices: 
>      # if GPU-aware scheduling is available, pin to the assigned GPU index 
>      return int(available_devices[0]) 
> ```
> 
> This should read
> 
> ```
>  if available_devices:
>     # if GPU-aware scheduling is available, pin to one of the assigned GPUs
>     if len(available_devices) >= hvd.local_size():
>         # pin to GPU based on local rank index
>         return int(available_devices[hvd.local_rank()])
>     else:
>         # pin to first available GPU
>         return int(available_devices[0])
> ```

I think this solves most issues. In the beginning, I was wondering why horovod didn't consider availabel_devices could be  larger than 1. So I suggested the env PR. Now, it looks like combine the two updates, horovod can work better.

Say we have 3 VMs and each have 4 GPUs. There are three kinds of spark settings.

Case 1. Each VM has 1 spark executor, and all the 4 GPUs are assinged to this executor.

Now, we want to init 4 horovod processes on this VM since it has 4 GPUs.

```
if len(available_devices) >= hvd.local_size():
        # pin to GPU based on local rank index
        return int(available_devices[hvd.local_rank()])
```

This works as available_devices will always be like ['0', '1', '2', '3']

Case 2. Each VM has 4 spark executors and each executor has been assigned with 1 GPU.

Now, we still want to init 4 horovod processes on each VM.

```
    else:
        # pin to first available GPU
        return int(available_devices[0])
```

This will work.

Case 3. Each VM has 1 spark executor but only 1 or even 0 GPU is assigned to it. (In this case, configs should be improved from the spark side, but it could happen.)

Now, we still want to init 4 horovod processes on each VM.

```
os.getenv('HOROVOD_USE_DEFAULT_GPU_INDEX') == '1'
```

This will do the trick.

thinkall:
> I'd say the problem is here:
> 
> ```
>  if available_devices: 
>      # if GPU-aware scheduling is available, pin to the assigned GPU index 
>      return int(available_devices[0]) 
> ```
> 
> This should read
> 
> ```
>  if available_devices:
>     # if GPU-aware scheduling is available, pin to one of the assigned GPUs
>     if len(available_devices) >= hvd.local_size():
>         # pin to GPU based on local rank index
>         return int(available_devices[hvd.local_rank()])
>     else:
>         # pin to first available GPU
>         return int(available_devices[0])
> ```


One thing I'd like to mention is that to add `hvd.local_size()` in `_get_assigned_gpu_or_default`, we'll need to update all the related codes as in below figure.

image


And the `_get_assigned_gpu_or_default` will be like below:
```
def _get_assigned_gpu_or_default(default, hvd=None):
    from horovod.spark.task import get_available_devices
    available_devices = get_available_devices()
    if available_devices:
        if hvd and len(available_devices) >= hvd.local_size():
            # if GPU-aware scheduling is available, pin to one of the assigned GPUs
            # pin to GPU based on local rank index
            return int(available_devices[hvd.local_rank()])
        else:
            # pin to first assigned GPU
            return int(available_devices[0])
    else:
        # pin to default GPU index (local rank)
        return default
```

horovod/spark/task/task_info.py
thinkall: Docs and CHANGLOG are updated.

docs/spark.rst
```diff
@@ -354,4 +354,8 @@ for the spark task from which ``get_available_devices()`` is called.
 See `keras_spark3_rossmann.py <../examples/spark/keras/keras_spark3_rossmann.py>`__ for an example of using
 ``get_available_devices()`` with the Run API.
 
+In some cases, ``get_available_devices()`` may not return the correct list of GPUs if the GPUs resource are not configured correctly
+in the spark cluster. In this case, you can set environment variable ``HOROVOD_USE_DEFAULT_GPU_INDEX`` to ``1`` to use the default
```
EnricoMi, line 358: I do not thing this is legit use case. This smells like Horovod is working around mis-configuration of an external system.

If gpu configuration is wrong, make Spark not provide this mis-configuration. Then the current implementation will use local rank GPU index.

horovod/spark/common/util.py
```diff
@@ -763,12 +763,17 @@ def to_list(var, length):
     return var
 
 
-def _get_assigned_gpu_or_default(default):
+def _get_assigned_gpu_or_default(default, hvd=None):
```
EnricoMi, line 766: I think `default` is not (and never was) a good naming here. It should be `local_rank`. With `local_size` as the second argument, `hvd` is not needed and the method becomes more readable. This also means the method has to be renamed.

EnricoMi:
Case 1 and 2 are understood. Case 3 looks like two cases:

Case 3.1: There is one GPU assigned to the task but you want the local rank GPU to be used.
Case 3.2: There is no GPU assigned and you want the local rank GPU to be used.

Case 3.2 is the current behaviour, right?
Case 3.1 looks fishy, see https://github.com/horovod/horovod/pull/3737#discussion_r1026321143.

horovod/spark/common/util.py
thinkall: For the function `_get_assigned_gpu_or_default` itself, `default` is a passed in parameter, and `default` refers to `local_rank` is actually reflected in `spark/keras/remote.py` and other `remote.py` files. If we must change the parameter name here, maybe we'd better change the function name too. Then it will seem to be too much. So, I suggest we keep the `default` for now. If we really want to do the refactor, maybe we should do it in another PR.

docs/spark.rst
thinkall: Thanks for the comment. I agree. Polished the words, just introduce another option for users.

```
In some cases, you may want to set the GPU index manually. You can set environment variable ``HOROVOD_USE_DEFAULT_GPU_INDEX`` to
``1`` to use the default GPU index (local rank) for each task.
```

horovod/spark/common/util.py
thinkall: Thanks for suggestions. Updated code accordingly.

thinkall:
> Case 1 and 2 are understood. Case 3 looks like two cases:
> 
> Case 3.1: There is one GPU assigned to the task but you want the local rank GPU to be used, which could be a completely different GPU (not visible in Spark resources). Case 3.2: There is no GPU assigned and you want the local rank GPU to be used.
> 
> Case 3.2 is the current behaviour, right? Case 3.1 looks fishy, see [#3737 (comment)](https://github.com/horovod/horovod/pull/3737#discussion_r1026321143).

Yes, Case 3.2 is the current behaviour. Case 3.1 could happen too. I agree that should be handled first by the Spark cluster side. A good thing is that our solution for Case 3.2 would also benefit Case 3.1.

EnricoMi:
I am still not convinced that a wrong GPU configuration provided to Horovod on Spark should be worked around by Horovod.

thinkall:
> I am still not convinced that a wrong GPU configuration provided to Horovod on Spark should be worked around by Horovod.

I agree with you, @EnricoMi , Horovod has no responsibility for working around other's issue. In this PR, changes for `_get_assigned_gpu_or_default` are needed for improving Horovod. So I guess your concerns are mainly for adding this `HOROVOD_USE_DEFAULT_GPU_INDEX`. In my view, if we provide this option in Horovod, Horovod will become more robust. It's true that we can say the GPU configuration of a Spark cluster is wrong, but maybe it's not that "wrong" but only not perfect if we consider the fact that GPU supporting in Spark is not as good/easy as CPU supporting. In another case, what if users want to use the local rank? They can do that with horovod runner, but not spark estimator. With this env variable, it will be more flexible. Overall, I believe this one line code change is worth it. If you still insist, I'll remove this env and keep other changes. Thanks :-)

thinkall:
Hi @EnricoMi , may I ask for your decision again? Thanks.

EnricoMi:
@thinkall sorry for the long delay. I have reviewed the changes and made the following changes:
- reverted logic changes in _get_assigned_gpu_or_default
- renamed `HOROVOD_USE_DEFAULT_GPU_INDEX` to `HOROVOD_SPARK_USE_LOCAL_RANK_GPU_INDEX`
- renamed `_get_assigned_gpu_or_default` to `_get_assigned_gpu_or_local_rank`
- moved usage of `HOROVOD_SPARK_USE_LOCAL_RANK_GPU_INDEX` into `_get_assigned_gpu_or_local_rank`
- updated docs

Let me know what you think.

thinkall:
> @thinkall sorry for the long delay. I have reviewed the changes and made the following changes:
> 
> * reverted logic changes in `_get_assigned_gpu_or_default`
> * renamed `HOROVOD_USE_DEFAULT_GPU_INDEX` to `HOROVOD_SPARK_USE_LOCAL_RANK_GPU_INDEX`
> * renamed `_get_assigned_gpu_or_default` to `_get_assigned_gpu_or_local_rank`
> * moved usage of `HOROVOD_SPARK_USE_LOCAL_RANK_GPU_INDEX` into `_get_assigned_gpu_or_local_rank`
> * updated docs
> 
> Let me know what you think.

Hi @EnricoMi , no problem for the delay :-) And your changes sound great to me! Thanks.

thinkall:
Fixed DCO failure and rebased on master.

EnricoMi (approved):

PR #3738, opened 2022-10-12
TensorFlow: add PartialDistributedOptimizer API

MrAta:

## Description
This PR builds upon #3695  and adds a new API called `PartialDistributedOptimizer(optimizer, [local_layers], ...)` for model parallel use cases where unlike `DistributedOptimizer()` ranks only sync a subset of gradients and skip the gradients of their "local layers".


plliao (approved):

test/parallel/test_tensorflow2_keras.py
```diff
@@ -523,7 +523,7 @@ def compute_and_apply_gradients_in_tf_function(var_list, **kwargs):
         aggregation_counter = hvd_optimizer._agg_helper.counter.numpy()
         assert aggregation_counter == total_num_of_steps % backward_passes_per_step
 
-    def test_distributed_optimizer_with_local_vars(self):
+    def test_partial_distributed_optimizer(self):
```
skyw, line 526: Should `test_distributed_optimizer_with_local_vars` be kept? The function it tests is not covered by `test_partial_distributed_optimizer`

MrAta: The test now calls `PartialDistributedOptimizer()` which in turn calls `register_local_var()` function, so it's the same.

maxhgerlach:
To me it feels as if it would make more sense to only add this new API to `horovod.keras` and not to `horovod.tensorflow` as well. Not everybody uses Keras layers with TensorFlow, but everybody uses variables. So I believe it would be nice to add this higher-level API for Keras, but stick with the variables-level of API and documentation for TensorFlow.

MrAta:
> To me it feels as if it would make more sense to only add this new API to `horovod.keras` and not to `horovod.tensorflow` as well. Not everybody uses Keras layers with TensorFlow, but everybody uses variables. So I believe it would be nice to add this higher-level API for Keras, but stick with the variables-level of API and documentation for TensorFlow.

@maxhgerlach that totally makes sense. So it should be only in `horovod/keras/__init__.py`, right?

maxhgerlach:
> @maxhgerlach that totally makes sense. So it should be only in horovod/keras/__init__.py, right?

That sounds good to me, @MrAta. I don't really use Keras personally, though. What do you think about this new API, @romerojosh?

romerojosh:
@MrAta Thanks for contributing this for Keras! I think the API is great (and also agree with @maxhgerlach that this should be put under the `keras` subdirectory.

MrAta:
thank @maxhgerlach @romerojosh, updated the PR accordingly.

horovod/keras/__init__.py
```diff
@@ -108,6 +110,79 @@ def DistributedOptimizer(optimizer, name=None,
         groups=groups,
     )
 
+def PartialDistributedOptimizer(optimizer, name=None,
+                         device_dense='', device_sparse='',
```
maxhgerlach, line 114: Nit: indentation is a bit inconsistent here and in the function body. (One pass of "Reformat Code" in PyCharm over the function should clean it up, for instance)

horovod/keras/__init__.py
```diff
@@ -108,6 +110,79 @@ def DistributedOptimizer(optimizer, name=None,
...
+                         compression=Compression.none,
+                         sparse_as_dense=False,
+                         gradient_predivide_factor=1.0,
+                         op=Average,
+                         num_groups=0,
```
maxhgerlach, line 119: I think we don't need to have the extra `num_groups` argument here that has been deprecated for a while in the original `DistributedOptimizer`.

(Even though nobody got around to actually removing it by 0.23)

maxhgerlach (approved):
Great work, thanks for the update @MrAta!

I've commented just two minor observations that could be addressed optionally.

PR #3742, opened 2022-10-13
Test installation via setup.py

EnricoMi:
Tests that `setup.py` can build the source package and install Horovod from it. This should catch #3741 and #3739 in the future.

Together with #3741, this builds (without compiling anything) and installs horovod:
https://github.com/horovod/horovod/actions/runs/3245078387/jobs/5322133719#step:4:1068


maxhgerlach (approved):
LGTM

PR #3750, opened 2022-10-18
Change linking order to avoid using gloo in pytorch dynamic libraries

Magnus1cheng:
Signed-off-by: cheng.chen <c392410997@gmail.com>

## Description

 {Pytorch_LIBRARIES} contains libtorch_cpu.so, which also includes gloo-related symbols, if  {Pytorch_LIBRARIES} comes before gloo/compatible_gloo, then some of gloo functions (under gloo:rendezvous namespace) would dynamic link to libtorch_cpu.so instead of static link to gloo/compatible_gloo in horovod. This would create incompatible related errors if pytorch and horovod using different gloo versions.
In my case, I use different version of gloo in pytorch and horovod, horovod generates `[1]<stderr>:*** Error in python3: malloc(): memory corruption: 0x000014578c00b6a0 ***`, and I found out error generated because https://github.com/horovod/horovod/blob/master/horovod/common/gloo/gloo_context.cc#L85 links to libtorch_cpu.so, so made error at runtime.
So make ${Pytorch_LIBRARIES} after gloo/compatible_gloo can make sure mpi_lib_v2.cpython-36m-x86_64-linux-gnu.so static link to the gloo/compatible_gloo of third-party in horovod.

## Review process to land 


maxhgerlach (approved):
LGTM and thank you for the contribution @Magnus1cheng! Debugging this problem sounds annoying..

GPU test runs somehow don't appear to be linked (yet) for me, but the pipeline has passed: https://buildkite.com/horovod/horovod/builds/8563

EnricoMi:
Here are the tests, all green (I had to rerun the failed "CI (Results)" workflow).

PR #3756, opened 2022-10-26
Replace set-output commands with env file

EnricoMi:
Setting output values via echo commands is deprecated, using `$GITHUB_OUTPUT` file instead.

Fixes these warnings: https://github.com/horovod/horovod/actions/runs/3325228573/jobs/5498623309#step:2:16

    Warning: The `set-output` command is deprecated and will be disabled soon.
    Please upgrade to using Environment Files. For more information see:
    https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


maxhgerlach (approved):
LGTM

Issue #3483, opened 2022-03-21
Horovod cannot be installed together with ML frameworks with pip>=22.0


EnricoMi:
Since `pip>=22.0`, the following way to install Horovod does not work any more:

    HOROVOD_WITH_TENSORFLOW=1 pip install tensorflow==2.7.0 horovod[tensorflow]

Even though `horovod[tensorflow]` depends on `tensorflow`, pip does not install `tensorflow before attempting to install horovod.

**This probably holds for any ML framework, not only tensorflow.**

With latest PIP:
```
$ python -m pip install --upgrade --force pip
...
Successfully installed pip-22.0.4

$ HOROVOD_WITH_TENSORFLOW=1 pip install tensorflow==2.7.0 horovod[tensorflow]
Collecting tensorflow==2.7.0
  Using cached tensorflow-2.7.0-cp38-cp38-manylinux2010_x86_64.whl (489.6 MB)
Collecting horovod[tensorflow]
  Using cached horovod-0.24.2.tar.gz (3.4 MB)
  Preparing metadata (setup.py) ... done
...
Building wheels for collected packages: horovod
  Building wheel for horovod (setup.py) ... error
  ...
      ModuleNotFoundError: No module named 'tensorflow'
      CMake Error at /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
        Could NOT find Tensorflow (missing: Tensorflow_LIBRARIES) (Required is at
        least version "1.15.0")
  ....
  ERROR: Failed building wheel for horovod
  Running setup.py clean for horovod
Failed to build horovod
Installing collected packages: horovod, tensorflow
  Running setup.py install for horovod ... error
...
```

Looks like an older PIP swaps the order of installing horovod and tensorflow when horovod fails, continues to install tensorflow and attempts horovod again, which then works fine:

```

$ python -m pip install --upgrade --force "pip<22.0"
...
Successfully installed pip-21.3.1

$ HOROVOD_WITH_TENSORFLOW=1 pip install tensorflow==2.7.0 horovod[tensorflow]
Collecting tensorflow==2.7.0
  Using cached tensorflow-2.7.0-cp38-cp38-manylinux2010_x86_64.whl (489.6 MB)
Collecting horovod[tensorflow]
  Using cached horovod-0.24.2.tar.gz (3.4 MB)
  Preparing metadata (setup.py) ... done
...
Building wheels for collected packages: horovod
[...]

PR #3758, opened 2022-10-27
Add pyproject.toml with build system dependencies for PEP-517 installs

### LINKED ISSUE #3697: Horovod fails to install via Poetry
**Environment:**
1. Framework: Tensorflow
2. Framework version: 1.8.1
3. Horovod version: 0.24.2
4. MPI version: -
5. CUDA version: -
6. NCCL version: -
7. Python version: 3.8
8. Spark / PySpark version: -
9. Ray version: -
10. OS and version: Ubuntu 22.04
11. GCC version: -
12. CMake version: - 

**Checklist:**
1. Did you search issues to find if somebody asked this question before? Yes

**Bug report:**
Trying to install horovod via poetry and keep receiving:
```
        ModuleNotFoundError: No module named 'tensorflow'
```

If I install via pip into the virtual env directly this works, however it adds extra load onto the team as they have to remember to do this on top of poetry install. 

Cmake Output:
```
cmake /tmp/pip-req-build-69ukzo_1 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_LIBRARY_OUTPUT_DIRECTORY_RELWITHDEBINFO=/tmp/pip-req-build-69ukzo_1/build/lib.linux-x86_64-3.8 -DPYTHON_EXECUTABLE:FILEPATH=/root/.cache/pypoetry/virtuale
nvs/kheironml-VsnhxLU2-py3.8/bin/python
```

And I can import tensorflow via that direct path and the virtual env.

itsdani:
This should fix missing modules errors when installed with `pip install --use-pep517 horovod`

## Description

Fixes #3697.


itsdani:
I think the build fails because it tries to build the wheel using PEP-517 (because it finds the pyproject.toml) with `HOROVOD_WITH_TENSORFLOW=1`, but the build-requirements in `pyproject.toml` doesn't include `tensorflow` as it's not always needed.

I think the solution is to define a custom wrapper for the setuptools build system ([see this in the setuptools docs](https://setuptools.pypa.io/en/latest/build_meta.html)), which can specify additional dynamic build-dependencies by providing a `get_requires_for_build_wheel` function. This could check which additional packages are needed during the build based on build parameters (I'm not sure if env vars or required extras), and return them.

Unfortunately I was unable to set up the build locally to reproduce the error in CI, and I don't have more time to work on this for a while.

maxhgerlach:
Yes, some dynamic identification of build dependencies would be necessary. 

I did some research in https://github.com/horovod/horovod/issues/3483#issuecomment-1074192411 and follow up comments and then it looked like it wouldn't work well to have differing build requirements depending on extras.

----

If none of the frameworks TensorFlow, PyTorch, or MXNet are available at build time, Horovod is installed in a minimal configuration without its C++ core and cannot be used for distributed training. I suspect that's what you built locally.

`HOROVOD_WITH_TENSORFLOW=1` enforces that TF must be available instead of (silently) skipping support.

itsdani:
Based on this POC the dynamic build-dependencies might work this way, but there are a few problems:

- In `setup.py` there is a `from horovod import __version__` line I had to remove, because this makes the build depend on horovod itself, and the build fails. The baked-in version in my POC is just a hack to test the dependency installs...
- The build tries to install the `tensorflow` dependency from the dynamic list, but it fails for some reason:
```
  Installing backend dependencies: started
  Running command /opt/anaconda3/envs/wmlce/bin/python /tmp/pip-standalone-pip-30bvicml/__env_pip__.zip/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-vxb6p8od/normal --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.org/simple -- tensorflow wheel
  ERROR: Could not find a version that satisfies the requirement tensorflow (from versions: none)
  ERROR: No matching distribution found for tensorflow
  Installing backend dependencies: finished with status 'error'
```

I failed to set up the local build for the project and this makes it hard to test anything. I also feel inadequate to write the correct dynamic dependency list code, because I'm mostly clueless about the project's build process.

maxhgerlach:
Thanks for giving this another shot, @itsdani!

The ppc64le build in the CI (which runs without explicit approval) is a bit particular in how the environment is set up, see https://github.com/horovod/horovod/blob/master/Jenkinsfile.ppc64le 

But I think this highlights that it's actually pretty important to use precisely the tensorflow (and/or pytorch, mxnet) package that's installed in the environment which Horovod is added to: Horovod needs to be compiled accordingly.

joeyearsley:
I've just ran into this issue again, locally I've managed to get it working by changing some of the environment variables.

itsdani:
>I've just rerun into this issue; locally, I've managed to get it working by changing some of the environment variables.
E.g.
HOROVOD_WITH_TENSORFLOW=1 to HOROVOD_WITH_TENSORFLOW=2.8.0

I think you just built it without tensorflow, `HOROVOD_WITH_TENSORFLOW` seems to be a flag that only does something if it's value is `1`.
https://github.com/horovod/horovod/blob/7e4d99386ee4196d767943e30945e45640049ebe/horovod/tensorflow/CMakeLists.txt#L7-L12

joeyearsley:
Locally, I've altered the CMake to ensure it matches a version string.

I don't think extras would work; in my opinion, they are only for Python optional extras intended to say if you want this functionality, install this lib in isolation. Not for defining build extras. 

Since we already require people to say explicitly they want tensorflow or pytorch links, I don't think it's much more to ask them to define the versions instead of a binary flag. 

I'll open my MR soon so you can see the changes.

itsdani:
Ah, I see, extras might be only run-time dependencies and we need build-time deps 😞 

I'm still unsure about this "versions in the env-var" thing, I feel like this still contradicts having an isolated build. Although I have to admit I'm interested in your solution 🤞

joeyearsley:
The PR: #3991 

Having tried many ways, this was the only pragmatic approach I could make work. 

It still fulfils an isolated build but tells the build venv which versions of the necessary DL libraries to install, ensuring we don't get linking issues when the finished build is moved from the isolated build venv into its end site-packages location (i.e. the none isolated poetry venv site-packages).

itsdani:
Ah very nice, I understand now, the custom build backend reports the required dependencies based on the env-var values, and then it can be used in the isolated environment.

PR #3767, opened 2022-11-16
Update usage of TensorFlow Status

ashahba:

- [X] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)? Y
- [] Did you update the docs? N/A
- [ ] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users? N/A

## Description

Fixes # (issue).

## Review process to land 

### DIFF
--- a/horovod/tensorflow/mpi_ops.cc
+++ b/horovod/tensorflow/mpi_ops.cc
@@ -110,7 +110,7 @@ namespace {
 Status ConvertStatus(const common::Status& status) {
   switch (status.type()) {
   case common::OK:
-    return Status::OK();
+    return OkStatus();
   case common::UNKNOWN_ERROR:
     return errors::Unknown(status.reason());
   case common::PRECONDITION_ERROR:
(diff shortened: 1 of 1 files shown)


maxhgerlach:
Thanks for looking into this. Horovod maintains compatibility with TF 1.15+. I think we could either go with `Status()` or if we feel that's not explicit enough, add a preprocessor branch with updated code for TF 2.10+.

Tixxx (changes requested):
Could you use Status() instead of OkStatus()? Status() is backward compatible.

ashahba:
> Thanks for looking into this. Horovod maintains compatibility with TF 1.15+. I think we could either go with `Status()`, or if we feel that's not explicit enough, add a preprocessor branch with updated code for TF 2.10+.

Thanks Max!
I'm open to ideas.

yimeisun123:
When using horovod with master of Tensorflow, it fails as Status::OK() is no longer supported by Tensorflow master.

ashahba:
I'll go with `Status()`

Tixxx (approved):

Tixxx:
will merge after CI passes

PR #3771, opened 2022-11-17
record local version for better debug

KimBioInfoStudio:

## Description

Add a git sha version for better debug

Fixes # (issue).

## Review process to land 

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -209,6 +209,10 @@ if not os.environ.get('HOROVOD_WITHOUT_PYTORCH'):
 def get_package_version():
     return __version__ + "+" + os.environ['HOROVOD_LOCAL_VERSION'] if 'HOROVOD_LOCAL_VERSION' in os.environ else __version__
 
+# record local version in __init__.py
+if 'HOROVOD_LOCAL_VERSION' in os.environ:
+    with open(os.path.join(os.path.dirname(os.path.abspath(__file__)), 'horovod/__init__.py'), 'a') as f:
+        f.write("__local_version__ = '{}'".format(os.environ['HOROVOD_LOCAL_VERSION']))
 
 setup(name='horovod',
       version=get_package_version(),


maxhgerlach:
Hi @KimBioInfoStudio,

would it help you out to install Horovod via something like `HOROVOD_LOCAL_VERSION=$(git rev-parse --short HEAD) pip install ...`?

Then  `horovod.__version__` would be similar to `0.26.1+f05338e` with the git SHA encoded as a local version.

EnricoMi:
I like that solution as it keeps calling `git` in user's control and out of `setup.py`.

KimBioInfoStudio:
Thx @maxhgerlach  @EnricoMi thx for the suggestions, my original plan is to add a private label within horvod itself, without any and methods we could get the local version, I do some modifications on the current pr, do u think this is valuable

EnricoMi:
@maxhgerlach installing with `HOROVOD_LOCAL_VERSION` modifies the version for pip (_cf._ `pip show horovod`) but not in `horovod.__version__`. For that, the `__init__.py`  has to be modified.

Ideally

    HOROVOD_LOCAL_VERSION=$(git rev-parse --short HEAD) python setup.py sdist

creates `./horovod-0.26.1+b7214314.tar.gz`, which has a modified `horovod.__init__.py`, so that

    pip install ./horovod-0.26.1+b7214314.tar.gz

then installs `horovod==0.26.1+b7214314`.

Installing with `pip install .` would have to modify the file in the current directory, which provokes merge conflicts and may leak into git.

EnricoMi:
@KimBioInfoStudio I presume `pip show horovod | grep Version` is not a solution for you as you want the SHA programatically?

maxhgerlach:
Thanks for the correction @EnricoMi! So there's some good motivation for modifying `__init__.py` during (some types of) installation.

EnricoMi:
There is a way to fetch the version from setup.py, see point 5 in https://packaging.python.org/en/latest/guides/single-sourcing-package-version/

The following patch allows the following:

    $ HOROVOD_LOCAL_VERSION=$(git rev-parse --short HEAD) python setup.py sdist
    $ HOROVOD_LOCAL_VERSION=$(git rev-parse --short HEAD) pip install --no-cache-dir -v dist/horovod-0.26.1+b7214314.tar.gz
    $ python -c "import horovod; print(horovod.__version__)"
    0.26.1+b7214314

```patch
diff --git a/horovod/__init__.py b/horovod/__init__.py
index 298461e9..2fea6c32 100644
--- a/horovod/__init__.py
+++ b/horovod/__init__.py
@@ -1,4 +1,16 @@
+import sys
+
 from horovod.runner import run
 
-__version__ = '0.26.1'
+# https://packaging.python.org/en/latest/guides/single-sourcing-package-version/
+if sys.version_info >= (3, 8):
+    from importlib import metadata
+else:
+    import importlib_metadata as metadata
+
+__release_version__ = '0.26.1'
 
+try:
+    __version__ = metadata.version('horovod')
+except Exception:
+    __version__ = __release_version__
diff --git a/setup.py b/setup.py
index e9ef7b1a..e947c681 100644
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ import textwrap
 from setuptools import setup, Extension, find_packages
 from setuptools.command.build_ext import build_ext
 
-from horovod import __version__
+from horovod import __release_version__
 
 _FRAMEWORK_METADATA_FILE = 'horovod/metadata.json'
 
@@ -207,7 +207,7 @@ if not os.environ.get('HOROVOD_WITHOUT_PYTORCH'):
 
 
 def get_package_version():
-    return __version__ + "+" + os.environ['HOROVOD_LOCAL_VERSION'] if 'HOROVOD_LOCAL_VERSION' in os.environ else __version__
+    return __release_version__ + "+" + os.environ['HOROVOD_LOCAL_VERSION'] if 'HOROVOD_LOCAL_VERSION' in os.environ else __release_version__
 
 
 setup(name='horovod',
@@ -234,7 +234,10 @@ setup(name='horovod',
       # which is undesirable.  Luckily, `install` action will install cffi before executing build,
       # so it's only necessary for `build*` or `bdist*` actions.
       setup_requires=require_list if is_build_action() else [],
-      install_requires=require_list,
+      install_requires=require_list + [
+        # https://packaging.python.org/en/latest/guides/single-sourcing-package-version/
+        'importlib-metadata >= 1.0 ; python_version < "3.8"',
+      ],
       tests_require=test_require_list,
       extras_require={
           'all-frameworks': all_frameworks_require_list,
```

Even

    python setup.py sdist

works, though it creates `dist/horovod-0.26.1.tar.gz` rather than `dist/horovod-0.26.1+b7214314.tar.gz`.

PR #3774, opened 2022-11-17
Upgrade Tensorflow and Spark test dependencies

EnricoMi:
Adds the following changes
- Upgrade Tensorflow to 2.10.1, 2.9.3, and 2.8.4
- Add Torch 1.13.0, remove 1.10.2
- Upgrade Spark to 3.2.3 (pending imminent release)


maxhgerlach (approved):
LGTM, thanks for keeping these updated.

I think dependencies for macOS haven't been updated here, but we can always do that separately.

PR #3777, opened 2022-11-17
Skip mxnet v1 tests for v2 pre-release

EnricoMi:
Interestingly, `2.0.0beta1` is smaller than `2.0.0`, so we run v1 tests with v2 beta1.


maxhgerlach (approved):

maxhgerlach:
@EnricoMi, something seems to be off here:

>  CI / Unit Test Results (pull_request) — No tests found 

https://github.com/horovod/horovod/pull/3777/checks?check_run_id=9565893209

PR #3786, opened 2022-11-30
Handle tf.IndexedSlices types when scaling local gradients in TF.

romerojosh:

## Description

This PR fixes the local gradient scaling introduced in #3719 when the gradients are sparse (e.g. of type `tf.IndexedSlices`).

### DIFF
--- a/test/parallel/test_tensorflow.py
+++ b/test/parallel/test_tensorflow.py
@@ -4198,15 +4198,27 @@ class TensorFlowTests(BaseTensorFlowTests):
         rank = hvd.rank()
         local_rank = hvd.local_rank()
 
-        mp_model = DummyMPModel2Devices()
-        optimizer = tf.keras.optimizers.SGD(learning_rate=1)
-        bce = tf.keras.losses.BinaryCrossentropy(from_logits=True)
-        labels = tf.constant(1., shape=[1, 1])
+        if tf.test.is_gpu_available(cuda_only=True):
+            with tf.device("/gpu:%d" % local_rank):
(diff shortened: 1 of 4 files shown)


MrAta:
Thanks for the fix @romerojosh!
Do we also need to apply the same check in `gradient_aggregation_eager.py::_allreduce_helper` and `gradient_aggregation.py::_allreduce_grads_helper` as well?

romerojosh:
This PR is good for review. The currently failing part of the CI is unrelated to this change and due to this issue reported by Max: https://github.com/horovod/horovod/issues/3797

test/parallel/test_tensorflow.py
```diff
@@ -4169,7 +4169,77 @@ def test_partial_distributed_gradient_tape(self):
...
+
+        if tf.test.is_gpu_available(cuda_only=True):
+            with tf.device("/gpu:%d" % local_rank):
+                # "Transpose" input from data parallel to model parallel
+                mp_inputs = hvd.alltoall(dp_inputs)
+                mp_loss = mp_train_step(mp_inputs)
+        else:
+            # "Transpose" input from data parallel to model parallel
+            mp_inputs = hvd.alltoall(dp_inputs)
+            mp_loss = mp_train_step(mp_inputs)
```
maxhgerlach, line 4242: In its current form the test only asserts that the code doesn't crash, but that's better than having no test at all.

maxhgerlach (approved):
LGTM, thanks for the fix, @romerojosh!

I think the test case can be expanded eventually if there's further interest in developing this class of features.

PR #3815, opened 2023-01-13
Add prescale_factor and postscale_factor for Reducescatter

### LINKED ISSUE #3708: Reducescatter: Support `prescale_factor` and `postscale_factor` arguments
They should have the same effect as the arguments to Allreduce and the implementation should be equivalently efficient.

maxhgerlach:

## Description

This adds support for the `prescale_factor` and `postscale_factor` arguments to the Reducescatter op. They are equivalent to the arguments of the same name for the Allreduce op. Averaging is now implemented via postscaling with a factor of `1. / process_set.size()`.

In contrast to Allreduce, we can't do the prescaling on the buffer allocated for the output tensor. This is because with Reducescatter the output tensor is smaller than the input tensor, but all input elements need to be prescaled. To avoid an extra temporary allocation, I decided to use the fusion buffer to do this scaling even for unfused tensors (at least with MPI and NCCL; with Gloo we already use a temporary buffer and I haven't updated that part). Of course, this requires the user to enable a sufficiently large fusion buffer if they want to use Reducescatter with prescaling -- even if they do not actually use tensor fusion. I would be open to discussing alternative approaches if we have concerns about this.

Support implemented for:
- [x] TensorFlow
- [x] PyTorch
- [x] MXNet

Resolves #3708.

## Review process to land 

### DIFF
--- a/test/parallel/test_tensorflow.py
+++ b/test/parallel/test_tensorflow.py
@@ -126,7 +126,7 @@ class TensorFlowTests(BaseTensorFlowTests):
                 minval = -maxval if dtype not in [tf.uint8] else 0
                 tensor = self.random_uniform([17] * dim, minval, maxval)
                 tensor = tf.cast(tensor, dtype=dtype)
-                summed = hvd.allreduce(tensor, average=False)
+                summed = hvd.allreduce(tensor, op=hvd.Sum)
             multiplied = tensor * size
             difference = summed - multiplied
             difference = tf.cast(difference, tf.int32) if dtype == tf.uint8 else difference
@@ -158,7 +158,7 @@ class TensorFlowTests(BaseTensorFlowTests):
                 minval = -maxval if dtype not in [tf.uint8] else 0
                 tensor = self.random_uniform([17] * dim, minval, maxval)
                 tensor = tf.cast(tensor, dtype=dtype)
-                averaged = hvd.allreduce(tensor, average=True)
+                averaged = hvd.allreduce(tensor, op=hvd.Average)
             # handle int8, uint8 overflows when allreduce sums up and averages the values
             tensor = tf.cast((tensor*size)/size, dtype=dtype)
             difference = tf.cast(averaged, dtype=dtype) - tensor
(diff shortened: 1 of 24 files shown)


test/parallel/test_tensorflow.py
```diff
@@ -126,7 +126,7 @@ def test_horovod_allreduce_cpu(self):
                 minval = -maxval if dtype not in [tf.uint8] else 0
                 tensor = self.random_uniform([17] * dim, minval, maxval)
                 tensor = tf.cast(tensor, dtype=dtype)
-                summed = hvd.allreduce(tensor, average=False)
+                summed = hvd.allreduce(tensor, op=hvd.Sum)
```
romerojosh, line 129: Thanks for updating these!

romerojosh (approved):
LGTM! 👍 

I think using the fusion buffer to avoid the extra allocation is a good choice, and only a minor inconvenience if it crops up.

maxhgerlach:
Thank you for the review and weighing in on the fusion buffer abuse, @romerojosh!

I've also verified locally that Reducescatter-related TF tests pass with GPU MPI.
```
  $ docker run --gpus all -it --rm nvcr.io/nvidia/tensorflow:22.02-tf2-py3
  root@74b91a3bae34:/workspace# ompi_info --parsable --all | grep mpi_built_with_cuda_support:value
  mca:mpi:base:param:mpi_built_with_cuda_support:value:true
  root@74b91a3bae34:/workspace# pip list | grep horovod
  nvidia-horovod                0.23.0+nv22.2
  root@74b91a3bae34:/workspace# pip uninstall  -y nvidia-horovod
  root@74b91a3bae34:/workspace# HOROVOD_GPU_OPERATIONS=MPI HOROVOD_WITHOUT_GLOO=1 HOROVOD_WITHOUT_PYTORCH=1 HOROVOD_WITHOUT_MXNET=1 HOROVOD_WITH_TENSORFLOW=1 HOROVOD_LOCAL_VERSION=pr3815_gpu_mpi pip install -vU git+https://github.com/horovod/horovod.git@refs/pull/3815/head
  root@74b91a3bae34:/workspace# pip list | grep horovod
  horovod                       0.26.1+pr3815.gpu.mpi
  root@74b91a3bae34:/workspace# git clone https://github.com/horovod/horovod.git
  root@74b91a3bae34:/workspace# cd horovod/
  root@74b91a3bae34:/workspace/horovod# git fetch origin pull/3815/head:pr3815
  root@74b91a3bae34:/workspace/horovod# git checkout pr3815
  root@74b91a3bae34:/workspace/horovod# cd test/parallel/
  root@74b91a3bae34:/workspace/horovod/test/parallel# pip install 'mock' 'pytest' 'pytest-forked' 'pytest-subtests' 'parameterized'
  root@74b91a3bae34:/workspace/horovod/test/parallel# mpirun --allow-run-as-root -np 2 python -m pytest -v test_tensorflow.py -k scatter
  # ...
  ================================================================= 25 passed, 120 deselected, 1 warning in 26.15s ==================================================================
```

There was a minimal changelog-related merge conflict. I think in this case it will be fine to merge to master without waiting for the full pre-merge CI pipeline to pass again.

Screenshot 2023-01-31 at 11 11 29

Issue #3812, opened 2023-01-11
Which version of FlatBuffers is used to generate head file?


GHGmc2:
I built FlatBuffers from source with the commit in third_party: https://github.com/google/flatbuffers/tree/a9a295fecf3fbd5a4f571f53b01f63202a3e2113
```bash
cmake -G "Unix Makefiles"
make
make install
```

And then generated the head file from `fbs`:
```bash
flatc -c -o ~/ ./horovod/common/wire/message.fbs
```

And I found the the generated `message_generated.h` file is not the same as the one in Horovod repo: https://github.com/horovod/horovod/blob/master/horovod/common/wire/message_generated.h

May I know which version of FlatBuffers is used to get the head file in Horovod repo?

maxhgerlach:
What's the diff between your generated file and the version that's checked in?

I believe I've used flatbuffers 2.0 (that's the commit you posted https://github.com/google/flatbuffers/tree/a9a295fecf3fbd5a4f571f53b01f63202a3e2113) when I last updated the file, but the copyright header that's at the top of it is maintained manually.

Probably it would be good to have some build process with clearly versioned dependencies generate `message_generated.h`...

GHGmc2:
The file I got manually: 
[message_generated.h.manual.txt](https://github.com/horovod/horovod/files/10407608/message_generated.h.manual.txt)

Please help to confirm.

maxhgerlach:
Interesting! So the first part of the diff concerns the copyright / license comment header that we maintain manually. 

The second part showcases an actual discrepancy. But I don't think it's related to the version of the FlatBuffers compiler. I think it's more likely that the enum was renamed from `PROD` to `PRODUCT` in `message.fbs`, but then accidentally the header file wasn't regenerated.

As all tests pass, this doesn't appear to cause any actual problems now. But for the sake of consistency we should probably update `message_generated.h` so the enum names match those in: https://github.com/horovod/horovod/blob/bfe96f8cc239be5fbb75836db8682e040021d858/horovod/common/message.h#L43-L50

GHGmc2:
@maxhgerlach Thanks.

Can you help to confirm the other file: https://github.com/horovod/horovod/blob/master/horovod/tensorflow/custom_call_config.fbs ? I got more discrepancies with this file.

maxhgerlach:
[...]

PR #3816, opened 2023-01-13
Update message_generated.h to match flatbuffers schema

maxhgerlach:

## Description

Here I've just run `flatc` on `horovod/common/wire/message_generated.fbs`. The generated header file was out of sync regarding one of the `ReduceOp` enum names (as pointed out in #3812). 

## Review process to land 


romerojosh (approved):
Oof, nice catch. That seems to have been due to a last minute commit from my PR that added the new ops where I changed the op name from `Prod` to `Product`. Seems the flatbuffers implementation doesn't care about this, since the ordering (and index assignment) is consistent. Good to make consistent anyway!

EnricoMi (approved):
LGTM!

PR #3824, opened 2023-01-25
Reducescatter: Allocate output tensors before enqueuing the operation

maxhgerlach:

## Description

This change is motivated by the observations posted to discussion #3819: We can compute the shapes of the output tensors of (grouped) Reducescatter operations early, without any inter-process communication, even though the scattered tensors may not be distributed evenly. (Allgather behaves differently in this regard: There, each rank receives exactly the same output tensor, but Horovod cannot determine its size without communication.) We can take advantage of this fact and allocate memory for output tensors early, before the Horovod operation is enqueued, and without blocking the Horovod background thread.

This still happens in the C++ backend, but in framework-specific code. Allocations should be asynchronous with TensorFlow and PyTorch, but as far as I can tell the respective call is blocking with MXNet. In general I'm not closely familiar with the MXNet section of the code base. Probably it would be possible to be a bit smarter if `MXNET_ASYNC_GPU_ENGINE_SUPPORTED` is defined?


romerojosh (approved):
Glad you and @nvcastet were able to identify the performance issue. Changes LGTM @maxhgerlach!

Issue #3435, opened 2022-03-01
Ray tests are flaky


EnricoMi:
Ray tests have shown to be flaky, especially with GPU (Buildkite CI).

There are two places that cause these flakiness:
1. Some tests fetch `ray.available_resources()` at the beginning of the test and compare the `dict` against `ray.available_resources()` after the test: `assert check_resources(original_resources)`. It looks like that `dict` has race conditions: https://buildkite.com/horovod/horovod/builds/7306#cc0600a5-13ed-479d-ba1b-3bb0f6847992/332-436
```
       AssertionError: assert {'CPU': 4.0, ...147712.0, ...} == {'object_stor... 9999999984.0}
         Differing items:
         {'object_store_memory': 10000000000.0} != {'object_store_memory': 9999999984.0}
         Left contains 5 more items:
         {'CPU': 4.0,
          'GPU': 4.0,
          'accelerator_type:T4': 1.0,
          'memory': 189132147712.0,...
```
2. Some tests see more GPUs than there should be:
```
       assert len(all_envs[0]["CUDA_VISIBLE_DEVICES"].split(",")) == 4
       assert 8 == 4
         +8
         -4
```
First, the tests should work with any number of GPUs (larger than the minimum required GPUs of 4). Second, the test run on a machine with only 4 GPUs on Buildkite, with only one agent per machine. The init script would print a warning if it could not setup `CUDA_VISIBLE_DEVICES` properly, but there is no output: `/etc/buildkite-agent/hooks/environment:`
```bash
if [[ "${BUILDKITE_AGENT_NAME:-}" == *"-1" ]]; then
  echo "CUDA_VISIBLE_DEVICES=0,1,2,3"
  export CUDA_VISIBLE_DEVICES=0,1,2,3
else
  echo "WARN: unable to detect CUDA_VISIBLE_DEVICES based on BUILDKITE_AGENT_NAME: ${BUILDKITE_AGENT_NAME:-}"
fi
```

https://buildkite.com/horovod/horovod/builds/7296#7a1bf515-2ba2-4d43-908b-154dd3b76407/1-2
```
Running global environment hook
$ /etc/buildkite-agent/hooks/environment
```

EnricoMi:
@richardliaw @tgaddair the `check_resources` method seems to be very flaky (see 1. above), I am going to remove it entirely:
[...]

PR #3828, opened 2023-01-27
Reduce flakiness of Ray GPU tests that check CUDA_VISIBLE_DEVICES

maxhgerlach:
On Buildkite these tests fail quite often with messages like

```
=================================== FAILURES ===================================
___________________________ test_gpu_ids_num_workers ___________________________
ray_start_4_cpus_4_gpus = {'metrics_export_port': 57818, 'node_id': 'e3703e1d4e6396f6233c7bb316d604fc15ca225757ff96036253ef6b', 'node_ip_address': '172.16.32.2', 'object_store_address': '/tmp/ray/session_2023-01-27_12-44-34_471603_3376/sockets/plasma_store', ...}
 
    @pytest.mark.skipif(
        torch.cuda.device_count() < 4, reason="GPU test requires 4 GPUs")
    @pytest.mark.skipif(
        not torch.cuda.is_available(), reason="GPU test requires CUDA.")
    def test_gpu_ids_num_workers(ray_start_4_cpus_4_gpus):
        setting = RayExecutor.create_settings(timeout_s=30)
        hjob = RayExecutor(setting, num_workers=4, use_gpu=True)
        hjob.start()
        all_envs = hjob.execute(lambda _: os.environ.copy())
        all_cudas = {ev["CUDA_VISIBLE_DEVICES"] for ev in all_envs}
 
        assert len(all_cudas) == 1, all_cudas
>       assert len(all_envs[0]["CUDA_VISIBLE_DEVICES"].split(",")) == 4, all_envs[0]["CUDA_VISIBLE_DEVICES"]
E       AssertionError: 0,3,1,2,1,2,0,3
E       assert 8 == 4
E        +  where 8 = len(['0', '3', '1', '2', '1', '2', ...])
E        +    where ['0', '3', '1', '2', '1', '2', ...] = <built-in method split of str object at 0x7fb0095ba170>(',')
E        +      where <built-in method split of str object at 0x7fb0095ba170> = '0,3,1,2,1,2,0,3'.split
```
Taken from https://buildkite.com/horovod/horovod/builds/8886, where it failed 3x in a row on one instance.

So `CUDA_VISIBLE_DEVICES` contains four distinct GPU IDs `0,1,2,3`, however, each of them appears twice and they are not ordered. That shouldn't matter, though, as long as they are all valid GPU IDs accessible by the running job. So let's make the tests more permissive.

Not sure if the unusual value of that env variable is caused by Ray or Buildkite.

### DIFF
--- a/test/single/test_ray.py
+++ b/test/single/test_ray.py
@@ -175,7 +175,7 @@ def test_gpu_ids(ray_start_4_cpus_4_gpus):
     all_envs = hjob.execute(lambda _: os.environ.copy())
     all_cudas = {ev["CUDA_VISIBLE_DEVICES"] for ev in all_envs}
     assert len(all_cudas) == 1, all_cudas
-    assert len(all_envs[0]["CUDA_VISIBLE_DEVICES"].split(",")) == 4, all_envs[0]["CUDA_VISIBLE_DEVICES"]
+    assert len(set(all_envs[0]["CUDA_VISIBLE_DEVICES"].split(","))) == 4, all_envs[0]["CUDA_VISIBLE_DEVICES"]
     hjob.shutdown()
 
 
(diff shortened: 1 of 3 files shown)


maxhgerlach:
@EnricoMi, I had forgotten about https://github.com/horovod/horovod/issues/3435, where you brought this problem to attention. Do you think it may be indicative of a deeper issue? 

It's a bit disruptive if long-running flaky CI workflows need to be re-run often..

maxhgerlach:
Ah, never mind, CUDA actually doesn't support repeated device IDs.

```
$ CUDA_VISIBLE_DEVICES=0,0,1 python -c 'import tensorflow as tf; print(tf.config.list_physical_devices())'
2023-01-27 14:34:54.351339: E tensorflow/stream_executor/cuda/cuda_driver.cc:271] failed call to cuInit: CUDA_ERROR_INVALID_DEVICE: invalid device ordinal
2023-01-27 14:34:54.351400: I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:169] retrieving CUDA diagnostic information for host: 
# ...
[PhysicalDevice(name='/physical_device:CPU:0', device_type='CPU')]
```

So this should be fixed.

PR #3822, opened 2023-01-23
Fix broken TF DistributedOptimizer with Keras 2.11+

nvcastet:
Signed-off-by: Nicolas Castet <ncastet@nvidia.com>

## Checklist before submitting

- [x] Did you read the [contributor guide](https://github.com/horovod/horovod/blob/master/CONTRIBUTING.md)?
- [x] Did you update the docs?
- [ ] Did you write any tests to validate this change?  
- [x] Did you update the [CHANGELOG](https://github.com/horovod/horovod/blob/master/CHANGELOG.md), if this change affects users?

## Description

Horovod was only checking for `tf.keras.optimizers.Optimizer` in  `DistributedOptimizer` but supported optimizers are under `tf.keras.optimizers.legacy.Optimizer` for Keras 2.11+.

PR #3860, opened 2023-03-10
Re-enabling new keras optimizers

nvcastet:
## Description

Re-enabling TF Keras 2.11+ optimizers. ~Those and the legacy ones should work out the box with Horovod.~
Using new optimizers with `LossScaleOptimizer` is broken in 2.11 but fixed in 2.12.
See discussions at https://github.com/keras-team/keras/issues/17275
 and commit at https://github.com/keras-team/keras/commit/c98ea36b96f6036b1dec569e5e496aee06a44b29
 
 EDIT:
 They are quite a few changes in API between the 2 optimizer versions. But hopefully this PR properly integrates Horovod to the new ones while maintaining support for the old ones.

PR #3862, opened 2023-03-13
fix keras optimizer initialization for example script

tejaschumbalkar:

## Description

Running `tensorflow2_keras_synthetic_benchmark.py` script results in the below error
```
ValueError: Optimizer has to be an instance of tensorflow.keras.optimizers.legacy.Optimizer starting from Keras 2.11: SGD
```

Check was added as a part of #3822. This PR fixes the import for `optimizers` w.r.t. TF versions

Fixes # (issue).

## Review process to land 

### DIFF
--- a/examples/tensorflow2/tensorflow2_keras_synthetic_benchmark.py
+++ b/examples/tensorflow2/tensorflow2_keras_synthetic_benchmark.py
@@ -21,6 +21,12 @@ import tensorflow as tf
 import horovod.tensorflow.keras as hvd
 from tensorflow.keras import applications
 
+from packaging import version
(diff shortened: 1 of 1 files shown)


MrAta (commented):
@tejaschumbalkar this shouldn't be needed once #3860 is merged as it was introduced by the dual that pull.

maxhgerlach:
Thanks for spotting the inconsistency in this example! I agree that ideally we should rather fix the Keras 2.11 support via #3860. However, in case we don't complete that in time for a new release, it will be a good idea to at least include this PR.

maxhgerlach:
Closing since #3860 has landed on master.

PR #3865, opened 2023-03-16
Don't pass CUDA_VISIBLE_DEVICES

rb-determined-ai:
There is never a valid reason to pass CUDA_VISIBLE_DEVICES.  It may be configured with UUIDs, which aren't suitable for passing between nodes, or it may be configured with indicies, which also might be different between nodes.

## Review process to land 


rb-determined-ai:
The counter-argument to this PR might be: if `CUDA_VISIBLE_DEVICES` is needed by worker nodes, then running worker nodes without it is also not an option.

maxhgerlach (approved):
I think the change makes sense. 

Setting `CUDA_VISIBLE_DEVICES` for the Horovod processes can be a good idea if they are started on the local node, but AFAIK at least Open MPI `mpirun` will expose the existing environment to locally started processes. I think the exclusion rules would only affect remote processes.

PR #3874, opened 2023-03-21
Use Python 3.8 to build docs as well

ashahba:
Signed-off-by: Abolfazl Shahbazi <abolfazl.shahbazi@intel.com>
This is just to use `Python 3.8` everywhere since all GitHub actions are on that version also `3.7` end of life is approaching fast: https://endoflife.date/python

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -28,6 +28,6 @@ formats: []
 
 # Requirements for building docs
 python:
-  version: 3.7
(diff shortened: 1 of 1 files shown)


maxhgerlach (approved):
Thanks, it makes sense to update this. Doc build appears to work fine with Python 3.8.

EnricoMi (approved):

PR #3885, opened 2023-04-10
Force tf.logical_and in hvd allreduce condition running on CPU

plliao:

## Description

Forces tf.logical_and and related ops during hvd allreduce condition check running on CPU.

Why?

tf.logical_and will be scheduled on GPU by default. However, other hvd control ops such as process_set_included_op or size_op and TF cond ops only run on CPU. It will end up doing a memory copy from host to device to compute the tf.logical_and and doing another device to host copy to compute tf.cond. The PR force all condition related ops running on CPU to avoid unnecessary host to device and device to host memory copy.

We see 10% performance improvement when memory copy resource is under contention in our internal workflow . 

## Review process to land 


maxhgerlach (approved):
Great find! Thanks for improving code readability as well, @plliao.

PR #3894, opened 2023-04-19
Further fixes and updates for tf-head

maxhgerlach:
## Description

Let's collect more fixes for TensorFlow nightly builds here.

### DIFF
--- a/horovod/_keras/__init__.py
+++ b/horovod/_keras/__init__.py
@@ -84,6 +84,9 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
                 return super(self.__class__, self).variables()
             return self.get_weights()
 
+        if version.parse(tf.__version__) >= version.parse('2.12.0'):
+            variables = property(variables)
+
         def register_local_var(self, var):
             """Registers a source/variable as worker local. Horovod will not perform any global
             operations on gradients corresponding to these sources and will instead return the local
--- a/horovod/spark/keras/tensorflow.py
+++ b/horovod/spark/keras/tensorflow.py
@@ -31,7 +31,7 @@ from tensorflow.python.util import serialization
 
 
 def save_tf_keras_optimizer(optimizer, h5py_file):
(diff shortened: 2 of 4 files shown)


maxhgerlach:
There's a couple of opimizer-related errors in `test_spark_keras.py` with tf-nightly:

- `test_fit_model`
- `test_fit_model_multiclass`
- `test_keras_direct_parquet_train`
- `test_keras_model_checkpoint_callback`
- `test_model_serialization`
- `test_restore_from_checkpoint`

(see collected tracebacks in https://github.com/horovod/horovod/runs/12877848393)

I don't have a good understanding about how the optimizers are wrapped there. @chongxiaoc, do you maybe have an idea how to fix these?

maxhgerlach:
`DistributedOptimizer.variables` is expected to be a property from Keras 2.12, which is behind some of the Spark Keras failures. Let's see how my attempted fix fares.

maxhgerlach:
All TensorFlow/Keras tests pass now.

There's a failure in Spark Lightning MNIST on the CPU/GPU head pipelines. It should be investigated separately.

horovod/_keras/__init__.py
```diff
@@ -84,6 +84,9 @@ def variables(self):
                 return super(self.__class__, self).variables()
             return self.get_weights()
 
+        if version.parse(tf.__version__) >= version.parse('2.12.0'):
```
nvcastet, line 87: Thanks @maxhgerlach I could figure out this one.
So TF changed the attribute from method to property? :)

nvcastet (approved):

horovod/_keras/__init__.py
maxhgerlach: They did, yes: https://github.com/keras-team/keras/commit/50e0b18ceeb370157f36a387115c5ae1a16fca47

There's a little compatibility shim so code using an optimizer still works with both `.variables` and with `.variables()`.

Cheers!

Issue #3841, opened 2023-02-07
CI: Spark tests crash on GPU


maxhgerlach:
- GPU, Spark Lightning MNIST (test-gpu-gloo-py3_8-tf2_9_3-keras2_9_0-torch1_11_0-mxnet1_7_0_p1-pyspark3_3_1), https://buildkite.com/horovod/horovod/builds/8954#01862c7a-7e6a-44eb-8920-dd338006a42f
```
Tue Feb  7 15:26:02 2023[1]<stderr>:RuntimeError: CUDA error: invalid device ordinal
# ...
Tue Feb  7 15:26:15 2023[0]<stderr>:Segmentation fault (core dumped)
```
- GPU heads, Spark Torch MNIST (test-gpu-openmpi-gloo-py3_8-tfhead-keras_none-torchhead-mxnethead-pyspark3_3_1), https://buildkite.com/horovod/horovod/builds/8955#01862c83-2e49-48d0-afaf-36bd47b2a64b
```
Tue Feb  7 15:35:37 2023[1,1]<stderr>:RuntimeError: CUDA error: invalid device ordinal
```
- GPU heads, Spark Lightning MNIST (test-gpu-openmpi-gloo-py3_8-tfhead-keras_none-torchhead-mxnethead-pyspark3_3_1), https://buildkite.com/horovod/horovod/builds/8955#01862c82-eff4-4699-b37d-459f22d7fa2d
```
Tue Feb  7 15:35:05 2023[1,0]<stderr>:malloc(): invalid next->prev_inuse (unsorted)
Tue Feb  7 15:35:05 2023[1,1]<stderr>:malloc(): invalid next->prev_inuse (unsorted)
```

chongxiaoc:
This is the issue: `Tue Feb  7 15:26:13 2023[0]<stderr>:LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [2,3]`

irasit:
May need to use some thing similar to keras to get the visible gpu device. 
https://sourcegraph.com/github.com/horovod/horovod@master/-/blob/horovod/spark/keras/remote.py?L323&subtree=true

irasit:
```
Tue Feb  7 15:26:13 2023[0]<stderr>:  table = piece.read(columns=column_names - partition_names, partitions=self._dataset.partitions)
--
[...]

PR #3906, opened 2023-04-28
Fix build with tf-nightly

maxhgerlach:

## Description

TF 2.13 moves to an API returning status messages as `string_view`. This PR fixes the build accordingly  by introducing an extra string copy. We can't introduce `std::string_view` more widely without moving more code to C++-17.

## Review process to land 

### DIFF
--- a/horovod/common/common.cc
+++ b/horovod/common/common.cc
@@ -28,27 +28,27 @@ namespace common {
 
 Status::Status() = default;
 
-Status::Status(StatusType type, std::string reason)
-    : type_(type), reason_(std::move(reason)) {
+Status::Status(StatusType type, std::string_view reason)
+    : type_(type), reason_(reason) {
 }
 
 Status Status::OK() {
   return Status();
 }
 
-Status Status::UnknownError(const std::string& message) {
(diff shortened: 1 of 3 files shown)


maxhgerlach:
Build is fine now apparently, but there are errors with TensorFlow tests on tf-head.

```
[1]<stdout>:self = <tensorflow.core.function.capture.capture_container.FunctionCaptures object at 0x7fca30515af0>
[1]<stdout>:
[1]<stdout>:    @property
[1]<stdout>:    def capture_types(self):
[1]<stdout>:>     return self._by_val_tracetype | self._by_ref_tracetype
[1]<stdout>:E     TypeError: unsupported operand type(s) for |: 'MutationAwareDict' and 'collections.OrderedDict'
```

We could still merge this PR now to at least unblock torch and mxnet tests with nightly builds.

maxhgerlach:
Non-head GPU failure looks like known problem #3841

nvcastet (approved):

EnricoMi (commented):
LGTM!

EnricoMi (approved):
LGTM!

PR #3912, opened 2023-05-05
Update with_device functions in MXNet and PyTorch to skip unnecessary cudaSetDevice calls

romerojosh:

## Description

In CUDA 12.x, calls to `cudaSetDevice` create GPU contexts. It was found that the `with_device` objects used in the MXNet and PyTorch code can execute `cudaSetDevice(0)` in the destructor if called before the Horovod background thread sets a device (e.g. before launching any GPU collectives). With the new CUDA 12.x behavior, this will generate spurious contexts for all ranks on GPU 0, unnecessarily consuming GPU memory.

This PR updates the behavior of `with_device` to set the `restore_device_` value to the `device` value supplied to the constructor if no current GPU context is associated with the thread.

### DIFF
--- a/horovod/mxnet/cuda_util.cc
+++ b/horovod/mxnet/cuda_util.cc
@@ -13,9 +13,11 @@
 // limitations under the License.
 // =============================================================================
 
+#include <dlfcn.h>
 #include <stdexcept>
 
 #if HAVE_CUDA
+#include "cuda.h"
 #include "cuda_runtime.h"
 #include <mxnet/base.h>
 #endif
@@ -27,13 +29,35 @@
 namespace horovod {
 namespace mxnet {
 
+#if HAVE_CUDA
+typedef CUresult (CUDAAPI *PFN_cuCtxGetDevice)(CUdevice* device);
+static void* cudalib = nullptr;
+static PFN_cuCtxGetDevice pfn_cuCtxGetDevice = nullptr;
+
+static void initialize_driver_api() {
+  // Clear previous errors
+  (void) dlerror();
+
+  cudalib = dlopen("libcuda.so", RTLD_LAZY);
+  if (!cudalib) {
+    throw std::logic_error("Internal error. Could not dlopen libcuda.so.");
+  }
+
+  pfn_cuCtxGetDevice = (PFN_cuCtxGetDevice) dlsym(cudalib, "cuCtxGetDevice");
+  if (!pfn_cuCtxGetDevice) {
+    throw std::logic_error("Internal error. Could not load cuCtxGetDevice.");
+  }
+}
+#endif
+
 with_device::with_device(int device) {
   if (device == CPU_DEVICE_ID) {
     restore_device_ = CPU_DEVICE_ID;
   } else {
 #if HAVE_CUDA
+    if (!cudalib) initialize_driver_api();
     CUdevice cudev;
-    auto err = cuCtxGetDevice(&cudev);
+    auto err = pfn_cuCtxGetDevice(&cudev);
     if (err == CUDA_ERROR_NOT_INITIALIZED ||
         err == CUDA_ERROR_INVALID_CONTEXT) {
        // If device has never been set on this thread,
@@ -42,7 +66,8 @@ with_device::with_device(int device) {
      } else if (err == CUDA_SUCCESS) {
        restore_device_ = static_cast<int>(cudev);
      } else {
-       HVD_GPU_DRIVER_CHECK(err);
+       throw std::logic_error("Internal error. cuCtxGetDevice returned error code " +
+                              std::to_string(err));
      }
      CUDA_CALL(cudaSetDevice(device));
 #else
--- a/horovod/mxnet/cuda_util.cc
+++ b/horovod/mxnet/cuda_util.cc
@@ -32,8 +32,19 @@ with_device::with_device(int device) {
     restore_device_ = CPU_DEVICE_ID;
   } else {
 #if HAVE_CUDA
-    CUDA_CALL(cudaGetDevice(&restore_device_));
-    CUDA_CALL(cudaSetDevice(device));
+    CUdevice cudev;
+    auto err = cuCtxGetDevice(&cudev);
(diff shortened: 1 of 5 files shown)


maxhgerlach:
Github builds with docker-compose don't work right now, but the ppc64le build ran into an actual linker error:

> E   ImportError: /opt/anaconda3/envs/wmlce/lib/python3.9/site-packages/horovod/torch/mpi_lib_v2.cpython-39-powerpc64le-linux-gnu.so: undefined symbol: cuCtxGetDevice

https://powerci.osuosl.org/job/Horovod_PPC64LE_GPU_PIPELINE/view/change-requests/job/PR-3912/2/console

horovod/mxnet/cuda_util.cc
```diff
@@ -32,8 +32,19 @@ with_device::with_device(int device) {
...
 #if HAVE_CUDA
-    CUDA_CALL(cudaGetDevice(&restore_device_));
-    CUDA_CALL(cudaSetDevice(device));
+    CUdevice cudev;
+    auto err = cuCtxGetDevice(&cudev);
+    if (err == CUDA_ERROR_NOT_INITIALIZED ||
+        err == CUDA_ERROR_INVALID_CONTEXT) {
+      // If device has never been set on this thread,
+      // restore to supplied device.
+      restore_device_ = device;
```
EnricoMi, line 41:
```suggestion
       // If device has never been set on this thread,
       // restore to supplied device.
       restore_device_ = device;
```

horovod/torch/cuda_util.cc
```diff
@@ -32,7 +32,18 @@ with_device::with_device(int device) {
...
   } else {
 #if HAVE_GPU
-    C10_CUDA_CHECK(cudaGetDevice(&restore_device_));
+    CUdevice cudev;
+    auto err = cuCtxGetDevice(&cudev);
+    if (err == CUDA_ERROR_NOT_INITIALIZED ||
+        err == CUDA_ERROR_INVALID_CONTEXT) {
+      // If device has never been set on this thread,
+      // restore to supplied device.
+      restore_device_ = device;
```
EnricoMi, line 41:
```suggestion
       // If device has never been set on this thread,
       // restore to supplied device.
       restore_device_ = device;
```

EnricoMi (commented):
minor whitespace suggestions

romerojosh:
Hmm, the driver linking error is unfortunate. I thought maybe that would happen. I can try and fix the build to link in the driver libraries. Alternatively, there is a much simpler option here that doesn't require adding CUDA driver APIs at all. 

The current usage of this `with_device` class is unnecessary. The purpose of that object is to "reset" the Horovod thread to the device it was previously using. However, the Horovod thread is not really associated with a specific device; we just set the device as required based on the tensors we are processing (see: https://github.com/horovod/horovod/blob/master/horovod/common/ops/gpu_operations.cc#L36). In fact, we never "reset" the device after performing a GPU collective operation. As a simpler solution, I propose removing the CUDA device reset in `with_device` altogether (or replace all usage of `with_device` with `cudaSetDevice()` calls.

maxhgerlach:
> The current usage of this with_device class is unnecessary. The purpose of that object is to "reset" the Horovod thread to the device it was previously using. However, the Horovod thread is not really associated with a specific device; we just set the device as required based on the tensors we are processing

I am not super familiar with these concerns. But apparently `with_device` is not only called from the Horovod background thread, but also from Torch "foreground" threads. For instance within `TorchOpContext::AllocateOutput` (1), which is directly called by `DoReducescatter` (2) and related functions. `DoBroadcast` etc. have a similar usage (3). Would we need to be more careful there then, so we don't interfere with other framework operations?


(1): https://github.com/horovod/horovod/blob/1ea00673d9a2e2cbf47278e2353039c3c6916cab/horovod/torch/adapter_v2.cc#L146
(2): https://github.com/horovod/horovod/blob/1ea00673d9a2e2cbf47278e2353039c3c6916cab/horovod/torch/mpi_ops_v2.cc#L764
(3): https://github.com/horovod/horovod/blob/1ea00673d9a2e2cbf47278e2353039c3c6916cab/horovod/torch/mpi_ops_v2.cc#L544

romerojosh:
Ah, good catch. I wasn't aware of any usage of the allocation functions in the `mpi_ops` definitions (seems it is only ReduceScatter that does it). Also, I thought PyTorch broadcast was implemented similar to MXNet, where the tensor creation was handled in Python like this: https://github.com/horovod/horovod/blob/1ea00673d9a2e2cbf47278e2353039c3c6916cab/horovod/mxnet/mpi_ops.py#L424-L428
but apparently not. 

Ok, I'll move forward with the current implementation then.

maxhgerlach:
> I wasn't aware of any usage of the allocation functions in the mpi_ops definitions (seems it is only ReduceScatter that does it).

True, it felt a bit unconventional to rework it like that in #3824, but also more straight forward than exposing `ReducescatterOp::ComputeOutputShapeForRank` to Python (or duplicating the logic there).

maxhgerlach:
Jenkins passed! 👍 

The `docker-compose` fix from #3913 seems to be picked up already (maybe because this is a fork PR and the GitHub CI pipeline runs on master config?), but not the docs build fix. I think the failing readthedocs build should not block the Buildkite GPU pipelines, so this might run through without rebasing to master first.

maxhgerlach (approved):
LGTM, provided that the GPU test pipeline passes.

maxhgerlach:
Hmm, the method using `cuCtxGetDevice` apparently doesn't work when one tries to use a CUDA-aware Horovod binary in a CPU-only context. CPU-running tests for the GPU aware docker image then fail with `ImportError: libcuda.so.1: cannot open shared object file: No such file or directory`: https://github.com/horovod/horovod/actions/runs/4897708229/jobs/8747477066?pr=3912

I think that's not an untypical use case. Would it be feasible to only load the driver library dynamically if a GPU device is available at run time?

romerojosh:
Ok. I think the way around this is to load the driver library and functions at runtime using `dlopen/dlsym`. I will implement this.

Issue #3927, opened 2023-05-18
ROCm-Compilation fails for 0.28.0


will-HPE:
**Environment:**
1. Framework: (TensorFlow, Keras, PyTorch, MXNet)
2. Framework version:
3. Horovod version: 0.28.0
4. MPI version:
5. CUDA version:
6. NCCL version:
7. Python version:
8. Spark / PySpark version:
9. Ray version:
10. OS and version:
11. GCC version:
12. CMake version:

**Checklist:**
1. Did you search issues to find if somebody asked this question before? Yes
2. If your question is about hang, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/running.rst)? N/A
3. If your question is about docker, did you read [this doc](https://github.com/horovod/horovod/blob/master/docs/docker.rst)? N/A
4. Did you check if you question is answered in the [troubleshooting guide](https://github.com/horovod/horovod/blob/master/docs/troubleshooting.rst)?

**Bug report:**
Please describe erroneous behavior you're observing and steps to reproduce it.
When building horovod-0.28.0 an update to torch/cuda_util.cc causes torch/hip_util.cc to fail build:
```
In file included from /tmp/pip-req-build-favtnjl1/horovod/torch/../common/ops/collective_operations.h:26,
                     from /tmp/pip-req-build-favtnjl1/horovod/torch/mpi_ops_v2_hip.cc:33:
    /tmp/pip-req-build-favtnjl1/horovod/torch/../common/ops/../half.h: In function 'void horovod::common::HalfBits2Float(const short unsigned int*, float*)':
    /tmp/pip-req-build-favtnjl1/horovod/torch/../common/ops/../half.h:76:11: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
       76 |   *res = *reinterpret_cast<float const*>(&f);
          |           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    /tmp/pip-req-build-favtnjl1/horovod/torch/hip_util.cc:34:28: error: expected ')' before '*' token
       34 | typedef hipError_t (CUDAAPI *PFN_cuCtxGetDevice)(hipDevice_t* device);
          |                    ~       ^~
          |                            )
    /tmp/pip-req-build-favtnjl1/horovod/torch/hip_util.cc:36:8: error: 'PFN_cuCtxGetDevice' does not name a type; did you mean 'hipCtxGetDevice'?
[...]

PR #3928, opened 2023-05-19
Fix torch build on ROCm by avoiding dynamically loaded cuCtxGetDevice

will-HPE:
Macro testing on `__HIP_PLATFORM_HCC__` as workaround ROCm's not supporting PFN_cuCtxGetDevice (does support cuCtxGetDevice -> hipCtxGetDevice).

## Description

Fixes # (issue).
https://github.com/horovod/horovod/issues/3927
## Review process to land 

### DIFF
--- a/horovod/torch/cuda_util.cc
+++ b/horovod/torch/cuda_util.cc
@@ -49,7 +49,6 @@ static void initialize_driver_api() {
   }
 }
 #endif
-#endif
 
 with_device::with_device(int device) {
   if (device == CPU_DEVICE_ID) {
--- a/horovod/torch/cuda_util.cc
+++ b/horovod/torch/cuda_util.cc
@@ -29,25 +29,27 @@
 namespace horovod {
 namespace torch {
 
+#ifndef __HIP_PLATFORM_HCC__
 #if HAVE_GPU
-typedef CUresult (CUDAAPI *PFN_cuCtxGetDevice)(CUdevice* device);
+typedef CUresult(CUDAAPI* PFN_cuCtxGetDevice)(CUdevice* device);
 static void* cudalib = nullptr;
 static PFN_cuCtxGetDevice pfn_cuCtxGetDevice = nullptr;
 
 static void initialize_driver_api() {
   // Clear previous errors
-  (void) dlerror();
+  (void)dlerror();
 
   cudalib = dlopen("libcuda.so", RTLD_LAZY);
   if (!cudalib) {
     throw std::logic_error("Internal error. Could not dlopen libcuda.so.");
   }
 
-  pfn_cuCtxGetDevice = (PFN_cuCtxGetDevice) dlsym(cudalib, "cuCtxGetDevice");
+  pfn_cuCtxGetDevice = (PFN_cuCtxGetDevice)dlsym(cudalib, "cuCtxGetDevice");
   if (!pfn_cuCtxGetDevice) {
     throw std::logic_error("Internal error. Could not load cuCtxGetDevice.");
   }
 }
+#endif
 #endif
 
 with_device::with_device(int device) {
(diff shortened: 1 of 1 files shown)


will-HPE:
Sorry, I keep using ctrl-enter as enter and that submits on here.
So the submission was premature.

paklui:
Hi @will-HPE , small comment, I think `__HIP_PLATFORM_HCC__` may not get defined for ROCm, I think starting ~ ROCm 5.x? it uses `__HIP_PLATFORM_AMD__` instead of `__HIP_PLATFORM_HCC__`. https://docs.amd.com/bundle/HIP-Programming-Guide-v5.1/page/Transitioning_from_CUDA_to_HIP.html#d3937e736 . Do you think about below? Thanks

```
$diff ~/cuda_util.cc.modified ~/cuda_util.cc.will.unmod
32c32,33
< #if HAVE_CUDA
---
> #ifndef __HIP_PLATFORM_HCC__
> #if HAVE_GPU
51a53
> #endif
58c60
< #if HAVE_CUDA
---
> #ifndef __HIP_PLATFORM_HCC__
```

paklui:
I tired the patch and it appears to work

horovod/torch/cuda_util.cc
```diff
@@ -29,46 +29,52 @@
 namespace horovod {
 namespace torch {
 
+#ifndef __HIP_PLATFORM_HCC__
 #if HAVE_GPU
```
maxhgerlach, line 33:
```suggestion
#if !HAVE_ROCM
```

horovod/torch/cuda_util.cc
```diff
@@ -29,46 +29,52 @@
...
   }
 
-  pfn_cuCtxGetDevice = (PFN_cuCtxGetDevice) dlsym(cudalib, "cuCtxGetDevice");
+  pfn_cuCtxGetDevice = (PFN_cuCtxGetDevice)dlsym(cudalib, "cuCtxGetDevice");
   if (!pfn_cuCtxGetDevice) {
     throw std::logic_error("Internal error. Could not load cuCtxGetDevice.");
   }
 }
 #endif
+#endif
```
maxhgerlach, line 53:
```suggestion
#endif
```

horovod/torch/cuda_util.cc
```diff
@@ -29,46 +29,52 @@
...
 
 with_device::with_device(int device) {
   if (device == CPU_DEVICE_ID) {
     restore_device_ = CPU_DEVICE_ID;
   } else {
 #if HAVE_GPU
-    if (!cudalib) initialize_driver_api();
+#ifndef __HIP_PLATFORM_HCC__
```
maxhgerlach, line 60:
```suggestion
#if !HAVE_ROCM
```

maxhgerlach (changes requested):
Thank you for providing this fix, @will-HPE, and thank you for testing it on your system, @paklui!

Please note that you would need to sign off your commit for DCO (see https://github.com/horovod/horovod/pull/3928/checks?check_run_id=13614457547) before we can merge it.

One other note: In the Horovod codebase we typically use macros `HAVE_CUDA` and `HAVE_ROCM` that are set in the build process. Would it be possible to use these rather than `__HIP_PLATFORM_HCC__`? (See my suggested code changes.)

will-HPE:
Max,

I think I've completed the requested changes. But, I may have done them sub-optimally. Could you take another look and see if everything is as it should be?

maxhgerlach (approved):

maxhgerlach:
> I think I've completed the requested changes. But, I may have done them sub-optimally. Could you take another look and see if everything is as it should be?

Thanks for the update including the sign-off line on the commit, @will-HPE! Apparently GitHub didn't include one with the suggested edits, so I overrode the DCO check manually. I think in this case the attribution should be clear.

I've triggered the Horovod CI pipeline, which will build and test on Nvidia GPUs, happy to merge if it asses. For confirming that the build works on AMD GPUs we rely on user input, unfortunately.

PR #3937, opened 2023-06-01
Docs: Fix invalid link to local section

maxhgerlach:

## Description

Fix (warning in) readthedocs build.

A warning appeared regarding an invalid label, which started to fail the build. I couldn't figure out easily how to fix the link without introducing extra duplicate label warnings (seem to be related to the nested file structure with `mpi.rst` and `mpi_include.rst`), so I just removed it. In the latest or stable docs the link currently doesn't work at all anyway.

## Review process to land 

### DIFF
--- a/docs/mpi.rst
+++ b/docs/mpi.rst
@@ -61,7 +61,7 @@ multiprocessing issues that Open MPI has with RDMA which typically results in se
 does not have noticeable performance impact since most of the heavy communication is done by NCCL, which will use RDMA
 via RoCE or InfiniBand if they're available (see `Horovod on GPU <gpus.rst>`_).  Notable exceptions from this rule are
 models that heavily use ``hvd.broadcast()`` and ``hvd.allgather()`` operations.  To make those operations use RDMA,
-read the :ref:`Open MPI with RDMA<open-mpi-with-rdma>` section below.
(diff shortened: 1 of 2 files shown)


docs/mpi.rst
```diff
@@ -61,7 +61,7 @@ multiprocessing issues that Open MPI has with RDMA which typically results in se
 does not have noticeable performance impact since most of the heavy communication is done by NCCL, which will use RDMA
 via RoCE or InfiniBand if they're available (see `Horovod on GPU <gpus.rst>`_).  Notable exceptions from this rule are
 models that heavily use ``hvd.broadcast()`` and ``hvd.allgather()`` operations.  To make those operations use RDMA,
-read the `Open MPI with RDMA <#open-mpi-with-rdma>`_ section below.
+read the **Open MPI with RDMA** section below.
```
EnricoMi, line 64: Not nice but works:
```suggestion
read the `Open MPI with RDMA <#id1>`_ section below.
```

EnricoMi (approved):
Not sure what is better, referencing `#id1` or no reference.

docs/mpi.rst
maxhgerlach: Great, then let's go with this quick fix (as long as it works).

Issue #3933, opened 2023-05-29
Missing `-iface` argument in mpirun command generated by Horovod runner


alumik:
**Environment:**
1. Framework: TensorFlow, PyTorch
2. Framework version: 2.12.0, 2.0.1
3. Horovod version: 0.28.0
4. MPI version: MPICH (HYDRA) 4.1.1
5. CUDA version: 11.8
6. NCCL version: 2.16.5
7. Python version: 3.11.3
8. Spark / PySpark version: N/A
9. Ray version: N/A
10. OS and version: Ubuntu 22.04
11. GCC version: 11.3.0
12. CMake version: 3.26.3

**Bug report:**

**Issue Description:**

The current implementation of the Horovod runner translates the command

```
horovodrun -n 3 --network-interface enp94s0 -H server2:3
```

into an mpirun command. However, it seems that the generated mpirun command

```
mpirun -l -np 3 -ppn 3 -hosts server2 -bind-to none -map-by slot -genv NCCL_SOCKET_IFNAME=enp94s0
```

is missing the `-iface enp94s0` argument. This omission can cause errors in setups with multiple servers.

**Steps to Reproduce:**

- Install MPICH and Horovod.
- Run the command `horovodrun -n 3 --network-interface enp94s0 -H server2:3 echo hello`.

**Error Message:**

```
[proxy:0:0@server1] HYDU_sock_connect (lib/utils/sock.c:110): unable to get host address for server1
[proxy:0:0@server1] main (proxy/pmip.c:105): unable to connect to server server2 at port 39647 (check for firewalls!)
[mpiexec@server2] ui_cmd_cb (mpiexec/pmiserv_pmci.c:51): Launch proxy failed.
[mpiexec@server2] HYDT_dmxu_poll_wait_for_event (lib/tools/demux/demux_poll.c:76): callback returned error status
[mpiexec@server2] HYD_pmci_wait_for_completion (mpiexec/pmiserv_pmci.c:181): error waiting for event
[mpiexec@server2] main (mpiexec/mpiexec.c:247): process manager error waiting for completion
```

**Expected Fix:**

The mpirun command generated by the Horovod runner should include the `-iface` argument to ensure proper network interface binding and avoid the mentioned error.

**References:**

Possible culprit code snippet:

https://github.com/horovod/horovod/blob/b93a87a6c79233d85113b8a42f5bd513d6c0de91/horovod/runner/mpi_run.py#L169-L170

**Workaround:**

[...]

PR #3946, opened 2023-06-21
Fix missing -iface argument in generated mpirun command

alumik:

## Description

This pull request addresses GitHub issue #3933, which reports a missing `-iface` argument in the mpirun command generated by the Horovod runner. The omission of this argument can lead to errors in setups with MPICH and multiple servers.

## Review process to land 

### DIFF
--- a/horovod/runner/mpi_run.py
+++ b/horovod/runner/mpi_run.py
@@ -166,8 +166,15 @@ def mpi_run(settings, nics, env, command, stdout=None, stderr=None):
         joined_ssh_args = ' '.join(ssh_args)
         mpi_ssh_args = f'-bootstrap=ssh -bootstrap-exec-args \"{joined_ssh_args}\"' if impi_or_mpich else f'-mca plm_rsh_args \"{joined_ssh_args}\"'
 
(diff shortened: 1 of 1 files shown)


maxhgerlach:
Thank you for the contribution, @alumik! Changes look good, let's just wait for the full CI pipeline.

maxhgerlach:
A couple of integration tests in `test.integration.test_static_run.StaticRunTests` fail. https://github.com/horovod/horovod/runs/14481555612

maxhgerlach:
Apparently `nics` generally has more than one entry here. I don't personally know how `horovodrun` handles that variable. Best way to go forward might be to investigate that test locally, then decide how to proceed.

alumik:
I have made some changes so that the `-iface` argument is only added if the user has explicitly set either the `--network-interface` or `--network-interfaces` options.

maxhgerlach:
Sounds great! I triggered another pipeline.

maxhgerlach:
`settings.nics` might still be set in those tests https://github.com/horovod/horovod/runs/14501458784

maxhgerlach:
Ah, the joys of untyped Python. 😉 

Thanks for persevering!

maxhgerlach (approved):

PR #3959, opened 2023-07-15
Add backward_passes_per_step for gradient aggregation with partial distributed optimizer

MrAta:

## Description
Add backward_passes_per_step for gradient aggregation with partial distributed optimizer.

Fixes # (issue).
N/A

## Review process to land 

### DIFF
--- a/horovod/_keras/__init__.py
+++ b/horovod/_keras/__init__.py
@@ -195,7 +195,7 @@ def create_distributed_optimizer(keras, optimizer, name, device_dense, device_sp
                                     if isinstance(grad, tf.IndexedSlices):
                                         grad = tf.IndexedSlices(grad.values / horovod_size, grad.indices, grad.dense_shape)
                                     else:
-                                        grad /= horovod_size
+                                        grad /= float(horovod_size)
                                     v2g[v_ref] = grad
 
                         return [v2g[rv.ref()] for rv in vars]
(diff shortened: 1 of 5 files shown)


MrAta:
Hi @romerojosh @maxhgerlach 

While adding this I was thinking where  the `PartialDistributedOptimzier` should be placed as we have three `DistributedOptmizer` defined in `horovod/tensorflow/__init__.py`, `horovod/tensorflow/keras/__init__.py`, and `horovod/keras/__init__.py`? 

Given that it's just a wrapper around DistributedOptimizer, feels like it could be at all three places; but I don't quite remember why I added it in `tensorflow/keras/__init__.py` when sending #3738 .

Can you please provide your input here? especially on the difference between horovod/keras and horovod/tensorflow/keras.

As an example, looking  #2346 (and related commits) seems the `backward_passes_per_step` is never given to the DistributedOptimizer in horovod/keras one.

maxhgerlach:
Sorry for the long wait, @MrAta!

Could you rebase your branch to master? That should fix the CI problems.

The proposed changes look good to me

> While adding this I was thinking where the `PartialDistributedOptimzier` should be placed as we have three `DistributedOptmizer` defined in `horovod/tensorflow/__init__.py`, `horovod/tensorflow/keras/__init__.py`, and `horovod/keras/__init__.py`?
> 
> Given that it's just a wrapper around DistributedOptimizer, feels like it could be at all three places; but I don't quite remember why I added it in `tensorflow/keras/__init__.py` when sending #3738 .
> 
> Can you please provide your input here? especially on the difference between horovod/keras and horovod/tensorflow/keras.
> 
> As an example, looking #2346 (and related commits) seems the `backward_passes_per_step` is never given to the DistributedOptimizer in horovod/keras one.

I don't have a strong opinion here, but I don't actively use much of Keras myself. Generally it seems that not every feature from `hvd.tensorflow` has been replicated in Keras-specific wrappers. I assume some redundancies could be removed from the code base nowadays if there is interest in systematizing it more.

MrAta:
Thanks for the update @maxhgerlach just rebased.

Re the redundancies, I will reach out to you and @romerojosh on slack to discuss it more. Thanks!

maxhgerlach (approved):
Thank you

PR #3977, opened 2023-08-31
CI: Bump container images to CUDA 11.6.2 and remove CUDA 10.x container configs

maxhgerlach:
Nvidia doesn't provide the CUDA 11.6.1 container anymore -> minor bump to 11.6.2

Also, Nvidia doesn't provide any CUDA 10.x containers anymore. To continue testing TF 1.15.5 we can switch to the `nvidia-tensorflow` package from https://github.com/NVIDIA/tensorflow. In doing so, we drop testing with some old CUDA-10-specific versions of mxnet or torch. This tensorflow wheel is built for Python 3.8 (previously we used Python 3.7 for TF 1.15), which requires updating pyspark from 2.4 to 3.0. Ultimately, this means that some minimum versions in CI would differ between CPU and GPU paths.

### DIFF
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -57,8 +57,8 @@ tests=$(if [[ -n "${PIPELINE_MODE:-}" ]] && ( [[ "${BUILDKITE_BRANCH:-}" == "${B
   printf "test-cpu-openmpi-gloo-py3_7-tfmin-kerasmin-torchmin-mxnetmin-pysparkmin "
 
   # then we vary the frameworks for gpu
+  # There is no CUDA10 nvidia/cuda image any more, so we move to nvidia-tensorflow==1.15.5+nv... (with CUDA 11.6) from mainline tensorflow-gpu==1.15.5, which requires us to deviate from CPU versions for python, torch, and mxnet.
   printf "test-gpu-gloo-py3_8-tf1_15_5-keras2_2_4-torch1_12_1-mxnet1_8_0_p0-pyspark3_4_0 "
-  # The container isn't provided for CUDA 10 anymore. The lowest version of mxnet available for cu112 is 1.8.0.post0.
   printf "test-gpu-gloo-py3_8-tf2_10_1-keras2_10_0-torch1_12_1-mxnet1_8_0_p0-pyspark3_4_0 "
   printf "test-gpu-gloo-py3_8-tf2_11_1-keras2_11_0-torch1_13_1-mxnet1_8_0_p0-pyspark3_4_0 "
   printf "test-gpu-openmpi-gloo-py3_8-tf2_12_0-keras2_12_0-torch2_0_0-mxnet1_9_1-pyspark3_4_0 "
--- a/.buildkite/gen-pipeline.sh
+++ b/.buildkite/gen-pipeline.sh
@@ -64,7 +64,7 @@ tests=$(if [[ -n "${PIPELINE_MODE:-}" ]] && ( [[ "${BUILDKITE_BRANCH:-}" == "${B
   printf "test-gpu-openmpi-gloo-py3_8-tf2_12_0-keras2_12_0-torch2_0_0-mxnet1_9_1-pyspark3_4_0 "
   printf "test-gpu-openmpi-gloo-py3_8-tfhead-keras_none-torchhead-mxnethead-pyspark3_4_0 "
   # these are the lowest framework versions that Horovod compiles with, but they are not tested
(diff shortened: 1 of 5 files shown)


maxhgerlach:
Hmm, the `nvidia-tensorflow` package (1.15.5) on the CUDA 11.6.2 container leads to this error unfortunately:
```
tensorflow.python.framework.errors_impl.InternalError: cudaGetDevice() failed. Status: CUDA driver version is insufficient for CUDA runtime version
```
https://buildkite.com/horovod/horovod/builds/9377#018a5184-7d24-431c-97b9-2f138347ad8c

From https://docs.nvidia.com/deeplearning/frameworks/support-matrix/index.html:
> Release 22.03 is based on [NVIDIA CUDA® 11.6.1](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html), which requires [NVIDIA Driver](http://www.nvidia.com/Download/index.aspx?lang=en-us) release 510 or later.

So downgrading to 22.03 may be a quick fix for now.

.buildkite/gen-pipeline.sh
```diff
@@ -55,20 +57,14 @@ tests=$(if [[ -n "${PIPELINE_MODE:-}" ]] && ( [[ "${BUILDKITE_BRANCH:-}" == "${B
...
-  # mxnet-cu100==1.7.0 does not contain mkldnn headers, and there is no 1.7.0.postx that would have them
-  # there is no mxnet-1.6.0.post0 and mxnet-1.6.0 does not work with horovod
-  # https://github.com/apache/incubator-mxnet/issues/16193
-  # so we test with mxnet 1.5.1
-  printf "test-gpu-gloo-py3_7-tf1_15_5-keras2_2_4-torch1_8_1-mxnet1_5_1_p0-pyspark3_4_0 "
-  # here we deviate from mxnet==1.7.0.post2 as there is none for cu101, only post1
-  printf "test-gpu-gloo-py3_8-tf2_10_1-keras2_10_0-torch1_12_1-mxnet1_7_0_p1-pyspark3_4_0 "
+  printf "test-gpu-gloo-py3_8-tf1_15_5-keras2_2_4-torch1_12_1-mxnet1_8_0_p0-pyspark3_4_0 "
+  # The container isn't provided for CUDA 10 anymore. The lowest version of mxnet available for cu112 is 1.8.0.post0.
+  printf "test-gpu-gloo-py3_8-tf2_10_1-keras2_10_0-torch1_12_1-mxnet1_8_0_p0-pyspark3_4_0 "
```
EnricoMi, line 62: Can we move the comment and rephrase a bit?
```suggestion
  # There is no CUDA10 nvidia/cuda image any more, so we move to nvidia-tensorflow==1.15.5+nv... with CUDA 11.6, which requires us to deviate from CPU versions for python, torch and mxnet.
  printf "test-gpu-gloo-py3_8-tf1_15_5-keras2_2_4-torch1_12_1-mxnet1_8_0_p0-pyspark3_4_0 "
  printf "test-gpu-gloo-py3_8-tf2_10_1-keras2_10_0-torch1_12_1-mxnet1_8_0_p0-pyspark3_4_0 "
```

docker-compose.test.yml
```diff
@@ -166,48 +168,42 @@ services:
...
-  # torch==1.8.1 is the latest we can test in this setup
-  # mxnet-cu100==1.7.0 does not contain mkldnn headers, and there is no 1.7.0.postx that would have them
-  # there is no mxnet-1.6.0.post0 and mxnet-1.6.0 does not work with horovod
-  # https://github.com/apache/incubator-mxnet/issues/16193
-  # so we test with mxnet 1.5.1
-  test-gpu-gloo-py3_7-tf1_15_5-keras2_2_4-torch1_8_1-mxnet1_5_1_p0-pyspark3_4_0:
+  # Mainline tensorflow-gpu==1.15.5 is compiled against and linked to CUDA 10.0, but appropriate containers aren't
+  # available anymore. Hence, we use the updated Python 3.8 wheel provided by Nvidia, see
+  # https://github.com/NVIDIA/tensorflow.
+  test-gpu-gloo-py3_8-tf1_15_5-keras2_2_4-torch1_12_1-mxnet1_8_0_p0-pyspark3_4_0:
```
EnricoMi, line 174: We also have to deviate from CPU versions for torch and mxnet.

EnricoMi (approved):
LGTM! minor comment change suggestions

docker-compose.test.yml
maxhgerlach: thanks, updated

EnricoMi (approved):
LGTM!

PR #4004, opened 2023-11-15
Prevent excessive parallel jobs during compilation

njzjz:

## Description

The number of job slots should be fewer than the number of available logical CPUs in the system.

## Review process to land 

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,7 +115,7 @@ class custom_build_ext(build_ext):
                       '-DCMAKE_LIBRARY_OUTPUT_DIRECTORY_{}={}'.format(config.upper(), build_dir),
                       '-DPYTHON_EXECUTABLE:FILEPATH=' + sys.executable]
 
-        make_args = ['-j8'] if not os.environ.get('MAKEFLAGS') else []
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -115,7 +115,7 @@ def build_extensions(self):
                       '-DCMAKE_LIBRARY_OUTPUT_DIRECTORY_{}={}'.format(config.upper(), build_dir),
                       '-DPYTHON_EXECUTABLE:FILEPATH=' + sys.executable]
 
-        make_args = ['-j8'] if not os.environ.get('MAKEFLAGS') else []
+        make_args = ['-j' + str(min(8, len(os.sched_getaffinity(0))))] if not os.environ.get('MAKEFLAGS') else []
```
EnricoMi, line 118: Looks like `sched_getaffinity` is only available on some UNIX platforms. What is the return value of it when not available?

njzjz: I see. The method doesn't exist in  other platforms, per https://github.com/python/cpython/issues/81781. I'll fix that.

PR #4010, opened 2023-12-20
[fix]func return no values, and catch SystemExit

gl-001:

## Description
```python
def fn():
    from absl import app
    app.run(main)

horovod.run(
    fn,
    np=1,
    hosts="localhost:4",
    verbose=3,
    use_mpi=True
)
```
In the case above will raise Exception, whether the fn is error or not.
Fixes # (issue).
when the fn use absl.app.run, will raise SystemExit.
## Review process to land 


EnricoMi:
Horovod requires `func` to not call `sys.exit` but simply return a value, so that `SystemExit` should be handled in `func`:

```python
def fn():
    from absl import app

    try:
        app.run(main)
    except SystemExit as e:
        if e.code:
            raise e
```

PR #4026, opened 2024-03-01
pass `*args, **kwargs` to `Optimizer.zero_grad`

njzjz:

## Description

PyTorch 1.7 added `set_to_none`  to `Optimizer.zero_grad` (see https://github.com/pytorch/pytorch/commit/c51588113779c0dbdeddff75ddf76f9bfe8d1d92). Horovod needs to be compatible with PyTorch 1.5, so I pass `*args, **kwargs` to `zero_grad` in this PR, which is safe in both old and new PyTorch versions.

## Review process to land 


njzjz:
Any reviewers here? This PR is just a simple fix.

PR #4031, opened 2024-03-15
Resolve TF saved model not portable issue with  tf.keras.optimizers

supercharleszhu:

- [X] Did you write any tests to validate this change?  

## Description

Fixes # (4028).
Root Cause: the `self._allreduce_grad` will be treated as a tf.function when we initialize the optimizer outside tf.function([code](https://github.com/horovod/horovod/blob/9f88e1df14daec6f4e64bfad981536eb6d034cbd/horovod/tensorflow/__init__.py#L626)). Since all the concrete function will be saved into saved_model.pb (see code [here](https://github.com/tensorflow/tensorflow/blob/eb2351e70baad69552ae9f415f926bc22aa2fe86/tensorflow/python/saved_model/save.py#L969)), it will cause the HorovodAllRuduce Op to be saved into the graph, which might not be registered in other environment. 

When using the [tf.keras.optimzer.legacy](https://github.com/supercharleszhu/tensorflow/blob/cd257c973ee383fc7904f4ae3a7fb1c5ca4e3b9b/tensorflow/python/keras/optimizer_v2/optimizer_v2.py#L533), It will not reach this path because when we call model.fit, the optimizer.minimize will call [_compute_gradients](https://github.com/supercharleszhu/tensorflow/blob/cd257c973ee383fc7904f4ae3a7fb1c5ca4e3b9b/tensorflow/python/keras/optimizer_v2/optimizer_v2.py#L533) which is not overriden by `compute_gradient` function in DistributedOptimizer, so the distributed optimizer is not taking effect at all!
# Resolution
We don't need to explicitly register the `allreduce_grad` as a tf function, as in graph mode, it will be trace in the outer tf.function. In this case, the horovod ops will not be saved explicitly

## Review process to land 

### DIFF
--- a/horovod/tensorflow/__init__.py
+++ b/horovod/tensorflow/__init__.py
@@ -622,10 +622,7 @@ def _make_cached_allreduce_grads_fn(name, device_dense, device_sparse,
                     if grad is not None else grad
                     for i, grad in enumerate(grads)]
 
-    if _executing_eagerly():
-        return _make_subgraph(allreduce_grads)
-    else:
-        return allreduce_grads
+    return allreduce_grads
 
 
 def _make_allreduce_grads_fn(name, device_dense, device_sparse,


EnricoMi:
Needs #4033.

supercharleszhu:
@EnricoMi  Has the CI been fixed? Seems that above 2 prs are checked in.

PR #4033, opened 2024-03-21
Fix MXNet MNIST download url

EnricoMi:
The MNIST download url provided by MXNet is dead, added zip file to Horovod datasets S3 and use that.


EnricoMi:
Blocked by CUDA upgrade of Buildkite workers. After upgrading to CUDA 12.4, tests fail with

    RuntimeError: "mul_cpu" not implemented for 'Half'

https://buildkite.com/horovod/horovod/builds/9580#018e6547-5449-46c1-a325-e6ff0f9c798e/6-14130

PR #4035, opened 2024-03-21
Move to v6.17.0 queue with CUDA 12.4

EnricoMi:
Updated Buildkite instances to CUDA 12.4 and Buildkite stack v6.17.0.


EnricoMi:
Contained in #4033 to test.

PR #4298, opened 2025-07-29
lazily load shared libs to avoid tf/torch/mxnet conflict

hjp709394:

## Description

Fixes # (issue).

When user import horovod, mpi_ops of torch, tensorflow, mxnet will be imported along. And import mpi_ops.py has two side effects: 
1. create a global HorovodBasics and load c++ libraries. 
``` torch/mpi_ops.py
_basics = _HorovodBasics(__file__, 'mpi_lib_v2')
...
_setup_process_sets(_basics)
```
2. set the process_sets.py's module level reference to a _basiscs object.
``` process_set.py
def _setup(basics):
    # type: (Optional[HorovodBasics]) -> None
    """" Horovod internal, to be called after the Horovod C++ module has been loaded. """
    global _basics
    _basics = basics
```

These two side effects create issue when more than one frameworks is used in the training infra. For instance: if both tensorflow and torch are imported, then two HorovodBasics (tensorflow and torch each has one) instances will be created and two c++ mpi lib will be loaded. And whoever is imported later will set process_sets.py's _basics reference. 

process_sets.py rely on the _basics to be correctly referenced and be init (through hvd.init()) correctly to figure out workers within cluster. 
Then apply gradient synchronization among the workers (indicated by process_set) in each step inside DistributedOptimizer:
```
def allreduce(tensor, average=None, device_dense='', device_sparse='',
              compression=Compression.none, op=None,
              prescale_factor=1.0, postscale_factor=1.0,
              name=None, process_set=global_process_set,
              ignore_name_scope=False):
```

The one that fail to set process_sets.py's _basics will not obtain process_set correctly then fail to apply gradient sync. 

This issue happens only for a cluster that import more than one horovod of tensorflow, torch and mxnet. Only the one that got imported last will work correctly.

This diff fix the issue by:
1. _setup_process_sets(self) inside HorovidBasics.init to reset the _basics reference of process_sets.py. So that the correct HorovodBasics reference is used by process_set. 'correct' here means the one with hvd.init() is explicitly called. 
2. lazy executes 'ctypes.CDLL' to avoid multiple c++ mpi libs are loaded and mess up NCCL set up. 

With the fix, as long as one hvd.init is called, the gradient synchronization will work as expected. Multiple hvd.init will introduce undefined behavior. 

Under a internal environment where both tf & torch are imported and tf horovod is imported earlier than torch, meaning tensorflow's grad-sync does not work. Before the fix, we can see horovod_timeline:
Screenshot 2025-07-29 at 5 05 11 PM
where no nccl_allreduce happen after each step. 

after the fix, we see the horovod_timeline:
Screenshot 2025-07-29 at 5 05 59 PM
where nccl_allreduce happens after each step. 

## Review process to land 

5. At least one member of the [technical steering committee](https://github.com/horovod/horovod/blob/master/GOVERNANCE.md) must review and approve.
6. If any member of the technical steering committee requests changes, they must be addressed.

### DIFF
--- a/horovod/common/basics.py
+++ b/horovod/common/basics.py
@@ -23,6 +23,7 @@ class MPI:
         ...
 
 from horovod.common.process_sets import ProcessSet, global_process_set, _init_process_sets
+from horovod.common.process_sets import _setup as _setup_process_sets
 from horovod.common import util as util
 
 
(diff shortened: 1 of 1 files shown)


horovod/common/basics.py
```diff
@@ -23,22 +23,32 @@ class Comm:
...
+
+        # These ReduceOp must be kept in sync with horovod/common/message.h
+        # We can not get these values from ctypes (self.MPI_LIB_CTYPES.horovod_reduce_op_average() etc.) dynamically
+        # because ctypes hasn't load shared lib yet, but other components need to import these ReduceOp now.
+        AVERAGE = 0
+        SUM = 1
+        ADASUM = 2
+        MIN = 3
+        MAX = 4
+        PRODUCT = 5
```
irasit, line 44: Please put these definition outside of this class's init function. Can also be out of the class.

